PR #48, opened 2015-02-09
[enh] Add app settings to redirect request

zamentur:
This commit add app settings to redirect a page with ssowat. There is also a fix on the modules api, to access it without been connected.


lib/yunohost/app.py
```diff
@@ -953,7 +959,17 @@ def app_ssowatconf(auth):
 
     msignals.display(m18n.n('ssowat_conf_generated'), 'success')
 
-
+def app_listmodules():
```
Kloadut: Is this function working ? Is it worth merging along above modifications ?

bin/yunohost-api
```diff
@@ -3,6 +3,9 @@
 
 import sys
 import os.path
+import yaml
+from os import listdir
+from os.path import isfile, join
```
jeromelebleu: Why importing those modules since it's not used in this file?

julienmalik:
Please resubmit this PR against "unstable"

PR #50, opened 2015-02-22
[enh] Add app settings to redirect request

zamentur:
This commit add app settings to redirect a page with ssowat.
Exemple of use:
>  sudo yunohost app setting $app redirected_urls -v "{'/ynhpanel.js':'$path/menu.js'}"

This allow to redirect the ynh_panel.js, like that I can display a menu on all pages in place of the square yunohost panel.


jeromelebleu:
As discussed, a new category - e.g. *ssowat* - with actions for adding/changing/deleting SSOwat rules - instead of using app_setting - should be added. So it will be subject to change!

PR #51, opened 2015-02-22
[enh] An api to list adminjs modules in /usr/share/yunohost/admin/modules/ which doesn't need authentication

zamentur:
Whith this pull request you can call yunohost app listmodules to get modules listed in /usr/share/yunohost/admin/modules/.

The web api is called like that https://yunohost.local/yunohost/api/modules . There is no need to authenticate (because the adminjs password is not available when adminjs is loading). The result is like this:
{"modules": ["helloworld", "menu", "usersubscribe"]}

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -538,6 +538,12 @@ app:
                 apps:
                     nargs: "+"
 
+        ### app_listmodules()
+        listmodules:
+            action_help: List all web admin modules
+            api: GET /modules
+            configuration:        
+                authenticate: False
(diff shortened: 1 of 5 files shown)


Psycojoker:
@zamentur is this PR still needed?

julienmalik:
I also propose we close this to clean up the PR queue. This is old and not mergeable anymore.

zamentur:
https://dev.yunohost.org/issues/47
I close this PR beacause this code is too old. The idea was to create a plug in system in the yunohost-admin. To do that a list of each plugin was needed.

PR #55, opened 2015-03-16
Reboot function in the admin/tools menu

polytan02:
Reboot function (to be done properly !), to add the possibility of remotely rebooting the yunohost system.  
The function added is JUST AN EXAMPLE and needs to be done properly.

The insterest is to have a dedicated entry in the web interface.

### DIFF
--- a/lib/yunohost/tools.py
+++ b/lib/yunohost/tools.py
@@ -469,3 +469,11 @@ def tools_upgrade(auth, ignore_apps=False, ignore_packages=False):
     if is_api:
         from yunohost.service import service_log
         return { "log": service_log('yunohost-api', number="100").values()[0] }
+
+
+def tools_reboot_server():
+    """
+    Simple function to reboot the server. 
+    Interesting to use via the web interface.
+    """
+    os.system('reboot')


opi:
Due to #94 from @JocelynDelalande , I guess we must use something else than `os.system` to reboot the server.

But I'd like to have this tool included, especially in InternetCube purpose, as one needs to reboot periodically.

Once this PR is merged, I could implement the admin view ( https://github.com/YunoHost/yunohost-admin/pull/90 )

opi:
@JocelynDelalande is `subprocess.call(['sudo', 'reboot'])` right ?

JocelynDelalande:
> is subprocess.call(['sudo', 'reboot']) right ?

Yes, but you may want to use `subprocess.check_call(['sudo', 'reboot'])` that will raise a `subprocess.CalledProcessError` in case the underlying command do not return a zero-status.

@opi why are you suggesting sudo whereas @polytan02 does not ?

opi:
About `sudo`, it may be a bad habbit from app install script, and maybe not required in moulinette context.
I'll let moulinette-yunohost regular contributors answer.

zamentur:
The actionmap changes are missing.
For the name I suggest simply reboot:

`yunohost tools reboot`

Psycojoker:
If this is for the `yunohost-api` process then it's already running has root and YunoHost forces you to run as root do we don't need a `sudo` here:

```
admin@browny:~$ yunohost app list
Error: yunohost command must be run as root or with sudo.
```

Psycojoker:
I'm interested in seeing this merged and I'm proposing this PR to be a minor decision https://github.com/YunoHost/yunohost-project-organization#d%C3%A9cision-mineure

I'm ready to take on it to finish it.

julienmalik:
Seems a reasonable feature :+1:
Of course, currently it is not mergeable as-is.
OK with @zamentur on the name.

opi:
- should we use `systemctl reboot` ? (don't really know pros/cons)
- should we have a `halt` method too ? I think of internetcu.be boxes that might be turned off properly

Psycojoker:
> should we use systemctl reboot ? (don't really know pros/cons)

I have no idea neither.

> should we have a halt method too ? I think of internetcu.be boxes that might be turned off properly

That could be interesting indeed, dangerous but cool (we need to have a good UX for that to avoid drama). I would do that in another PR however (I love micro PR).

mbugeia:
> should we use systemctl reboot ? (don't really know pros/cons)

That basically the same, check on your yunohost, the reboot command link to systemctl as shutdown and halt.

Psycojoker:
I have reopen this PR here to fix the conflicts https://github.com/YunoHost/yunohost/pull/190

Psycojoker:
Therefor I'm closing this one.

PR #58, opened 2015-04-14
Add alias module

mbugeia:
Hi, this is a extension to handle email alias, based on this post: https://forum.yunohost.org/t/add-a-redirection-mail-alias-without-attaching-it-to-a-yunohost-account/182

This will just create an email alias in LDAP that forwards messages to an email list. It accept email list delimited by commas to add/remove email in bulk. This is a first version, it seems to work well on my yunohost but there are probably untested case.

There is an init function to initialize the LDAP schema on a existing installation but maybe there is a better way to do this. Please tell me if you see any improvement to be done.

A cool feature would be an "all" alias, an email list for all yunohost user on a domain.

### DIFF
--- a/data/other/ldap_scheme.yml
+++ b/data/other/ldap_scheme.yml
@@ -28,6 +28,12 @@ parents:
             - organizationalUnit
             - top
 
+    ou=alias:
+        ou: alias
+        objectClass:
+            - organizationalUnit
+            - top
+
 children:
     cn=admins,ou=groups:
         cn: admins
--- a/lib/yunohost/alias.py
+++ b/lib/yunohost/alias.py
@@ -1,0 +1,237 @@
+# -*- coding: utf-8 -*-
+
+""" License
+
+    Copyright (C) 2014 YUNOHOST.ORG
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published
+    by the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


locales/en.json
```diff
@@ -193,5 +193,18 @@
     "pattern_backup_archive_name" : "Must be a valid filename with alphanumeric and -_. characters only",
 
     "format_datetime_short" : "%m/%d/%Y %I:%M %p"
```
Kloadut: lack a "," :)

Kloadut:
What a gorgeous contribution <3

Now we will have to plug this work on the Web interface and remove the old way to handle aliases (`yunohost user update myuser --add-mailalias alias@domain.org`)

Epic job though :+1:

lib/yunohost/alias.py
```diff
@@ -0,0 +1,237 @@
...
+
+    if auth.remove('uid=%s,ou=alias' % alias):
+        pass
+    else:
+        raise MoulinetteError(169, m18n.n('alias_deletion_failed'))
+
+    msignals.display(m18n.n('alias_deleted'), 'success')
+
+
+def alias_update(auth, alias, name=None, add_mailforward=None, remove_mailforward=None):
```
Kloadut: Why is this function needed ? You have alias_create and delete

julienmalik:
wow ! I should look into the pr list more often. this is awesome and I dreamed of having something like that several times.
sadly it now needs a rebase.

it would be such a shame to leave all this work sleeping here...

data/other/ldap_scheme.yml
```diff
@@ -28,6 +28,12 @@ parents:
             - organizationalUnit
             - top
 
+    ou=alias:
+        ou: alias
```
uda: I would suggest using `aliases` since all ou's use plural names

julienmalik:
This PR has been rebased/reworked in #176

PR #82, opened 2015-09-03
[enh] Add dropdb helper.

opi:


jeromelebleu:
I'll close this, please provide a helper script instead please, as we discussed once... See: https://dev.yunohost.org/issues/97

PR #89, opened 2015-09-13
Fix ipv6 dyndns

JocelynDelalande:
Was quite broken :)

In fact, ynh dyndns client was setting all *AAAA* records to `"::"` which is [supposed to be](http://tools.ietf.org/html/rfc4291#section-2.5.2) *"The Unspecified Address"*, a non-routable address.

However, in practice, this *unspecified address* is considered as loopback (try `ping6 ::` to figure out), and causing non-delivery when an IPv6-enabled MTA try to reach your dyndns domain, with that kind of error : 

     The mail system <john.doe@example.nohost.me>: mail for example.nohost.me loops back to myself

So it's a severe bug, **preventing all **.nohost.me* domains to work reliably** . Thus, I suggest, after review, to backport this fix on stable and testing as well.

It's good step towards #25 btw.


JocelynDelalande:
Any review/feedback on that ?

opi:
Add some pseudo-random ping to @jvaubourg @Kloadut @jeromelebleu ;)

Kloadut:
woa didn't see it.

LGTM :+1:

jvaubourg:
Very nice, this was a huge bug for me.

I'm not a Python expert, I can't really review this code. But I seen this comment: *Get the IPv6 address given by radvd and sanitize it*. With the VPN Client app (e.g. Internet Cube), the IP addresses to use in the DNS correspond to the IPs delivered by the VPN, and not the wired network.

The only solution I know is to use the routing table. For example, with `ip r g 2001:913::8` or `ip r g 80.67.188.188` (or other remote addresses), we exactly know the IPs and the interfaces used for going out.

JocelynDelalande:
> But I seen this comment: Get the IPv6 address given by radvd and sanitize it. 
> With the VPN Client app (e.g. Internet Cube), the IP addresses to use in the DNS correspond to the IPs delivered by the VPN, and not the wired network.

I just kept the logic as it was : get the first global v6 address found.

> The only solution I know is to use the routing table. For example, with ip r g 2001:913::8 or ip r g 80.67.188.188 (or other remote addresses), we exactly know the IPs and the interfaces used for going out.

That's a clever one.

My feeling is that changing the address selection logic normally goes to another PR. That said, this PR may break "working" IPv6 installations by creating unexpected DNS records, and worse, leak an address that the VPN mechanism should have kept secret.

So, I 'm working on integrating what you suggest.

JocelynDelalande:
@jvaubourg actually, as we rely on */etc/yunohost/interface* it should be fine, what is the value of */etc/yunohost/interface* on a yunohost installation using the VPN app ?

opi:
on my olimex board using vpnclient_ynh : 

```
admin@olinux:~$ cat /etc/yunohost/interface                                                                                        
eth0                                                                             
```

JocelynDelalande:
> on my olimex board using vpnclient_ynh :
> 
> admin@olinux:~$ cat /etc/yunohost/interface                                                                                        
> eth0

Ok, I'm wondering if it's not a bug then, do any of you know what is the purpose of /etc/yunohost/interface ? shoult it contain tun0 while using vpnclient_ynh ?

Kloadut:
The file is not used anymore. We should suppress it :)

JocelynDelalande:
> The file is not used anymore. We should suppress it :)

Actually, it's used by this dyndns script :/

Kloadut:
@JocelynDelalande Yup !

I am making a fix to get the IPv6 address the same way than the current IPv4, by reaching ip6.yunohost.org

JocelynDelalande:
@Kloadut I don't know… It is something we can do without relying on an external service (getting public ipv6), let's do it the propper way, don't you think ?

Kloadut:
I guess we should, if we manage to do the same for IPv4

Kloadut:
With ip6.yunohost.org we can forget about the interface name fetching

JocelynDelalande:
> I guess we should, if we manage to do the same for IPv4

No, we can't, because we're using all the time a hack called NAT, but in IPv6 we don't use such hacks, that's the beauty :)

jvaubourg:
@JocelynDelalande I agree with you, we don't have to use an external service for that with IPv6. But, in fact, it is the easier way to find the public IPv6 to use in AAAA (considering that the privacy extension is not enabled, as expected on a server). Finding a way to do `ip route get 80.67.188.188` with */proc* files is just a bit of a mess...

Kloadut:
If you successfully managed to do it, I can't be against it :-)

In the mean time, I pushed my working version (not released yet), feel free to adapt the part that you need.
https://github.com/YunoHost/moulinette-yunohost/commit/1ceb1259bc5f65cc94ce7447aa71117f9ad33dd2

JocelynDelalande:
> @JocelynDelalande I agree with you, we don't have to use an external service for that with IPv6. 
But, in fact, it is the easier way to find the public IPv6 to use in AAAA (considering that the privacy extension is not enabled, as expected on a server). 

well, the way you suggested with routing decision and source adress seems just fine to me, no ? (I mean, appart the fact `ip route` output is a bit shitty to parse).

> Finding a way to do ip route get 80.67.188.188 with /proc files is just a bit of a mess...

s/just a bit of a mess/impossible/ ?

JocelynDelalande:
@Kloadut on a more meta note, no big deal but I would appreciate that you do not copy/paste changes of a PR to some new commit. As a contributor :

- I like to review and amend my PR myself, at my pace, based on peer comments, taking time if it requires some discussion
- I like being credited for my work (commit message)
- I hate fixing conflict merge ;)

Thanks :)

PR #90, opened 2015-09-16
Minor changes in Python source files

RyDroid:
New option can be useful if an alias is defined with an option inside.

### DIFF
--- a/bin/yunohost
+++ b/bin/yunohost
@@ -1,6 +1,25 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+""" License
+
+    Copyright (C) 2015 YunoHost
+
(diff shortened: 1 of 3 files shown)


jeromelebleu:
Thanks for your PR, but I don't really understand the goal of the *--cache* and *--no-debug* options since it's the default behaviour... Can you explain please?

RyDroid:
Some person may define (by themselves or through a distribution package different than than the original) aliases like alias yunohost='yunohost --no-cache'. This can be hard to find out and user may not be able to change the alias permanently in potential YunoHost service more restricted than IaaS.

RyDroid:
Ooops. Do not merge currently, arguments matching tests should be done iteratively.

PR #91, opened 2015-09-16
HTTP links -> HTTPS links

RyDroid:


opi:
Not sure we have to change anything from original licence file : https://www.gnu.org/licenses/agpl.txt

Maybe you should pull request upstream to gnu.org ;)

PR #99, opened 2015-09-30
[fix] Open 1900 port to accept  UPNP responses

ZeHiro:
Fix https://github.com/YunoHost/moulinette-yunohost/issues/97


jeromelebleu:
Thanks for this point, I would suggest to not open it by default and instead open it when the UPnP is activated - as many people don't use it...

ZeHiro:
You're right. I'll propose something else.

jeromelebleu:
Fixed with https://github.com/YunoHost/yunohost/commit/4e72595aaa64f1372532203ad9728d3c7c63a220.

PR #104, opened 2015-10-03
Major update: Add a configuration file generator

Kloadut:
# The idea

We wanted to propose configuration updates on YunoHost instances for security and quality purposes. If the user has not modified her/his configuration files by her/himself, the configuration is automatically updated.
In the other hand, if a user has modified a configuration manually, we wanted to prevent this modification to be overwritten.

# Usage

You can regenerate all the configuration at once, or one service at a time.
```
sudo yunohost service regenconf
# Or
sudo yunohost service regenconf -s nginx
```

Note that you can pass a `--force` argument as well to force configuration overriding.

# Functioning

The action will compare a configuration file with one of a proposed [set of templates](https://github.com/YunoHost/moulinette-yunohost/tree/regen-conf/data/templates) and try to replace it, storing its modification at the same time.
* If the action finds no difference between the proposed configuration file and the current one, it only stores a hash of the file in `/etc/yunohost/services.yml`
* If the action finds a difference between the proposed configuration file and the current one, but the current one already had its hash stored and was not modified since, the proposed file overrides it.
* If the action finds a difference between the proposed configuration file and the current one, either because the current file was not registered before or because a manual modification was applied since then, the action displays a warning in addition to a diff between the two files. In this case, the user is free to apply the proposed modifications manually, or override her/his modifications by passing `--force` to the action.

# Side effect

We managed to merge all YunoHost packages into one: `yunohost`, which inherit all dependencies to services (`postfix`, `slapd`, etc.), the set of configuration templates, and the python script of the `moulinette` specific to YunoHost.


jeromelebleu:
Youhou! \O/

PR #105, opened 2015-10-03
App debug

Psycojoker:
New command for the app debug page.

Current limitations:

* nginx logs files aren't filtered on the current domain

New entry in the manifest.json: "services" which waits for a list of strings that are the services linked to an app.

#swag

### DIFF
--- a/lib/yunohost/app.py
+++ b/lib/yunohost/app.py
@@ -731,9 +731,8 @@ def app_debug(app):
     return {
         'name': manifest['name'],
         'services_logs': [{
-                'service': x,
-                'log': service_log(x),
-            } for x in manifest.get("services", [])]
+                x: service_log(x),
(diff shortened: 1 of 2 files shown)


lib/yunohost/app.py
```diff
@@ -722,6 +724,23 @@ def app_clearaccess(auth, apps):
     app_ssowatconf(auth)
 
 
+def app_debug(app):
```
Kloadut: Would be great with a commentary :-)

lib/yunohost/app.py
```diff
@@ -722,6 +724,29 @@ def app_clearaccess(auth, apps):
...
+    """
+    Display debug informations for an app
+
+    Keyword argument:
+        app
```
Kloadut: Thanks <3

PR #109, opened 2015-10-12
[fix] Use official.json app list.

opi:


Kloadut:
There is an alias `official.json -> list.json` on the server for compatibility

opi:
yep, saw that alias. First my patch was for cli documentation for actionmap. thx anyway

PR #114, opened 2015-11-04
WIP Letsencrypt integration

mbugeia:
This is a non-working WIP for Let's encrypt

### DIFF
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -69,8 +69,7 @@ def domain_list(auth, raw=False, filter=None, limit=None, offset=None):
     else:
         return { 'domains': result_list }
 
-
-def domain_add(auth, domain, dyndns=False):
+def domain_add(auth, domain, dyndns=False, noletsencrypt=False):
     """
     Create a custom domain
 
@@ -125,24 +124,29 @@ def domain_add(auth, domain, dyndns=False):
         try: os.listdir(ssl_domain_path)
         except OSError: os.makedirs(ssl_domain_path)
 
-        command_list = [
-            'cp %s/openssl.cnf %s'                               % (ssl_dir, ssl_domain_path),
-            'sed -i "s/yunohost.org/%s/g" %s/openssl.cnf'        % (domain, ssl_domain_path),
-            'openssl req -new -config %s/openssl.cnf -days 3650 -out %s/certs/yunohost_csr.pem -keyout %s/certs/yunohost_key.pem -nodes -batch'
-            % (ssl_domain_path, ssl_dir, ssl_dir),
-            'openssl ca -config %s/openssl.cnf -days 3650 -in %s/certs/yunohost_csr.pem -out %s/certs/yunohost_crt.pem -batch'
-            % (ssl_domain_path, ssl_dir, ssl_dir),
-            'ln -s /etc/ssl/certs/ca-yunohost_crt.pem %s/ca.pem' % ssl_domain_path,
-            'cp %s/certs/yunohost_key.pem    %s/key.pem'         % (ssl_dir, ssl_domain_path),
-            'cp %s/newcerts/%s.pem %s/crt.pem'                   % (ssl_dir, serial, ssl_domain_path),
-            'chmod 755 %s'                                       % ssl_domain_path,
-            'chmod 640 %s/key.pem'                               % ssl_domain_path,
-            'chmod 640 %s/crt.pem'                               % ssl_domain_path,
-            'chmod 600 %s/openssl.cnf'                           % ssl_domain_path,
-            'chown root:metronome %s/key.pem'                    % ssl_domain_path,
-            'chown root:metronome %s/crt.pem'                    % ssl_domain_path,
-            'cat %s/ca.pem >> %s/crt.pem'                        % (ssl_domain_path, ssl_domain_path)
-        ]
+        if noletsencrypt:
+            command_list = [
+                'cp %s/openssl.cnf %s'                               % (ssl_dir, ssl_domain_path),
+                'sed -i "s/yunohost.org/%s/g" %s/openssl.cnf'        % (domain, ssl_domain_path),
+                'openssl req -new -config %s/openssl.cnf -days 3650 -out %s/certs/yunohost_csr.pem -keyout %s/certs/yunohost_key.pem -nodes -batch'
+                % (ssl_domain_path, ssl_dir, ssl_dir),
+                'openssl ca -config %s/openssl.cnf -days 3650 -in %s/certs/yunohost_csr.pem -out %s/certs/yunohost_crt.pem -batch'
+                % (ssl_domain_path, ssl_dir, ssl_dir),
+                'ln -s /etc/ssl/certs/ca-yunohost_crt.pem %s/ca.pem' % ssl_domain_path,
+                'cp %s/certs/yunohost_key.pem    %s/key.pem'         % (ssl_dir, ssl_domain_path),
+                'cp %s/newcerts/%s.pem %s/crt.pem'                   % (ssl_dir, serial, ssl_domain_path),
+                'chmod 755 %s'                                       % ssl_domain_path,
+                'chmod 640 %s/key.pem'                               % ssl_domain_path,
+                'chmod 640 %s/crt.pem'                               % ssl_domain_path,
+                'chmod 600 %s/openssl.cnf'                           % ssl_domain_path,
+                'chown root:metronome %s/key.pem'                    % ssl_domain_path,
+                'chown root:metronome %s/crt.pem'                    % ssl_domain_path,
+                'cat %s/ca.pem >> %s/crt.pem'                        % (ssl_domain_path, ssl_domain_path)
+            ]
+        else:
+            command_list = [
+                'yunohost domain letsencrypt -c %s' % domain
+            ]
 
         for command in command_list:
             if os.system(command) != 0:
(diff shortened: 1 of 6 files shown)


stevenroose:
Just to make sure: does this work with multiple domains? Is it possible to automatically do this for all new domains?

Kloadut:
Awesome

src/yunohost/domain.py
```diff
@@ -218,3 +237,57 @@ def domain_remove(auth, domain, force=False):
...
+
+    """
+    if domain not in domain_list(auth)['domains']:
+        raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
+
+    if create and not renew and not revoke:
+
+        # backup self signed certificate if exist
+        if os.path.exists('/etc/yunohost/certs/%s/cert.pem' % domain):
+            os.system('mkdir -p /etc/yunohost/certs/%s/yunohost_self_signed' % domain)
```
Kloadut: POpen usage is recommended.

os.system is deprecated

mbugeia:
@stevenroose Yes, the way I try to do it should work with multiple domain.
The default for a new domain is to use let's encrypt unless if --noletsencrypt option is used.
I also included theses lines https://github.com/mbugeia/yunohost/blob/letsencrypt/data/hooks/conf_regen/16-letsencrypt#L26-L31 to perform automatic cert generation for existing domains, but I'm not sure it should be automatic so I will probably remove them.

pyrech:
Awesome work ;-) Any news on it? Is it near to be merged?

src/yunohost/domain.py
Psycojoker: Actually, you should use os.makedirs for this case https://docs.python.org/2/library/os.html#os.makedirs

src/yunohost/domain.py
```diff
@@ -218,3 +237,57 @@ def domain_remove(auth, domain, force=False):
...
+            os.system('sudo mv /etc/yunohost/certs/%s/*.pem /etc/yunohost/certs/%s/*.cnf /etc/yunohost/certs/%s/yunohost_self_signed/' % domain)
```
Psycojoker: Here https://docs.python.org/2/library/shutil.html#shutil.move

src/yunohost/domain.py
```diff
@@ -218,3 +237,57 @@ def domain_remove(auth, domain, force=False):
...
+            os.system('sudo rm -f /etc/yunohost/certs/%s/*.pem /etc/yunohost/certs/%s/*.cnf' % domain)
```
Psycojoker: And here https://docs.python.org/2/library/shutil.html#shutil.rmtree

src/yunohost/domain.py
```diff
@@ -218,3 +237,57 @@ def domain_remove(auth, domain, force=False):
...
+            os.system('sudo /root/letsencrypt/letsencrypt-auto -a webroot --renew-by-default --agree-dev-preview --agree-tos --webroot-path /etc/letsencrypt/webrootauth -m root@%s -d %s auth' % domain)
+            # restore right for metronome
+            os.system('sudo chown root:metronome /etc/letsencrypt/archive/%s/*' % domain)
+            # create cron
+            os.system('echo "@monthly root yunohost domain letsencrypt -r %s" > /etc/cron.d/letsencrypt-%s' % domain)
+            # symbolic link for cert and key
+            os.system('ln -s /etc/letsencrypt/live/%s/privkey.pem /etc/yunohost/certs/%s/key.pem' % domain)
+            os.system('ln -s /etc/letsencrypt/live/%s/fullchain.pem /etc/yunohost/certs/%s/crt.pem' % domain)
+            msignals.display(m18n.n('domain_letsencrypt_created'), 'success')
+        except:
```
Psycojoker: While I understand the current state of the PR, I very very strongly recommend NOT to keep a catch all exception mechanism here that doesn't provide any debug information, especially since the previous code is based on os.system that wil NOT raised if there is an error return by the bash script (and will continue to happily execute the next python code while it should have stopped before).

Ideally it would be great to have everything ported to full python use os and shutil and having a proper error mechanism to avoid debugging hell.

src/yunohost/domain.py
```diff
@@ -218,3 +237,57 @@ def domain_remove(auth, domain, force=False):
...
+            raise MoulinetteError(errno.EIO, m18n.n('domain_letsencrypt_create_failed'))
+
+    elif renew and not create and not revoke:
+        try:
+            os.system('sudo /root/letsencrypt/letsencrypt-auto -a webroot --renew-by-default --agree-dev-preview --agree-tos --webroot-path /etc/letsencrypt/webrootauth -m root@%s -d %s auth' % domain)
+            # restore right for metronome
+            os.system('sudo chown root:metronome /etc/letsencrypt/archive/%s/*' % domain)
+            msignals.display(m18n.n('domain_letsencrypt_renewed'), 'success')
+        except:
```
Psycojoker: Same comment than previously.

src/yunohost/domain.py
```diff
@@ -218,3 +237,57 @@ def domain_remove(auth, domain, force=False):
...
+            raise MoulinetteError(errno.EIO, m18n.n('domain_letsencrypt_renew_failed'))
+    elif revoke and not create and not renew:
+        try:
+            os.system('sudo /root/letsencrypt/letsencrypt-auto -a webroot --renew-by-default --agree-dev-preview --agree-tos --webroot-path /etc/letsencrypt/webrootauth -m root@%s -d %s auth' % domain)
+            msignals.display(m18n.n('domain_letsencrypt_revoked'), 'success')
+        except:
```
Psycojoker: And again.

src/yunohost/domain.py
Psycojoker: A quick hack in the mean time would be to have a

```python
assert os.system(sutff) == 0
```

But it should remain as a quick hack.

src/yunohost/domain.py
```diff
@@ -206,7 +210,22 @@ def domain_remove(auth, domain, force=False):
...
+                'yunohost domain letsencrypt revoke %s' % domain,
+                'rm -rf /etc/letsencrypt/archive/%s /etc/letsencrypt/live/%s' % domain,
+                'rm -f /etc/letsencrypt/renewal/%s.conf' % domain,
+                'rm -f /etc/cron.d/letsencrypt-{domain}' % domain
+            ])
+
+        for command in command_list:
+            if os.system(command) != 0:
+                msignals.display(m18n.n('path_removal_failed', command[7:]),
+                                 'warning')
```
Psycojoker: All this block should be converted to pure python and I think that instead of storing a list of functions to call, calling them directly will simplify the code and make it more maintainable.

Also "yunohost domain letsencrypt revoke" is a function stored in this python file right now, it would made more sens to call it directly (and make code more obvious) :)

src/yunohost/domain.py
```diff
@@ -125,24 +124,29 @@ def domain_add(auth, domain, dyndns=False):
...
+                'chmod 640 %s/crt.pem'                               % ssl_domain_path,
+                'chmod 600 %s/openssl.cnf'                           % ssl_domain_path,
+                'chown root:metronome %s/key.pem'                    % ssl_domain_path,
+                'chown root:metronome %s/crt.pem'                    % ssl_domain_path,
+                'cat %s/ca.pem >> %s/crt.pem'                        % (ssl_domain_path, ssl_domain_path)
+            ]
+        else:
+            command_list = [
+                'yunohost domain letsencrypt -c %s' % domain
+            ]
```
Psycojoker: Again for the command list pattern.

Psycojoker:
For a more general comment: thanks a **lot** for doing this work <3

mbugeia:
Hi,
@pyrech This probably won't be merged soon since it doesn't work.

@Psycojoker Thanks for the review. I agree with all your point. As you have probably see this was a quick and dirty PR ;) I was not aware of shutil lib which can replace some of the os.system call (and other should be replaced either by subprocess call or native os.xxx function like os.symlink).

Otherwise this was done before the public LE beta and some client usage have changed, this thread https://forum.yunohost.org/t/how-to-install-letsencrypt-certificates/1075 is more up to date. And there are some mistake like nginx need to be reload and metronome restarted after creating/renewing a cert.

Moreover some things have to be discussed before a proper LE integration like:
 - Is this a good idea to use the official LE client? It's very heavy and there is a bunch of other good client. Full list here: https://community.letsencrypt.org/t/list-of-client-implementations/2103
https://github.com/diafygi/acme-tiny/ and https://github.com/kuba/simp_le seems very good for yunohost usage.
 - Is revoke function really necessary with easy renewal and 90 days validity?
 - Is default to LE when creating a domain a good idea? Adding a domain will fail if DNS are not already propagated or configured for example, this can be a issue for yunohost dydns and need to be tested.

Unfortunatly I won't have time to work on this soon (I am about to move and change jobs, + christmas) :/ So if someone want to give it a try and maybe reuse some of my code go for it.

jeromelebleu:
Hello,

@mbugeia thanks for your work! It's a good starting point.

The Let's Encrypt support will be implemented soon in the branch [domain-enh](https://github.com/YunoHost/yunohost/tree/domain-enh) under a new *domain_cert* action - or a new *certificate* category.
The goal is to have a specific action to generate SSL Certificates - using pure Python code - which offers to select what to use - e.g. self-signed, existing CA or... Let's Encrypt!

alexAubin:
Hey there,

I just discovered the issue tracker and dev chat of yunohost, as well as this topic. Didn't realize some people were already working on let's encrypt integration before my tuto/package :P 

Is there any progress on this ? Though I'm new, I'd be willing to try to contribute to a proper integration. Here's my opinion on @mbugeia questions : 
- I agree that the LE official client is pretty heavy. Didn't try acme_tiny or simp_le yet, but they sound great ;
- Not sure about the revoke, but I think it does no harm to have it ;
- LE shouldn't be the default

alexAubin:
See #180

mbugeia:
Close because @alexAubin PR #180 is way more relevant.

PR #115, opened 2015-11-14
Make use of logging instead of msignals.display

jeromelebleu:
It is a work in progress with several changes:
 * replace msignals.display calls with properly configured logger (done for firewall, hook and service categories)
 * refactor and print usage of additional arguments of *yunohost* and *yunohost-api* binaries
 * enhance outputs/logs and make the debug easier 
 * re-introduce the good-old bash completion
 * ...


jeromelebleu:
Mmmh.... mmmokay, I take the risk of merging it in a really democratic way.

opi:
"enhance outputs/logs and make the debug easier" & "re-introduce the good-old bash completion" looks gorgeous !

jeromelebleu:
Unfortunately bash completion is not working behind one level - for now... e.g. `yunohost fire` + `TAB` doesn't work.

PR #118, opened 2015-12-28
hooks: Use a more elegant grep command for mysql process check

sbadia:
Hi here,

Just one less pipe :)

Cheers


jeromelebleu:
Subtle...!

PR #119, opened 2016-01-22
New command to generate DNS configuration for a given domain name

Psycojoker:
Actually, this command being so stupidly simple, it could be backported to testing.

I have no idea on how I'm supposed to handle DKIM (it is supposed to be integrated now but I don't know where/how).

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -291,6 +291,17 @@ domain:
                     extra:
                         pattern: *pattern_domain
 
+        ### domain_generate_dns_config()
+        generate_dns_configuration:
+            action_help: Generate DNS configuration for a domain
+            api: GET /domains/<domain>/dns_config
+            configuration:
+                lock: false
+                authenticate: all
+            arguments:
+                domain:
+                    help: Target domain
+
         ### domain_info()
 #        info:
 #            action_help: Get domain informations
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -297,7 +297,8 @@ domain:
             api: GET /domains/<domain>/dns
             configuration:
                 lock: false
-                authenticate: all
+                authenticate:
+                    - api
             arguments:
                 domain:
                     help: Target domain
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -291,8 +291,8 @@ domain:
                     extra:
                         pattern: *pattern_domain
 
-        ### domain_generate_dns_config()
-        generate_dns_configuration:
+        ### dns_conf()
+        dns-conf:
             action_help: Generate DNS configuration for a domain
             api: GET /domains/<domain>/dns
             configuration:
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -291,7 +291,7 @@ domain:
                     extra:
                         pattern: *pattern_domain
 
-        ### dns_conf()
+        ### domain_dns_conf()
         dns-conf:
             action_help: Generate DNS configuration for a domain
             api: GET /domains/<domain>/dns
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -294,7 +294,7 @@ domain:
         ### domain_generate_dns_config()
         generate_dns_configuration:
             action_help: Generate DNS configuration for a domain
-            api: GET /domains/<domain>/dns_config
+            api: GET /domains/<domain>/dns
             configuration:
                 lock: false
                 authenticate: all
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -245,7 +245,7 @@ def domain_generate_dns_configuration(auth, domain):
 
     result += "muc 1800 IN CNAME @\npubsub 1800 IN CNAME @\nvjud 1800 IN CNAME @\n\n"
 
-    result += "@ 1400 IN MX 10 {domain}.tld.\n".format(domain=domain)
+    result += "@ 1400 IN MX 10 {domain}.\n".format(domain=domain)
 
     if ip6 is None:
         result += '@ 1400 IN TXT "v=spf1 a mx ip4:{ip4} -all"\n'.format(ip4=ip4)
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -243,6 +243,8 @@ def domain_generate_dns_configuration(auth, domain):
 
     result += "\n_xmpp-client._tcp 14400 IN SRV 0 5 5222 {domain}.\n_xmpp-server._tcp 14400 IN SRV 0 5 5269 {domain}.\n".format(domain=domain)
 
+    result += "muc 1800 IN CNAME @\npubsub 1800 IN CNAME @\nvjud 1800 IN CNAME @\n\n"
+
     result += "@ 1400 IN MX 10 {domain}.tld.\n".format(domain=domain)
 
     if ip6 is None:
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -220,7 +220,7 @@ def domain_remove(auth, domain, force=False):
     msignals.display(m18n.n('domain_deleted'), 'success')
 
 
-def domain_generate_dns_configuration(auth, domain):
+def domain_dns_conf(auth, domain):
     """
     Generate DNS configuration for a domain
 
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -239,7 +239,7 @@ def domain_generate_dns_configuration(auth, domain):
     except Exception:
         pass
     else:
-        result += "@ 1400 IN A {ip6}\n* 900 IN AAAA {ip6}\n".format(ip6=ip6)
+        result += "@ 1400 IN AAAA {ip6}\n* 900 IN AAAA {ip6}\n".format(ip6=ip6)
 
     result += "\n_xmpp-client._tcp 14400 IN SRV 0 5 5222 {domain}.\n_xmpp-server._tcp 14400 IN SRV 0 5 5269 {domain}.\n".format(domain=domain)
 
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -230,7 +230,7 @@ def domain_generate_dns_configuration(auth, domain):
 
     ip4 = urlopen("http://ip.yunohost.org").read().strip()
 
-    result = "@ 1400 IN A {ip4}\n* 900 IN A {ip4}\n".format(ip4=ip4)
+    result = "@ 1400 IN A {ip4}\n* 1400 IN A {ip4}\n".format(ip4=ip4)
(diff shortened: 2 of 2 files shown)


data/actionsmap/yunohost.yml
```diff
@@ -291,6 +291,17 @@ domain:
                     extra:
                         pattern: *pattern_domain
 
+        ### domain_generate_dns_config()
+        generate_dns_configuration:
+            action_help: Generate DNS configuration for a domain
+            api: GET /domains/<domain>/dns_config
```
opi: Maybe it's just a personnal taste, but I would prefer a dash instead of an underscore in url. or maybe just `/domain/<domain>/dns`.

src/yunohost/domain.py
```diff
@@ -218,3 +218,36 @@ def domain_remove(auth, domain, force=False):
...
+    result = "@ 1400 IN A {ip4}\n* 900 IN A {ip4}\n".format(ip4=ip4)
+
+    ip6 = None
+
+    try:
+        ip6 = urlopen("http://ip6.yunohost.org").read().strip()
+    except Exception:
+        pass
+    else:
+        result += "@ 1400 IN A {ip6}\n* 900 IN AAAA {ip6}\n".format(ip6=ip6)
```
opi: `IN A {ip6}` ?

src/yunohost/domain.py
```diff
@@ -218,3 +218,36 @@ def domain_remove(auth, domain, force=False):
...
+
+    result += "\n_xmpp-client._tcp 14400 IN SRV 0 5 5222 {domain}.\n_xmpp-server._tcp 14400 IN SRV 0 5 5269 {domain}.\n".format(domain=domain)
```
opi: Why didn't you set the `muc`, `pubsub`, `vjud` subdomain, required for some XMPP features.

Psycojoker: I've based my work on https://github.com/labriqueinternet/configuration_scripts/blob/master/neutrinet.sh#L278 which doesn't included them. Don't ask my why, I'm really not into XMPP.

I'm adding them.

src/yunohost/domain.py
Psycojoker: Fixed, thx <3

data/actionsmap/yunohost.yml
Psycojoker: done

src/yunohost/domain.py
Psycojoker: and if I'm reading DNS correctly, the wildcard declaration covers those cases already. But I'm not really an expert in DNS.

src/yunohost/domain.py
```diff
@@ -218,3 +218,38 @@ def domain_remove(auth, domain, force=False):
...
+    except Exception:
+        pass
+    else:
+        result += "@ 1400 IN AAAA {ip6}\n* 1400 IN AAAA {ip6}\n".format(ip6=ip6)
+
+    result += "\n_xmpp-client._tcp 14400 IN SRV 0 5 5222 {domain}.\n_xmpp-server._tcp 14400 IN SRV 0 5 5269 {domain}.\n".format(domain=domain)
+
+    result += "muc 1800 IN CNAME @\npubsub 1800 IN CNAME @\nvjud 1800 IN CNAME @\n\n"
+
+    result += "@ 1400 IN MX 10 {domain}.tld.\n".format(domain=domain)
```
M5oul: Is `.tld` necessary here?

Psycojoker: arg no >.<

Psycojoker:
Example of generated DNS config for domain <ocde>yolo.swag</code> (ip has been replace for obvious reason):

```
@ 1400 IN A 123.123.123.123
* 1400 IN A 123.123.123.123

_xmpp-client._tcp 14400 IN SRV 0 5 5222 yolo.swag.
_xmpp-server._tcp 14400 IN SRV 0 5 5269 yolo.swag.
muc 1800 IN CNAME @
pubsub 1800 IN CNAME @
vjud 1800 IN CNAME @

@ 1400 IN MX 10 yolo.swag.
@ 1400 IN TXT "v=spf1 a mx ip4:123.123.123.123 -all"
```

data/actionsmap/yunohost.yml
```diff
@@ -291,6 +291,17 @@ domain:
                     extra:
                         pattern: *pattern_domain
 
+        ### domain_generate_dns_config()
+        generate_dns_configuration:
```
jeromelebleu: As for the API route, it's quite a long and not really nice cli command - e.g. it will be `yunohost domain generate_dns_configuration`... Note that it's possible to use -, so what about *dns-conf* or something like that?

Psycojoker: Changed :)

data/actionsmap/yunohost.yml
```diff
@@ -291,6 +291,17 @@ domain:
                     extra:
                         pattern: *pattern_domain
 
+        ### dns_conf()
+        dns-conf:
```
opi: Actually we use a `domain_xxx` pattern, so why don't call this function `domain_dns` ? It'll be consistent with the API route too (`GET /domain/<domain>/dns`).
Sorry I didn't catch it earlier.

opi: Forget my previous comment; This comment `### dns_conf()` mislead me.

data/actionsmap/yunohost.yml
```diff
@@ -291,6 +291,17 @@ domain:
...
+            action_help: Generate DNS configuration for a domain
+            api: GET /domains/<domain>/dns
+            configuration:
+                lock: false
+                authenticate: all
```
jeromelebleu: One more thing - which is not documented too ;) - regarding `authenticate: all`, it will imply the admin password to be asked in the cli too - which is maybe too much for this action... You can omit it and remove the first `auth` argument in the function - which is passed only if `authenticate` is `all`. It will result in a required authentication just for the api - which is the default: https://github.com/YunoHost/yunohost/blob/unstable/data/actionsmap/yunohost.yml#L38. I know, it's quite dirty, that's why a rewriting is needed!

Psycojoker: done

Psycojoker:
I've just add to configure the DNS for a member who is at OVH. OVH being OVH, they have decided to do things their bucause fuck the world. Here is how it should look:
```dns
$TTL 86400
@	IN SOA dns14.ovh.net. tech.ovh.net. (2016012901 86400 3600 3600000 86400)
                         IN NS     ns14.ovh.net.
                         IN NS     dns14.ovh.net.
                         IN MX 5   $DOMAIN.
                         IN A      $IPV4
                         IN AAAA   $IPV6
                         IN TXT    "v=spf1 a mx ip4:$IPV4 ip6:$IPV6 -all"
*                        IN A      $IPV4
*                        IN AAAA   $IPV6
_xmpp-client._tcp        IN SRV    0 5 5222 $DOMAIN.
_xmpp-server._tcp        IN SRV    0 5 5269 $DOMAIN.
```

I don't know how to handle this wonderful special case. Another command? An option? Anyway, is it conceivable to have a template engine at disposition (like jinja2)? Because concatening strings is driving me a bit crazy.

opi:
what about adding a *registrar* agrument ? 
Something like `yunohost domain dns --domain=yolo.com --registrar=ovh`
It sucks, but registrar sucks as well

zamentur:
We should add the DKIM TXT instruction too.

The instruction is listed in:
/etc/dkim/DOMAIN.mail.txt

Psycojoker:
I wanted to had that too, but not one was able to tell how it works at the time I've written this command.

I'm adding this to my todolist.

Psycojoker:
It's possible to create if tags with .format, this will allow wayyyyy less horrible code http://makina-corpus.com/blog/metier/the-worlds-simplest-python-template-engine?utm_source=twitterfeed&utm_medium=twitter

M5oul:
Thanks for that beautiful command. I get this result:

```bash
# yunohost domain dns-conf moul.re
@ 1400 IN A 78.227.107.45
* 1400 IN A 78.227.107.45
@ 1400 IN AAAA 2a01:e34:ee36:b2d0:4a:9ff:fe82:8f7d
* 1400 IN AAAA 2a01:e34:ee36:b2d0:4a:9ff:fe82:8f7d

_xmpp-client._tcp 14400 IN SRV 0 5 5222 moul.re.
_xmpp-server._tcp 14400 IN SRV 0 5 5269 moul.re.
muc 1800 IN CNAME @
pubsub 1800 IN CNAME @
vjud 1800 IN CNAME @

@ 1400 IN MX 10 moul.re.
@ 1400 IN TXT "v=spf1 a mx ip4:78.227.107.45 ip6:2a01:e34:ee36:b2d0:4a:9ff:fe82:8f7d -all"
```

Is wildcards entries are relevant? I don't know in which use case they are used. May be separate them from @ entries as in [documentation I have updated](https://yunohost.org/#/dns_config_en).

M5oul:
The result of this command use different [TTL](https://en.wikipedia.org/wiki/Time_to_live#DNS_records). May be, we could use a default value as in [updated documentation](https://yunohost.org/#/dns_config_en) and if we want to modify this value we could use an option `--ttl`.

opi:
`--ttl` option is relevant, and should be easy to implement.

Psycojoker:
Yes, wildcards are totally relevant, adding them cost nothing and allows the user to add any subdomain he wants as installing one app per subdomain is a popular way of doing things (and actually a way more frequent modern way of doing this than installing an app per path).

I have no idea on which default value TTL should have tbh, for me it's more or less random or "the last one I've set". Okay for the option.

opi:
can't agree more with @Psycojoker about adding wildcards

M5oul:
Wow, wildcard entries are very usefull! Thanks for explanation.

M5oul:
I would like to implement the `--ttl` option. I will open a new pull request on `unstable` branch. How could I test my implementation? Where are located this scripts on the system?

Psycojoker:
use ynh-dev with the docker settings, you'll find this command in /yunohost/yunohost and you can use "./ynh-dev deploy" to test your modifications https://yunohost.org/#/dev_fr

M5oul:
Why can't I test modifications on a deployed testing YunoHost? Do I really need a dev environnement with `ynh-dev`? What's different? If so, I will try again to deploy `ynh-dev` with Docker.

Psycojoker:
Oh, you can totally do that, the file should be there for you <code>/usr/lib/moulinette/yunohost/domain.py</code>

M5oul:
Wow! thanks :smile_cat:

opi:
You'll need to edit actionmap file too, located at `/usr/share/moulinette/actionsmap/yunohost.yml` to add the `ttl` argument.
Then either run the API without cache (`yunohost-api --no-cache`) or manually remove the actionmap cache located at `/var/cache/moulinette/actionsmap/yunohost.pkl`.

Psycojoker:
oh, yeah, don't forget to do that (remove the action map), otherwise your actions won't take effect https://github.com/YunoHost/ynh-dev/blob/master/ynh-dev#L152

PR #121, opened 2016-02-09
Don't emit Nginx version

infertux:
This hides the Nginx version in error pages to make it harder for script kiddies to target vulnerabilities specific to a version.

Before:

```html
<html>
<head><title>302 Found</title></head>
<body bgcolor="white">
<center><h1>302 Found</h1></center>
<hr><center>nginx/1.6.2</center>
</body>
</html>
```

After:

```html
<html>
<head><title>302 Found</title></head>
<body bgcolor="white">
<center><h1>302 Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

I also did some whitespace cleanup.

This is my first PR to get to know the _yunohost_ codebase. I used https://yunohost.org/#/docker to test my changes and it worked relatively well but I'm wondering if you have any kind of automated testsuite or build system? Thank you for this great project!

### DIFF
--- a/data/hooks/conf_regen/15-nginx
+++ b/data/hooks/conf_regen/15-nginx
@@ -24,6 +24,7 @@ cd /usr/share/yunohost/templates/nginx
 
 # Copy plain single configuration files
 files="ssowat.conf
+global.conf
 yunohost_admin.conf
 yunohost_admin.conf.inc
 yunohost_api.conf.inc
--- a/data/templates/nginx/global.conf
+++ b/data/templates/nginx/global.conf
@@ -1,0 +1,1 @@
+server_tokens off;
--- a/data/templates/nginx/server.conf.sed
+++ b/data/templates/nginx/server.conf.sed
@@ -2,7 +2,7 @@ server {
     listen 80;
     listen [::]:80;
     server_name {{ domain }};
-    
+
     access_by_lua_file /usr/share/ssowat/access.lua;
 
     include conf.d/{{ domain }}.d/*.conf;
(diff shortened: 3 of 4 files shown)


opi:
Looks good to me.
( and no, there's actually no build system or automated testsuite. Some happy volunteers are working hard to make this happen )

julienmalik:
Looks good to me too, except for the name of the new file.
Let's try to avoid common/utils/... and other meaningless names !

Otherwise :+1: 
we did similar modifications for other services (postfix, ...)

opi:
> except for the name of the new file.

what about `server.conf` ? ( btw, I don't understand the purpose of the `server.conf.sed` file in `data/templates/nginx` )

infertux:
Thanks for the reply, that's great to hear about the build system. 

Regarding the file name, Nginx's default config file - `/etc/nginx/nginx.conf` - contains `include /etc/nginx/conf.d/*.conf;` so I figured we could have `/etc/nginx/conf.d/common.conf` to override any default config that is _common_, i.e. shared through all vhosts/sites.

Though I'm happy to rename it if you have a better idea @julienmalik?

julienmalik:
can we merge as-is ? and eventually file an issue if you also feel the name should change ? I currently have no better alternative to propose (global.conf maybe?).

julienmalik:
Also I'm surprised I discovered we don't override the default /etc/nginx/nginx.conf in regenconf. That would be a simple place to put it.

infertux:
I prefer `global.conf` - I renamed it. Feel free to merge if you're happy with it.

julienmalik:
@opi @jeromelebleu  : let's merge ?

jeromelebleu:
Yes, and I agree to see later - as needed - to rename this file or move it somewhere else...

PR #122, opened 2016-02-14
[fix] Error "seems to be not correctly installed" on app install

zamentur:
It fix the #206 error see https://dev.yunohost.org/issues/206#note-3

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -475,7 +475,10 @@ def app_install(auth, app, label=None, args=None):
             hook_add(app_id, app_tmp_folder +'/hooks/'+ file)
 
     now = int(time.time())
-    app_setting(app_id, 'id', app_id)
+
+    # We can't use app_setting cause it creates an 'app_not_correctly_installed'
+    # message. See https://dev.yunohost.org/issues/206#note-3
+    _set_app_settings(app, {'id':app_id})
     # TODO: Move install_time away from app_setting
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -148,8 +148,12 @@ def app_list(offset=None, limit=None, filter=None, raw=False):
 
     for applist in applists:
         if '.json' in applist:
+            list_name=applist.split('.json', 1)[0]
             with open(repo_path +'/'+ applist) as json_list:
-                app_dict.update(json.loads(str(json_list.read())))
+                list_content=json.loads(str(json_list.read()))
+                for k in list_content.keys():
+                    list_content[k]['store']=list_name
```
jeromelebleu: What's the goal of this key? It seems not to be related to the other part, isn't it?

zamentur: Exact, it was added because I push on my unstable branches. Github
automatically adds commit to the last PR not merged...

Those changes give the possibility to know from which list comes an
application (official, community, internetcube). It is important because
when a user add a list, he can't know if the app is official...

jeromelebleu: Could you please remove this commit from your PR? Except that, I think it could be merged.

jeromelebleu:
Done with https://github.com/YunoHost/yunohost/commit/b27aa331921f4b6d58e257a53121b37fe8350d6d. Thanks for the report and the idea!

PR #130, opened 2016-03-09
Diagnostic

opi:
Current output is documented here https://dev.yunohost.org/issues/39#note-3 (and will be updated according to futur code changes)

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -32,11 +32,20 @@ import requests
 import json
 import errno
 import logging
+from collections import OrderedDict
+
 import apt
 import apt.progress
 
-from moulinette.core import MoulinetteError
+from moulinette.core import MoulinetteError, init_authenticator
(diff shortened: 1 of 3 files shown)


src/yunohost/tools.py
```diff
@@ -418,5 +413,90 @@ def tools_upgrade(auth, ignore_apps=False, ignore_packages=False):
...
+        for application in applications:
+            if application['installed']:
+                diagnosis['applications'][application['id']] = application['label'] if application['label'] else application['name']
+
+    # Private data
+    if private:
+        diagnosis['private'] = OrderedDict()
+        # Public IP
+        diagnosis['private']['public_ip'] = {}
+        try: diagnosis['private']['public_ip']['IPv4'] = get_public_ip(4)
```
Psycojoker: split line here ("try:\n            diagnosis...") and same for following lines.

PR #131, opened 2016-03-09
Display a more useful error message to a user with how to perform the recommanded action.

Psycojoker:
In this kind of situation where things are obvious for us by maybe not for the
user we should always display how to the things we recommand to do or even put
a link to the documentation if needed. Those simple actions will **greatlly**
increase the ease of usage of YunoHost and make a lot of users more happy and
reduce support workload.


jeromelebleu:
Done with https://github.com/YunoHost/yunohost/commit/3909aaddf1bab9dc05ef8c4f47ae5e04eaea2395. Thanks!

PR #132, opened 2016-03-10
[enh] More precise for regenconf --force when a file has changed sinc…

julienmalik:
…e last generation

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -126,7 +126,7 @@
     "service_no_log" : "No log to display for service '{:s}'",
     "service_cmd_exec_failed" : "Unable to execute command '{:s}'",
     "services_configured": "Configuration successfully generated",
-    "service_configuration_conflict": "The file {file:s} has been changed since its last generation. Please apply the modifications manually or use the option --force (it will erase all the modifications previously done to the file).",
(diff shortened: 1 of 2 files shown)


julienmalik:
do not merge yet.

actually the message is not that helpful, since we do not provide any way to "apply modification manually" : users can't apply their modifications unless they first run with --force. and then they need the old diff around, or they need to remember what they did to their file...

julienmalik:
see also https://dev.yunohost.org/issues/172

julienmalik:
IMHO we should be ultra-verbose in that case and better drive users into how to reapply their custom modifications over the new yunohost-maintainer version.

regenconf --force should :
- check if it needs to save the  user-modified file (already done ?)
- save the current user-modified file (already done I think in /home/yunohost.backup/conffiles)
- report the location of the saved user-modified file explicitely in the logs
- show diff between the user-modified file and the new "maintainer version."

julienmalik:
follow-up in https://dev.yunohost.org/issues/172

jeromelebleu:
Not valid anymore due to https://github.com/YunoHost/yunohost/pull/138.

PR #134, opened 2016-03-15
[enh] Add --installed and --with-backup filter for app_list method. #227

opi:


src/yunohost/app.py
```diff
@@ -183,16 +186,24 @@ def app_list(offset=None, limit=None, filter=None, raw=False):
...
                 if (filter and ((filter in app_id) or (filter in app_info_dict['manifest']['name']))) or not filter:
-                    installed = _is_installed(app_id)
+                    app_installed = _is_installed(app_id)
+
+                    # Only installed apps filter
+                    if installed and not app_installed:
+                        continue
+
+                    # Only backupable apps filter
+                    if with_backup and not os.path.isfile(apps_setting_path + app_id + '/scripts/backup'):
```
jeromelebleu: Does it make sense to also check for a restore script? Otherwise, the backup is not really useful except for acknowledges users, isn't it?

opi: I hope that every app which implements a backup script also have a restore script, but who knows !

PR #135, opened 2016-03-19
[enh] add example to use '-a' option with 'yunohost app install' comm…

M5oul:
…and.

I can't use a dot to split in two sentences.

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -436,7 +436,7 @@ app:
                     help: Custom name for the app
                 -a:
                     full: --args
-                    help: Serialize arguments for app installation
+                    help: Serialized arguments for app script (i.e. "domain=domain.tld&url_path=/path")
 
         ### app_remove() TODO: Write help
         remove:


data/actionsmap/yunohost.yml
```diff
@@ -436,7 +436,7 @@ app:
                     help: Custom name for the app
                 -a:
                     full: --args
-                    help: Serialize arguments for app installation
+                    help: Serialized arguments for app script (i.e. "domain=domain.tld&url_path=/path")
```
opi: why use "url_path" instead of the commonly used "path" option ? ;)

opi:
thx

data/actionsmap/yunohost.yml
M5oul: I prefer use [`url_path`](https://github.com/M5oul/sonerezh_ynh/blob/master/scripts/install#L6) and [`src_path`](https://github.com/M5oul/sonerezh_ynh/blob/master/scripts/install#L34) variables to be more specific and avoid confusions between `path` and `final_path` variables in the install script.

opi: But "domain" and "path" are kind of reserved variable names, and are automatically saved on app_install ( https://github.com/yunohost/yunohost/blob/unstable/src%2Fyunohost%2Fapp.py#L876 ).
Using `url_path` could lead to some variable name mixin as in https://github.com/M5oul/sonerezh_ynh/blob/master/scripts%2Fupgrade#L9
But your app, your choice ;)

M5oul: Yep, you are right, we should call this variable `path`.
Fixed with https://github.com/YunoHost/yunohost/commit/1dd1ae483f1308a836af3cf4a8c9403ba26a591e.

opi: thx for that. Better cli help description is always helpfull

PR #136, opened 2016-03-26
Fix dyndns update

jeromelebleu:


opi:
shall we try to hotfix the stable release ?

jeromelebleu:
Sure! Is it okay for you? Or is there anything else to fix regarding DynDNS?

opi:
imho, we need to (hot)fix what's really broken, and focus on releasing 2.4 stable release as soon as possible

PR #139, opened 2016-04-18
[fix] Prevent non-updated multiinstance apps to be installed if no re…

opi:
…quirements for yunhost


jeromelebleu:
Done with https://github.com/YunoHost/yunohost/commit/aca251e98173d042e254f09d53c985a5fb39fb18 which checks for a valid requirements too, as you pointed out. Thanks! :)

PR #140, opened 2016-04-19
[enh] Add a yunopaste script to paste data to YunoHost Haste server

jeromelebleu:
Thanks to @opi, we now have our Haste server back. As also suggested by @opi, we could easily provide a script for users to paste commands output - and especially YunoHost ones - to it. This script attempts to do that.

### DIFF
--- a/bin/yunopaste
+++ b/bin/yunopaste
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+set -e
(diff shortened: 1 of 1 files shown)


opi:
tested & approved

jeromelebleu:
\o/ thanks!

PR #144, opened 2016-05-02
Alter SSOwat conf output

gkrnours:
This patch will align the SSOwat conf produced by yunohost by the one asked by a coming patch to SSOwat.

The idea is to put applications details not in the users dict but in a separate dict. The user dict now only have a list of app id. This allow to have more details than name:url and remove duplicate information in the conf in case of more than one user.

That way of storing data reduce the cost of storing more information on app at the price of indirect access to the app details in SSOwat.

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -965,7 +965,11 @@ def app_ssowatconf(auth):
 
     users = {}
     for username in user_list(auth)['users'].keys():
-        users[username] = app_map(user=username)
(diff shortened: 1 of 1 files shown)


opi:
Related to https://github.com/Kloadut/SSOwat/pull/58 and https://dev.yunohost.org/issues/302#change-919

alexAubin:
[Semi-automatic message] 

PR #145, opened 2016-05-04
[fix] Can install app on domain root even if another app is installed…

opi:
… in a sub folder.


opi:
Before that, on an testing instance:
```
 yunohost app install a_random_app
Choose a domain for your app: testing.local
Choose a path for your app (default: /app): /app
```
then when playing with `app checkurl`:
```
yunohost app checkurl testing.local/ -a roundcube
```
return no errors. 

I guess we wan't to prevent a user to install an app on domain root if there are already installed apps  on sub folders.

PR #146, opened 2016-05-10
Migration script from 2.2 to 2.4 #308

opi:


lib/yunohost/tools.py
```diff
@@ -469,3 +474,44 @@ def tools_upgrade(auth, ignore_apps=False, ignore_packages=False):
     if is_api:
         from yunohost.service import service_log
         return { "log": service_log('yunohost-api', number="100").values()[0] }
+
+
+def tools_upgrade_v24(auth):
+    """
+    YunoHost upgrade to new Yunohost version (on jessie)
+    """
+    import platform
```
Psycojoker: Imports would be better at the beginning of the file.

opi: even for a one-shot command, on a 'old-stable' branch  ? ;)

Psycojoker: It's better yes. The cases where it's kinda ok to put import somewhere else is when you do that in a "except" block where the import is guaranteed not to fail.

lib/yunohost/tools.py
```diff
@@ -469,3 +474,44 @@ def tools_upgrade(auth, ignore_apps=False, ignore_packages=False):
...
+
+    # Get Debian major version
+    debian_major_version = platform.linux_distribution()[1].split('.')[0]
+    if debian_major_version is not '8':
+        msignals.display(m18n.n('upgrade_24_not_jessie'), 'error')
+        return
+
+    # Upgrade with current sources
+    os.system('apt-get update')
+    os.system('yes "q" | DEBIAN_FRONTEND=noninteractive DEBIAN_PRIORITY=critical apt-get -y --force-yes -qq -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" upgrade')
```
Psycojoker: os.system won't raised an exception if the program fails (add to patch that for prosopopee today).

A way to ensure you don't continued in a broken state woudl be to do a 

```python
assert os.system(command) == 0, "The command '%s' failed, abort" % command
```

To check that the return code of the command is 0.

lib/yunohost/tools.py
```diff
@@ -469,3 +474,44 @@ def tools_upgrade(auth, ignore_apps=False, ignore_packages=False):
...
+    os.system('yes "q" | DEBIAN_FRONTEND=noninteractive DEBIAN_PRIORITY=critical apt-get -y --force-yes -qq -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" dist-upgrade')
+    os.system('yes "q" | DEBIAN_FRONTEND=noninteractive DEBIAN_PRIORITY=critical apt-get remove -y --force-yes amavisd-new')
+    os.system('yes "q" | DEBIAN_FRONTEND=noninteractive DEBIAN_PRIORITY=critical apt-get -y --force-yes autoremove')
+    os.system('yunohost service regen-conf -f')
+
+    msignals.display(m18n.n('system_upgraded'), 'success')
+
+    # Prepare systemctl
+    with open('/etc/cron.d/yunohost-regenconf', 'w+') as f:
+        f.write('00 * * * * root PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin systemctl start yunohost-api && rm -f /etc/cron.d/yunohost-regenconf\n' )
```
Psycojoker: Shouldn't it be `* * * * *` or am I missing something?

lib/yunohost/tools.py
```diff
@@ -410,11 +413,25 @@ def tools_upgrade(auth, ignore_apps=False, ignore_packages=False):
...
 
+    # Special case for v2.2 to v2.4 upgrade
+    if not v24 and not is_api:
+        try:
+            i = msignals.prompt(m18n.n('upgrade_24_confirm',
+                                       answers='y/N'))
+        except NotImplemented:
+            pass
+        else:
+            if i == 'y' or i == 'Y':
```
Psycojoker: You can do a `i in ('y', 'Y')' or even a `i.lower() == 'y'` (if you are sure that `i` will have a `.lower` method).

PR #151, opened 2016-05-16
[fix] Update argument with empty value adding for OrderedDict usage

jeromelebleu:


src/yunohost/app.py
```diff
@@ -1535,7 +1535,7 @@ def _parse_args_from_manifest(manifest, action, args={}, auth=None):
                 raise MoulinetteError(errno.EINVAL,
                     m18n.n('app_argument_required', name=arg_name))
             elif not arg_value:
```
Psycojoker: `not arg_value` is *nearly* equivalent to `(arg_value is None or arg_value == '')`, I don't know if it's something you want.

(The other situation in which the 2 differs is when arg_value is an empty datastructure like dict/list or zero or a custom object modified to behave like that)

jeromelebleu: This is the desired behavior - but thanks for the review.

PR #153, opened 2016-05-29
Fix minor grammar errors in English translation

Bugsbane:
Fixed a number of minor English spelling and grammar errors in the messages (for example the word "list" implies plural, so no need to double pluralize by saying "apps list").

For what it's worth, I'm a native English speaker who has worked as a copywriter.


opi:
Thanks a lot, looks good to me. I'll let @jeromelebleu do the merge, as he knows the translation system better than me (translate.yunohost.org).

@Bugsbane do you wan't to review other english messages (YunoHost admin, user panel, ...) ? Help from native english speakier is always welcome, especially in a french driven project ;)

jeromelebleu:
Thanks a lot! I would like to ask you the same question as @opi, if we can abuse your contribution... :)

By the way, I would like to take this opportunity to remove *successfully* from each message, do you have any objection?

locales/en.json
```diff
@@ -219,7 +219,7 @@
     "user_info_failed" : "Unable to retrieve user information",
     "user_home_creation_failed" : "Unable to create user home folder",
 
-    "package_unexpected_error" : "An unexpected error occured with the package '{pkgname}'",
+    "package_unexpected_error" : "An unexpected error occured in the package '{pkgname}'",
```
jeromelebleu: This message is printed when *pkgname* - which is generally a Debian package - cannot be installed / upgraded / removed and unexpectedly. Is *in the package* mean that? I really don't want to question your English but just learn it! :)

Bugsbane: Ok. Good to know @jeromelebleu. I may know English better, but you guys understand the technical reality the error message represents better than I in many cases. In this case it sounds like both "in" and "with" are roughly equally (in)correct. The really accurate way of stating should actually be more along the lines of:
"An unexpected error occurred processing the package '{pkgname}'"

jeromelebleu:
Done with https://github.com/YunoHost/yunohost/commit/502ae7577360430e2369124586b9450e2568bcab. Thanks! :)

Bugsbane:
@opl / @jeromelebleu I'm happy to help where I can. I don't have a ton of time, so I don't think I could go through the entire wiki / docs etc, but I went over the YunoHost admin and about 1/2 of what's just listed as "Yunohost" at translate.yunohost.org. I'll try to go through the rest when I get some time..

jeromelebleu:
No problem, just do what you want / can! :) Thanks anyway!

But in the future, I think it would be better to use as most as possible [Weblate](https://translate.yunohost.org) for English translations... I accidentally *broke* it since someone else has also changed some English strings there... ^^

Bugsbane:
> I accidentally broke it since you also changed some English strings there...

Sorry! Like I said. I may know English, but I'm no coder. No breakage intended...

PR #156, opened 2016-06-03
Some extra filters to reply-headers for outbound mail

retmarut:


jeromelebleu:
Thanks for that!

What's the goal of putting those filters there? From what I know and understand reading the [documentation](http://www.postfix.org/postconf.5.html#smtp_reply_filter), it *is a last-resort tool to work around server replies that break interoperability with the Postfix SMTP client* so it concerns more inbound email, isn't it?

Also, please note that if you move the file away from the `plain` folder, [this](https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/conf_regen/19-postfix) script needs to be modified too to replace and copy this new configuration file.

julienmalik:
Agreed with @jeromelebleu 

This should be in http://www.postfix.org/header_checks.5.html instead
Especially I think this is for smtp_header_checks ("FILTERS WHILE DELIVERING MAIL") : http://www.postfix.org/postconf.5.html#smtp_header_checks

Please re-submit

M5oul:
@retmarut up?

frju365:
seems to be same as the https://github.com/YunoHost/yunohost/pull/374

PR #157, opened 2016-06-09
[ref] Invert no-stats option to with-stats in monitor_enable

jeromelebleu:
Since monitoring statistics was not finished - and by the way not usable, it inverts the behavior of monitor_enable to disable it by default. See [#395](https://dev.yunohost.org/issues/395) for more details.


Psycojoker:
Looks good from a python point of view.

PR #159, opened 2016-06-11
Clean app fetchlist

Psycojoker:
Small code cleaning, use buildin python tools and avoid using exceptions when not necessary.


src/yunohost/app.py
```diff
@@ -89,27 +90,24 @@ def app_fetchlist(url=None, name=None):
...
+                              m18n.n('custom_appslist_name_required'))
 
-    list_file = '%s/%s.json' % (repo_path, name)
-    if os.system('wget "%s" -O "%s.tmp"' % (url, list_file)) != 0:
-        os.remove('%s.tmp' % list_file)
+    try:
+        urlretrieve(url, '%s/%s.json' % (repo_path, name))
+    except Exception as e:
+        # I don't know how to put e into the MoulinetteError stuff
+        print e
```
jeromelebleu: Just use the `logger` object instead to log the exception traceback - and error message too, i.e. `logger.debug('cannot fetch app list', exc_info=1)`.
Or, if you want to display an English error message to the user - who will probably not understand what it is about - you can update the `appslist_retrieve_error` key [here](https://github.com/YunoHost/yunohost/blob/unstable/locales/en.json#L14), append for example `: {error}` and pass this argument value with `m18n.n('appslist_retrieve_error', error=str(e))`.

Psycojoker: I've done the later one.

src/yunohost/app.py
```diff
@@ -100,14 +101,13 @@ def app_fetchlist(url=None, name=None):
...
         raise MoulinetteError(errno.EBADR, m18n.n('appslist_retrieve_error'))
 
-    # Rename fetched temp list
-    os.rename('%s.tmp' % list_file, list_file)
```
zamentur: Are you sure this line is not needed ?

Psycojoker: Yes because I'm not creating any .tmp file since I'm using urlretreive directly.

src/yunohost/app.py
```diff
@@ -89,27 +90,22 @@ def app_fetchlist(url=None, name=None):
...
+    try:
+        urlretrieve(url, '%s/%s.json' % (repo_path, name))
+    except Exception as e:
+        raise MoulinetteError(errno.EBADR, m18n.n('appslist_retrieve_error'), error=str(e))
 
-    os.system("touch /etc/cron.d/yunohost-applist-%s" % name)
-    os.system("echo '00 00 * * * root yunohost app fetchlist -u %s -n %s > /dev/null 2>&1' >/etc/cron.d/yunohost-applist-%s" % (url, name, name))
+    open("/etc/cron.d/yunohost-applist-%s" % name, "w").write('00 00 * * * root yunohost app fetchlist -u %s -n %s > /dev/null 2>&1\n' % (url, name))
```
julienmalik: where (or how) is the stream closed ?

Psycojoker: What are you calling a stream here?

If it's the open file it's closed automatically after this line because python is based on reference couting and will automatically garbage the file item once it's not referenced anymore.

julienmalik: you don't use : 
```
with open("filename", "w") as f :
  f.write( stuff )
```
while it is generally good practice.

Psycojoker: Yes, because I'm lazy and that this is not needed here.

julienmalik: This might be needed later when someone adds code after this line. Also, as we usually like to copy-paste code, this might spread elsewhere where it would have bad consequences.

Psycojoker: Adding lines after this line won't change the behavior, the file object is not referenced anywhere so it's automatically destroyed after this line.

Psycojoker: And worst case it's destroyed at the end of the function if it's referenced somewhere.

Psycojoker:
Hello,

I'm re-open this one. I'm consider it to be a minor decision https://github.com/YunoHost/yunohost-project-organization#d%C3%A9cision-mineure

Psycojoker:
I've rebased the branch.

src/yunohost/app.py
```diff
@@ -89,27 +90,22 @@ def app_fetchlist(url=None, name=None):
...
-        os.remove('%s.tmp' % list_file)
-        raise MoulinetteError(errno.EBADR, m18n.n('appslist_retrieve_error'))
+    elif name is None:
+        raise MoulinetteError(errno.EINVAL,
+                              m18n.n('custom_appslist_name_required'))
 
-    # Rename fetched temp list
-    os.rename('%s.tmp' % list_file, list_file)
+    try:
+        urlretrieve(url, '%s/%s.json' % (repo_path, name))
```
Psycojoker: I actually think that we should introduce a time out mechanism but I don't know if this is related to this PR.

Psycojoker:
It's been one week but since I don't consider this PR as prio I won't relaunch it until we finish other big features.

alexAubin:
Bumping for glory

Psycojoker:
Looks good (but since it's my PR I can't approve).

We could also go crazy on "handle every case where we could get an exception" for IO operation but that's already an improvement.

src/yunohost/app.py
```diff
@@ -87,27 +88,29 @@ def app_fetchlist(url=None, name=None):
...
-        if name is None:
-            raise MoulinetteError(errno.EINVAL,
-                                  m18n.n('custom_appslist_name_required'))
+    elif name is None:
+        raise MoulinetteError(errno.EINVAL,
+                              m18n.n('custom_appslist_name_required'))
 
+    # Download file
+    try:
+        applist = requests.get(url, timeout=30).text
```
opi: using `requests` instead of `wget` does not trigger 404 error nor no json file. Try `yunohost app fetchlist -u https://app.yunohost.org/404.json -n 404list` for example

alexAubin: Indeed, thanks for the catch !

PR #160, opened 2016-06-11
[enh] Refactor applist management

Psycojoker:
Hello,

Right now `yunohost app fetchlist` only fetch the official app list by default and there is not easy way to fetch all list at once. I've chosen to modify this behavior to fetch all list by default if no url and name are provided.

I've included retrocomabitle code since YunoHost doesn't store each list url.

This PR depends on https://github.com/YunoHost/yunohost/pull/159

You should probably read this diff to avoid having a mix with the other commits https://github.com/YunoHost/yunohost/compare/b787e3a...470636b00c

Cheers


src/yunohost/app.py
```diff
@@ -81,37 +82,69 @@ def app_listlists():
 
 def app_fetchlist(url=None, name=None):
     """
-    Fetch application list from app server
+    Fetch application list(s) from app server
 
     Keyword argument:
         name -- Name of the list (default yunohost)
         url -- URL of remote JSON list (default https://app.yunohost.org/official.json)
```
opi: Doc needs probably to be updated

src/yunohost/app.py
```diff
@@ -81,37 +82,69 @@ def app_listlists():
...
     """
     # Create app path if not exists
-    try: os.listdir(repo_path)
-    except OSError: os.makedirs(repo_path)
+    if not os.path.exists(repo_path):
+        os.makedirs(repo_path)
 
-    if url is None:
-        url = 'https://app.yunohost.org/official.json'
-        name = 'yunohost'
```
opi: As you removed default url/name, I wonder when this list is initially added. We may fix postinstall.

src/yunohost/app.py
```diff
@@ -81,37 +82,69 @@ def app_listlists():
...
+            base_filename = filename[:-len(".json")]
+            url_filename_path = os.path.join(repo_path, base_filename + ".url")
+
+            if os.path.exists(url_filename_path):
+                url = open(url_filename_path, "r").read().strip()
+            else:  # XXX backward compatible code, YunoHost never store the url of the list else where
+                cron_file_path = "/etc/cron.d/yunohost-applist-%s" % base_filename
+
+                if not os.path.exists(cron_file_path):
+                    logger.warning("neither %s nor %s exist, is the app list '%s' correctly installed on the system?" %\
```
jeromelebleu: This will display this English message to the user. You should add a new translation key with this message and make use of it with `m18n.n(...)`.

jeromelebleu:
You should also mention that you've chosen to store the repository url in a file named `name.url`! Also, I don't see where you write this file - except in *backward compatible code* - for a new repository...

src/yunohost/app.py
Psycojoker: Indeed, I have no idea on how the postinstall code works.

Psycojoker:
Where should I mention that?

Indeed, the "add new url" process is not finished, I've forgot to test that.

I'm doing all of this once/if the PR 159 is merged to work on a fresh rebase.

src/yunohost/app.py
```diff
@@ -100,14 +101,13 @@ def app_fetchlist(url=None, name=None):
             raise MoulinetteError(errno.EINVAL,
                                   m18n.n('custom_appslist_name_required'))
 
-    list_file = '%s/%s.json' % (repo_path, name)
-    if os.system('wget "%s" -O "%s.tmp"' % (url, list_file)) != 0:
-        os.remove('%s.tmp' % list_file)
+    try:
+        urlretrieve(url, '%s/%s.json' % (repo_path, name))
+    except Exception as e:
+        # I don't know how to put e into the MoulinetteError stuff
```
zamentur: I think it's something like:
```
except Exception as e:
        raise MoulinetteError(errno.EBADR, m18n.n('appslist_retrieve_error', error=e.strerror))
```

alexAubin:
I'll have a look after we merge #159

opi:
#159 is now merged

alexAubin:
After discussing on the chat, we need to thing what we want to do about that, especially about how we want to keep track of "registered" lists and how to migrate away from the old scheme. We should also stop that dirty trick with the cron job who re-write itself. Let's have a single default cron job in `/etc/cron.daily/` who simply calls `yunohost app fetchlist` with no arguments and fetch all lists that are tracked.

If we don't plan to merge #195 soon, I'd personally go for a small hard-coded migration for this specifically (something like, at the beginning of app_fetchlist, `if _using_old_scheme(): _migrate_to_new_fetchlist_scheme()`)

alexAubin:
So this is mostly done. Still need to test a few things though (such as the cron job working as expected, and the postinstall correctly adding the default list + cron job)(also need to test the migration in real life situation).

Major changes : 
- Basically reworked `app_listlists`, `app_fetchlist`, `app_removelist` from ~scratch ;
- Applists url are now stored in a `.url` in same directory as `.json` ;
- `app_fetchlist` fetches all registered lists ;
- Got rid of the cron jobs who re-wrote themselves, now there's a single `yunohost-fetch-applists` cronjob in `/etc/cron.daily` simply calling `yunohost app fetchlist` ;
- Migration should be transparent : calling  `app_listlists` or `app_fetchlist` will trigger, if needed, a migration to register properly the lists (with a `.url`), remove the legacy cron jobs, and install the new cronjob ;
- I added unit tests for basically all of this. Basically first time writing unit tests, hope I got them right ;

src/yunohost/app.py
```diff
@@ -140,7 +144,7 @@ def app_fetchlist(url=None, name=None):
             raise MoulinetteError(errno.EBADR,
                                   m18n.n('appslist_retrieve_error', error=str(e)))
 
-        if (applist_request.status_code != 200):
+        if applist_request.status_code != 200:
```
Psycojoker: Shouldn't we continue if there are more than one app list to get? To avoid blocking all next lists if one fails.

alexAubin: Yeah I was wondering about this. Shall we replace all the MoulinetteError inside the loop with a logger.error() + a continue ?

alexAubin:
So btw for people who didn't follow the discussion on the chat with Bram : instead of having multiple .url to remember the url of the list, we decided to have a single .json file stored in `/etc/yunohost/` to store info about the known applists (except the applist themselves, which are still stored in serapate jsons, at the same place as before).

alexAubin:
Still need to test if the postinstall change works, but in the meantime reviews are welcomed !

src/yunohost/app.py
Psycojoker: Yeah, that sound better, and also change the return code (if we know how to do that...).

alexAubin: Didn't look into changing the return code for now.

I moved to logger.error's, which made me wonder if/how we want to check stdout/stderr in the tests (e.g. does it contains the appropriate error string/key)

alexAubin: Follow-up : https://github.com/YunoHost/yunohost/pull/261

src/yunohost/app.py
```diff
@@ -65,66 +67,113 @@ def app_listlists():
...
-    elif name is None:
-        raise MoulinetteError(errno.EINVAL,
-                              m18n.n('custom_appslist_name_required'))
+    # Migrate applist system if needed
+    # XXX move that to a migration once they are finished
+    if _using_legacy_applist_system():
+        _migrate_applist_system()
+
+    # Read the list of applist...
+    applists = _read_applist_list()
```
opi: these last 3 lines (+comments) should be replaced by a call to app_listlists() to avoid code duplication.

src/yunohost/app.py
```diff
@@ -65,66 +67,113 @@ def app_listlists():
...
+
+        url = applists[name]["url"]
+
+        logger.debug("Attempting to fetch list %s at %s" % (name, url))
+
+        # Download file
+        try:
+            applist_request = requests.get(url, timeout=30)
+        except requests.exceptions.SSLError:
+            logger.error(m18n.n('appslist_retrieve_error',
```
opi: this error message does not mention which list is concerned

src/yunohost/app.py
```diff
@@ -65,66 +67,113 @@ def app_listlists():
...
+                                error="SSL connection error"))
+            continue
+        except Exception as e:
+            logger.error(m18n.n('appslist_retrieve_error',
```
opi: same issue with the error message

src/yunohost/app.py
```diff
@@ -65,66 +67,113 @@ def app_listlists():
...
+                                error=str(e)))
+            continue
+        if applist_request.status_code != 200:
+            logger.error(m18n.n('appslist_retrieve_error',
```
opi: same issue with error message

src/yunohost/app.py
```diff
@@ -65,66 +67,113 @@ def app_listlists():
...
-        f.write(applist)
+        # Validate app list format
+        # TODO / Possible improvement : better validation for app list (check
+        # that json fields actually look like an app list and not any json
+        # file)
+        applist = applist_request.text
+        try:
+            json.loads(applist)
+        except ValueError, e:
+            logger.error(m18n.n('appslist_retrieve_bad_format'))
```
opi: same issue with error message

src/yunohost/app.py
alexAubin: Well I wanted to try to keep `applist_list()` as much as a front-end function as possible. For instance to display the "lastUpdate" time in human readable format. The function to grab the list of applist should be `_read_applist_list()`. Maybe we could move the `_migrate_applist_system()` call in `_read_applist_list()` though ?

src/yunohost/app.py
alexAubin: Fixed in 69013c0

alexAubin:
I just noticed that the permissions of the cron job might not be good (no +x), might need to fix that

src/yunohost/app.py
alexAubin: In fact I realized why I didn't do that in the first place (putting a call to `_migrate_applist_system()` in `_read_applist_list()`) : that's because `_migrate_applist_system()` calls `_register_new_applist` which calls `_read_applist_list()`, so that would create an infinite recursive call...

src/yunohost/app.py
```diff
@@ -1659,6 +1721,129 @@ def _parse_app_instance_name(app_instance_name):
...
+def _register_new_applist(url, name):
+    """
+    Add a new applist to be fetched regularly.
+    Raise an exception if url or name conflicts with an existing list.
+    """
+
+    applist_list = _read_applist_list()
+
+    # Check if name conflicts with an existing list
+    if name in applist_list.keys():
```
alexAubin: Ah indeed... Well, atm you can simply remove the list with removelist, then redo a fetchlist with the new url.

src/yunohost/app.py
```diff
@@ -1659,6 +1721,129 @@ def _parse_app_instance_name(app_instance_name):
...
+    """
+    Read the json corresponding to the list of applists
+    """
+
+    # If file does not exists yet, return empty dict
+    if not os.path.exists(APPLISTS_JSON):
+        return {}
+
+    # Read file content
+    with open(APPLISTS_JSON, "r") as f:
```
alexAubin: I started a related PR here : https://github.com/YunoHost/yunohost/pull/274 
This is going to take quite some time so let's consider these PR independently, then we'll move to the common io helpers once they're here.

alexAubin:
Exceptions handling is really starting to feel like madness... But hopefully #274 should really help to have simpler code once it's there...

So, bumping again, hopefully for the last time :/

Psycojoker:
2017-03-20-013259_1678x762_scrot

Well...

alexAubin:
@Psycojoker : da0b003 should fix it :)

src/yunohost/app.py
```diff
@@ -65,66 +69,122 @@ def app_listlists():
...
+            continue
+        except Exception as e:
+            logger.error(m18n.n('appslist_retrieve_error',
+                                applist=name,
+                                error=str(e)))
+            continue
+        if applist_request.status_code != 200:
+            logger.error(m18n.n('appslist_retrieve_error',
+                                applist=name,
+                                error="404, not found"))
```
zamentur: It could be a 500 or 403.
By redirections are followed by request.get, so it couldn't be a 301.

alexAubin:
Comment from Bram on the chat (which I agree with :)) : 
> 01:19:17 Bram>     open("/etc/cron.d/yunohost-applist-%s" % name, "w").write('00 00 * * * root yunohost app fetchlist -u %s -n %s > /dev/null 2>&1\n' % (url, name))  
01:19:30 Bram> je me dit que ça serait peut être malin qu'on fasse un randint pour la minute                                                                                      
01:19:47 Bram> sinon là on dit basiquement à TOUS les yunohost "YOLO, DDoS nous pile à la même minute à minuit !"

src/yunohost/app.py
```diff
@@ -65,66 +69,122 @@ def app_listlists():
...
+    applists = _read_applist_list()
+
+    # Determine the list of applist to be fetched
+    applists_to_be_fetched = []
+
+    # If a url and and a name is given, try to register new list,
+    # the fetch only this list
+    if url is not None:
+        if name:
+            _register_new_applist(url, name)
```
alexAubin: I was trying to implement this, and realized that it might not be such a good idea : during the postinstall, we add the official.json list. If we implement this and someone does an offline postinstall, then the official.json won't be added, which we don't want to happen. Same goes if yunohost.org (or whatever list server) is down when trying to add a list which the user knows really exists.

Or we can have an option to disable the check... But I think people should (by default) be able to add a list independently from the internet connection and the availability of the remote server. If the list really doesn't exist, the renew cron job will anyway send an email to root when after attempting to fetch the list, and he'll end up removing the list.

alexAubin:
(Calling for merging approval again)

src/yunohost/app.py
zamentur: ok

zamentur:
So there is 2 positive review (alexAubin and me).
As it is a medium decision we need at least one positive review more.

alexAubin:
Agreed, we should definitely squash :P

PR #161, opened 2016-06-16
Harden TLS

phrag:
Strongest Ciphers / EECDH

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,21 @@
+yunohost (2.4.0.6) stable; urgency=low
+
+  [ Bugsbane ]
+  * [i18n] Fixed minor English grammar errors
+
+  [ Jérôme Lebleu ]
+  * [fix] Harden backup hooks with set options and use ynh_backup
+  * [fix] Set default value for YNH_APP_BACKUP_DIR in ynh_backup helper
+  * [fix] Raise proper MoulinetteError exception in hook_exec
+  * [fix] Use the classic way to create read-only bind mount in ynh_backup
+  * [fix] Escape arguments and env values in hook_exec (bugfix #377)
+
+  [ opi ]
+  * [enh] Comments will save us.
+  * [enh] Use 'source' instead of dot notation, more explicit.
+
+ -- Jérôme Lebleu <jerome@yunohost.org>  Mon, 30 May 2016 12:15:21 +0200
+
 yunohost (2.4.0.5) stable; urgency=low
 
   * [enh] Call iptables/ip6tables with --wait option (close #325)


M5oul:
I noticed that [Calomel extension](https://addons.mozilla.org/en-US/firefox/addon/calomel-ssl-validation) now say that SHA-1 is not secure with 4/15 as note.

M5oul:
I tried it.
It did not changed the SHA issue.

data/templates/nginx/plain/yunohost_admin.conf
```diff
@@ -21,9 +21,13 @@ server {
...
     ssl_prefer_server_ciphers on;
     ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-    ssl_ciphers ALL:!aNULL:!eNULL:!LOW:!EXP:!RC4:!3DES:+HIGH:+MEDIUM;
+    ssl_ciphers "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
+    ssl_ecdh_curve secp384r1;
+    ssl_stapling on;
+    ssl_stapling_verify on;
 
     add_header Strict-Transport-Security "max-age=31536000;";
+    add_header X-Content-Type-Options nosniff;
```
julienmalik: didn't know about this one, but seems relevant : http://stackoverflow.com/questions/18337630/what-is-x-content-type-options-nosniff

julienmalik:
You only modified yunohost_admin.conf, but the server conf (server.tpl) also does embed ssl directives.

Best would probably be to factorize generic ssl options (all ssl_xxx except the certs) in a common file, to be used in domain specific conf and yunohost_admin.conf

julienmalik:
What recommendation did you follow to set up these new cipher list and the rest ?

I think we used the Mozilla SSL conf generator : https://mozilla.github.io/server-side-tls/ssl-config-generator/
which is good since it quickly shows the corresponding supported clients.

The choice to harden ssl configuration usually has consequences on the number of supported clients (browsers), and on performance (client or server side). Ideally that's something that should be chosen by the admin, but yunohost is not yet ready for this kind of configuration.

zamentur:
Typically, it's here that we should think to introduce a settings management
In this case it could be:

- Harden encryption (some old device couldn't reach this server)
- Classic encryption

Psycojoker:
This one seems to be depending on https://github.com/YunoHost/yunohost/pull/229

alexAubin:
Closing because it seems to be superseded by https://github.com/YunoHost/yunohost/pull/259

PR #163, opened 2016-06-23
[fix] modifying a global variable was breaking apps installation on yunohost-api because the global modification was persistent

Psycojoker:
Hello,

Currently `app_tmp_folder` is a global variable that is modified during app related process (install, upgrade) here https://github.com/YunoHost/yunohost/compare/fix_435_340_326_maybe_311?expand=1#diff-c98834a0df97adaa917878824a06b38fL1174

This doesn't create any problem in CLI because once the command is finished, the program ends and at a new run the variable will be reseted.

But on the web interface, execpt if you restart the moulinette, this variable is still modified and you end up with those kind of errors:

    CalledProcessError: Command '['wget', '-qO', '/var/cache/yunohost/from_file/opensondage_ynh-0af12d46ca521ccd0e652901e1462578bc27f9b9/zerobin_ynh-6819a3e219ad3204b75551c3debda2390b483103/zerobin_ynh-6819a3e219ad3204b75551c3debda2390b483103.zip', 'https://github.com/YunoHost-Apps/shellinabox_ynh/archive/7a3a13cda984173eaa873711a933a2c07f13fe3d.zip']' returned non-zero exit status 1

Notice the: `/var/cache/yunohost/from_file/opensondage_ynh-0af12d46ca521ccd0e652901e1462578bc27f9b9/zerobin_ynh-6819a3e219ad3204b75551c3debda2390b483103/zerobin_ynh-6819a3e219ad3204b75551c3debda2390b483103.zip` ?

It's this variable that has received several modifications at the line indicated and thus generated an unexisting path.

This path intent to remove the modification of the global variable and thus preventing kind of errors to occur.

It should not create new bug (I've been greping all around to be sure not to miss something and it runs in my vagrant), but I'm not 100% sure about that since using a global variable and modifying it tends to create very hard to predict behavior. Sadly the person who have written this code is not here anymore to check that.

On a side no I think that we should rename for uppercase all our global variables for readability, I'm ready to send a patch for that.

This should fix those bugs: https://dev.yunohost.org/issues/435 https://dev.yunohost.org/issues/340 https://dev.yunohost.org/issues/326 and maybe https://dev.yunohost.org/issues/311

Kind regards,

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -362,9 +362,9 @@ def app_upgrade(auth, app=[], url=None, file=None):
         new_app_dict     = app_info(app_instance_name, raw=True)
 
         if file:
-            manifest = _extract_app_from_file(file)
+            manifest, extracted_app_folder = _extract_app_from_file(file)
         elif url:
-            manifest = _fetch_app_from_git(url)
+            manifest, extracted_app_folder = _fetch_app_from_git(url)
         elif new_app_dict is None or 'lastUpdate' not in new_app_dict or 'git' not in new_app_dict:
             logger.warning(m18n.n('custom_app_url_required', app=app_instance_name))
             continue
@@ -373,7 +373,7 @@ def app_upgrade(auth, app=[], url=None, file=None):
                    and (new_app_dict['lastUpdate'] > current_app_dict['settings']['install_time'])) \
               or ('update_time' in current_app_dict['settings'] \
                    and (new_app_dict['lastUpdate'] > current_app_dict['settings']['update_time'])):
-            manifest = _fetch_app_from_git(app_instance_name)
+            manifest, extracted_app_folder = _fetch_app_from_git(app_instance_name)
         else:
             continue
 
@@ -401,7 +401,7 @@ def app_upgrade(auth, app=[], url=None, file=None):
 
         # Execute App upgrade script
         os.system('chown -hR admin: %s' % install_tmp)
-        if hook_exec(app_tmp_folder +'/scripts/upgrade', args=args_list, env=env_dict) != 0:
+        if hook_exec(extracted_app_folder +'/scripts/upgrade', args=args_list, env=env_dict) != 0:
             logger.error(m18n.n('app_upgrade_failed', app=app_instance_name))
         else:
             now = int(time.time())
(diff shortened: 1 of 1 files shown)


Psycojoker:
Oh yeah, regarding the strategy I've chosen: explicitly passed the new modified path around and use this one except of using a global variable for that.

src/yunohost/app.py
```diff
@@ -1254,11 +1253,11 @@ def _fetch_app_from_git(app):
                 branch = app[tree_index+6:]
             try:
                 subprocess.check_call([
-                    'git', 'clone', '--depth=1', url, app_tmp_folder])
+                    'git', 'clone', '--depth=1', url, extracted_app_folder])
```
jeromelebleu: I think - since it's not easy to review that with this view - `extracted_app_folder` has not been declared here...

Psycojoker: It's normally declared here https://github.com/YunoHost/yunohost/pull/163/files#diff-c98834a0df97adaa917878824a06b38fR1247

jeromelebleu:
As the extracted folder path is now returned, it should be enhanced at an upper level. I think this folder should be created randomly and safely using [`tempfile.mkdtemp`](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp). It will prevent any other related strange issues - or at least reduce them. What do you think?

I understand that you didn't wanted to modify to much things, but I've seen some remaining lines that could still cause some issues...

src/yunohost/app.py
jeromelebleu: Except that it's not the same block, isn't it?

Psycojoker: Ah, indeed, good catch.

Psycojoker:
> As the extracted folder path is now returned, it should be enhanced at an upper level. I think this folder should be created randomly and safely using tempfile.mkdtemp. It will prevent any other related strange issues - or at least reduce them. What do you think?

Regarding other problems, normally `app_tmp_folder` is wiped out everytime so I
can't see a conflict happening in the current form of the code but maybe you
have though about some other situations I don't have in mind.

See https://github.com/YunoHost/yunohost/blob/f682468aee0699d1549af8dbdf0345ef73c30d59/src/yunohost/app.py#L1147 https://github.com/YunoHost/yunohost/blob/f682468aee0699d1549af8dbdf0345ef73c30d59/src/yunohost/app.py#L1218

But that would probably indeed cleaner (but there are so many things to clean
in the code already :/ and indeed I've tried to make this PR as small as
possible since this change is non trivial and a bit dangerouse)

I can do that if you want, but I would prefer to do that in another pull
request since this modification is already a very sensible modification, I
don't want to risk breaking other things by doing too much at the same time.

> I understand that you didn't wanted to modify to much things, but I've seen some remaining lines that could still cause some issues...

Which lines do you have in mind?

src/yunohost/app.py
Psycojoker: Fixed in a way that will prevent those kind of errors.

PR #165, opened 2016-06-28
First version of "journals" for YunoHost

Psycojoker:
Hello,

A feature I've been talking about in the past: store in an individual "journals" files some operations like "installing an app" "remove an app" etc...

The idea is to ease debug (it's easy to find what has happen after) but also to empower web users who can't get access to those logs once they have close the tab or other similar behavior.

For now I'm only generating journals for app installation/remove/remove when installation failed/upgrade. We should probably introduce new journals like domain create (openssl stuff), apt related operations and other I can't thnink about right now.

This PR needs a bit of finishing work but it fully functional but I'm still opening it to get feedback,

Here is the result of both commands:

2016-06-28-061340_925x335_scrot

2016-06-28-061414_923x270_scrot
End of previous command:
2016-06-28-061529_744x339_scrot

Weird format is due to mustache limitations.

Kinds regards,

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1392,7 +1392,7 @@ journals:
     category_help: Manage debug journals
     actions:
 
-        ### domain_list()
+        ### journals_list()
         list:
             action_help: List journals
             api: GET /journals
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1401,3 +1401,11 @@ journals:
                     full: --limit
                     help: Maximum number of journals per categories
                     type: int
+
+        ### journals_display()
+        display:
+            action_help: List journals
+            api: GET /journals/<file_name>
+            arguments:
+                file_name:
+                    help: Journal file name
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1384,3 +1384,15 @@ hook:
                 -d:
                     full: --chdir
                     help: The directory from where the script will be executed
+
+#############################
+#          Journals         #
+#############################
+journals:
+    category_help: Manage debug journals
+    actions:
+
+        ### domain_list()
+        list:
+            action_help: List journals
+            api: GET /journals
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1396,3 +1396,8 @@ journals:
         list:
             action_help: List journals
             api: GET /journals
+            arguments:
+               -l:
+                    full: --limit
+                    help: Maximum number of journals per categories
+                    type: int
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -334,6 +334,7 @@ def app_upgrade(auth, app=[], url=None, file=None):
 
     """
     from yunohost.hook import hook_add, hook_remove, hook_exec
+    from yunohost.journals import Journal
 
     try:
         app_list()
@@ -401,7 +402,8 @@ def app_upgrade(auth, app=[], url=None, file=None):
 
         # Execute App upgrade script
         os.system('chown -hR admin: %s' % install_tmp)
-        if hook_exec(app_tmp_folder +'/scripts/upgrade', args=args_list, env=env_dict) != 0:
+        journal = Journal(["upgrade", app_instance_name], "app", args=args_list, env=env_dict)
+        if hook_exec(app_tmp_folder +'/scripts/upgrade', args=args_list, env=env_dict, journa=journal) != 0:
             logger.error(m18n.n('app_upgrade_failed', app=app_instance_name))
         else:
             now = int(time.time())
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -600,6 +600,7 @@ def app_remove(auth, app):
 
     """
     from yunohost.hook import hook_exec, hook_remove
+    from yunohost.journals import Journal
 
     if not _is_installed(app):
         raise MoulinetteError(errno.EINVAL,
@@ -624,7 +625,9 @@ def app_remove(auth, app):
     env_dict["YNH_APP_INSTANCE_NAME"] = app
     env_dict["YNH_APP_INSTANCE_NUMBER"] = str(app_instance_nb)
 
-    if hook_exec('/tmp/yunohost_remove/scripts/remove', args=args_list, env=env_dict) == 0:
+    journal = Journal(["remove", app], "app", args=args_list, env=env_dict)
+
+    if hook_exec('/tmp/yunohost_remove/scripts/remove', args=args_list, env=env_dict, journal=journal) == 0:
         logger.success(m18n.n('app_removed', app=app))
 
     if os.path.exists(app_setting_path): shutil.rmtree(app_setting_path)
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -446,6 +446,7 @@ def app_install(auth, app, label=None, args=None):
 
     """
     from yunohost.hook import hook_add, hook_remove, hook_exec
+    from yunohost.journals import Journal
 
     # Fetch or extract sources
     try: os.listdir(install_tmp)
@@ -529,7 +530,12 @@ def app_install(auth, app, label=None, args=None):
     try:
         install_retcode = hook_exec(
             os.path.join(app_tmp_folder, 'scripts/install'),
-            args=args_list, env=env_dict)
+            args=args_list, env=env_dict,
+            journal = Journal(
+                ["install", app_instance_name],
+                "app", args=args_list, env=env_dict
+            ),
+        )
     except (KeyboardInterrupt, EOFError):
         install_retcode = -1
     except:
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -551,7 +551,12 @@ def app_install(auth, app, label=None, args=None):
             # Execute remove script
             remove_retcode = hook_exec(
                 os.path.join(app_tmp_folder, 'scripts/remove'),
-                args=[app_instance_name], env=env_dict_remove)
+                args=[app_instance_name], env=env_dict_remove,
+                journal = Journal(
+                    ["remove", app_instance_name, "failed install"],
+                    "app", args=[app_instance_name], env=env_dict_remove,
+                ),
+            )
             if remove_retcode != 0:
                 logger.warning(m18n.n('app_not_properly_removed',
                                       app=app_instance_name))
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+
+""" License
+
+    Copyright (C) 2016 YunoHost
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published
+    by the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program; if not, see http://www.gnu.org/licenses
+
+"""
+
+""" yunohost_journals.py
+
+    Manage debug journals
+"""
+
+import os
+
+from moulinette.utils.log import getActionLogger
+
+JOURNALS_PATH = '/var/log/journals/'
+
+logger = getActionLogger('yunohost.journals')
+
+
+def journals_list():
+    """
+    List domains
+
+    Keyword argument:
+        filter -- LDAP filter used to search
+        offset -- Starting number for domain fetching
+        limit -- Maximum number of domain fetched
+
+    """
+    return {}
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -25,6 +25,9 @@
 """
 
 import os
+import yaml
+
+from datetime import datetime
 
 from moulinette.utils.log import getActionLogger
 
@@ -54,8 +57,12 @@ class Journal(object):
     def __init__(self, name, category, on_stdout=None, on_stderr=None, on_write=None, **kwargs):
         self.name = name
         self.category = category
-        self.first_write = False
+        self.first_write = True
         self.started_at = None
+
+        self.path = os.path.join(JOURNALS_PATH, category)
+
+        self.fd = None
 
         self.on_stdout = [] if on_stdout is None else on_stdout
         self.on_stderr = [] if on_stderr is None else on_stderr
@@ -63,8 +70,34 @@ class Journal(object):
 
         self.additional_information = kwargs
 
+    def __del__(self):
+        if self.fd:
+            self.fd.close()
+
     def write(self, line):
-        print "[journal]", line.rstrip()
+        if self.first_write:
+            self._do_first_write()
+            self.first_write = False
+
+        self.fd.write("%s: " % datetime.now().strftime("%F_%X").replace(":", "-"))
+        self.fd.write(line.rstrip())
+        self.fd.write("\n")
+        self.fd.flush()
+
+    def _do_first_write(self):
+        self.started_at = datetime.now()
+
+        if not os.path.exists(self.path):
+            os.makedirs(self.path)
+
+        file_name = "%s_%s.journal" % (self.name if isinstance(self.name, basestring) else "_".join(self.name), self.started_at.strftime("%F_%X").replace(":", "-"))
+
+        serialized_additional_information = yaml.safe_dump(self.additional_information, default_flow_style=False)
+
+        self.fd = open(os.path.join(self.path, file_name), "w")
+
+        self.fd.write(serialized_additional_information)
+        self.fd.write("\n---\n")
 
     def stdout(self, line):
         for i in self.on_stdout:
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -48,3 +48,41 @@ def journals_list():
         return {}
 
     return {}
+
+
+class Journal(object):
+    def __init__(self, name, category, on_stdout=None, on_stderr=None, on_write=None, **kwargs):
+        self.name = name
+        self.category = category
+        self.first_write = False
+        self.started_at = None
+
+        self.on_stdout = [] if on_stdout is None else on_stdout
+        self.on_stderr = [] if on_stderr is None else on_stderr
+        self.on_write = [] if on_write is None else on_write
+
+        self.additional_information = kwargs
+
+    def write(self, line):
+        print "[journal]", line.rstrip()
+
+    def stdout(self, line):
+        for i in self.on_stdout:
+            i(line)
+
+        self.write(line)
+
+    def stderr(self, line):
+        for i in self.on_stderr:
+            i(line)
+
+        self.write(line)
+
+    def as_callbacks_tuple(self, stdout=None, stderr=None):
+        if stdout:
+            self.on_stdout.append(stdout)
+
+        if stderr:
+            self.on_stderr.append(stderr)
+
+        return (self.stdout, self.stderr)
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -95,7 +95,7 @@ def journals_display(file_name):
 
                 infos, logs = content.split("\n---\n", 1)
                 infos = yaml.safe_load(infos)
-                logs = [x.split(": ", 1) for x in logs.split("\n") if x]
+                logs = [{"datetime": x.split(": ", 1)[0].replace("_", " "), "line": x.split(": ", 1)[1]}  for x in logs.split("\n") if x]
 
                 return {
                     "started_at": journal_datetime,
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -76,6 +76,40 @@ def journals_list(limit=None):
     return result
 
 
+def journals_display(file_name):
+    if not os.path.exists(JOURNALS_PATH):
+        # TODO raise exception
+        return {}
+
+    for category in os.listdir(JOURNALS_PATH):
+        for journal in filter(lambda x: x.endswith(".journal"), os.listdir(os.path.join(JOURNALS_PATH, category))):
+            if journal != file_name:
+                continue
+
+            with open(os.path.join(JOURNALS_PATH, category, file_name), "r") as content:
+                content = content.read()
+
+                journal = journal[:-len(".journal")]
+                journal = journal.split("_")
+                journal_datetime = datetime.strptime(" ".join(journal[-2:]), "%Y-%m-%d %H-%M-%S")
+
+                infos, logs = content.split("\n---\n", 1)
+                infos = yaml.safe_load(infos)
+                logs = [x.split(": ", 1) for x in logs.split("\n") if x]
+
+                return {
+                    "started_at": journal_datetime,
+                    "name": " ".join(journal[:-2]),
+                    "file_name": file_name,
+                    "path": os.path.join(JOURNALS_PATH, category, file_name),
+                    "metadata": infos,
+                    "logs": logs,
+                }
+
+    # TODO raise exception
+    return {}
+
+
 class Journal(object):
     def __init__(self, name, category, on_stdout=None, on_stderr=None, on_write=None, **kwargs):
         self.name = name
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -74,7 +74,9 @@ class Journal(object):
         self.name = name
         self.category = category
         self.first_write = True
-        self.started_at = None
+
+        # this help uniformise file name and avoir threads concurrency errors
+        self.started_at = datetime.now()
 
         self.path = os.path.join(JOURNALS_PATH, category)
 
@@ -101,8 +103,6 @@ class Journal(object):
         self.fd.flush()
 
     def _do_first_write(self):
-        self.started_at = datetime.now()
-
         if not os.path.exists(self.path):
             os.makedirs(self.path)
 
--- a/src/yunohost/journals.py
+++ b/src/yunohost/journals.py
@@ -104,7 +104,7 @@ class Journal(object):
             self._do_first_write()
             self.first_write = False
 
-        self.fd.write("%s: " % datetime.now().strftime("%F_%X").replace(":", "-"))
+        self.fd.write("%s: " % datetime.now().strftime("%F %X"))
         self.fd.write(line.rstrip())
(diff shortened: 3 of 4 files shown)


src/yunohost/app.py
```diff
@@ -401,7 +402,8 @@ def app_upgrade(auth, app=[], url=None, file=None):
 
         # Execute App upgrade script
         os.system('chown -hR admin: %s' % install_tmp)
-        if hook_exec(app_tmp_folder +'/scripts/upgrade', args=args_list, env=env_dict) != 0:
+        journal = Journal(["upgrade", app_instance_name], "app", args=args_list, env=env_dict)
+        if hook_exec(app_tmp_folder +'/scripts/upgrade', args=args_list, env=env_dict, journa=journal) != 0:
```
zamentur: I think there is a missing "l" no?  "journa=journal"

data/actionsmap/yunohost.yml
```diff
@@ -1384,3 +1384,28 @@ hook:
                 -d:
                     full: --chdir
                     help: The directory from where the script will be executed
+
+#############################
+#          Journals         #
+#############################
+journals:
```
zamentur: Why we don't use the more common "log" designation ?

src/yunohost/journals.py
```diff
@@ -0,0 +1,179 @@
...
+logger = getActionLogger('yunohost.journals')
+
+
+def journals_list(limit=None):
+    """
+    List domains
+
+    Keyword argument:
+        filter -- LDAP filter used to search
+        offset -- Starting number for domain fetching
```
zamentur: False documentation

data/actionsmap/yunohost.yml
```diff
@@ -1384,3 +1384,28 @@ hook:
...
+            arguments:
+               -l:
+                    full: --limit
+                    help: Maximum number of journals per categories
+                    type: int
+
+        ### journals_display()
+        display:
+            action_help: List journals
+            api: GET /journals/<file_name>
```
zamentur: How to display the journal of an app from the APPID ? 
It could be useful to let the user access it from the webadmin app page.

src/yunohost/app.py
Psycojoker: Indeed.

data/actionsmap/yunohost.yml
Psycojoker: Because we already uses logs elsewhere and that a log is a continuous stream of output where here we have atomic operations.

I'm not 100% happy with "journal" but it's seems the closest to what I wanted and that wasn't log. I'm open to suggestion (hoping that this doesn't end up in a bikeshding).

data/actionsmap/yunohost.yml
Psycojoker: Indeed.

data/actionsmap/yunohost.yml
```diff
@@ -1384,3 +1384,28 @@ hook:
...
+    actions:
+
+        ### journals_list()
+        list:
+            action_help: List journals
+            api: GET /journals
+            arguments:
+               -l:
+                    full: --limit
+                    help: Maximum number of journals per categories
```
Psycojoker: Journals are arranged in categories : right now there is only the "app" category iirc (it's been a while since I've made this PR).

Psycojoker:
I won't work on this branch soon, I'm waiting for other more important PR to be merged first.

Psycojoker:
Finally took the time to finish this PR, should be ready for review (but not prior).

Psycojoker:
Hello,

Calling for a lighter merge on this one because Aleks is busy on stretch (and I prefer him to be on that part of the code than reviewing this) and that it seems that no one else has time for this PR https://github.com/YunoHost/project-organization/blob/master/yunohost_project_organization_fr.md#validation-all%C3%A9g%C3%A9-en-cas-de-manque-de-relecteurs

zamentur:
There is also `yunohost service log`

Psycojoker:
Parsing logs really seems like a unrelying way of doing this tbh :/

I have to admit that a "no, we should redo everything, it's not the good way to do that" on a finished PR that has been there for month and on which you've already commented and made reviews is ... tiring. Especially since it's already done, will provide value for existing user and is actually solving an existing problem of finding "what had happen when I did that" (which is something else than reading the raw logs). It feels a bit like working with mozilla management :p

zamentur:
Sorry Bram , you are true your pr is here since a long time, and I was quite embarrassing when i reread it. In the past i didn't thinking to that point, "yunohost tools yunohost-logs" pr exist since the 5th january, it was reading this one I realized the issue. And I already had explained my confusion between log and journal.
https://github.com/YunoHost/yunohost/pull/405

I may not realised what exactly journals are. May be i am wrong with my definition : part of log corresponding to an administrator actions. 

I had suggested to do something like that: 

```
awk  '$0 ~ /^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-2][0-9]:[0-5][0-9]:[0-5][0-9]/
      {
        if ($1" "$2 >= "2018-02-04 06:53:43")     p=1;
        if ($1" "$2 >= "2018-02-04 15:29:40")  p=0;
      }
    p { print $0 }' yunohost-cli.log*
```

**For me it's not a problem if the decision is to keep the code with specific file containing a copy of the log, instead of doing a complicated awk call.**  It's an implementation, not really matter for me.

But the choice of the user cli interactions with `yunohost journal` and `yunohost tools yunohost-logs` are a bit more problematic.

I suggest to rename `yunohost tools yunohost-logs`  as `yunohost journal display ` and to consider the possibility to change `journal` as `log` or `history`

So with this suggestion the cli could be:

```
yunohost [tools?] log list => List of META actions
yunohost [tools?] log display => Give the comple yunohost log
yunohost [tools?] log display install_ttrss_[...] => Give the log of a META ACTIONS with context
```

zamentur:
I replaced Journal by a UnitOperationHandler to catch directly Logrecords from all Loggers and I added a is_unit_operation decoractor to simplify the deployment of this log stuff.

Psycojoker:
I might be mistaking but it seems to me that you totally removed the possibility to get all the logs for a certain "unit", like get all the logs related to an application operation itself to be able to display those information directly on the application page to just having unit of logs which was really not the spirit of this pull request :/

This is not about having "a list of logs" but having categorized logs linked to different part of the system, I'm not sure you understood the initial intent :/

This is really to solved the problem of "I want to find the installation/remove/upgrade/wathever log of this application/other stuff that has failed", not making logs accessible in the admin UI. That's possibly the reason why you wanted to merged both PR btw, which are really 2 different things ("logs" vs "journals").

zamentur:
You 're right i accidentally removed some information from metadata and filename.

I restore it.

I started configuring the system on

- app_removelist
- app_change_url
- app_install
- app_remove
- app_upgrade
- app_makedefault
- domain_add
- domain_remove
- dyndns_subscribe
- dyndns_update
- user_create
- user_delete
- user_update
- tools_maindomain
- tools_postinstall
- tools_upgrade
- tools_shutdown
- tools_reboot

We might configure it on:
app_addaccess
app_removeaccess
app_clearaccess
domain_cert_install 
domain_cert_renew
tools_migrations_migrate
service_regenconf

zamentur:
- [ ] Create name from locale to display them correctly
- [ ] Test all actions
- [ ] Update Webadmin PR

zamentur:
> This is not about having "a list of logs" but having categorized logs linked to different part of the system, I'm not sure you understood the initial intent :/
> 
> This is really to solved the problem of "I want to find the installation/remove/upgrade/wathever log of this application/other stuff that has failed", not making logs accessible in the admin UI. That's possibly the reason why you wanted to merged both PR btw, which are really 2 different things ("logs" vs "journals").

I write a 'related_to' field in metadata. The idea is to reference in this field all entities related to the operation. Like that it will be possible to list all operation related to an entity (an app, a service, a domain, a system part, a user?).

app install operation
```
related_to:
    app: strut
    domain: example.org
```

change url operation (not yet implemented as it)
```
related_to:
    app: strut
    domain: 
     - old.org
     - new.org
```

I separated .journal in 2 files, a .yml containing metadata and a .log containing associated log. With this change I was able to add some metadata during the record of logs and at the end (to add error/success status and end date)

By convention, .log file is place in the same dir as the .yml file, but we could imagine an other path to be specified. It could solve the nextcloud log issue by referencing the log file related to nextcloud app located in /home/yunohost.app/data/nextcloud/nextcloud.log (or something like that). But UnitOperation terminology is unsuitable in this case because it's a 'usage' log not an operation on the system !

data/actionsmap/yunohost.yml
```diff
@@ -1614,16 +1614,23 @@ log:
...
             action_help: Display a log content
-            api: GET /logs/<file_name_list>
+            api: GET /logs/<file_name>
             arguments:
-                file_name_list:
-                    help: Log filenames for which to display the content
-                    nargs: "*"
+                file_name:
+                    help: Log filename which to display the content
+                -n:
```
Psycojoker: Maybe add an option to get the full log or to avoid limiting it? Like `--full`

Psycojoker:
Fianlly took the time to review the general principal and looks really good, thanks :)

Something I really wanted to be able to do (but haven't done it) was to print at the end of a failed app operation something like "to see the log related to this operation you can run `yunohost journals display $stuff` and you can find all journals by using `yunhost journals list`" or something like that with and example on how to use yunopaste to share it (or having a button for that on the admin if like we already have).

I don't think it's yet possible since you haven't implemented filtering :confused: ?

zamentur:
Example of usage:
Wrap all the command with a unit operation
```
@is_unit_operation()
def domain_add(auth, domain, dyndns=False):
```

Several unit operations inside one command
```
def app_upgrade(auth, app=[], url=None, file=None):
    [...]
    for app_instance_name in apps:
        [...]
        # Start register change on system
        related_to = [('app', app_instance_name)]
        uo = UnitOperation('app_upgrade', related_to, env=env_dict)
        [CHANGE]

        if hook_exec(extracted_app_folder + '/scripts/upgrade', args=args_list)
            uo.error(msg)
        else:
            [...]
            uo.success()

```

Unit operations start after some checks or behaviour
```
@is_unit_operation(auto=False)
def app_change_url(uo, auth, app, domain, path):
    [Non changing code]
    if domain != old_domain:
        uo.related_to.append(('domain', old_domain))
    uo.extra.update({'env': env_dict})
    uo.start()
    [CHANGE]
```

zamentur:
List to things to test

- [x]  tools_postinstall
- [ ]  app_removelist
- [ ]  app_change_url
- [ ]  app_install
- [ ]  app_remove
- [ ]  app_upgrade
- [ ]  app_makedefault
- [ ]  domain_add
- [ ]  domain_remove
- [ ]  dyndns_subscribe
- [ ]  dyndns_update
- [ ]  user_create
- [ ]  user_delete
- [ ]  user_update
- [ ]  tools_maindomain
- [ ]  tools_upgrade
- [ ]  tools_shutdown
- [ ]  tools_reboot
- [ ]  app_addaccess
- [ ]  app_removeaccess
- [ ]  app_clearaccess
- [ ]  domain_cert_install
- [ ]  domain_cert_renew
- [ ]  tools_migrations_migrate
- [ ]  service_regenconf
- [ ]  backup_restore
- [ ]  Unit operation that fail

src/yunohost/log.py
```diff
@@ -0,0 +1,424 @@
...
+    "log_" + operation_key in locales files.
+
+    auto        If true, start the recording. If False, the unit operation object
+    created is given to the decorated function as the first argument and you can
+    start recording at the good time.
+    """
+    def decorate(func):
+        def func_wrapper(*args, **kwargs):
+            # For a strange reason we can't use directly the arguments from
+            # is_unit_operation function. We need to store them in a var before.
```
Psycojoker: You should be able to do here because you don't re-assign the arguments name here so I guess you've runt in this problem before but here you are safe.

This is due to the way python optimize things.

Here is a sample code to demonstrate this:

```python
def pouet(a):
    def prout():
        def pif():
            a + 1
            # uncommenting this will make it fails
            # a = a + 1
            print a
        pif()
    prout()

pouet(42)
```

src/yunohost/log.py
```diff
@@ -0,0 +1,424 @@
...
+    return infos
+
+def is_unit_operation(entities='app,domain,service,user', exclude='auth,password', operation_key=None, auto=True):
+    """
+    Configure quickly a unit operation
+
+    This decorator help you to configure quickly the record of a unit operations.
+
+    Argument:
+    entities    A list seperated by coma of entity types related to the unit
```
Psycojoker: Why a list of comma separated entities in a string and not a python list? (or a tuple) It's meant for that :yum:

src/yunohost/log.py
```diff
@@ -0,0 +1,424 @@
...
+    operation. The entity type is searched inside argument's names of the
+    decorated function. If something match, the argument value is added as
+    related entity.
+
+    exclude     Remove some arguments from the context. By default, arguments
+    called 'password' and 'auth' are removed. If an argument is an object, you
+    need to exclude it or create manually the unit operation without this
+    decorator.
+
+    operation_key   Key describing the unit operation. If you want to display a
```
Psycojoker: Alternative proposition: "a key to a translatable strings store in locales/*.json to describe your operation log. Please ensure that this key is present in locales/en.json otherwise it won't be translatable."

src/yunohost/log.py
```diff
@@ -0,0 +1,424 @@
...
+        def func_wrapper(*args, **kwargs):
+            # For a strange reason we can't use directly the arguments from
+            # is_unit_operation function. We need to store them in a var before.
+            entities_list = entities.split(',')
+            exclude_list = exclude.split(',')
+            op_key = operation_key
+            related_to = []
+
+            if op_key is None:
+                op_key = func.__name__
```
Psycojoker: Shouldn't it be "log_" + func.__name__ here? Following "operation_key" key logic.

src/yunohost/log.py
```diff
@@ -0,0 +1,424 @@
...
+
+            # In case the function is called directly from an other part of the
+            # code
+            if len(args) > 0:
+                from inspect import getargspec
+                keys = getargspec(func).args
+                if 'uo' in keys:
+                    keys.remove('uo')
+                for k, arg in enumerate(args):
+                    kwargs[keys[k]] = arg
```
Psycojoker: Without context this looks like heavy magic, a comment wouldn't be too much.

src/yunohost/log.py
```diff
@@ -0,0 +1,424 @@
...
+                args = ()
+
+            # Search related entity in arguments of the decorated function
+            for entity in entities_list:
+                entity = entity.split(':')
```
Psycojoker: Again, we can have a list of string or a list of tuple, that seems weird to invent a format store in a string :confused:

Psycojoker: If you keep that you probably want to do a `.split(":", 1)`

src/yunohost/log.py
```diff
@@ -252,8 +252,7 @@ def func_wrapper(*args, **kwargs):
                 result = func(*args, **kwargs)
             except Exception as e:
                 uo.error(e)
-                t, v, tb = exc_info()
-                raise t, v, tb
+                raise
```
zamentur: Have you check it make the same thing ? Because I think I added that due to the miss of the stacktrace.

src/yunohost/log.py
zamentur: +1

src/yunohost/service.py
```diff
@@ -708,15 +731,47 @@ def _tail(file, n):
...
 
     except IOError as e:
         logger.warning("Error while tailing file '%s': %s", file, e, exc_info=1)
         return []
 
+    if len(lines) < to_read:
+        previous_log_file = _find_previous_log_file(file)
+        if previous_log_file is not None:
+            lines = _tail(previous_log_file, to_read - len(lines)) + lines
+    return lines[-to_read:offset and -offset or None]
```
alexAubin: Got a critical error here because offset was undefined. Not sure what the proper fix is, but in the meantime i just replaced the line with `return lines` instead.

Psycojoker: That's a conflict with master code because I've removed the notion of offset mostly everywhere because we never used it.

alexAubin:
In 8eaceef, I propose to introduce a `--share` option to `yunohost log display` to be able to share the log easily via a python-equivalent of yunopaste.

This is motivated by the following experience : 
- trying the use case of "failure to install an app", the user gets a message about `yunohost log display`, which is fine but : 
    - this doesn't display all the lines, which might be necessary for people in the support to really understand what is going on
    - because of the whole dictionnary thing, keys are in alphabetical order, and therefore the logs end up in the middle of other keys, instead of having for instance 1st : the metadata, and 2nd : the full log
    - there is no indication to use yunopaste to share the log
    - yunopaste sometimes complains that the POST data are too long, which is a motivation to introduce a python-equivalent using `requests` which is hopefully more robust (though I havent tested)

Using `yunohost log display <somename> --share` yields an url pointing to something like this : https://paste.yunohost.org/raw/oneyicusit

This is still not perfect (e.g. we might want to reorder the metadata keys order) but imho that should definitely help with the use case of "app install crashed and user wants to share the log".

One remaining thing to do would be to use this mechanism also from the webadmin ? (Doing some tests, it looks to me as if the 'Share' button does some buggy behavior and yields a different format compared to the CLI...)

src/yunohost/log.py
```diff
@@ -0,0 +1,436 @@
...
+            self.logger.removeHandler(self.file_handler)
+
+        is_api = msettings.get('interface') == 'api'
+        desc = _get_description_from_name(self.name)
+        if error is None:
+            if is_api:
+                msg = m18n.n('log_link_to_log', name=self.name, desc=desc)
+            else:
+                msg = m18n.n('log_help_to_get_log', name=self.name, desc=desc)
+            logger.info(msg)
```
alexAubin: I don't know what to do about this ... but printing this message after each command looks a bit too much... In particular, this also apply to "sub" operations. To illustrate, when adding a domain I get this : 

```
 > yunohost domain add test.lol
Administration password: 
Success! Successfully installed a self-signed certificate for domain test.lol!
Info: To view the log of the operation 'Install self signed certificate on 'test.lol' domain', use the command 'yunohost log display 20180725-144845-selfsigned_cert_install-test.lol'
Warning: The configuration file '/etc/nginx/conf.d/yunohost_admin.conf' has been manually modified and will not be updated
Success! The configuration has been updated for service 'metronome'
Info: To view the log of the operation 'Regenerate system configurations 'nginx'', use the command 'yunohost log display 20180725-144847-service_regen_conf-nginx'
Success! The SSOwat configuration has been generated
Success! The domain has been created
Info: To view the log of the operation 'Add 'test.lol' domain into system configuration', use the command 'yunohost log display 20180725-144845-domain_add-test.lol'
```

So here I got three `To view the log of the operation [...]` about logs which I won't go checking because everything worked anyway ... As a user, what I really want to know is if everything worked correctly and if I need to do something in particular, so in that perspective those messages (at least the first two) looks like noise :/

src/yunohost/log.py
```diff
@@ -0,0 +1,436 @@
...
+        This name is used as filename, so don't use space
+        """
+        if self._name is not None:
+            return self._name
+
+        name = [self.started_at.strftime("%Y%m%d-%H%M%S")]
+        name += [self.operation]
+        if self.related_to:
+            if isinstance(self.related_to[0], tuple):
+                name += [self.related_to[0][1]]
```
alexAubin: This behavior is kind of an issue imho. c.f. my other comment about adding a domain. I got the following message : 
```
Info: To view the log of the operation 'Regenerate system configurations 'nginx'', use the command 'yunohost log display 20180725-144847-service_regen_conf-nginx
```
so I had a look at the log and see a whole bunch of lines about metronome. So I was confused ... then I realize that nope, this isn't a regen of nginx. This is a regen of nginx+metronome+dnsmasq(+rmilter) ... 

This is a bit tricky do define what to do for this ... Naively I would do something like : "if there is only one related thing (e.g. service), use that, otherwise don't use anything". So in that case the description would just be "Regenerate configurations" and it's up to user / support to check the metadata. (Note that if you regen all config, some magic is used so that the description is "Regenerate configurations 'all'".)

src/yunohost/log.py
```diff
@@ -0,0 +1,436 @@
...
+            else:
+                msg = m18n.n('log_help_to_get_log', name=self.name, desc=desc)
+            logger.info(msg)
+        else:
+            if is_api:
+                msg = "<strong>" + m18n.n('log_link_to_failed_log',
+                                        name=self.name, desc=desc) + "</strong>"
+            else:
+                msg = m18n.n('log_help_to_get_failed_log', name=self.name, desc=desc)
+            logger.warning(msg)
```
alexAubin: About this I don't know either what we should do ... I just tested a dummy cert-install like this : 

```
 > yunohost domain cert-install gruh.schpof
Info: Now attempting install of certificate for domain gruh.schpof!
Error: No DNS 'A' record found for gruh.schpof. You need to make your domain name point to your machine to be able to install a Let's Encrypt certificate! (If you know what you are doing, use --no-checks to disable those checks.)
Warning: The operation 'Install Let's encrypt certificate on 'gruh.schpof' domain' has failed ! To get help, please share the full log of this operation using the command 'yunohost log display 20180725-154526-letsencrypt_cert_install-gruh.schpof --share'
```
So if I'm a use not paying much attention, what I understand is that I should upload the log and get some help on the forum/chat about this. But the true meaningful message here is that there is no A record in the DNS and the user should check its DNS config. (That's especially relevant for the webadmin where, if the user don't click on the top bar, he'll only see the very last message)

Thing is that we have at least two kind of errors. 'Expected' errors like this (e.g. invalid arguments, invalid DNS conf, ...) where we shouldn't print this message, and 'Unexpected' errors (e.g. unexpected exception or failed app install) where we want to show this. But not sure how to easily differentiate the two :/

alexAubin:
So after reviewing this (mostly from a user / 'black box' point of view) this is pretty cool :+1: 

Apart from the few points I mentioned before I'm okay with basically everything :+1: 

ljf will be back in one week so I'll try to see with him (and other people interested ?) to discuss the last remaining items (including Bram's comments from June 22). Hopefully that should be the last iteration before the release :sweat_smile:

src/yunohost/log.py
zamentur: No because we use the key to make the filename too

src/yunohost/log.py
zamentur: I improve the comment about this part of the code

alexAubin:
Alrighty guyz, we should be done with this PR now :sweat_smile:, or at least there is nothing else to iterate on with @zamentur 

If you have some final feedback, please send them ~soon :sweat_smile:. The plan is to make a testing with this PR and a few other after we release as stable (as discussed in the last meeting)

src/yunohost/domain.py
```diff
@@ -113,20 +117,24 @@ def domain_add(auth, domain, dyndns=False):
...
+        from sys import exc_info;
+        t, v, tb = exc_info()
+
         # Force domain removal silently
         try:
             domain_remove(auth, domain, True)
         except:
             pass
-        raise
+        raise t, v, tb
```
Psycojoker: You don't need "t, v, tb" here, raise alone will reraise and keep the stack.

src/yunohost/tools.py
```diff
@@ -852,26 +868,36 @@ def tools_migrations_migrate(target=None, skip=False, auto=False, accept_disclai
...
                 if mode == "forward":
-                    migration.migrate()
+                    m = migration["module"].MyMigration()
+                    m.uo = uo
+                    m.migrate()
                 elif mode == "backward":
-                    migration.backward()
+                    m = migration["module"].MyMigration()
+                    m.uo = uo
+                    m.backward()
```
Psycojoker: Can you really avoid using one letter variable name plz? Especially when it wasn't one and a perfectly fine one before :/

src/yunohost/utils/yunopaste.py
```diff
@@ -0,0 +1,29 @@
...
+        raise MoulinetteError(errno.EIO,
+                              "Something wrong happened while trying to paste data on paste.yunohost.org : %s" % str(e))
+
+    if r.status_code != 200:
+        raise MoulinetteError(errno.EIO,
+                              "Something wrong happened while trying to paste data on paste.yunohost.org : %s" % r.text)
+
+    try:
+        url = json.loads(r.text)["key"]
+    except:
```
Psycojoker: Hidden exception that will make debugging harder.

src/yunohost/utils/yunopaste.py
```diff
@@ -0,0 +1,29 @@
...
+
+    try:
+        r = requests.post("%s/documents" % paste_server, data=data)
+    except Exception as e:
+        raise MoulinetteError(errno.EIO,
+                              "Something wrong happened while trying to paste data on paste.yunohost.org : %s" % str(e))
+
+    if r.status_code != 200:
+        raise MoulinetteError(errno.EIO,
+                              "Something wrong happened while trying to paste data on paste.yunohost.org : %s" % r.text)
```
Psycojoker: You should also display the error code in the exception.

src/yunohost/utils/yunopaste.py
```diff
@@ -0,0 +1,29 @@
...
+import errno
+
+from moulinette.core import MoulinetteError
+
+def yunopaste(data):
+
+    paste_server = "https://paste.yunohost.org"
+
+    try:
+        r = requests.post("%s/documents" % paste_server, data=data)
```
Psycojoker: No timeout set, we should ALWAYS have a timeout on requests calls to avoid freezing the whole YunoHost.

Psycojoker:
Some global comments:

I'm quite busy right now so I don't have the time to implement small details I've raised.

Shouldn't we do a release without journals first? To have a quieter one as we've discussed duirng the meeting.

In general, using "with block" would make the API way more explicit/simple, that would move from:

```python
uo = UnitOperation('app_fetchlist')
uo.start()
....
....
....
....
uo.success()
```

To:

```python
with UnitOperation('app_fetchlist') as uo:
    ....
    ....
    ....
    ....
```

That's not super important but that would split the concepts present in the code and avoid mixing the uo stuff with the rest of the code. That's not blocking I think, we can do that later.

Last and most important comment: I think that "uo" is actually a terrible variable name: it's not intuitive, doesn't mean anything related to any close concept present in our fields, it really requires and ask someone to learn a new concept that would be way more simpler to grasp if it was simply called "log" or "journal" or something similar. Also we should never have a 2 letters variable name if it's not intuitive, that makes the code harder to read and to grasp for new comers and we should avoid that a lot because readability is way more important that saving a few seconds (and we have text-completion in editors today, no excuse).

Psycojoker:
I haven't had the time to test the API and review log.py, will do that later.

src/yunohost/domain.py
zamentur: I think I change that after a test in which the initial error line 120 was not reraise. I think if we put raise, it will raise the error from the remove operation no ?

zamentur:
> In general, using "with block" would make the API way more explicit/simple

With statement will add an indentation in more on a large part of code. I prefer without personally.

PR #167, opened 2016-07-07
[mod] link to 2.4 migration guide on forum on app incompatibility error

Psycojoker:
Could not find string on weblate https://translate.yunohost.org/search/?q=requires+a+more+recent&search=ftx&source=on&target=on&context=on&location=on&comment=on&type=all&ignored=False&lang=


opi:
I guess old-stable branch is not on weblate

jeromelebleu:
Merged with https://github.com/YunoHost/yunohost/commit/b591c48a33bfa5246abceedff46a88740192b59b.

PR #169, opened 2016-07-13
Replace deprecated psutil.BOOT_TIME attribute

jellium:
The `psutil.BOOT_TIME` attribute appears to be deprecated, and returns an `AttributeError` when called. 
Even though the psutil documentation states that old attributes should be callable, one should use `psutil.boot_time()` function instead.
In some version of psutil (e.g. 2.1.1), both appear callable. In old versions (e.g. 1.2.1), `psutil.BOOT_TIME` is but `psutil.boot_time()` is not. In the newest version (4.3.0), only the latter can be called.

This bug would introduce an Internal Server Error (Error 500) when accessing the Monitor tool in the YunoHost admin panel.

### DIFF
--- a/src/yunohost/monitor.py
+++ b/src/yunohost/monitor.py
@@ -283,7 +283,7 @@ def monitor_system(units=None, human_readable=False):
         elif u == 'process':
             result[u] = json.loads(glances.getProcessCount())
         elif u == 'uptime':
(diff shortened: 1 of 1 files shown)


jeromelebleu:
Thanks about that, actually, https://github.com/giampaolo/psutil/blob/release-2.1.1/psutil/__init__.py#L1964.

jeromelebleu:
Merged with https://github.com/YunoHost/yunohost/commit/53e19fa32ddd484e777e06bd1711cea29294f43e.

PR #170, opened 2016-07-21
[fix] Raise error on malformed SSOwat persistent conf.

opi:


opi:
See error trace: https://paste.yunohost.org/ihedaketuf.vhdl

src/yunohost/user.py
```diff
@@ -177,14 +177,17 @@ def user_create(auth, username, firstname, lastname, mail, password,
...
-                del ssowat_conf['redirected_urls']['/']
-
-            with open('/etc/ssowat/conf.json.persistent', 'w+') as f:
-                json.dump(ssowat_conf, f, sort_keys=True, indent=4)
+        if 'redirected_urls' in ssowat_conf and '/' in ssowat_conf['redirected_urls']:
+            del ssowat_conf['redirected_urls']['/']
 
-        except IOError: pass
+        with open('/etc/ssowat/conf.json.persistent', 'w+') as f:
+            json.dump(ssowat_conf, f, sort_keys=True, indent=4)
```
Psycojoker: I would be great to still handle `IOError` exception here since this is something that can fail (no more place on hardrive of whatever) and give an appropriate message exception (the previous behavior of failing silently was really something we don't want to do).

The rest of the PR looks fine for me, thx <3

opi: Maybe I'm wrong, but the `IOError` seems to be handled sooner: https://github.com/YunoHost/yunohost/pull/170/files/937f26bdaaca64849570c472914e8d8a2e7f1fbf#diff-f4b72011f32f548ac5655aec63be4d7fR183 (when the file is open in read-only mode. I don't know if it catch a *no space left* error)

Psycojoker: Yes, but that's another `IOError` actually. Here we make 2 IO operations, one line 189 (opening) and one line 190 (json.dump which writes to the file) so that's another situation that it would be great to handle.

opi: Got it !
I guess we'd like to have to different error message then, for both *read* and *write* errors.

Psycojoker: That would be the best!
Note that the current state is already very good ;)

src/yunohost/app.py
```diff
@@ -806,9 +806,9 @@ def app_makedefault(auth, app, domain=None):
     try:
         with open('/etc/ssowat/conf.json.persistent') as json_conf:
             ssowat_conf = json.loads(str(json_conf.read()))
-    except ValueError:
+    except ValueError as e:
         raise MoulinetteError(errno.EINVAL,
```
zamentur: Why raise an error, why not just a warning and ignore the bad persistent conf ?

opi: Because bad conf is probably a user generated conf, for a good reason, and we souldn't override it. I'd rather raise an error for the user to carefully watch and fix his custom configuration.

Psycojoker:
I think that we should add to the error message "To fix that you need to verify the syntax of the file /etc/ssowat/conf.json.persistant".

I'm calling for a small decision on this one.

opi:
feel free to edit error message to add info about ssowat persistent configuration file. (note that it's "persistent", not "persist**a**nt" ;p )

Psycojoker:
I've made an attempt to do that, probably too verbose.

opi:
Way too verbose indeed ;) i'll try something when back on my laptop (phone is lame for creativity)

Psycojoker:
@opi up :p

opi:
What about "Error while reading/saving SSOwat persistent configuration: {error:s}. Edit /etc/ssowat/conf.json.persistent file to fix the JSON syntax" ?

Psycojoker:
We need another positive opinion.

PR #173, opened 2016-08-12
[fix] ensure that multi_instance key value is always a boolean

Psycojoker:
We've chosen to use boolean everywhere for multi_instance apps, but some app still use string, we need to ensure that we have boolean value everywhere in the mean time (and we'll remove that code in the future).

This is done because for now yunohost-admin expect string and that breaks app that uses a boolean (they aren't detected as multi instance apps by yunohost-admin but ones that uses strings are).

I'm going to do a pr for yunohost-admin too.


opi:
Looks good to me (as a Python newbie)

PR #174, opened 2016-08-16
[fix] Support git ynh app with submodules #533

zamentur:
I tried to use git submodule for assure the iso sources but yunohost doesn't support it with a non github url because the git clone is not recursive.

Since git 2.9 there is an option to force submodule to just download the last commit --shallow-submodules https://git-scm.com/docs/git-clone , but with the current yunohost (debian jessie) git is in version 2.1 so we should wait before adding this option.

This Merge Request is linked to the ticket https://dev.yunohost.org/issues/533

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1259,8 +1259,13 @@ def _fetch_app_from_git(app):
                 url = url[:tree_index]
                 branch = app[tree_index+6:]
             try:
+                # We use currently git 2.1 so we can't use --shallow-submodules
+                # option. When git will be in 2.9 (with the new debian version)
+                # we will be able to use it. Without this option all the history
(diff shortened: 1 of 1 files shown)


Psycojoker:
Looks good to me.

zamentur:
Moul had made an answer about this pr. I need to take time to explore his proposal.

zamentur:
The Moul answer is here https://dev.yunohost.org/issues/533#change-1803

zamentur:
I tried subtrees, it works without this change. But I suggest to merge this contribution to be able to support submodules too.

Psycojoker:
I'm calling for a small decision on this one.

zamentur:
I have merged it (5 positive vote since 5 days)

PR #176, opened 2016-09-12
Mail aliases

julienmalik:
Based on the work from @mbugeia in #58 , rebased onto current unstable.

List of modifications applied after this initial work :
- ou=aliases (instead of ou=alias) to follow ou=users, ou=domains, ...
- Removing alias_update for the moment (one can delete and re-create)
- Main modification : do not make alias entries in LDAP inherit from inetOrgPerson. inetOrgPerson are for people (with first name, surname, ...). Instead, create a specific dedicated mailAlias objectclass. Alias objects in ldap inherits from both mailAlias and mailAccount. mailAlias is the "structural" class for holding the aliases. It enforces that the maildrop be non empty. mailAccount holds the mail address/mail forward addresses. Using mailAccount (as it is done for users) makes it transparent for most downstream applications (postfix, all apps searching for mail accounts in ldap) so that no modification is required for them.
- add support for mail-forward with mail addresses containing a '+' separator
- remove the need to call alias_init : the creation of the ou=aliases in the ldap db is done automatically at first use.


locales/en.json
```diff
@@ -237,5 +237,17 @@
...
+    "pattern_listemail" : "Must be a valid email address list delimited by commas",
+    "alias_init" : "Initializing aliases in LDAP",
+    "alias_created" : "Alias successfully created",
+    "alias_creation_failed" : "Unable to create alias",
+    "alias_deletion_failed" : "Unable to delete alias",
+    "alias_deleted" : "Alias successfully deleted",
+    "alias_unknown" : "Unknown alias",
+    "alias_updated" : "Alias successfully updated",
+    "alias_update_failed" : "Unable to update alias",
+    "alias_info_failed" : "Unable to retrieve alias information"
```
julienmalik: are we required to lexicographically order this file ? or is this done by weblate ?

opi:
could you provide any usecase + examples of command line usage please ?

locales/en.json
Psycojoker: This is a json object, fields are unsorted by default so this won't have an impact.

To sort them, if you handle vim, you can select all the lines of the files excepts the `{}` with V then do `:sort<enter>` (you'll see something like that `'<'>` automatically added after the `:`).

data/actionsmap/yunohost.yml
```diff
@@ -226,6 +226,80 @@ user:
...
+                        required: True
+                        pattern: *pattern_email
+                -f:
+                    full: --mailforward
+                    help: List of email to forward, separated by commas without space
+                    extra:
+                        ask: ask_alias_mailforward
+                        required: True
+                        pattern: &pattern_listemail
+                            - !!str ^([\w.-]+(\+[\w.-]+)*@([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+([^\W\d_]{2,}),?)+$
```
Psycojoker: This regex might prove to be too restrictive, it's basically impossible to validate an email using a regex.

src/yunohost/alias.py
```diff
@@ -0,0 +1,188 @@
...
+    if len(alias['maildrop']) > 0:
+        result_dict['mail-forward'] = alias['maildrop'][0:]
+
+    if result:
+        return result_dict
+    else:
+        raise MoulinetteError(167, m18n.n('alias_info_failed'))
+
+
+def _ensure_ou_created(auth):
```
Psycojoker: Would be more obvious if named something like `_ensure_ldap_ou_is_created`.

That would be a good candidate for the `migration` PR to avoid having to check every time.

src/yunohost/alias.py
```diff
@@ -0,0 +1,188 @@
...
+    rdn = 'ou=aliases'
+    attr_dict = {
+           'objectClass'   : ['organizationalUnit', 'top'],
+    }
+    if not auth.search('dc=yunohost,dc=org', rdn, attr_dict['objectClass']) :
+        if auth.add(rdn, attr_dict):
+            msignals.display(m18n.n('alias_init'), 'success')
+    return
```
Psycojoker: FYI `return` is useless here.

src/yunohost/alias.py
```diff
@@ -0,0 +1,188 @@
...
+    alias_attrs = { 'mail': 'alias',
+                    'maildrop': 'mail-forward' }
+    attrs = []
+    result_list = []
+
+    # Set default arguments values
+    if offset is None:
+        offset = 0
+    if limit is None:
+        limit = 1000
```
Psycojoker: Both `offset` and `limit` can be setted in the function arguments (except if moulinette is doing weird things that I'm not aware of).

Psycojoker: And `filter` too.

src/yunohost/alias.py
julienmalik: I know, just seemed more clear to me when writing it...

src/yunohost/alias.py
```diff
@@ -0,0 +1,188 @@
...
+
+def alias_list(auth, fields=None, filter=None, limit=None, offset=None):
+    """
+    List alias
+
+    Keyword argument:
+        filter -- LDAP filter used to search
+        offset -- Starting number for alias fetching
+        limit -- Maximum number of alias fetched
+        fields -- fields to fetch
```
Psycojoker: I have no idea of what `fields` is supposed to do from this sentence and even from the code. I'm not expecting other people to guess that.

src/yunohost/alias.py
julienmalik: ack

about the migration, yes, but let's modify this when it is implemented. until then, (and actually after too), we don't care at all. we don't want to create millions of aliases in seconds.

src/yunohost/alias.py
```diff
@@ -0,0 +1,188 @@
...
+
+    result = auth.search('ou=aliases,dc=yunohost,dc=org', filter, attrs)
+
+    if len(result) > offset and limit > 0:
+        for alias in result[offset:offset+limit]:
+            entry = {}
+            for attr, values in alias.items():
+                try:
+                    entry[alias_attrs[attr]] = values[0:]
+                except:
```
Psycojoker: Whats the reason for silently hiding errors here? This generally make debugging a nightmare :/

src/yunohost/alias.py
```diff
@@ -0,0 +1,188 @@
...
+        raise MoulinetteError(errno.EINVAL, m18n.n('alias_unknown'))
+
+    result_dict = {
+        'alias': alias['mail'][0]
+    }
+
+    if len(alias['maildrop']) > 0:
+        result_dict['mail-forward'] = alias['maildrop'][0:]
+
+    if result:
```
Psycojoker: `result` is already tested line 161, are you sure that you want to test it here again? The code doesn't seems to work on it.

src/yunohost/alias.py
Psycojoker: That's quite nitpicking but that would be nicer as a function decorator.

src/yunohost/alias.py
julienmalik: this is pure copy-paste from the user command. what's the problem with all three arguments being used together ?

src/yunohost/alias.py
julienmalik: true. pure copy paste from the user command. I don't know who uses that (the web admin ?)

data/actionsmap/yunohost.yml
julienmalik: also mirroring the user command. While I do agree, I think that's the topic of another PR/discussion.

src/yunohost/alias.py
Psycojoker: That's not a problem, it's just that you can avoid 6 lines of code by replacing the function definition by `alias_list(auth, fields=None, filter='(&(objectclass=mailAccount)(objectclass=mailAlias))', limit=1000, offset=0)` which is nicer.

data/actionsmap/yunohost.yml
Psycojoker: Okay, opening a ticket on the redmine.

julienmalik:
In the current state, aliases do not appear in the automatic roundcube address book.
This is because the address book looks in ou=users (https://github.com/YunoHost-Apps/roundcube_ynh/blob/master/conf/config.inc.php#L108)

It may not be desirable to show all aliases here. My opinion is that it is better to NOT show them. Possibly we could add another address book in roundcube with aliases. In which case, we may want to add/manage a special attribute for each alias to tell if they should appear in the address book or not (iRedMail does that)

src/yunohost/alias.py
```diff
@@ -0,0 +1,238 @@
...
+        'uid'           : alias,
+        'mail'          : alias
+    }
+
+    attr_dict['maildrop'] = mailforward.split(",")
+
+    if auth.add(rdn, attr_dict):
+        msignals.display(m18n.n('alias_created'), 'success')
+        return { 'alias' : alias, 'name' : name, 'mailforward' : attr_dict['maildrop'] }
+
```
zamentur: It could be useful to add hooks on each create, delete, update action

uda:
@julienmalik mailing list are not considered aliases, the postfix and other's implementation does it using the existing alias mechanism.
I can think of a ton of features / enhancements to this one... but most are nice to have, not requirements for a group forwarding feature.

julienmalik:
@uda (& all) : please consider contributing your ideas to https://dev.yunohost.org/issues/558

data/templates/slapd/mailserver.schema
```diff
@@ -82,7 +82,19 @@ objectclass ( 1.3.6.1.4.1.40328.1.1.2.2
...
+        AUXILIARY
         DESC 'Mail Group'
         MUST ( mail )
         )
+
+# Mail Alias Objectclass
+objectclass ( 1.3.6.1.4.1.40328.1.2.4.4
+        NAME 'mailAlias'
+        SUP top
+        STRUCTURAL
```
julienmalik: - missing DESC
- after second thought, the existing but unused mailGroup was probably meant to be used for aliases. Look into deleting it ? or reusing it ?

src/yunohost/alias.py
```diff
@@ -0,0 +1,186 @@
...
+    # Adapt values for LDAP
+    rdn = 'mail=%s,ou=aliases' % alias
+    attr_dict = {
+        'objectClass'   : ['mailAccount', 'mailAlias'],
+        'mail'          : alias,
+        'maildrop'      : mailforward.split(",")
+    }
+
+    if not auth.add(rdn, attr_dict):
+        raise MoulinetteError(169, m18n.n('alias_creation_failed'))
```
julienmalik: double check all those errno. they seem arbitrary (probably from copy paste). if not, random numbers in code is wrong, so change it anyway

src/yunohost/alias.py
```diff
@@ -0,0 +1,186 @@
...
+    _ensure_ldap_ou_is_created(auth)
+
+    alias_attrs = [
+        'mail', 'maildrop'
+    ]
+
+    if len(alias.split('@')) is 2:
+        filter = 'mail=' + alias
+    else:
+        # TODO better error message
```
julienmalik: address remaining TODO

mbugeia:
What do you think about adding an "all" option that would create (and automatically update on user create/update/remove) an alias that forward to all user of a yunohost instance ?
That was my use case when I wanted the alias feature.

src/yunohost/alias.py
```diff
@@ -0,0 +1,186 @@
...
+
+from moulinette.core import MoulinetteError
+from yunohost.domain import domain_list
+
+def alias_list(auth):
+    """
+    List aliases
+
+    """
+    _ensure_ldap_ou_is_created(auth)
```
Psycojoker: This is the kind of things we'll need to move in a migration framework (I have proto in a branch for that).

Psycojoker:
Once the todo are cleared I'm okay with this PR :)

src/yunohost/alias.py
julienmalik: Cannot agree more, but before such migration framework is merged, it does the trick quite well/transparently.

julienmalik:
Thanks for the small fixes, all good to me !

julienmalik:
@mbugeia can you raise this point on the redmine issue ?

julienmalik:
All comments/todos addressed.

julienmalik:
The last commit enables catch-all alias, that is, an alias that catches mails sent to any invalid user/alias. This is done by using `@domain.tld` for the alias name.

The commit also relaxes the email pattern, to match "anything with an @ sign, but not at the end or at the beginning". Actually difficult to make it 100% safe, ideas welcome.
Ideally this should better be "anything with a single @ sign, but not at the end or the beginning". Ideas welcome, I don't know how to say that in regex...

For alias, the regex is slightly different, and allows for @ sign to be at the beginning (to support catch-all alias)

julienmalik:
I am officially calling for a merge.

To exercise the process, this is at least a [standard/medium](https://github.com/YunoHost/yunohost-project-organization#d%C3%A9cision-standardmoyenne) decision.

julienmalik:
Please also review the associated [issue](https://dev.yunohost.org/issues/558), if anything comes to mind now.

As of today we are in dev group : juju, Bram, ljf, Moul.
@opi we are waiting for you :)


As an exercise, I computed the official percentage. So if I'm not too wrong :
We need at least 2 devs giving their opinion in two weeks time, 3 to anticipate the merge.
If @opi were with us ;) we would need 3 devs giving their opinion in two weeks time, 4 to anticipate.


In any case I will actually wait for 4 to 5 approvals since it's the beginning :)

Psycojoker:
Since the PR as already been started we can consider that we are only doing a "décalage" of one week (and that people have already review and commented on it).

data/templates/slapd/mailserver.schema
```diff
@@ -80,9 +80,13 @@ objectclass ( 1.3.6.1.4.1.40328.1.1.2.2
             )
         )
 
-# Mail Group Objectclass
```
opi: are we sure nothing use "Mail Group" objectclass ?

data/other/ldap_scheme.yml
```diff
@@ -28,6 +28,12 @@ parents:
             - organizationalUnit
             - top
 
+    ou=aliases:
```
opi: is it enough to create this OU during regen-conf ? (that being said, I could have tested it before asking)

julienmalik: From what I remember when managing this, the ldap_scheme.yml is only used during initial Yunohost installation. not regen-conf.

To create the OU on existing installation, I implemented _ensure_ldap_ou_is_created in the alias.py and this does the trick. So it is actually created on first use, at the cost of an additional small ldap request.

Bram already commented that this should ultimately be managed in a migration management framework. I concur.

data/templates/slapd/mailserver.schema
opi: :)

data/other/ldap_scheme.yml
opi: you're right. Yesterday I read _ensure_ldap_ou_is_created function too quickly. thanks

julienmalik:
ping : seems reviewed & ok, but I would prefer to have formal approval, and only @Psycojoker gave it

data/actionsmap/yunohost.yml
```diff
@@ -226,6 +226,97 @@ user:
...
+    category_help: Manage alias
+    actions:
+
+        ### alias_list()
+        list:
+            action_help: List existing aliases
+            api: GET /alias
+            configuration:
+                authenticate: all
+                authenticator: ldap-anonymous
```
opi: First time this function is called (by `yunohost alias list`), even after a `regen-conf` of slapd, the anonymous ldap connection breaks `_ensure_ldap_ou_is_created`.
See this log: https://paste.yunohost.org/cusucofise.vala

data/actionsmap/yunohost.yml
```diff
@@ -226,6 +226,97 @@ user:
...
+                authenticate: all
+            arguments:
+                alias:
+                    help: Alias to update
+                    metavar: ALIAS
+                    extra:
+                        required: True
+                        pattern: *pattern_alias
+                --add-mailforward:
+                    help: List of mailforward addresses to add, separated by commas without space
```
opi: actually `--add-mailforward` accpets a space separated list of email, which breaks the ability to do `yunohost alias update --add-mailforward forward@domain.com existing-alias@domain.com`, because both addresses are considered as mailforward

julienmalik: thanks for the report, i need to investigate this.

data/actionsmap/yunohost.yml
opi: removing "authenticator: ldap-anonymous" is fine for me, but it can be painful for a cli administrator to enter password each time just to list aliases

julienmalik: ok. doable actually : alias_info and alias_list should *not* call _ensure_ldap_ou_is_created, and just gracefully exit if the ou is not created (alias_list should simply return an empty list, alias_info should output the same error as if the admin requested info on a non-existing alias).

thanks for catching this.

alexAubin:
This PR is opened for quite a while. If we need feedback on this, I can test it on my side. Otherwise if we keep this for later I suggest to flag it as "postponed" ? (But maybe put it in milestone 2.6.x or something, not in Horizon, so that we don't forget about it :P)

Psycojoker:
@alexAubin imo, right now we should focus on fixing bugs in the testing branch before looking at new features, this way we can do a stable release then we can move back on cool new features :)

julienmalik:
@alexAubin agreed with @Psycojoker let's release with LE, then move on to the next feature. I am currently busy with other stuff for yunohost, but I'm not forgetting this. I really want this to happen :)

The two points raised by @opi are real bugs that need to be fixed

mbugeia:
@julienmalik Do you think it would be a good idea to attach an user to a created alias ? This way we could have admin alias (for administrative stuff like all@domain.tld) and per user alias that could be edited in SSOwat (like current alias système).

julienmalik:
@mbugeia  With the current implementation, you can already attach an existing yunohost user email to an alias (or to several). just use --add-mailforward theuser@maindomain.tld

julienmalik:
w.r.t ssowat, you make me think that I did not fully test the overlap between the alias defined this way and the alias defined in ssowat. if only we could write unit tests :)

mbugeia:
Oh I was talking about an user that can modify the alias in yunohost. In the current implementation, only yunohost administrator can create alias, I think it would be great if each user can create one too.

src/yunohost/alias.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # Adapt values for LDAP
+    rdn = 'mail=%s,ou=aliases' % alias
+    attr_dict = {
+        'objectClass': ['mailAccount', 'mailAlias'],
+        'mail': alias,
+        'maildrop': mailforward.split(",")
+    }
+
+    success = auth.add(rdn, attr_dict)
```
julienmalik: this ensures that no two global aliases can have the same id.
but we must make sure there is also no conflict with the per-user aliases (those defined in ssowat).

also ssowat and the user command should check conflict with the alias command.

alexAubin:
So from what I understood, talking about this quickly during today's meeting, this is still not ready to be reviewed / need some work. Therefore, reflaging this as "work needed"

alexAubin:
[Semi-automatic message] 

PR #177, opened 2016-09-18
dns-conf is broken for subdomains

Psycojoker:
Gandi refuses it with a '.'

### DIFF
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -273,7 +273,7 @@ def domain_dns_conf(domain, ttl=None):
             '(?=.*(;[\s]*|")p=(?P<p>[^";]+))'), dkim_content, re.M|re.S
         )
         if dkim:
-            result += '\n{host}. {ttl} IN TXT "v={v}; k={k}; p={p}"'.format(
+            result += '\n{host} {ttl} IN TXT "v={v}; k={k}; p={p}"'.format(
                 host='{0}.{1}'.format(dkim.group('host'), domain), ttl=ttl,
                 v=dkim.group('v'), k=dkim.group('k'), p=dkim.group('p')
             )


src/yunohost/domain.py
```diff
@@ -273,7 +273,7 @@ def domain_dns_conf(domain, ttl=None):
             '(?=.*(;[\s]*|")p=(?P<p>[^";]+))'), dkim_content, re.M|re.S
         )
         if dkim:
-            result += '\n{host}. {ttl} IN TXT "v={v}; k={k}; p={p}"'.format(
+            result += '\n{host} {ttl} IN TXT "v={v}; k={k}; p={p}"'.format(
```
opi: By removing the last dot you decide to use the "sub-domain notation", so `{host}` must not contain full domain, but only subdomain (`dkim.group('host')`).

Psycojoker: Are you sure?

Tbh we've tried to do that on the domain name of a member and gandi rejected the file telling that the syntax was invalid, I didn't had the time to dig this. I'm at the work but I'll try to remember to check the domain for which we did that but from what I understand we'll probably need to handle both full domain and subdomain case no?

opi: Here's 2 copy/paste:

- From Gandi
```
mail._domainkey 1800 IN TXT "v=DKIM1; k=rsa; p=XXXX"
_dmarc 1800 IN TXT "v=DMARC1; p=none"
```

- From BookMyName
```
mail._domainkey           900  TXT    "v=DKIM1; k=rsa; p=XXX"
_dmarc                    900  TXT    "v=DMARC1; p=none"
```

The only difference is the keyword `IN` (which is 'record class', always IN for INternet), but syntax is the same, only subdomain without trailing period.

Quote from wikipedia : "host names that do not end in a period are relative to the zone origin" (https://en.wikipedia.org/wiki/Zone_file#File_format)

To be really nerdy, we can read corresponding RFC: https://tools.ietf.org/html/rfc1034#page-8

Psycojoker:
This PR should be modified, this command simply doesn't work for subdomains.

Psycojoker:
Yes, this can be address by parsing the domain for now.

 TLD like ".co.uk" will be annoying to parse, do someone know if we can we easily get a list of all available tld (ideally GET on something somewhere since now people can add more) to grab the list of annoying TLD with 2 "." in them?

opi:
I'm not sure that we (yunohost) have to know if a domain is "a domain" or "a subdomain". Can't we find a way that dns-conf output is nice in any case ? btw, what is wrong by now ? ;)

Psycojoker:
For now, we output in all cases something like this (without ipv6 here):

```dns
@ 3600 IN A 123.123.123.123
* 3600 IN A 123.123.123.123

_xmpp-client._tcp 3600 IN SRV 0 5 5222 domain.tld.
_xmpp-server._tcp 3600 IN SRV 0 5 5269 domain.tld.
muc 3600 IN CNAME @
pubsub 3600 IN CNAME @
vjud 3600 IN CNAME @

@ 3600 IN MX 10 domain.tld.
@ 3600 IN TXT "v=spf1 a mx ip4:123.123.123.123 -all"
mail._domainkey.browny.pink 3600 IN TXT "v=DKIM1; k=rsa; p=key"
_dmarc 3600 IN TXT "v=DMARC1; p=none"
```

While for a subdomain we should output something like this:

```dns
subdomain 3600 IN A 123.123.123.123
*.subdomain 3600 IN A 123.123.123.123

# here I don't know if we can show this block
_xmpp-client._tcp.subdomain 3600 IN SRV 0 5 5222 domain.tld.
_xmpp-server._tcp.subdomain 3600 IN SRV 0 5 5269 domain.tld.
muc 3600 IN CNAME subdomain
pubsub 3600 IN CNAME subdomain
vjud 3600 IN CNAME subdomain

subdomain 3600 IN MX 10 domain.tld.
subdomain 3600 IN TXT "v=spf1 a mx ip4:123.123.123.123 -all"

mail._domainkey.subdomain. 3600 IN TXT "v=DKIM1; k=rsa; p=key"
_dmarc.subdomain. 3600 IN TXT "v=DMARC1; p=none"
```

**There are probably errors in this second example but that's to give you an idea on what is different from a topdomain**.

opi:
is something like this would work in any case ?

```
domain.tld 3600 IN A 123.123.123.123
*.domain.tld 3600 IN A 123.123.123.123

# here I don't know if we can show this block
_xmpp-client._tcp.domain.tld. 3600 IN SRV 0 5 5222 domain.tld.
_xmpp-server._tcp.domain.tld. 3600 IN SRV 0 5 5269 domain.tld.
muc.domain.tld. 3600 IN CNAME domain.tld
pubsub.domain.tld. 3600 IN CNAME domain.tld
vjud.domain.tld. 3600 IN CNAME domain.tld

domain.tld. 3600 IN MX 10 domain.tld.
domain.tld. 3600 IN TXT "v=spf1 a mx ip4:123.123.123.123 -all"

mail._domainkey.domain.tld. 3600 IN TXT "v=DKIM1; k=rsa; p=key"
_dmarc.domain.tld. 3600 IN TXT "v=DMARC1; p=none"
```

( note the trailing dot, which tell that it is a fqdn )

opi:
my bad, we should handle the case where user ask for dns-conf for a subdomain which main domain is already on yunohost. and then only have xmpp/mail related stuff (and not the A/Wildcard records)

Psycojoker:
I can't find it again but I've end up on a website saying that for dmarc you should write `_dmarc.subdomain.` without things after that but that might be crap.

Regarding your proposition, I suck too much at DNS to confirm that for you :/

Do anyone have a better understanding of that?

julienmalik:
See this discussion : https://lists.tetaneutral.net/pipermail/technique/2016-October/002478.html for the use of $ORIGIN which might help in our case.

At least the Gandi interface says 
```
This interface is for viewing and/or editing the zone file
 in the  format used by BIND version 9.
 To avoid frequent errors, we suggest
 that  you consult the following guide: 
https://en.wikipedia.org/wiki/Zone_file.
```
which suggests that the $ORIGIN trick probably work there. Don't know about other registrars (I doubt we can find something which works for all of them anyway)

Psycojoker:
It would be great if we could fix this but I'm not motivated :(

alexAubin:
I don't know much about DNS and not sure to understand the $ORIGIN trick, but we might as well just have an option for the user to tell wether this domain should be understood as a sub-domain (DNS-wise).

On the admin interface, this could be something as having a checkbox on the "dns configuration" page, saying something like : 

>toto.jaimelecafe.com
> ----------------------
>[ ] Check this case if you also control jaimelecafe.com and want to handle the DNS configuration toto.jaimelecafe.com as a subdomain of jaimelecafe.com

alexAubin:
[Semi-automatic message] 

PR #179, opened 2016-10-14
Fix backup when /home/yunohost.backup/archives is on a different device

JimboJoe:
if /home/yunohost.backup/archives is on a different device (for instance, a NAS), backups fail with the message:
...
 File "/usr/lib/moulinette/yunohost/backup.py", line 303, in backup_create
    '{:s}/{:s}.info.json'.format(archives_path, name))
OSError: [Errno 18] Invalid cross-device link

A simple solution is to use shutil.move because, as stated in the documentation, "If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied (using shutil.copy2()) to dst and then removed.", which is the needed behavior.
This PR has been successfully tested.

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -299,7 +299,7 @@ def backup_create(name=None, description=None, output_directory=None,
                                   m18n.n('backup_creation_failed'))
 
         # Move info file
-        os.rename(tmp_dir + '/info.json',
+        shutil.move(tmp_dir + '/info.json',
                   '{:s}/{:s}.info.json'.format(archives_path, name))
 
     # Clean temporary directory


mbugeia:
This should be merged, shutil is the proper way.

Psycojoker:
I didn't wanted to re-open too much PR at the same time but I'm ok for this one too (considerer minor).

Psycojoker:
It's been one week but we miss an opinion. @zamentur @julienmalik @opi could you give one please?

zamentur:
May be this fix this issue too : https://dev.yunohost.org/issues/618

JimboJoe:
I don't think so, the description explicitly says "note : changing line 305 by shutil.move doesn't change.".

PR #180, opened 2016-10-28
[enh] Certificate management integration (e.g. Let's Encrypt ...)

alexAubin:
Hi guys,

here's a first proposition of integration of Let's Encrypt (and more generally certificate management).

It adds the following features : 

**Check certificate status**

```
usage: yunohost domain cert-status [-h] [--full] [domainList [domainList ...]]

positional arguments:
  domainList  Domains to check

optional arguments:
  -h, --help  show this help message and exit
  --full      Show more details
```

**Install Let's Encrypt certificates**

```
usage: yunohost domain cert-install [-h] [--force] [--no-checks]
                                    [--self-signed]
                                    [domainList [domainList ...]]

positional arguments:
  domainList     Domains for which to install the certificates

optional arguments:
  -h, --help     show this help message and exit
  --force        Install even if current certificate is not self-signed
  --no-checks    Does not perform any check that your domain seems correcly
                 configured (DNS, reachability) before attempting to install.
                 (Not recommended)
  --self-signed  Install self-signed certificate instead of Let's Encrypt
```

**Renew Let's Encrypt certificate**

```
usage: yunohost domain cert-renew [-h] [--force] [--email] [--no-checks]
                                  [domainList [domainList ...]]

positional arguments:
  domainList   Domains for which to renew the certificates

optional arguments:
  -h, --help   show this help message and exit
  --force      Ignore the validity treshold (30 days)
  --email      Send an email to root with logs if some renewing fails
  --no-checks  Does not perform any check that your domain seems correcly
               configured (DNS, reachability) before attempting to renew. (Not
               recommended)
```

**Some more details / explanations**
- Let's Encrypt certificates are managed using [acme-tiny](https://github.com/diafygi/acme-tiny). The choice was driven by the fact that it is a lightweight module, somewhat easy to use, and can be installed through `pip install acme-tiny` ;
- I edited the `ssowatconf` function to automatically authorize the ACME challenge uri `/.well-known/blah`if it is present in the nginx conf ;
- `cert-install` automatically adds a small cron-job (2 lines of code), calling `yunohost domain cert-renew --email` every week, which will renew Let's Encrypt certificates which are about to expire, and send an email to `root` if some renewing fails ;
- By default, `cert-install` and `cert-renew` include some checks that the domain seems to have a reasonable setup to attempt Let's Encrypt certificate installation (DNS poiting to global IP, + being able to GET a page) ;
- I moved the generation of self-signed certificate into a new file, and `cert-install` can be used to regenerate a self-signed certificate (in case you don't want Let's Encrypt anymore for some reason). The short/mid-term goal is also to refactorize this part of code to get rid of `os.system()` calls and use the OpenSSL.crypto module instead.

There are more features we can think about including in the future, such as supporting other ACME authorities, supporting DNS-based challenge, set level of security (e.g. deactivate TLS v1.0 and 1.1), import certificates from non-ACME authorities, provide a diagnosis for domain (e.g. inform user a Let's Encrypt cert can be installed, display current SSL Labs rating) etc... But for now, let's focus on the basic stuff :) 

**Remaining work**
- How do we manage acme-tiny (and python-tabulate) as dependencies. At the moment this code does not work if you didn't do a `pip install acme-tiny` and `pip install tabulate` for example.
- How do we make sure the transition from letsencrypt_ynh is smooth
- ???

Cheers !


M5oul:
```bash
sudo yunohost domain -h
    cert-install        Install Let's Encrypt certificates for given domains
                        (all by default).
    cert-status         List status of current certificates (all by default).
    list                List domains
    remove              Delete domains
    add                 Create a custom domain
    dns-conf            Generate DNS configuration for a domain
    cert-renew          Renew the Let's Encrypt certificates for given domains
                        (all by default).
```

Could `cert-*` sub-commands be grouped?
On the actionmap, there are all followed.
Why do they appear in this order?

M5oul:
>How do we manage acme-tiny (and python-tabulate) as dependencies. At the moment this code does not work if you didn't do a pip install acme-tiny and pip install tabulate for example.

As said on the chat room, there is three solutions:

##### Debian packages
That solution would be the best choice.
[tabulate](https://packages.debian.org/search?keywords=python-tabulate) and [acme-tiny](https://packages.debian.org/search?searchon=names&keywords=acme-tiny) Debian packages are available on Jessie backport and Debian Stretch.
May be we could install them using [this](https://serverfault.com/questions/22414/how-can-i-run-debian-stable-but-install-some-packages-from-testing) or [this](https://unix.stackexchange.com/questions/107689/how-to-install-a-single-jessie-package-on-wheezy) solutions.

##### Subtrees
For now, we could add them as subtrees from respective repositories: [tabulate](https://bitbucket.org/astanin/python-tabulate/overview) and [acme-tiny](https://github.com/diafygi/acme-tiny).
They need to move them to right place.

##### Pip
It's kind of ugly solution:
```bash
sudo pip install tabulate acme-tiny
```

----
I suggest to use subtrees solution for now, and Debian packages on Debian Stretch.

src/yunohost/certificate.py
```diff
@@ -0,0 +1,641 @@
...
+
+
+
+###############################################################################
+#   Back-end stuff                                                            #
+###############################################################################
+
+def _install_cron() :
+
+    cron_job_file = "/etc/cron.weekly/certificateRenewer"
```
opi: Maybe rename cron file to something more consistent with other yunohost cron. Currently we have `/etc/cron.d/yunohost-applist-yunohost`,  `/etc/cron.d/yunohost-firewall` and `/etc/cron.d/yunohost-dyndns`.
I don't know if it worth to move this cron to `/etc/cron.d` as well.

I vote for `/etc/cron.d/yunohost-certificate-renew` ;)

src/yunohost/certificate.py
```diff
@@ -0,0 +1,641 @@
...
+        raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_domain_dns_ip_differs_from_public_ip', domain=domain))
+
+    # Check if domain seems to be accessible through HTTP ?
+    if not _domain_is_accessible_through_HTTP(public_ip, domain) :
+        raise MoulinetteError(errno.EINVAL,  m18n.n('certmanager_domain_http_not_working', domain=domain))
+        
+def _dns_ip_match_public_ip(public_ip, domain) :
+
+    try :
+        r = requests.get("http://dns-api.org/A/"+domain)
```
opi: Please use https url ;)

opi: I don't know if using 'dnspython' lib is easy, but I'd like to avoid using a third party API. Note that 'dnspython' is package in Debian's `python-dns`.

alexAubin:
@M5oul 
> Could cert-* sub-commands be grouped?
> On the actionmap, there are all followed.
> Why do they appear in this order?

That seems related to the yaml file being loaded in a dictionnary in moulinette ? See [here](https://github.com/YunoHost/moulinette/blob/unstable/generate_function_doc.py#L61) ?

src/yunohost/certificate.py
```diff
@@ -0,0 +1,641 @@
...
+        dns_ip = json.loads(r.text)[0]["value"]
+    except :
+        raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_error_parsing_dns', domain=domain, value=r.text))
+
+    if (dns_ip != public_ip) :
+        return False
+    else :
+        return True
+
+def _domain_is_accessible_through_HTTP(ip, domain) :
```
Psycojoker: This test looks really weak, I don't think that this is the good way to do that but I don't have any other in mind right now.

Psycojoker:
Regarding dependancies:

* `sudo pip install` is a total no-go for me
* if we don't have any other solution, including a copy of acme-tiny inside yunohost temporary (until we build or there is a debian package we/debian distribute) seems acceptable
* we don't need python-tabulate because this breaks yunohost-api that expect json to be usable on the admin interface (I'm going to push a fix to remove that), not an ascii table

Psycojoker:
As a reference, the things discusted in the previous meeting:

- intégration acme-tiny
- taille de clé: 2048 –> 4096
- [TODO/Bram] aller parler à des gens qui aiment la sécu pour voir si ça vaut la peine effectivement d'augmenter cette taille de clef
- Référence: https://github.com/alexAubin/letsencrypt_ynh/issues/2
- utiliser dns-python au lieu de la requêtte sur l’API du service externe
- gestion de la migration depuis letsencrypt_ynh : remove de l’app (en vérifiant qu’il s’agit de la dernière version de l’app qui contient la dernière version du script de remove)
- lors du check http, vérifier qu'on a bien le header X-SSO-WAT dans la réponse pour éviter de se taper les interfaces web des boxs 
- problème du rate limit sur les .nohost.me et .noho.st https://letsencrypt.org/docs/rate-limits/ https://github.com/publicsuffix/list/pull/250

alexduf:
(sorry I'm just a random guy interested in that PR)

2048 bits is good enough until roughly 2020.
The 2014 ANSSI recommendations state that for a key that lives after that date, you'd better have a key length of 3072 bits.
given the fact that the certificate rotates every three months with Let's Encrypt, I think it's safe to use 2048 for the next few years.

Now it doesn't really hurt to have a bigger key for low traffic sites so 3072 may be a good idea.

Here's more detailed information: https://www.keylength.com/en/8/
and the French security agency recommendations: http://www.ssi.gouv.fr/

alexAubin:
@alexduf Thanks for the feedback on this ! Do you think there's any drawback in using directly a 4096 bits key ? I can't think of anything except the impact on CPU, which from what I understood is completely negligible.

mbugeia:
@opi  for the CPU concern it does not affect only the server but also the client, so it will drain the battery faster on a mobile.

I'm for using 2048 bits keys since the key exchange is slower with larger keys and this will slow down page loading on the client.

opi:
> The 2014 ANSSI recommendations state that for a key that lives after that date, you'd better have a > key length of 3072 bits.
> given the fact that the certificate rotates every three months with Let's Encrypt, I think it's safe to use > 2048 for the next few years.

This 2048-or-more key is only used with self-signed certificates, which are valid for 10 years. I think we should follow ANSSI recommandation and use a 3072 key.

julienmalik:
@opi you meant 2048 i guess ?

opi:
@julienmalik nope

Psycojoker:
Mozilla appears to recommend 2048 https://wiki.mozilla.org/Security/Server_Side_TLS

On my side I'm kinda in a "like for cake, the bigger the better" but that might not be the smartest approach here.

M5oul:
I am testing it, but I have this issue:
```bash
yunohost domain cert-status
  File "/usr/bin/yunohost", line 212, in <module>
    password=opts.password, parser_kwargs={'top_parser': parser}
  File "/usr/lib/python2.7/dist-packages/moulinette/__init__.py", line 138, in cli
    moulinette.run(args, output_as=output_as, password=password)
  File "/usr/lib/python2.7/dist-packages/moulinette/interfaces/cli.py", line 352, in run
    ret = self.actionsmap.process(args, timeout=5)
  File "/usr/lib/python2.7/dist-packages/moulinette/actionsmap.py", line 467, in process
    return func(**arguments)
TypeError: domain_cert_status() got an unexpected keyword argument 'domain_list'
```

alexAubin:
Indeed, must have forgot to propagate the change in the actionmap at some point. Last commit should fix it !

M5oul:
Nice, it works very well.
It's magic, it's almost immediate :smile: 
Great work :+1:

M5oul:
Status output is a bit ugly in one-liner json format.
`tabulate` module could have been a good idea.

mbugeia:
@M5oul Does the json output really need to be readable by human ?
`whatever-that-output-json | python -m json.tool` would do the trick either way.

M5oul:
> Does the json output really need to be readable by human ?

It's great to easily read the status of certificates.

Yep, this python trick is great.

Finally, certificate management should be integrated on the web admin.

alexAubin:
Personally I would find it convenient to have a human readable output (which does not require any additional trickery) for people who tend to administrate their server in cli rather than through the web interface. (I don't really care if it's using `python-tabulate` or not though, since it seems difficult to integrate this dependency). 

From what I understood, returning the raw json was to allow the integration with the web interface ... though to me it seems weird that we cannot have both a) a command or option to get something human-readable, and  b) a command or option to get something parsable by the web interface ?

mbugeia:
Isn't the default output of yunohost command not yaml like in `yunohost user list` ? Adn --json to output json

alexAubin:
> Isn't the default output of yunohost command not yaml like in yunohost user list ? Adn --json to output json

Eh, you're right, only needed to tweak the return value a bit (see last commit). Now I get an output similar to `yunohost user list` :smile:. Thanks !

likeitneverwentaway:
Can you add an option so that we can choose who the certificate is "delivered to" ? That way we could have our name (instead of the domain as of right now) displayed when checking the certificate with a browser. Thanks!

alexAubin:
Not sure to know what you mean ? Are you talking about the CN (Common Name) ? I think this one must be the domain url for the certificate to work. We might be able to choose the "Organization" though, which for now is not set.

See [this certificate for github.com](http://i.imgur.com/noaygN9.png) for instance

We could have something like `yunohost domain cert-install --organization="Super Awesome Corp"`. Not sure to know if that really matters though :stuck_out_tongue:

likeitneverwentaway:
Yes that's what I was referring to, I know a few people who set it up with their name for their blog. Sure it's not that important, don't bother 😛

opi:
Let'sEncrypt doesn't allow to define the Organization field

M5oul:
### Procedure I followed to test it
- Setup a Debian [DigitalOcean](https://www.digitalocean.com) VM.
- Setup on the DNS zone of a registrar `AAAA` and/or `A` records which point to this VM:
 - This record is mandatory, as LE won't works.
 - With OVH, if it's a new record, it's almost immediate. With other registrars like Gandi, it's one hour.
 - If you are in second case, you may use `*.nohost.me` records which takes almost 5 min.
- On the VM, [install YunoHost with the install_script](https://github.com/YunoHost/install_script#basic-usage).
- Setup [`ynh-dev` tool](https://github.com/YunoHost/ynh-dev#21-setup):
 ```bash
 apt-get install git
 wget https://github.com/YunoHost/ynh-dev/raw/master/ynh-dev
 bash ynh-dev create-env /vagrant
 ```

- Use `yunohost` git repository with [Let's Encrypt integration](https://github.com/alexAubin/yunohost/tree/certmanager):
 - Hardcore way:
 ```bash
 cd /vagrant
 rm -rf yunohost
 git clone https://github.com/AlexAubin/yunohost -b certmanager
 bash ynh-dev use-git yunohost
 ```
 - git light way:
 ```bash
 cd /vagrant/yunohost
 git remote add LE https://github.com/AlexAubin/yunohost
 git pull LE certmanager
 cd ..
 bash ynh-dev use-git yunohost
 ```

- Now, the YunoHost instance is ready to test Let's Encrypt:
 ```bash
 yunohost domain -h
 ```

Have fun :smiley_cat:

alexAubin:
Regarding the transition from the old letsencrypt_ynh to the new feature, I added a check that the letsencrypt app is installed. If it is, it blocks all the new commands (cert-status, cert-install, cert-renew) and displays a warning/error message to the user, and suggests instructions : 
```
 > yunohost domain cert-status
Warning:  
Warning: Yunohost detected that the 'letsencrypt' app is installed, 
Warning: which conflits with the new certificate management features
Warning: directly integrated in Yunohost. If you wish to use these  
Warning: new features, please run the following commands to migrate 
Warning: your installation :
Warning:  
Warning:    yunohost app remove letsencrypt
Warning:    yunohost domain cert-install
Warning:  
Warning: N.B. : this will attempt to re-install certificates for    
Warning: all domains with a Let's Encrypt certificate or self-signed
Warning: certificate.
Warning:  
Error: Command aborted because the letsencrypt app is conflicting with the yunohost certificate management features.
```
I couldn't come up with any better solution for now... In particular I don't think it is a good idea to do this automatically - the uninstall of the letsencrypt app should really be performed manually by the user. I just hope the remove script won't break anything badly on particular setups :stuck_out_tongue:.

Feel free to suggest a better solution or message to display.

alexAubin:
The previous commits should address the remaining todo, namely : 
- Don't rely on an external API to check the DNS A record / Use dnspython instead.
- Check for the presence of the ssowat header when checking server is accessible through http
- Use 3072 bit keys

(We might want to add the python-dnspython packet as a dependency, regarding the first bullet).

Feel free to test all the stuff using the nice instructions from @M5oul, or/and speak up about some feature you think should go in that first round ! :smile_cat:

alexAubin:
I cleaned a bit the way the different certificates (or the history of it) for a given domain are stored in `/etc/yunohost/certs/`.
- Now each time a certificate is added (self-signed or lets-encrypt), it goes in `domain.tld-history/`
- `domain.tld/` is a symlink pointing to the current/active certificate in `domain.tld-history/`
- If a new cert is created, but `domain.tld/` was not a symlink, we backup what it was to a new folder `domain.tld-backups/`. This covers current installs (i.e. to ensure smooth migration) or cases where the user tweaked stuff manually (to install a cert from another authority for example).

See picture below : 


That way, if anything bad happens, the user can still (manually at least) revert to an old certificate since nothing is deleted. On the mid/long-term, we could have a option to purge certificates which expired though.

src/yunohost/certificate.py
```diff
@@ -0,0 +1,717 @@
...
+        _enable_certificate(domain, new_cert_folder)
+
+        # Check new status indicate a recently created self-signed certificate,
+        status = _get_status(domain)
+
+        if status and status["CA_type"]["code"] == "self-signed" and status["validity"] > 3648:
+            logger.success(m18n.n("certmanager_cert_install_success_selfsigned", domain=domain))
+        else :
+            logger.error("Installation of self-signed certificate installation for %s failed !", domain)
+            logger.error(str(e))
```
Psycojoker: "e" is undeclared here, what is it supposed to be?

Having the reason of failure seems a really important information to avoid nightmare debugging.

src/yunohost/certificate.py
```diff
@@ -0,0 +1,717 @@
...
+
+        shutil.copyfile(os.path.join(ssl_dir, "openssl.cnf"), os.path.join(new_cert_folder, "openssl.cnf"))
+
+        # FIXME : should refactor this to avoid so many os.system() calls...
+        # We should be able to do all this using OpenSSL.crypto and os/shutil
+        command_list = [
+            'sed -i "s/yunohost.org/%s/g" %s/openssl.cnf' % (domain, new_cert_folder),
+            'openssl req -new -config %s/openssl.cnf -days 3650 -out %s/certs/yunohost_csr.pem -keyout %s/certs/yunohost_key.pem -nodes -batch 2>/dev/null'
+            % (new_cert_folder, ssl_dir, ssl_dir),
+            'openssl ca -config %s/openssl.cnf -days 3650 -in %s/certs/yunohost_csr.pem -out %s/certs/yunohost_crt.pem -batch 2>/dev/null'
```
Psycojoker: Instead of hidding stderr we should probably store them somewhere and show them if the command failed (and log them somewhere anyway), otherwise this would make debugging really hard when something bad happen.

src/yunohost/certificate.py
```diff
@@ -0,0 +1,717 @@
...
+
+def _backup_current_cert(domain):
+    logger.info("Backuping existing certificate for domain %s", domain)
+
+    cert_folder_domain = os.path.join(CERT_FOLDER, domain)
+
+    date_tag = datetime.now().strftime("%Y%m%d.%H%M%S")
+    backup_folder = "%s-backups/%s" % (cert_folder_domain, date_tag)
+
+    shutil.copytree(cert_folder_domain, backup_folder)
```
Psycojoker: Does this switch to a subdirectory works? Shouldn't we do a makedir before?

alexAubin: Yup, `shutil.copytree()` creates the destination directories if needed

src/yunohost/certificate.py
```diff
@@ -656,29 +658,14 @@ def _check_domain_is_correctly_configured(domain):
...
-        raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_no_A_dns_record', domain=domain))
-
-    if "value" not in dns_ip:
-        raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_error_parsing_dns', domain=domain, value=result.text))
-
-    dns_ip = dns_ip["value"]
+        resolver = dns.resolver.Resolver()
+        # These are FDN's DNS
+        resolver.nameservers = [ "80.67.169.12", "80.67.169.40" ]
+        answers = resolver.query(domain, "A")
```
M5oul: Why only DNS `A` record is checked?
IPv6 (`AAAA`) is the future. Why not checking IP6?
Checking one of them is enough.

src/yunohost/certificate.py
alexAubin: As suggested on the chat, I used subprocess to execute the command and grab stdout/stderr, which is now displayed as info if the return code is 0, or as a warning otherwise. So it's here in the logs, and displayed to the user in case of error :smiley:.

src/yunohost/certificate.py
alexAubin: That was a typo from copy-pasting, I removed the line. Unfortunately, I don't think there's much more info we can give to the user, since it's just a sanity check that the new current cert is a valid self-signed cert...

alexAubin:
I added more DNS resolvers from the link you suggested. The resolver.query() appears to correctly fallback to the other resolvers in the list if the first one is down (or a wrong dummy IP for instance). 

Regarding the use of `/etc/resolv.conf`, on my install it has been automatically configured to 127.0.0.1, which is probably related to dnsmasq ? When you create a dummy domain, it actually creates a dns record in its database (?) and will answer to query such as "host my.dummy.domain" with my current IP - despite the fact that it is not actually registered on an external/real DNS. So, not what we want :stuck_out_tongue:.

src/yunohost/certificate.py
M5oul: I think it's great to handle it right now as some people can only use IPv6 and won't benefit LE.
You could easily have an IPv6's VM using DigitalOcean. Couldn't you get an IPv6's VM working?

Psycojoker: Tbh, this will probably be like 1% of our users at most...

alexduf: _(sorry again, still the random guy interested in that PR)_

ipv6 is roughly 10% of web traffic (a bit more according to [google](https://www.google.com/intl/en/ipv6/statistics.html))

That being said, shouldn't the check be `A` or `AAAA` rather than `A` only or `AAAA` only?

julienmalik: I am not sure whether Yunohost has already been tested on IPv6-only hosts.

But more important, related to LE : last time I checked (few months ago) LE did not support doing the challenge via IPv6, so it was not possible to set up a LE certificate on an IPv6-only host. In which case, leaving an IPv4-only check here probably makes lot of sense, since this would prevent user from attempting to generate a cert for an IPv6 only host that LE does not support

alexAubin: Apparently this summer they announced [full support for IPv6](https://letsencrypt.org/2016/07/26/full-ipv6-support.html)

julienmalik: oops I'm not up-to-date : https://letsencrypt.org/2016/07/26/full-ipv6-support.html 
seems like LE does support IPv6-only hosts

Psycojoker:
I think that we are missing the `python-dnspython` dependancy in `debian/control` file for this PR to work.

opi:
Note that new dependency in "Depends" section seems to be installed only when doing a `apt dist-upgrade`.

opi:
great, `yunohost tools upgrade` is similar to `apt-get dist-upgrade`, see https://github.com/YunoHost/yunohost/blob/testing/src/yunohost/tools.py#L368

( See Python's APT  `cache.upgrade(True)` documentation: https://apt.alioth.debian.org/python-apt-doc-exp/library/apt.cache.html#apt.cache.Cache.upgrade )

So basically, we're good for every people who upgrade their yunohost from web administration or `yunohost tools upgrade` command line. We just have to remember that trick, to prevent writing any `apt-get upgrade` on our release documentation.

src/yunohost/certificate.py
M5oul: It would be great to add a comment on [IPv6-only ticket](https://dev.yunohost.org/issues/229).
In fact, It could be done later.

opi:
@Psycojoker `python-dsnpyton` is required since https://github.com/YunoHost/yunohost/commit/5d800e6b99c6115b7a5a5568880994f35099d48d , for more than 2 years from now. We can consider this dependnecy to be met now.

src/yunohost/certificate.py
```diff
@@ -0,0 +1,751 @@
...
+    cert_file = os.path.join(CERT_FOLDER, domain, "crt.pem")
+
+    if not os.path.isfile(cert_file):
+        return {}
+
+    try:
+        cert = crypto.load_certificate(crypto.FILETYPE_PEM, open(cert_file).read())
+    except Exception as exception:
+        import traceback
+        traceback.print_exc(file=sys.stdout)
```
julienmalik: same comment as before, this should probably be factorized in a function

src/yunohost/certificate.py
```diff
@@ -0,0 +1,751 @@
...
+            "verbose": "Self-signed",
+        }
+
+    elif cert_issuer.startswith("Let's Encrypt"):
+        CA_type = {
+            "code": "lets-encrypt",
+            "verbose": "Let's Encrypt",
+        }
+
+    elif cert_issuer.startswith("Fake LE"):
```
julienmalik: what's this ? could you add a small comment ?

Psycojoker:
@opi yup, realised it later, forgot to update my comment :)

src/yunohost/certificate.py
```diff
@@ -0,0 +1,751 @@
...
+        raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_error_no_A_record', domain=domain))
+
+    dns_ip = str(answers[0])
+
+    return dns_ip == public_ip
+
+
+def _domain_is_accessible_through_HTTP(ip, domain):
+    try:
+        r = requests.head("http://" + ip, headers={"Host": domain})
```
opi: This check doesn't work when an app is installed on domain root and set as public because SSOWat is bypassed and nginx try to redirect to http**S** url.  Using `https` and set `verify=False` works for me.

src/yunohost/certificate.py
alexAubin: Not sure why I added this. I think it was mainly intended for the case where we would like to have, for instance a --fake option (or a debug flag) to test the install using the Fake/Staging Let's Encrypt CA instead of the production one, to be able to run the code multiple times without the fear of the rate limit (20 certificate per period of 7 days per subdomains IIRC). This could also be used before an actual install to test that the ACME challenge is working... Any thoughts on this ?

src/yunohost/certificate.py
```diff
@@ -731,8 +731,11 @@ def _dns_ip_match_public_ip(public_ip, domain):
 
 def _domain_is_accessible_through_HTTP(ip, domain):
     try:
-        r = requests.head("http://" + ip, headers={"Host": domain})
-        # Check we got the ssowat header in the response
+        # Check HTTP reachability
+        requests.head("http://" + ip, headers={"Host": domain})
```
opi: so this one is useless, isn't it ?

alexAubin: Nope, because this checks the webserver is reachable on port 80, which is the one used by Let's Encrypt. There might be some sneaky cases where 443 is reachable correctly but 80 is not (e.g. the user didn't map correctly the port of his/her internet box :stuck_out_tongue:)

src/yunohost/certificate.py
```diff
@@ -0,0 +1,751 @@
...
+
+            logger.success(m18n.n("certmanager_cert_renew_success", domain=domain))
+
+        except Exception as e:
+            import traceback
+            from StringIO import StringIO
+            stack = StringIO()
+            traceback.print_exc(file=stack)
+            logger.error("Certificate renewing for %s failed !", domain)
+            logger.error(stack.getvalue())
```
Psycojoker: I think that we should have an util in moulinette to dump the stack.

src/yunohost/certificate.py
```diff
@@ -0,0 +1,760 @@
...
+
+def _domain_is_accessible_through_HTTP(ip, domain):
+    try:
+        # Check HTTP reachability
+        requests.head("http://" + ip, headers={"Host": domain})
+
+        # Check we got the ssowat header (in HTTPS)
+        r = requests.head("https://" + ip, headers={"Host": domain}, verify=False)
+        if "x-sso-wat" not in r.headers.keys():
+            return False
```
zamentur: This method to check is not sure, some apps (or persistent conf) used the ssowat skipped_uris  properties. f my memory serves me well, with this properties there is no  x-sso-wat header

src/yunohost/certificate.py
```diff
@@ -0,0 +1,760 @@
...
+    # Sign the certificate
+    logger.info("Now using ACME Tiny to sign the certificate...")
+
+    domain_csr_file = "%s/%s.csr" % (TMP_FOLDER, domain)
+
+    signed_certificate = sign_certificate(ACCOUNT_KEY_FILE,
+                                          domain_csr_file,
+                                          WEBROOT_FOLDER,
+                                          log=logger,
+                                          CA=CERTIFICATION_AUTHORITY)
```
zamentur: May be we should catch properly this kind of error here ?

ValueError: Error signing certificate: 429 {
  "type": "urn:acme:error:rateLimited",
  "detail": "Error creating new cert :: Too many certificates already issued for exact set of domains: testle5.nohost.me",
  "status": 429
}

src/yunohost/app.py
```diff
@@ -1031,6 +1031,23 @@ def _get_setting(settings, name):
...
+        if not os.path.isfile(nginx_acme_challenge_conf_file):
+            continue
+
+        # Check the file contains the ACME challenge uri
+        if not '/.well-known/acme-challenge' in open(nginx_acme_challenge_conf_file).read():
+            continue
+
+        # If so, then authorize the ACME challenge uri to unprotected regex
+        unprotected_regex.append(domain + "/%.well%-known/acme%-challenge/.*$")
+
```
alexAubin: Have you tried it ? If this really works, that would simplify quite a lot of things, yeah !

src/yunohost/certificate.py
alexAubin: I think we need to decide what to do with this piece of feature : 
- do some extensive and proper checks that this appears to be the good server (requires move time to dev). Using the staging/test CA of Let's Encrypt could be a proper check, though it doesn't tell what exactly is wrong
- do some checks with particular cases where it might not work. Users can still bypass them with --no-checks (in the CLI at least). That might be confusing for some users though.
- do some minimal checks that we are pretty sure are going to work for all functional setup.

I'd be in favor of the minimal checks, given that it's not a real big deal if the cert-install fails (ok, you get a message error, but that doesn't increase the rate limit, you just need to retry after you checked your config). In the initial PR, i was simply checking the http request worked (i.e. not the header presence). I doubt there will be many cases where the DNS is properly configured, and HTTP requests get answered, but it's actually not the server we're looking for.

On the other hand if would be good to have a separate feature/improvement to get a full diagnostic of features working for a given domain - not limited to DNS and HTTP but also score on mail tester for instance, working XMPP, working mucs, etc ?

src/yunohost/app.py
alexAubin: Seems to work ! Added in dc731c3. We need to validate that we're okay with having this skipped regex for all domains enabled by default though. I don't think it poses any security threat...

src/yunohost/certificate.py
alexAubin: Done, included the call to acme-tiny in a try/catch, with a particular handling for this exception pointing to https://letsencrypt.org/docs/rate-limits/. We might want to catch other exceptions which might occur often, but it's tricky to know this in advance :confused:

src/yunohost/certificate.py
alexAubin: Removed the check for the ssowat header, as long with the https request (which was causing the InsecureRequestWarning due to `verify=False`)

Psycojoker: I agree on keeping a simple check for now and doing the other more complex one on a more specific other feature. This PR is already big enough, let's not make it harder that it is already.

src/yunohost/certificate.py
```diff
@@ -731,13 +731,7 @@ def _dns_ip_match_public_ip(public_ip, domain):
 
 def _domain_is_accessible_through_HTTP(ip, domain):
     try:
-        # Check HTTP reachability
         requests.head("http://" + ip, headers={"Host": domain})
```
opi: no matter what `domain` value is, this requests will not return any exception ;)

Maybe we should request /yunohost/admin or /yunohost/api to be sure that domain & ip will match. See https://dev.yunohost.org/issues/638 for example

locales/en.json
```diff
@@ -237,5 +236,20 @@
     "yunohost_ca_creation_failed": "Unable to create certificate authority",
     "yunohost_configured": "YunoHost has been configured",
     "yunohost_installing": "Installing YunoHost...",
-    "yunohost_not_installed": "YunoHost is not or not correctly installed. Please execute 'yunohost tools postinstall'."
+    "yunohost_not_installed": "YunoHost is not or not correctly installed. Please execute 'yunohost tools postinstall'.",
```
julienmalik: oh ok I now understand the sentence... bad wording imho, but I understand this was present before your PR

src/yunohost/certificate.py
```diff
@@ -470,11 +469,20 @@ def _fetch_and_enable_new_certificate(domain):
...
+        signed_certificate = sign_certificate(ACCOUNT_KEY_FILE,
+                                              domain_csr_file,
+                                              WEBROOT_FOLDER,
+                                              log=logger,
+                                              CA=CERTIFICATION_AUTHORITY)
+    except ValueError as e:
+        if ("urn:acme:error:rateLimited" in str(e)) :
+            raise MoulinetteError(errno.EINVAL,  m18n.n('certmanager_hit_rate_limit', domain=domain))
+        else :
+            raise
```
alexAubin: Indeed. Fixed in e66a708.

src/yunohost/certificate.py
alexAubin: As discussed on the chat, even requesting `/yunohost/admin` or `/yunohost/api` doesn't give a clear indication. Properly diagnosing the status/configuration of a domain seems to be a complicated problem :stuck_out_tongue:. Let's agree that these are some minimal checks for now, and can be refined later.

src/yunohost/certificate.py
zamentur: I agree too

Psycojoker:

zamentur:
I have done a test with the presence of letsencrypt_ynh it was ok, the webadmin ask me to uninstall it before. For the cli I have add the answer:
```
root@yunohost:/vagrant/yunohost# yunohost domain cert-install testle4.grimaud.me
Error: certmanager_old_letsencrypt_app_detected
```

I have done a test with nginx disable:
```
root@yunohost:/vagrant/yunohost# yunohost domain cert-install testle4.exemple.com
Error: Certificate installation for testle4.exemple.com failed !
Error: [Errno 22] certmanager_domain_http_not_working
```

M5oul:
Why displaying `WARNING` for self-signed certif?
```bash
yunohost domain cert-status
certificates: 
  le.moul.re: 
    CA_type: Self-signed
    summary: WARNING
    validity: 3649
```

M5oul:
The `A` test check failed for a domain names I used.
With `host` command the DNS was spread on the DNS resolver of the VM.
Finally, I succeed to install the cert with `yunohost domain cert-install --no-check --force`.
But, from web admin interface, I could stayed blocked as the button was disabled and I couldn't bypassed it.

alexAubin:
> Why displaying WARNING for self-signed certif?

This is with the assumption that domains in Yunohost are supposed to be suitable for "public" use, not for a private use (i.e. just you, + maybe a few friends ?). If your domain is for private use, good for you : add the cert exception popping in your browser, or download the CA certificate - done, ignore the warning. 

But if your domain is meant to be visited by some other, nontechnical people, (95% of the use case IMHO ?) you don't want them to run into the spooky warning of all modern browser about self-signed cert, and there's no reason to have a self-cert if you can install a LE cert with one command or two clicks. Hence the choice to report a warning. The meaning of the warning is not really explicit in the CLI, but it is a bit more in the web interface.

alexAubin:
> The A test check failed for a domain names I used.
> With host command the DNS was spread on the DNS resolver of the VM.
> Finally, I succeed to install the cert with yunohost domain cert-install --no-check --force.
> But, from web admin interface, I could stayed blocked as the button was disabled and I couldn't bypassed it.

To me that sounds related to the propagation to the FFDN resolvers. We might actually have a lot of impatient user who run into this if they want their cert ASAP after setting up their domain :stuck_out_tongue:. 

On the CLI you should have seen this message : "The DNS 'A' record for domain {domain:s} is different from this server IP. If you recently modified your A record, please wait for it to propagate (some DNS propagation checkers are available online). (If you know what you are doing, use --no-checks to disable those checks.)".

On the web interface, we could change the message to tell the user to try again later ? But that's in another PR anyway.

(Pro Tip :tm: : https://dnschecker.org/ allows to see the worldwide propagation of a DNS change)

alexduf:
well done @alexAubin and all! I am super excited to this PR merged ! :)

PR #181, opened 2016-10-31
[fix] helper: setting_set: removing a setting with '-d' option wasn't handled:

M5oul:
- more details here:
https://github.com/YunoHost-Apps/pluxml_ynh/pull/24#issuecomment-257302520


M5oul:
Ok, finally `ynh_app_setting_delete` helper exist :smile:

PR #184, opened 2016-11-17
[fix] Can't restore app on a root domain

zamentur:
Fix this issue https://dev.yunohost.org/issues/593

It fix an unlisted other bug. Before the fix it was possible to install an app on /a/b and next install an other app on /a .

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -909,10 +909,6 @@ def app_checkurl(auth, url, app=None):
         raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
 
     if domain in apps_map:
-        # Domain already has apps on sub path
-        if path == '/':
-            raise MoulinetteError(errno.EPERM,
-                                  m18n.n('app_location_install_failed'))
         # Loop through apps
         for p, a in apps_map[domain].items():
             # Skip requested app checking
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -909,10 +909,6 @@ def app_checkurl(auth, url, app=None):
         raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
 
     if domain in apps_map:
-        # Domain already has apps on sub path
-        if path == '/':
-            raise MoulinetteError(errno.EPERM,
-                                  m18n.n('app_location_install_failed'))
```
zamentur: When the restore script is running the app settings are already set up inside /etc/yunohost/apps/APPID/settings.yml . So `domain in apps_map` is always true, because the app is "installed" on this domain...
When we use `yunohost checkurl unstable.local/ -a APPID` we want to check no **other** app than APPID are installed on the root path or a subpath.
But with this line we check all app even APPID itself !

src/yunohost/app.py
```diff
@@ -922,7 +918,7 @@ def app_checkurl(auth, url, app=None):
             if path == p:
                 raise MoulinetteError(errno.EINVAL,
                                       m18n.n('app_location_already_used'))
-            elif path.startswith(p):
+            elif path.startswith(p) or p.startswith(path):
                 raise MoulinetteError(errno.EPERM,
                                       m18n.n('app_location_install_failed'))
```
zamentur: `path.startswith(p) or p.startswith(path)` works with all case, even with "/" case ( "/"&"/a/" ,  "/a/"&"/", "/a/b/"&"/a/", "/a/"&"/a/b/").
This line could not be reach if p and path are from the same APPID (see Line 915 `if app is not None and a['id'] == app:`) .

Psycojoker:
It's been a week, ping @opi @julienmalik

Psycojoker:
I guess we can merge it then.

PR #185, opened 2016-11-23
Allow applications to ship a script to change its url

Psycojoker:
This PR allows an application to ship a <code>change_url</code> script that allow to move an application from one domain to another.

This branch is a working POC, I'm opening it to not forget it but I'm not calling for a decision yet as we already have important PR to finish first.


Psycojoker:

alexAubin:
Seems to address [issue 32](https://dev.yunohost.org/issues/32)

alexAubin:
This PR looks pretty cool. Is it still active, or are we focusing on other stuff before ? (Shall we test it ?)

Psycojoker:
> are we focusing on other stuff before ?

Yes. I was waiting for the LE admin IF to be released before
reactivating other things like this PR, but if you feel like it we can
start working on this one again.

-- 

Laurent Peuch -- Bram

Psycojoker:
Should be ready for review (but I haven't tested my new modifications).

Psycojoker:
Up

Psycojoker:
Example app https://github.com/YunoHost/change_url_app_ynh/blob/master/scripts/change_url

JimboJoe:
Are `YNH_APP_OLD/NEW_PATH_TRIM(M)ED` really necessary?
There is a pending app helper to normalize path here: https://github.com/YunoHost/yunohost/pull/234

It does also open the door to store "installation only" files on the system (e.g. some apps provide 2 nginx.conf models, depending on a root or subdirectory installation). Should we file an issue for this part?

Psycojoker:
> Are YNH_APP_OLD/NEW_PATH_TRIM(M)ED really necessary?
> There is a pending app helper to normalize path here: #234

This PR is quite old, at that time we didn't have those utilities, so yeah, this might not be necessary anymore.

> It does also open the door to store "installation only" files on the system (e.g. some apps provide 2 nginx.conf models, depending on a root or subdirectory installation). Should we file an issue for this part?

Yes, currently we don't store the `conf` folder of the app which indeed cause troubles here, this is probably something we'll want to change but we'll have to ensure retro-compatibility or find another way (the user will have to upgrade the app before being able to run this script since no app support it yet, we could save those file during upgrade and/or during new installation for eg).

data/actionsmap/yunohost.yml
```diff
@@ -535,6 +535,31 @@ app:
                     full: --file
                     help: Folder or tarball for upgrade
 
+        ### app_change_url()
+        change_url:
```
alexAubin: Shouldn't we call this "changeurl" (without _) to be consistent with other existing actions such as checkurl, clearaccess, removelist, ... ?

src/yunohost/app.py
```diff
@@ -341,6 +341,105 @@ def app_map(app=None, raw=False, user=None):
...
+    # (or checkurl renamed in reserve_url)
+    old_domain = app_setting(app, "domain")
+    old_path = app_setting(app, "path")
+
+    old_path_trimed = old_path.rstrip("/")
+
+    if (domain.strip().lower(), path.rstrip("/").strip()) == (old_domain, old_path_trimed.strip()):
+        raise MoulinetteError(errno.EINVAL, m18n.n("app_change_url_identical_domains", domain=domain, path=path))
+
+    app_checkurl(auth, '%s%s' % (domain, path), app)
```
alexAubin: I encountered a bug here if path does not start with "/". It should be handled correctly (maybe there's also a bug if path ends with a '/')

src/yunohost/app.py
```diff
@@ -341,6 +341,105 @@ def app_map(app=None, raw=False, user=None):
...
+        # grab nginx errors
+        # the "exit 0" is here to avoid check_output to fail because 'nginx -t'
+        # will return != 0 since we are in a failed state
+        nginx_errors = subprocess.check_output("nginx -t; exit 0",
+                                               stderr=subprocess.STDOUT,
+                                               shell=True).rstrip()
+
+        raise MoulinetteError(errno.EINVAL, m18n.n("app_change_url_failed_nginx_reload", nginx_errors=nginx_errors))
+
+    logger.success("Successfly changed '%s' url to '%s%s'" % (app, domain, path))
```
alexAubin: Typo + should be internationalized ?

src/yunohost/app.py
```diff
@@ -341,6 +341,105 @@ def app_map(app=None, raw=False, user=None):
...
+        return
+
+    # this should idealy be done in the change_url script but let's avoid common mistakes
+    app_setting(app, 'domain', value=domain)
+    app_setting(app, 'path', value=path)
+
+    app_ssowatconf(auth)
+
+    # avoid common mistakes
+    if os.system("service nginx reload") != 0:
```
alexAubin: Shouldn't you use _run_service_command("reload", "nginx") from yunohost.service here ? Imho the whole thing with nginx -t should be factorized out because that's a recurring need (this kind of helper would be good to have for certificate.py for instance)

data/actionsmap/yunohost.yml
Psycojoker: Don't know, we also have a `regen-conf`, I don't think that we have a very consistent naming scheme :x 

But that can be changed yes.

src/yunohost/app.py
Psycojoker: Ah, very possible for the `_run_service_command`, wasn't aware of that.

For the `nginx -t`, I'll wait for another part of the code to need it.

alexAubin:
I added a draft for unit tests for this feature. To run them, you'll need to create an `apps` folder in the `/tests/` folder, then inside, `git clone https://github.com/YunoHost/change_url_app_ynh`. (In the future, imho we should have a single repo name something like `test_apps` where we store all the small apps we use for tests, so we can git clone all of them with a single command (and not create many repo for every single tests apps)

I found that if the given path does not ends with a '/', the app_map is a bit broken (others apps path always have a / at the end by default).
Also it's not clear to me wether we should be able to install and change_url at the root (i.e. with path = "/")

data/actionsmap/yunohost.yml
Psycojoker: Done.

Shouldn't we rename the script "change_url" to "changeurl" too to be consistent?

alexAubin: Hm, dunno, I guess if it's not too much work then why not. But I'm more a core dev than and app packager so I don't care as much about the script name xD

src/yunohost/app.py
```diff
@@ -341,6 +341,109 @@ def app_map(app=None, raw=False, user=None):
...
+
+    if (domain, path) == (old_domain, old_path_trimed):
+        raise MoulinetteError(errno.EINVAL, m18n.n("app_change_url_identical_domains", domain=domain, path=path))
+
+    # WARNING / FIXME : checkurl will modify the settings
+    # (this is a non intuitive behavior that should be changed)
+    # (or checkurl renamed in reserve_url)
+    app_checkurl(auth, '%s%s' % (domain, path), app)
+
+    if not os.path.exists(os.path.join(APPS_SETTING_PATH, app, "scripts", "change_url")):
```
alexAubin: Shouldn't this check happen at the beginning of the function ? Here we already called app_checkurl and compared the old and new domains.

src/yunohost/app.py
```diff
@@ -341,6 +341,109 @@ def app_map(app=None, raw=False, user=None):
...
+    if not installed:
+        raise MoulinetteError(errno.ENOPKG,
+                              m18n.n('app_not_installed', app=app))
+
+    old_domain = app_setting(app, "domain")
+    old_path = app_setting(app, "path")
+
+    # Normalize path and domain format
+    domain = domain.strip().lower()
+    old_path_trimed = '/' + old_path.strip("/").strip() + '/'
```
alexAubin: Not sure why we need to have a special _trimmed variable ? Shouldn't we just normalize old_path ?

M5oul:
Could you implement this script on a simple app such as Jirafeau.
That way we could test this awesome feature!

alexAubin:
As said [here](https://github.com/YunoHost/yunohost/pull/185#issuecomment-285882246), we already have a simple/dummy test application made by Bram, which I moved [here](https://github.com/YunoHost/test_apps)

Also if you're able to setup the unit test stuff (not so well documented for now :s), [this unit test](https://github.com/YunoHost/yunohost/blob/2b82a2e74dc625680459c4ce5825f19c2a49374b/src/yunohost/tests/test_changeurl.py#L45-L52) automatically install the app, check it's available at the initial url, call the change url feature, then check it's available at the new url

JimboJoe:
As discussed earlier, we'll need [this improvement](https://dev.yunohost.org/issues/871) to be implemented: some apps will need their conf files to operate a URL change (e.g. reference nginx configuration files depending on root/non-root target).

alexAubin:
I was wondering why the current test app works if we don't have the `conf` folder available, and that's because in fact we do a simple sed in the existing nginx conf. In fact, for many relatively simple apps, I think the conf folder is not absolutely needed.

The fix to have the `conf` folder doesn't look trivial and will/might require a proper migration of all the installed apps on existing instances (which is in turn, in itself, not trivial). Imho this should be adressed in another PR / roadmap.

src/yunohost/tests/test_changeurl.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+import time
+import requests
+
+from moulinette.core import init_authenticator
+from yunohost.app import app_install, app_changeurl, app_remove, app_map
```
zamentur: I got ImportError: No module named yunohost.app
I don't know why( maybe the mistake come from my yuno instance)

src/yunohost/tests/test_changeurl.py
```diff
@@ -0,0 +1,61 @@
...
+def setup_function(function):
+    pass
+
+
+def teardown_function(function):
+    app_remove(auth, "change_url_app")
+
+
+def install_changeurl_app(path):
+    app_install(auth, "./tests/apps/change_url_app_ynh",
```
zamentur: There is no file in./ tests/apps

alexAubin: Yes, it's not well documented for now. You'll need to git clone https://github.com/YunoHost/test_apps ./apps from inside the tests folder.

src/yunohost/tests/test_changeurl.py
alexAubin: Are you running pytest from inside `/vagrant/yunohost/src/yunohost/` ?

zamentur: yep

src/yunohost/tests/test_changeurl.py
zamentur: change_url script should fix the path before to do operation.
See https://github.com/YunoHost-Apps/wordpress_ynh/blob/master/scripts/.fonctions#L39
 Or it might be done inside the python code

alexAubin: Not sure to understand the point ? You mean use [ynh_normalize_url_path](https://github.com/YunoHost/yunohost/pull/234) ? We already normalize the path from inside the python code [here](https://github.com/YunoHost/yunohost/blob/2b82a2e74dc625680459c4ce5825f19c2a49374b/src/yunohost/app.py#L367) and there are plans to use [this python helper](https://github.com/YunoHost/yunohost/blob/5125c0466a52ef8fe7ab2e2fd563f71b153f7474/src/yunohost/tools.py#L595) from PR #267 once it'll be available.

zamentur: sorry I didn't see this part of code (code review overdose :p )
LGTM

src/yunohost/app.py
```diff
@@ -341,6 +341,105 @@ def app_map(app=None, raw=False, user=None):
...
+    installed = _is_installed(app)
+    if not installed:
+        raise MoulinetteError(errno.ENOPKG,
+                              m18n.n('app_not_installed', app=app))
+
+    if not os.path.exists(os.path.join(APPS_SETTING_PATH, app, "scripts", "change_url")):
+        raise MoulinetteError(errno.EINVAL, m18n.n("app_change_no_change_url_script", app_name=app))
+
+    old_domain = app_setting(app, "domain")
+    old_path = app_setting(app, "path")
```
alexAubin: You mean some app don't have a path setting at all because they're implicitly on '/' ? Shall we have old_path = '/' by default ? Or shall we refuse to change_url() at all if there's no path setting available ?

zamentur: Some example: odoo, gitlab and many others aren't designed to be hosted on a subpath. So packager just ask for one question: on which domain.

This kind of apps could run change_url to change the domain only.

More generally, it asks the question of changing install parameters.

data/actionsmap/yunohost.yml
```diff
@@ -535,6 +535,31 @@ app:
                     full: --file
                     help: Folder or tarball for upgrade
 
+        ### app_changeurl()
+        changeurl:
```
alexAubin: Following the discussion about the name convention with ljf (https://dev.yunohost.org/issues/872), I think we need to revert back to `change-url` (instead of `changeurl`) (sorry for inducing you in error about that Bram :/)

src/yunohost/app.py
alexAubin: Hm okay

I don't know what to do about this. The status of a 'domain-only' or 'full-domain' (?) app is not so well-defined. If there's no path setting, we could assume that the app is domain-only / full-domain and that there's no path needed and therefore refuse the 'new_path' argument if it's given by the user ?

alexAubin: After discussing irl with ljf : it turns out that even 'full-domain' apps have a path settings because they use checkurl. Therefore we don't have a way to easily determine if an app is 'full-domain'.

On the other hand, for now we can just let the app packager handle this in the script : if the app is full domain and the new path variable is different from `/`, simply refuse to change url and inform the user that this is a 'full-domain' app.

JimboJoe: I agree, let's leave it to the app developer.

zamentur:
Medium decision: there is 2 positive review, we need one more approval (3 needed or 4 to merge quickly)

alexAubin:
Merging as proposed in the mail from April 12th to yunoteam.

maniackcrudelis:
Hi guys,

there is a small error somewhere in the code.
If the script change_url failed, it doesn't return a error code and finish by "Success !"

maniackcrudelis:
Sorry, little mistake, the script I've tested have not failed directly.
So, with an exit 1, I have an "Error : Failed to change".

But, the command still return 0

PR #186, opened 2016-11-24
[fix] Can't get mailbox used space if dovecot is down

zamentur:
See https://dev.yunohost.org/issues/640
To solve it there is a try except and a warning display to the user about dovecot status.

There are small fixes in more: it removes a bad eval and the way to parse `doveadm` result is more reliable (the percentage was sometimes the number of message).

Optionaly: I suggest a linked PR on yunohost-admin with https://github.com/YunoHost/yunohost-admin/pull/137 .

Here are some examples of the new api

**Classic case** 
```
root@yunohost:/vagrant# yunohost user info test
firstname: Jean
fullname: Jean Dupont
lastname: Dupont
mail: test@testle4.nohost.me
mailbox-quota: 
  limit: 500M
  use: 5M (1%)
username: test
```

**No quota case : the percentage in use is not displayed**
```
root@yunohost:/vagrant# yunohost user info test
firstname: Jean
fullname: Jean Dupont
lastname: Dupont
mail: test@testle4.nohost.me
mailbox-quota: 
  limit: No quota
  use: 0.0K
username: test
```

**Dovecot shutdown (or failed) case**
```
root@yunohost:/vagrant# yunohost user info test
firstname: Jean
fullname: Jean Dupont
lastname: Dupont
mail: test@testle4.nohost.me
mailbox-quota: 
  limit: 500M
  use: ?
username: test
```

### DIFF
--- a/src/yunohost/user.py
+++ b/src/yunohost/user.py
@@ -92,7 +92,7 @@ def user_list(auth, fields=None, filter=None, limit=None, offset=None):
 
 
 def user_create(auth, username, firstname, lastname, mail, password,
-        mailbox_quota=0):
+        mailbox_quota="0"):
     """
     Create user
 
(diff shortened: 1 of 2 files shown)


Psycojoker:
Shouldn't we start dovecot if it's down? At least from the web interface, I'm not expecting users to have any of what this mean.

julienmalik:
@Psycojoker would seem awkward to me that calling user_info can potentially restart a service. if it's down then probably there is something wrong, and anyway, calling for a restart does not imply it will be running afterwards.

Gracefully handling the case where dovecot is down seems relevant. And here, users get a warning telling them that dovecot seems down.

Psycojoker:
@julienmalik then I think that we should have an explicit error message telling how the user can restart it/diagnostic it, otherwise it will just mean "FAILED" to most users :/

Psycojoker:

alexAubin:
To me, a warning is fine. I think the question whether or not the restart dovecot or show a better warning/explicit message and/or diagnostic should be treated as another issue : we need a tool with proper diagnostic features (e.g. are domains correctly configured and reachable ? are all enabled services up and running ? etc...), automatically notifying the admin if something's wrong. But that's beyond the scope of this PR.

alexAubin:
I think this can be flagged as a small decision ? We can merge this before the end of the week imho.

src/yunohost/user.py
```diff
@@ -398,20 +398,44 @@ def user_info(auth, username):
...
+                storage_use = _convertSize(storage_use)
+
+                if is_limited:
+                    has_percent = re.search(r'%=(\d+)', cmd_result)
+
+                    if has_percent:
+                        percentage = int(has_percent.group(1))
+                        storage_use += ' (%s%%)' % percentage
+
+        except subprocess.CalledProcessError:
```
Psycojoker: Quoting python documentation `If the return code was non-zero it raises a CalledProcessError.`. This mean that **any** non zero value will raise this and our code consider that this always mean that dovecot is down which could not be the case, this will create a very annoying to debug situation :/

alexAubin: Maybe we could use `service_status("dovecot")` to check it's running ?

Psycojoker: This seems to be working on my internet cube, so yes, it's a good solution :+1:

PR #187, opened 2016-11-24
[fix] Disable removal of main domain and clarify tools_maindomain

alexAubin:
Attempt to fix [issue 639](https://dev.yunohost.org/issues/639).

I refactorized `tools_maindomain()` in the same time, which had a lot of weird stuff (old_domain option not actually used ? dyndns option not documented, only used in postinstall ? duplicated and outdated code with domain_add ? automatically create the new main domain if it does not exist ?)

I still need to check that this doesn't break postinstall.

### DIFF
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -160,7 +160,7 @@ def domain_remove(auth, domain, force=False):
         raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
 
     # Check domain is not the main domain
-    if (domain == _get_maindomain()) :
+    if domain == _get_maindomain():
         raise MoulinetteError(errno.EINVAL, m18n.n('domain_cannot_remove_main'))
 
     # Check if apps are installed on the domain
@@ -289,6 +289,7 @@ def get_public_ip(protocol=4):
         logger.debug('cannot retrieve public IPv%d' % protocol, exc_info=1)
         raise MoulinetteError(errno.ENETUNREACH,
                               m18n.n('no_internet_connection'))
+
 
 def _get_maindomain():
     with open('/etc/yunohost/current_host', 'r') as f:
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -113,6 +113,7 @@ def domain_add(auth, domain, dyndns=False):
                 raise MoulinetteError(errno.EINVAL,
                                       m18n.n('domain_dyndns_root_unknown'))
 
+
     try:
         yunohost.certificate._certificate_install_selfsigned([domain], False)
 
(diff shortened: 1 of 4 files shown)


Psycojoker:

alexAubin:
Tested post-install. It works for me.

src/yunohost/tools.py
```diff
@@ -104,52 +104,45 @@ def tools_adminpw(auth, new_password):
         logger.success(m18n.n('admin_password_changed'))
 
 
-def tools_maindomain(auth, old_domain=None, new_domain=None, dyndns=False):
+def tools_maindomain(auth, new_domain=None):
     """
-    Main domain change tool
+    Main domain consultaton or change tool
```
opi: Typo

src/yunohost/tools.py
```diff
@@ -104,52 +104,45 @@ def tools_adminpw(auth, new_password):
         logger.success(m18n.n('admin_password_changed'))
 
 
-def tools_maindomain(auth, old_domain=None, new_domain=None, dyndns=False):
+def tools_maindomain(auth, new_domain=None):
```
opi: The `new_domain` wording is no more relevant. What about `main_domain` or `domain`?

src/yunohost/tools.py
```diff
@@ -104,52 +104,45 @@ def tools_adminpw(auth, new_password):
...
 
     """
-    if not old_domain:
-        with open('/etc/yunohost/current_host', 'r') as f:
-            old_domain = f.readline().rstrip()
-
-        if not new_domain:
-            return { 'current_main_domain': old_domain }
 
+    # If no new domain specified, we return the current main domain
```
opi: \o/

src/yunohost/tools.py
alexAubin: Not sure about this ... `new_domain` makes it explicit that this value is meant to be the new main domain ? To me, just `domain` would be ambiguous.

src/yunohost/domain.py
```diff
@@ -301,3 +306,12 @@ def get_public_ip(protocol=4):
         logger.debug('cannot retrieve public IPv%d' % protocol, exc_info=1)
         raise MoulinetteError(errno.ENETUNREACH,
                               m18n.n('no_internet_connection'))
+
+def _get_maindomain():
+    with open('/etc/yunohost/current_host', 'r') as f:
+        maindomain = f.readline().rstrip()
```
Psycojoker: Why only `rstrip` and not `strip`?

Psycojoker:
Code seems ok otherwise

src/yunohost/domain.py
alexAubin: No idea, I just used the previous code and wrapped it in a function for it to be semantic. We can move to `strip()` if you want, but that shouldn't make any difference (except maybe if the user edited the file by hand) ?

Psycojoker: Ok, let's keep it that way then.

alexAubin:
(Had to do some dirty rebase magic to resolve conflicts before merging)

PR #188, opened 2016-11-24
Use the root user for app installation

Psycojoker:
Hello,

I'm opening this PR to avoid forgetting it but I'm not calling for a decision yet.

It's the implementation of "don't use admin user for installing apps because it doesn't brings anything and that forgetting sudo is the first bug ever while developping apps", also will greatly reduce broken situation du to missing admin user (well ... hide them).

We'll probably need to talk again about this choice.

### DIFF
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -292,7 +292,7 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
 
 
 def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
-              chdir=None, env=None):
+              chdir=None, env=None, user=None):
     """
     Execute hook from a file with arguments
 
@@ -303,6 +303,7 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
         no_trace -- Do not print each command that will be executed
         chdir -- The directory from where the script will be executed
         env -- Dictionnary of environment variables to export
+        user -- User with which to run the command
 
     """
     from moulinette.utils.process import call_async_output
@@ -327,7 +328,11 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
         cmd_script = path
 
     # Construct command to execute
-    command = ['sudo', '-n', '-u', 'admin', '-H', 'sh', '-c']
+    if user is not None:
+        command = ['sudo', '-n', '-u', user, '-H', 'sh', '-c']
+    else:
+        command = ['sh', '-c']
+
     if no_trace:
         cmd = '/bin/bash "{script}" {args}'
     else:
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -307,7 +307,6 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
 
     """
     from moulinette.utils.process import call_async_output
-    from yunohost.app import _value_for_locale
 
     # Validate hook path
     if path[0] != '/':
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -292,7 +292,7 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
 
 
 def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
-              chdir=None, env=None, user=None):
+              chdir=None, env=None, user="admin"):
     """
     Execute hook from a file with arguments
 
@@ -328,10 +328,10 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
         cmd_script = path
 
     # Construct command to execute
-    if user is not None:
-        command = ['sudo', '-n', '-u', user, '-H', 'sh', '-c']
-    else:
+    if user == "root":
         command = ['sh', '-c']
+    else:
+        command = ['sudo', '-n', '-u', user, '-H', 'sh', '-c']
 
     if no_trace:
         cmd = '/bin/bash "{script}" {args}'
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -275,7 +275,7 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
                 hook_args = pre_callback(name=name, priority=priority,
                                          path=path, args=args)
                 hook_exec(path, args=hook_args, chdir=chdir, env=env,
-                          no_trace=no_trace, raise_on_error=True)
+                          no_trace=no_trace, raise_on_error=True, user="root")
             except MoulinetteError as e:
                 state = 'failed'
                 logger.error(e.strerror, exc_info=1)
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -418,7 +418,7 @@ def app_upgrade(auth, app=[], url=None, file=None):
 
         # Execute App upgrade script
         os.system('chown -hR admin: %s' % INSTALL_TMP)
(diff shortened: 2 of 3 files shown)


Psycojoker:
I'm putting this as a medium decision because this is going to affect appliction installation (but exception for permissions stuff I don't see what it can really change, but I'm expecting you to find other situations)

Psycojoker:
Up. I'd least I'd like to have your opinion on the principle of moving to root for app installation.

@maniackcrudelis @scith I'd like your opinion on the principle of this change since you are in the apps team.

scith:
Hi, how may it affect app installation? Root has more privileges than admin, and we already use sudo most of the times.
Would there be an alternative to using root, while still taking into account the fact that many apps require sudo rights at least for install? If not, I agree with this

Psycojoker:
> Hi, how may it affect app installation? Root has more privileges than admin, and we already use sudo most of the times.

You won't have to put `sudo` everywhere from now on, which was (for the people I've talked to) the most common error they got when making applications.

> Would there be an alternative to using root, while still taking into account the fact that many apps require sudo rights at least for install? If not, I agree with this

You can still uses `su $user -c` when needed :)

Since admin has sudo right without a password it was basically having root access while having to put `sudo` everywhere, it's was complicating things more than anything else (and other system wide package managers uses root too).

maniackcrudelis:
It's ok for me, I think isn't a problem for packages, because the permissions are set when it's necessary.

src/yunohost/hook.py
```diff
@@ -330,7 +331,11 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
         cmd_script = path
 
     # Construct command to execute
-    command = ['sudo', '-n', '-u', 'admin', '-H', 'sh', '-c']
+    if user == "root":
+        command = ['sh', '-c']
```
julienmalik: This means we always consider this python code is executed as root. Shouldn't we add a sudo here just in case ?

Psycojoker: YunoHost can't be runned as another user than root.

Psycojoker: See https://github.com/YunoHost/yunohost/blob/unstable/bin/yunohost#L188

julienmalik:
Is it intentionnal that you handle install/upgrade/remove, and not backup/restore ?

Psycojoker:
No, that's a mistake.

mbugeia:
While I'm not really against this (after all apt use root for install), I have to disagree with the first statement of this PR:
"don't use admin user for installing apps because it doesn't brings anything" -> It does bring something: consciousness that you're doing an action that can break things and. And it respect the principle of least privilege.
"forgetting sudo is the first bug ever while developping apps" -> I disagree, this is the easiest problem to solve when developping apps.
"will greatly reduce broken situation du to missing admin user (well ... hide them)." -> IMO hiding a bug just makes things worst.

alexAubin:
Same as @mbugeia, though not against it either (c.f. apt).

zamentur:
I think hooks are missing in this PR.

There is pro and cons with this idea to run hook script with user root.

About backup, it may be a problem if there is no backup because the user admin (so the nscd, nslcd or slapd) is not available. 

At the same time, I'm agree too with the args from mbugeia about hiding a bug. But, an error during the app setup is not the good moment for warning the administrator about a missing admin user... About this bug, we should check that very earlier.

May be this PR could break some app package. Some program need to be run as a non root user so currently (without this pr) if you run in a script a program like this it works, but with this pr it could raise an error like "This programm need to be run as a non root user".

Psycojoker:
> I think hooks are missing in this PR.
Going to add it then.

> There is pro and cons with this idea to run hook script with user root.
I have no opinion on this one, I can easily add it if you want.

> About backup, it may be a problem if there is no backup because the user admin (so the nscd, nslcd or slapd) is not available.
Therefor it would be better to add it then? Making all app scripts running with root seems more coherent.

polytan02:
Je voulais dire que je suis d'accord avec le root pour une raison toute simple : on finit par mettre sudo partout. 
Pour moi sudo est utile quand on est en console et qu'on a besoin de lancer une ou deux commandes et qu'on ne peut/veut pas se connecter en root. Là on fait des scripts où tout doit être exécuté en root sinon ça ne marche pas.

Par contre, je suis d'accord avec @mbugeia sur un point inquiétant : certains qui font des apps ne se rendent pas compte des commandes qu'ils lancent ni du pouvoir qu'elles ont sur le systèmes (on parlait de rm -rf malheureux par exemple dans un script remove).

C'est aussi pour ça que je souhaitais pouvoir afficher le readme.md de l'app avant de l'installer.

Autre question : comment faire pour cocher "changes approved" ?

Psycojoker:
Work is ready to be rewied. Hooks run using root now and I've fixed the backup.py where root wasn't specified.

> Autre question : comment faire pour cocher "changes approved" ?
￼￼
You need to be a member of the core dev team for that.

Psycojoker:
We need one or two more approval to merge this PR (I always wonder why so few people have reviewed it :/)

alexAubin:
From my point of view, I don't know what to think about the purpose of PR (not trying to say it's bad, I genuinely don't know what to think). On one hand, we really do use sudo most of the time in app scripts and it's annoying to have to loose time because of forgetting to put "sudo" in install script ;). On the other hand, I agree with @mbugeia on the principle of least privilege. But if apt does the same thing, well, okay.

However what I'm worried about at this point is wether or not this breaks some apps. As @zamentur said, some apps or specific program need to be run as a non-root user. What's the plan to test apps ? Imho we can rely on the App CI if it runs on testing/unstable. But we should be careful and not release this too quickly.

Psycojoker:
```
01:58  <Aleks__> bah après sur le principe ok, apt installe en root
01:58  <Aleks__> la question imho c'est surtout est-ce que le gain "plus avoir à faire sudo" est supérieur à tous les 
                 risques qu'on encoure en changeant un truc critique comme ça :/
01:58  <Aleks__> et notamment casser les apps
01:59  <Aleks__> mais tant qu'on teste bien, why not
01:59 @<Bram> niveau casser les apps le risque est quand même fort bas ici
01:59  <Aleks__> ah, oké
01:59 @<Bram> et niveau avantage ça rend le dev d'app bien moins pénible
02:00 @<Bram> et je pense que faciliter le boulot des développeurs d'apps est quand même un très gros +
02:00 @<Bram> après oui, on va pas merger/déployer ça en mode swag du slip
02:00  <Aleks__> après j'suis biaisé par la dernière app que j'ai voulu packager (wekan) parce que meteor voulait pas etre 
                 root /o\
02:00 @<Bram> sinon
02:00 @<Bram> pour rigoler https://github.com/search?q=org%3AYunoHost-Apps+sudo&type=Code
02:01  <Aleks__> 655
02:01  <Aleks__> xD
02:01 @<Bram> et c'est 655 *fichiers*, pas 655 sudo :p
02:01 @<Bram> this is madness ™
02:01  <Aleks__> ogod
02:01 @<Bram> https://github.com/search?q=org%3AYunoHost-Apps+sudo&type=Commits
02:01 @<Bram> celui là aussi est super marrant comme résultat
```

And to insist, just look at any random script file of a YunoHost app, `sudo` is used for 80% of the lines and the % goes higher if you remove the variable declarations lines.

At this point it's just making the life of every package manager harder for nothing.

Psycojoker:
After having read a lot of random app scripts (and having made apps myself), I'm now really convinced that keeping `sudo` and not switching to `root` is ever a terrible bad design decision or just voluntary making the life of app developers painful.

Psycojoker:
@M5oul @maniackcrudelis already agreed on that PR apparently:

> It's ok for me, I think isn't a problem for packages, because the permissions are set when it's necessary.

PR #191, opened 2016-11-28
[fix] Fix ldap caching on postinstall, which might cause 'Unknown admin user'

alexAubin:
Follow-up of #172 after some investigation with @Psycojoker (thanks to him ! Had no idea about that nslcd/nscd thing 😛). Not entirely sure this is the real cause of the issue people reported, but there's a good 95% that it is.

Funnily enough, before this fix, if you attempted to log as admin (or `su admin`) before launching post-install, post-install would fail - because the cache of nscd remembers that user admin doesn't exists for the next minute at least. Maybe something like `su admin` happens somewhere in the procedure for raspberry install and others ? Or maybe the cache config is different ?

@Psycojoker suggested on the chat that if this doesn't solves all issues, we could check that nscd and nslcd (and slapd?) are correctly running.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -114,7 +114,7 @@
     "ip6tables_unavailable": "You cannot play with ip6tables here. You are either in a container or your kernel does not support it.",
     "iptables_unavailable": "You cannot play with iptables here. You are either in a container or your kernel does not support it.",
     "ldap_initialized": "LDAP has been initialized",
-    "ldap_init_failed_to_create_admin": "LDAP initialization failed to create admin. Aborting.",
+    "ldap_init_failed_to_create_admin": "LDAP initialization failed to create admin user.",
     "license_undefined": "undefined",
     "mail_alias_remove_failed": "Unable to remove mail alias '{mail:s}'",
     "mail_domain_unknown": "Unknown mail address domain '{domain:s}'",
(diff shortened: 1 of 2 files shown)


Psycojoker:
This bug being quite serious I think that it justify a release on itself, I would be really in favor on doing a bugfix release once this one is merged.

locales/en.json
```diff
@@ -114,6 +114,7 @@
     "ip6tables_unavailable": "You cannot play with ip6tables here. You are either in a container or your kernel does not support it.",
     "iptables_unavailable": "You cannot play with iptables here. You are either in a container or your kernel does not support it.",
     "ldap_initialized": "LDAP has been initialized",
+    "ldap_init_failed_to_create_admin": "LDAP initialization failed to create admin. Aborting.",
```
opi: 1/  "LDAP initialization failed to create admin **user**."  ?
2/ Remove the "Aborting" and use `installation_failed`string ?

Psycojoker:
I've test it using ynh-dev (including @alexAubin commit) and it works as expected for me.

Psycojoker:
Up @Ju @ljf @Moul !

I consider this to be a very critical bug, I'd really like to see it available in the next releases

M5oul:
I have no opinion about this fix.
I trust you.

locales/en.json
julienmalik: agreed

PR #193, opened 2016-11-28
Clean app upgrade

Psycojoker:
Hello,

This is mostly refactoring to make the code more readable. I've simply removed unused code and assign a temp variable for better readability.

We need to do way more refactoring on app.py but that's for later.


Psycojoker:
You need to stimulate an upgrade with a dummy app, not very convenient.

Psycojoker:
We need another approval.

src/yunohost/app.py
```diff
@@ -349,10 +349,12 @@ def app_upgrade(auth, app=[], url=None, file=None):
...
     if not app:
-        if (not url and not file):
-            app = os.listdir(apps_setting_path)
+        if not url and not file:
+            app = [app["id"] for app in app_list(installed=True)["apps"]]
     elif not isinstance(app, list):
-        app = [ app ]
+        app = [app]
+
+    logger.info("Will upgrade apps %s", ", ".join(app))
```
opi: nitpicking & personnal taste, but I'd prefer something like "Upgrading apps %s";

alexAubin:
I refactorized the upgradability check to be directly in app_info, in order to simplify the fix of https://dev.yunohost.org/issues/787. Still not really clean but definetely cleaner than current stuff imho :/

PR #194, opened 2016-11-28
[mod] try to clean a bit app_list code

Psycojoker:
This need to be rebased.

This is following a ticket on redmine asking to clean this part of the code. This is a first attemp, I think we can improve it more.

I'm not calling for a decision yet.

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -198,7 +198,7 @@ def app_list(offset=None, limit=None, filter=None, raw=False, installed=False, w
             app_dict[app]['repository'] = None
 
     # ???
-    if not (len(app_dict) > (0 + offset) and limit > 0):
+    if not (len(app_dict) > offset and limit > 0):
         return {'apps': list_dict} if not raw else list_dict
 
     # Sort app list
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -157,17 +157,12 @@ def app_list(offset=None, limit=None, filter=None, raw=False, installed=False, w
         with_backup -- Return only apps with backup feature (force --installed filter)
 
     """
-    if offset: offset = int(offset)
-    else: offset = 0
-    if limit: limit = int(limit)
-    else: limit = 1000
+    offset = int(offset) if offset else 0
+    limit = int(limit) if limit else 1000
     installed = with_backup or installed
 
     app_dict = {}
-    if raw:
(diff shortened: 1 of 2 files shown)


alexAubin:
Bumping for glory

Psycojoker:
We need another opinion.

src/yunohost/app.py
```diff
@@ -161,79 +156,99 @@ def app_list(offset=None, limit=None, filter=None, raw=False, installed=False, w
...
-            for app, info in json.loads(str(json_list.read())).items():
-                if app not in app_dict:
-                    info['repository'] = applist
-                    app_dict[app] = info
+            for app, info in json.load(json_list).items():
+                if app in app_dict:
+                    continue
 
+                info['repository'] = applist
+                app_dict[app] = info
```
zamentur: The last version (with "if not in") was easier.

src/yunohost/app.py
```diff
@@ -161,79 +156,99 @@ def app_list(offset=None, limit=None, filter=None, raw=False, installed=False, w
...
-            # Look for forks
+            # Handle multi-instance case like wordpress__2
             if '__' in app:
                 original_app = app[:app.index('__')]
                 if original_app in app_dict:
                     app_dict[app] = app_dict[original_app]
                     continue
-            with open( apps_setting_path + app +'/manifest.json') as json_manifest:
-                app_dict[app] = {"manifest":json.loads(str(json_manifest.read()))}
+                # FIXME : What if it's not !?!?
```
zamentur: This FIXME isn't needed. Because the goal of this if is to fill app_dict[app] if it is not already filled. But it is the case with the first setup of 'wordpress' (or other apps)

alexAubin: Uuuuh, I don't know if I get the point ?

Let's assume we have a "wordpress__2" installed, and let's imagine the first "wordpress" got uninstalled (c.f. recent tickets). Annnnd let's imagine wordpress is bad, so it got removed from  the official.json. Then the code doesn't add wordpress__2 to the app_list dictionnary (because original_app is not in app_dict). And then everything behave as if wordpress__2 is not even installed.

Psycojoker:
Rebased.

Psycojoker:
Seeing the low availability of ljf and the age of this PR and its low impact I suggest that we merge it soon.

PR #195, opened 2016-11-28
Migration framework

Psycojoker:
### Things that needs to be done

- [x] run "yunohost migrations migrate" at package upgrade
- [x] i18n

THIS PR NEEDS THIS MOULINETTE PR TO WORKS https://github.com/YunoHost/moulinette/pull/143

### Problem

We are often in need of modifying the way some data are store by yunohost (example: the fetchlist refactoring PR needed to modifying the way we store applist) or the state of an installed YunoHost or some config files.

Today, to solve this problem we either don't do anything (and avoir improving YunoHost) or add ugly backward compatibility code. This restrict us in the development process, making it less easy (and thus less funny) and prevent us from making YunoHost better.

### Solution

Based on my experience using the Django framework, I've imported the principle of data migrations and integrate it inside YunoHost while heavily based my work on their API.

This allows us to write migrations during .deb upgrade thus preventing the user from having to do it themselves and that being transparent.

A migration can be:

* modify the way applists are stored
* remove archivemount
* modifying the permissions of those certs that are breaking metronome

A migration is very simple to write and can, optionally implement a "backward" possibility if for some reason the users needs to revert its actions.

Here is an example:

```python
import subprocess
from yunohost.migrations import Migration


class MyMigration(Migration):
    "Remove archivemount because we don't use it anymore"

    def forward(self):
        subprocess.check_call("apt-get remove archivemount -y", shell=True)

    def backward(self):
￼        subprocess.check_call("apt-get install archivemount -y", shell=True)
```

In the idea of preventing the user from being stuck, I've implement the classical framework API if some actions needs to be done to debug.

The API currently looks like this:

```
yunohost tools migrations list
yunohost tools migrations migrate  # do the actual migration
yunohost tools migrations migrate -t 42  # run to the migration 42 (forward of backward depending on the state)
yunohost tools migrations migrate -t 42 -f # same but faking it to get to this state
yunohost tools migrations state  # show the current state
```

### How to test

Checkout this branch and run:

```
yunohost tools migrations list
yunohost tools migrations state
yunohost tools migrations migrate  # this will run the included migration that will remove archivemount
yunohost tools migrations state
yunohost tools migrations migrate -t 0
yunohost tools migrations state
yunohost tools migrations migrate -f
yunohost tools migrations state
```

You can also write another migration. For that you need to respect the file naming convention.

Any new migration needs to increase the number at the beginning of the file.

### Reviews

- [ ] Agree in principle (0/3)
- [ ] Simple tests (0/2)
- [ ] Quick review (0/2)
- [ ] Deep tests (0/1)
- [ ] Deep reviews (0/1)


Psycojoker:
Since it has been asked: this framework migration design is inspired (on the api side) by Django's one.

A migration file looks like this : 

```python
from yunohost.migrations import Migration


class MyMigration(Migration):
    def forward(self):
        # do some stuff
```

And need to be placed in `src/yunohost/data_migrations/` and the current state is saved in `/etc/yunohost/migrations_state.json`.

alexAubin:
What's the status on this ? Shall we try / review it ?

Psycojoker:
Last time I've tough about it, it needs testing and we also need to put exception handling in a lot of places.

But it's been a long time, we probably miss things.

This is a feature that we'll want to test **a lot**.

Psycojoker:
Should we try to have this ready for this release?

alexAubin:
I'd say it mostly depends of how much time is needed to test / validate it :D. But definetely soon™. Do you have a few steps in mind to test this with some dummy stuff ?

Psycojoker:
We can start by cleaning `/etc/yunohost/services.yaml` to remove useless stuff like that "ssl" service. There is also this permission things for certificates that we want to address.

data/actionsmap/yunohost.yml
```diff
@@ -1289,6 +1289,36 @@ tools:
 
 
 #############################
+#         Migrations        #
+#############################
+migrations:
```
zamentur: I am not sure to understand why a cli/api exposition is needed for this.
Maybe I miss a use case ?

Psycojoker: Debugging and be able to go backward/forward in the migrations if needed if everything has crashed.

alexAubin: That raises the question of : are those migrations meant to be run automatically, or manually ?

On one hand, we'd like them to be automatic. On the other hand, some stuff are tricky like, for instance, the old letsencrypt_ynh app -> new builtin certmanager system, for which migrating is not trivial.

Psycojoker: Automatically when the package is upgraded by apt, we can't ask the user to do that everytime and that would increase the cost of having a new migration a lot which would defeat the purpose of being able to reduce the cost of modification for us.

alexAubin:
Sounds like we won't have this for 2.6 so moving to 2.7

Psycojoker:
I realize that we'll need to set the default migration state when doing a new installation (or launch a command for that like `migrations_migrate(fake=True)`

src/yunohost/migrations.py
```diff
@@ -0,0 +1,170 @@
...
+
+def migrations_migrate(auth):
+    """
+    Perform migrations
+    """
+    state = migrations_state()
+
+    migrations = []
+
+    # loading all migrations that haven't already runned
```
julienmalik: runned is barbarism :)

data/actionsmap/yunohost.yml
```diff
@@ -1477,6 +1477,36 @@ tools:
...
+            api: POST /migrations/migrate
+            configuration:
+                authenticate: all
+                authenticator: ldap-anonymous
+
+
+        ### migrations_state()
+        state:
+            action_help: Perform migrations
+            api: GEt /migrations/state
```
julienmalik: GET

src/yunohost/migrations.py
```diff
@@ -0,0 +1,170 @@
...
+    for migration in migrations_list()::
+        # skip already runnned migrations
+        if state["last_runned_migration"] is not None and\
+           int(migration["number"]) <= int(state["last_runned_migration"]["number"]):
+            continue
+
+        migrations.append({
+            "number": migration["number"],
+            "name": migration["name"]
+            # this is python buildin method to import a module using a name, we use that to import the migration
```
julienmalik: builtin

src/yunohost/migrations.py
```diff
@@ -0,0 +1,170 @@
...
+            # has a python object so we'll be able to run it in the next lop
```
julienmalik: loop

src/yunohost/migrations.py
Psycojoker: That's way cooler :grin:

Psycojoker:
@julienmalik thanks, corrected :)

Psycojoker:
This should be good for review (but maybe not yet for prod).

I still have to write how to test it and so on.

This is really a ripoff of django migration framework (at least the principles, django's migration are way more powerful with way more things to handle).

Psycojoker:
I've updated the first post following how @alexAubin is currently prefixing his PRs.

Psycojoker:
Feedback from meeting: maybe migrations should not be a "first level command" on the cli interface (maybe put in "tools" or in another section in the `--help`?)

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+        if not fake:
+            logger.warn("Runing migration {number} {name}...".format(**migration))
+
+            try:
+                if mode == "forward":
+                    migration["module"].MyMigration().migrate()
+                elif mode == "backward":
+                    migration["module"].MyMigration().backward()
+                else:  # can't happend
+                    raise Exception()
```
julienmalik: without a message ? maybe better with RuntimeError("illegal state") or something similar ?

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+    for migration in migrations:
+        if not fake:
+            logger.warn("Runing migration {number} {name}...".format(**migration))
+
+            try:
+                if mode == "forward":
+                    migration["module"].MyMigration().migrate()
+                elif mode == "backward":
+                    migration["module"].MyMigration().backward()
+                else:  # can't happend
```
julienmalik: s/happend/happen

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+        # not data migrations present, return empty list
+        return migrations
+
+    migrations_path = data_migrations.__path__[0]
+
+    if not os.path.exists(migrations_path):
+        logger.warn("Can't access migrations files at path %s".format(migrations_path))
+        return migrations
+
+    for migration in filter(lambda x: re.match("^\d+_[a-zA-Z0-9_]+\.py$", x), os.listdir(migrations_path)):
```
julienmalik: you don't enforce the nupmber of starting digits here. is it safe to sort afterwards ?

data/actionsmap/yunohost.yml
```diff
@@ -1477,6 +1477,39 @@ tools:
...
+        migrate:
+            action_help: Perform migrations
+            api: POST /migrations/migrate
+            arguments:
+              -t:
+                  help: target migration number (or 0), latest one by default
+                  type: int
+                  full: --target
+              -f:
+                  help: fake run the migration, use it only if you know what you are doing
```
julienmalik: ah, and from the example, even less. because in my head, "-f" is "--force"

Psycojoker: probably going to renamed this to "skip" then, "fake" is a commonly used term in migrations framework but this culture is apparently not as spread as a I though

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+        logger.debug("Loading migration {number} {name}...".format(
+            number=migration["number"],
+            name=migration["name"],
+        ))
+
+        try:
+            module = import_module("yunohost.data_migrations.{file_name}".format(**migration))
+        except Exception:
+            import traceback
+            traceback.print_exc()
```
Psycojoker: yes, it prints the whole traceback, otherwise only having the exception name is the equivalent of a "fuck you" for debugging.

src/yunohost/migrations.py
Psycojoker: Well, that would be a situation where we, the yunohost team, accept a PR with a badly named file, I took the hypothesis that this won't happen.

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+    try:
+        import data_migrations
+    except ImportError:
+        # not data migrations present, return empty list
+        return migrations
+
+    migrations_path = data_migrations.__path__[0]
+
+    if not os.path.exists(migrations_path):
+        logger.warn("Can't access migrations files at path %s".format(migrations_path))
```
JimboJoe: Does the exist method really check you can access the files?
If not, the warning should stay factual: "Path %s for migration files doesn't exist".

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+        return
+
+    all_migration_numbers = [x["number"] for x in migrations]
+
+    if target is None:
+        target = migrations[-1]["number"]
+
+    # validate input, target must be "0" or a valid number
+    elif target != 0 and target not in all_migration_numbers:
+        raise MoulinetteError(errno.EINVAL, "Invalide number for target argument, available migrations numbers are 0 or {}".format(", ".join(map(str, all_migration_numbers))))
```
JimboJoe: "numbers" --> "identifiers"?

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+                number=migration["number"],
+                name=migration["name"],
+            ))
+            break
+
+        migrations.append({
+            "number": migration["number"],
+            "name": migration["name"],
+            # this is python builtin method to import a module using a name, we use that to import the migration
+            # has a python object so we'll be able to run it in the next loop
```
Psycojoker: It's generally super fast, especially for those migrations, the whole moulinette startup time is several magnitude of time greater than that.

src/yunohost/migrations.py
Psycojoker: I think that "number" would be a more obvious error message for the user, telling they that they need to uses ... a number.

data/actionsmap/yunohost.yml
```diff
@@ -1477,6 +1477,39 @@ tools:
...
+    category_help: Manage migrations
+    actions:
+
+        ### migrations_list()
+        list:
+            action_help: List migrations
+            api: GET /migrations
+
+        ### migrations_migrate()
+        migrate:
```
alexAubin: Can we rename this 'apply' ? :D 

`yunohost migrations apply` seems more beautiful than `yunohost migrations migrate`

data/actionsmap/yunohost.yml
```diff
@@ -1477,6 +1477,39 @@ tools:
...
+                  full: --target
+              -f:
+                  help: fake run the migration, use it only if you know what you are doing
+                  full: --fake
+                  action: store_true
+
+
+        ### migrations_state()
+        state:
+            action_help: Perform migrations
```
alexAubin: Copy/pasting typo, this probably does not 'perform migrations' :P

data/actionsmap/yunohost.yml
alexAubin: Havent looked yet at what it does, but `--skip` would be more intuitive for me too.

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+        logger.warn("no migrations to run")
+        return
+
+    logger.debug("last runed migrations is {}".format(last_runed_migration_number))
+
+    # we need to run missing migrations
+    if last_runed_migration_number < target:
+        logger.debug("migrating forward")
+        # drop all already runed migrations
+        migrations = filter(lambda x: x["number"] > last_runed_migration_number, migrations)
```
alexAubin: Should'nt you add x["number"] < target (or <=) also ?

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+        raise Exception()
+
+    # effectively run selected migrations
+    for migration in migrations:
+        if not fake:
+            logger.warn("Runing migration {number} {name}...".format(**migration))
+
+            try:
+                if mode == "forward":
+                    migration["module"].MyMigration().migrate()
```
alexAubin: Really a minor detail / suggestion, but maybe it could be more beautiful to call just `migration["module"].MyMigration().migrate(mode)`, then in the class, `if mode == "forward": self.forward()` and so on.

alexAubin:
Not tested yet, but code looks pretty much alright so far for me (just a few minor details to tweak/fix maybe).

About migrations themselves, I wonder if we shouldn't have a flag and mechanism to say that a migration can be runned automatically (i.e. right after an upgrade of the package) or must be runned manually (thinking about risky/critical migration like the future stretch migration where *we want a human being looking at what happens in direct* and not risk to mess up the whole setup automatically)

data/actionsmap/yunohost.yml
Psycojoker: "migrate" is the stand way of calling that in those framework :/

src/yunohost/migrations.py
Psycojoker: oh god right, thanks for noticing >.<'

src/yunohost/migrations.py
Psycojoker: Not sure about what would that bring :s

Psycojoker:
> About migrations themselves, I wonder if we shouldn't have a flag and mechanism to say that a migration can be runned automatically (i.e. right after an upgrade of the package) or must be runned manually (thinking about risky/critical migration like the future stretch migration where we want a human being looking at what happens in direct and not risk to mess up the whole setup automatically)

I think that we can keep the idea and implement it if it's needed? That doesn't seems hard.

Psycojoker:
Normally implemented mostly all fix/comments, thx everyone <3

src/yunohost/migrations.py
```diff
@@ -0,0 +1,249 @@
...
+
+    return migrations
+
+
+def migrations_migrate(target=None, skip=False):
+    """
+    Perform migrations
+    """
+
+    # state is a datastructure that represent the last run migration
```
JimboJoe: sorry to insist: represent**s**... :innocent:

Psycojoker: fixed, thx :)

debian/postinst
```diff
@@ -14,6 +14,9 @@ do_configure() {
       echo "Regenerating configuration, this might take a while..."
       yunohost service regen-conf --output-as none
 
+      echo "Launching migrations.."
+      yunohost migrations migrate
```
Psycojoker: I hope this is the good place to do it, I did it in the same place than the regen-conf assuming this was the correct location.

data/actionsmap/yunohost.yml
alexAubin: Okay, indeed after thinking about it, if you go 'forward', it wouldnt make much sense to have 'apply'. 'migrate' is okay :+1:

Psycojoker:
I've move the "migrations" command to a subcommand of "tools" following feedback and @alexAubin works on this PR https://github.com/YunoHost/moulinette/pull/143

I'm considering this PR done and ready to be merged since all feedback has been implemented and that I don't see anything more to be done.

alexAubin:
I realized when testing the `postinst` that we cannot run a migrations which use `apt-get` *inside* apt-get itself :)

```
Unpacking yunohost (2.6.7) over (2.6.7) ...
Setting up yunohost (2.6.7) ...
Regenerating configuration, this might take a while...
Warning: The configuration file '/etc/ssh/sshd_config' has been manually modified and will not be updated
Launching migrations..
Warning: Running migration 1 archivemount...
E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?
Error: Migration 1 archivemount has failed with exception Command 'apt-get remove archivemount -y' returned non-zero exit status 100, aborting
Restarting YunoHost firewall...
```

Psycojoker:
(I'm obviously approving final modifications)

PR #196, opened 2016-11-28
[enh] Automatically check for weak password

Psycojoker:
EDIT: LJF
## The problem

Some administrator (like CHATONS) could want to force user to choose good password.

## Solution

With this solution administrator can choose the level of password he wants for 'admin' and for 'user' profile.

Level of password:
-1 disabled all check
0 Just give advice if the password is weak
1 Don't accept password with 6 letters or listed in 100000 list of password
2 At least 8 with upper case and numeric
3 With special characters in more
4 12 characters

## How to test
apt install python-cracklib
mkdir -p /usr/local/share/dict/cracklib/
cp /vagrant/yunohost/data/other/password/* /usr/local/share/dict/cracklib/

## PR Status

Ready to be reviewed.
Related PR https://github.com/YunoHost/SSOwat/pull/104

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


julienmalik:
I would still like to see something like this in Yunohost.
But this needs thinking at a wider scale. password can be changed from several places (cli, admin, sso), and I would prefer to opt for a common strategy (a single place of code to maintain).

julienmalik:
Just rebased on current unstable

alexAubin:
I think this is an interesting feature that should definitely come sometime soon. I'd vote to put this in milestone 2.6.x.

julienmalik:
Just saw nextcloud implemented something interesting : a check against the HaveIBennPwned db

See https://help.nextcloud.com/t/nextcloud-will-check-passwords-against-database-of-haveibeenpwned/28110 (there's a link to the pr in it)

zamentur:
Here is the nextcloud implementation :
https://github.com/nextcloud/password_policy/pull/60/files#diff-f8b12744adebdfdefc9850bb7fc8f928R174

zamentur:
A repo with a lot of password:
https://github.com/danielmiessler/SecLists

zamentur:
Note : check directly with ldap could be an idea, but I found nothing especially good enough.

zamentur:
screen 1
screen2

src/yunohost/settings.py
```diff
@@ -90,9 +100,12 @@ def settings_set(key, value):
                 received_type=type(value).__name__, expected_type=key_type))
     elif key_type == "int":
         if not isinstance(value, int) or isinstance(value, bool):
```
alexAubin: Not sure to understand this condition ... if `value` is a `bool`, it's not an `int` anyway ? So the second condition is useless ? Or am I too sleepy ?

src/yunohost/settings.py
```diff
@@ -30,11 +30,21 @@
 # * enum (in form a python list)
 
 # we don't store the value in default options
+PWD_MODE = ["disabled", "weak", "strong"]
+PWD_CHOICES = ["error", "warn_only", "disabled"]
+PWD_DEFAULT_ERROR = {"type": "enum", "default": "error",
+                     "choices": PWD_CHOICES}
```
alexAubin: Those three variables don't seem to be used in fact ? Or am I missing something ?

src/yunohost/tools.py
```diff
@@ -127,6 +129,9 @@ def tools_adminpw(auth, new_password):
 
     """
     from yunohost.user import _hash_user_password
+    from yunohost.utils.password import LoggerPasswordValidator
+
+    LoggerPasswordValidator('admin').validate(new_password)
```
alexAubin: Those two lines are repeated at several places ... I'd rather have a call to `tools_validatepw` (or something like `_assert_password_is_strong_enough`)

src/yunohost/utils/password.py
```diff
@@ -0,0 +1,197 @@
...
+import cracklib
+
+import string
+ASCII_UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+ASCII_LOWERCASE = "abcdefghijklmnopqrstuvwxyz"
+PWDDICT_PATH = '/usr/local/share/dict/cracklib/'
+SMALL_PWD_LIST = ["yunohost", "olinuxino", "olinux", "raspberry", "admin",
+                  "root", "test", "rpi"]
+PWD_LIST_FILE = '100000-most-used'
+ACTIVATE_ONLINE_PWNED_LIST = False
```
alexAubin: This is set to False and doesn't seem to be set to True anywhere ... so, are we using it or not ? :s

src/yunohost/utils/password.py
```diff
@@ -0,0 +1,197 @@
...
+
+        logger = logging.getLogger('yunohost.utils.password')
+
+        status, msg = super(LoggerPasswordValidator, self).validate(password)
+        if status == "error":
+            raise MoulinetteError(1, m18n.n(msg))
+        elif status == "warning":
+            logger.info(m18n.n(msg))
+
+if __name__ == '__main__':
```
alexAubin: Should have a comment here explaining why this is needed and why not use `yunohost tools validatepw` instead

src/yunohost/utils/password.py
```diff
@@ -0,0 +1,197 @@
...
+            cracklib.VeryFascistCheck(password, None,
+                                      os.path.join(PWDDICT_PATH, pwd_dict))
+        except ValueError as e:
+            # We only want the dictionnary check of cracklib, not the is_simple
+            # test.
+            if str(e) not in ["is too simple", "is a palindrome"]:
+                return True
+
+
+class ProfilePasswordValidator(PasswordValidator):
```
alexAubin: Eh are those three different classes really needed ? Isnt there any way to merge one or all of them together ? What are the differences between them ?

src/yunohost/utils/password.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, validation_strength):
+        self.validation_strength = validation_strength
+
+    def validate(self, password):
+        """
+        Validate a password and raise error or display a warning
+        """
+        if self.validation_strength <= 0:
+            return ("success", "")
```
alexAubin: This contradicts the description in another file ? 

```
# -1 disabled, 0 alert if listed
```

alexAubin:
Discussing this with ljf, we ended up concluding that cracklib is too nazi and doesn't offer much flexibility with regard to how the "being in the list of most used password" is done. In particular, it detects passwords that are close to password in the list ... which ends up being annoying because legitime passwords trigger the check.

Simply using a list from https://github.com/danielmiessler/SecLists/tree/master/Passwords/Common-Credentials and grep turns out to be much simpler and at least we have full control on what's done and the outcome is much clearer.

src/yunohost/tools.py
```diff
@@ -32,6 +32,7 @@
 import subprocess
 import pwd
 import socket
+import cracklib
```
zamentur: We probably should remove that

src/yunohost/user.py
```diff
@@ -32,6 +32,7 @@
 import random
 import string
 import subprocess
+import cracklib
```
zamentur: Same here

src/yunohost/tools.py
alexAubin: Done

PR #198, opened 2016-11-30
[fix] Correctly handle custom backup directory

alexAubin:
This fixes https://dev.yunohost.org/issues/618

After fixing the reported bug, I realized that when using -o to set a custom directory, code didn't keep track of the path to the backup. I fixed it by adding a symlink to the archive + an exception handling later in case the link is broken.

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -120,8 +120,10 @@ def backup_create(name=None, description=None, output_directory=None,
             env_var['CAN_BIND'] = 0
     else:
         output_directory = archives_path
-        if not os.path.isdir(archives_path):
-            os.mkdir(archives_path, 0750)
+
+    # Create archives directory if it does not exists
+    if not os.path.isdir(archives_path):
+        os.mkdir(archives_path, 0750)
 
     def _clean_tmp_dir(retcode=0):
         ret = hook_callback('post_backup_create', args=[tmp_dir, retcode])
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -623,15 +623,16 @@ def backup_info(name, with_details=False, human_readable=False):
             m18n.n('backup_archive_name_unknown', name=name))
 
     # If symlink, retrieve the real path
-    if os.path.islink(archive_file) :
+    if os.path.islink(archive_file):
         archive_file = os.path.realpath(archive_file)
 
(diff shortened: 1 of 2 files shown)


src/yunohost/backup.py
```diff
@@ -602,10 +616,21 @@ def backup_info(name, with_details=False, human_readable=False):
 
     """
     archive_file = '%s/%s.tar.gz' % (archives_path, name)
-    if not os.path.isfile(archive_file):
+
+    # Check file exist (even if it's a broken symlink)
+    if not os.path.lexists(archive_file):
```
opi: `lexists` ?

M5oul: It seems to be `os.path.exists` according [to this](https://stackoverflow.com/questions/15077424/python-os-path-exists-vs-os-path-isdir).

So, a typo, an `l` surplus.

alexAubin: Not a typo 😸 

> os.path.exists(path)
> 
>     Return True if path refers to an existing path. Returns False for broken symbolic links. [...]
> 
> os.path.lexists(path)
> 
>     Return True if path refers to an existing path. Returns True for broken symbolic links. [..,]

src/yunohost/backup.py
```diff
@@ -602,10 +616,21 @@ def backup_info(name, with_details=False, human_readable=False):
...
         raise MoulinetteError(errno.EIO,
             m18n.n('backup_archive_name_unknown', name=name))
 
+    # If symlink, retrieve the real path
+    if os.path.islink(archive_file) :
+        archive_file = os.path.realpath(archive_file)
+
+        # Raise exception if link is broken (e.g. on unmounted external storage)
+        if (not os.path.exists(archive_file)) :
```
Psycojoker: chat

alexAubin:
Bumpity bump. Anyone wants to review this ?

To test it, try : 
```
yunohost backup create -o /some/place --ignore-apps
yunohost backup list
```
before and after the fix.

Psycojoker:
Tested, works as expected.

likeitneverwentaway:
Tested, working great.

zamentur:
I test it tomorrow

Psycojoker:
We have enough opinion and feedback to merge so I'm not going to wait since I have an available window of time to do a release :)

PR #199, opened 2016-12-02
[fix] Fix name of self-CA in certificate manager

alexAubin:
Fixes https://dev.yunohost.org/issues/652

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -817,9 +817,9 @@ def _name_self_CA():
             lines = f.readlines()
 
         for line in lines:
-            if (line.startswith("commonName_default")):
+            if line.startswith("commonName_default"):
                 return line.split()[2]
-    except :
+    except:
         pass
 
     logger.warning(m18n.n('certmanager_unable_to_determine_self_CA_name'))
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -813,7 +813,7 @@ def _name_self_CA():
     ca_conf = os.path.join(SSL_DIR, "openssl.ca.cnf")
 
     try :
-        with open("%s/openssl.ca.cnf" % SSL_DIR) as f:
+        with open(ca_conf) as f:
             lines = f.readlines()
 
         for line in lines:
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -812,17 +812,18 @@ def _domain_is_accessible_through_HTTP(ip, domain):
 def _name_self_CA():
     ca_conf = os.path.join(SSL_DIR, "openssl.ca.cnf")
 
-    try :
-        with open(ca_conf) as f:
-            lines = f.readlines()
+    if not os.path.exists(ca_conf) :
+        logger.warning(m18n.n('certmanager_self_ca_conf_file_not_found', file=ca_conf))
+        return ""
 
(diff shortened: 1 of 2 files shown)


opi:
I'd like other people to test this PR by generating self-signed certificate, but as it is a bugfix, I don't wan't to wait 2 weeks before merging. This need to be shipped in the next stable release, along LE integration work.

src/yunohost/certificate.py
```diff
@@ -810,9 +810,20 @@ def _domain_is_accessible_through_HTTP(ip, domain):
...
+    ca_conf = os.path.join(SSL_DIR, "openssl.ca.cnf")
+
+    try :
+        with open(ca_conf) as f:
+            lines = f.readlines()
+
+        for line in lines:
+            if line.startswith("commonName_default"):
+                return line.split()[2]
+    except:
```
Psycojoker: We should never silencely hide exception, this is always a debugging nightmare, we should either handle the case we think that could fail in the code and/or printing warning on exception catach

alexAubin: My try/except was aiming to catch the case where ca_conf doesn't exist, essentially... and to not block the rest of the program if something bad happens (hence the `logger.warning`). Would you like me to replace the try/except by something like `if os.path.exists(ca_conf)` ?

Psycojoker: You can do both and print the exception + the stack if something bad happen, but never silencly hide exception, this is really really bad for debugging :/

src/yunohost/certificate.py
```diff
@@ -810,9 +810,20 @@ def _domain_is_accessible_through_HTTP(ip, domain):
 
 
 def _name_self_CA():
-    cert = crypto.load_certificate(
-        crypto.FILETYPE_PEM, open(SELF_CA_FILE).read())
-    return cert.get_subject().CN
+    ca_conf = os.path.join(SSL_DIR, "openssl.ca.cnf")
```
julienmalik: Seems a little wrong to me to get the info you're looking for from this file.
Shouldn't it be interpreted from the cert itself ?

Is it what you are looking for ? :
```
openssl  x509 -text -noout -in /usr/share/yunohost/yunohost-config/ssl/yunoCA/ca/cacert.pem
```
then grep for ```Subject : CN=domain.com```

julienmalik: There is a ```Subject : CN=``` and a ```Issuer : CN=``` whenprinting the cert content.
So if both domains are the same, then it is self-signed ?

julienmalik: maybe doable with this crypto lib, I don't know.

alexAubin: This is what was done before, actually... But for some reason, opi had a setup where he changed the `cacert.pem` with something else. The "true" info is coming from the `openssl.ca.cnf`, from what I investigated (could be wrong, looks like there's some obscure magic going on in openssl).

opi: maybe we can forget about my strange setup ? ( bit of history: I used to have sefl-signed certificate, with CA on main domain; then change the main domain, then dirty hack to set up a gandi certificate, and then hack with letsencrypt by hand ).

src/yunohost/certificate.py
alexAubin: Should be fixed in 6aa64a0

PR #200, opened 2016-12-02
remove timeout from cli interface

Psycojoker:
This PR is needed for https://github.com/YunoHost/moulinette/pull/129


Psycojoker:
If you could review this PR along the [moulinette's one](https://github.com/YunoHost/moulinette/pull/129) that would be great

PR #203, opened 2016-12-05
[fix] Add missing dependency to nscd package #656

opi:
See https://dev.yunohost.org/issues/656

nscd is called during user_create and user_delete, but this package is only in Recommends for libnss-ldapd.


alexAubin:
c.f. https://github.com/YunoHost/yunohost/pull/172#issuecomment-264765164, this might be the root cause.

PR #204, opened 2016-12-05
[fix] don't remove trailing char if it's not a slash

Psycojoker:
Attempt to fix https://dev.yunohost.org/issues/655

ping @maniackcrudelis could you test and confirm that this works for you plz?

I've done a minimal solution, this part of the code should be cleaned a bit more.

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1011,19 +1011,19 @@ def app_ssowatconf(auth):
                 for item in _get_setting(app_settings, 'skipped_uris'):
                     if item[-1:] == '/':
                         item = item[:-1]
-                    skipped_urls.append(app_settings['domain'] + app_settings['path'][:-1] + item)
+                    skipped_urls.append(app_settings['domain'] + app_settings['path'].rstrip('/') + item)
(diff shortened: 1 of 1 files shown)


Psycojoker:
```
19:20  <ManiackC> @Bram à première vue ça marche niquel ton fix
19:20  <ManiackC> Sans le slash, ce n'est pas tronqué et avec il n'est pas en double
```

Psycojoker:
Looking again at it I think that `os.path.join` would be better but I still want to do the most minimalist patch possible to avoid breaking other things.

Psycojoker:
Ah didn't know you could give [rl]strip an argument, that looks indeed better.

likeitneverwentaway:
I had the problem with every installed app, this correctly fixes the ssoconf.

Psycojoker:
@alexAubin should be ok

@maniackcrudelis can you test again plz?

maniackcrudelis:
Tested, no regression on "ssowat/conf.json" with rstrip.

Psycojoker:
Thx <3

PR #206, opened 2016-12-11
[Fix] Add check and remove function for domain default app

likeitneverwentaway:
This fixes [125](https://dev.yunohost.org/issues/125) and [40](https://dev.yunohost.org/issues/125). I renamed the api call to make an app the default one for the domain "default" to "makedefault".

If I understand correctly we have to stick with conf.json.persistent rather than the generated conf, as it only contains info for paths after the app path itself, and not the domain

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -797,7 +797,7 @@ def app_makedefault(auth, app, domain=None):
 
     app_settings = _get_app_settings(app)
     app_domain = app_settings['domain']
-    app_path   = app_settings['path']
+    app_path = app_settings['path']
 
     if domain is None:
         domain = app_domain
@@ -820,7 +820,7 @@ def app_makedefault(auth, app, domain=None):
     if 'redirected_urls' not in ssowat_conf:
         ssowat_conf['redirected_urls'] = {}
 
-    ssowat_conf['redirected_urls'][domain +'/'] = app_domain + app_path
+    ssowat_conf['redirected_urls'][domain + '/'] = app_domain + app_path
 
     try:
         with open('/etc/ssowat/conf.json.persistent', 'w+') as f:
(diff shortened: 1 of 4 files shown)


Psycojoker:
@zamentur this API changes imply we'll also need to modify the admin interface, right?

src/yunohost/app.py
```diff
@@ -829,12 +829,110 @@ def app_makedefault(auth, app, domain=None):
         raise MoulinetteError(errno.EPERM,
                               m18n.n('ssowat_persistent_conf_write_error', error=e.strerror))
 
+    os.system('chmod 644 /etc/ssowat/conf.json.persistent')
```
Psycojoker: Uses https://docs.python.org/2/library/os.html#os.chmod instead and check for file existance before to avoid a bug.

src/yunohost/app.py
```diff
@@ -829,12 +829,110 @@ def app_makedefault(auth, app, domain=None):
...
+
+    Keyword argument:
+        app
+        domain
+
+    """
+    from yunohost.domain import domain_list
+
+    app_settings = _get_app_settings(app)
+    app_domain = app_settings['domain']
```
Psycojoker: You don't have any guarantee that [`domain` will be present in `app_settings`](https://github.com/likeitneverwentaway/yunohost/blob/252928717d2c0210fa00ac13f769e9be48333362/src/yunohost/app.py#L303), you need to do a `app_settings.get('domain')` instead and return if this app doesn't have a domain (some does).

src/yunohost/app.py
```diff
@@ -829,12 +829,110 @@ def app_makedefault(auth, app, domain=None):
...
+    app_path = app_settings['path']
+
+    if domain is None:
+        domain = app_domain
+    elif domain not in domain_list(auth)['domains']:
+        raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
+
+    if '/' in app_map(raw=True)[domain]:
```
Psycojoker: This way of testing if a domain can have an app installed on it should probably be tested in a helper.

src/yunohost/app.py
```diff
@@ -829,12 +829,110 @@ def app_makedefault(auth, app, domain=None):
...
+        raise MoulinetteError(errno.EEXIST,
+                              m18n.n('app_location_already_used'))
+
+    try:
+        with open('/etc/ssowat/conf.json.persistent') as json_conf:
+            ssowat_conf = json.loads(str(json_conf.read()))
+    except ValueError as e:
+        raise MoulinetteError(errno.EINVAL,
+                              m18n.n('ssowat_persistent_conf_read_error', error=e.strerror))
+    except IOError:
```
Psycojoker: We should display a warning, never silencly hide errors.

src/yunohost/app.py
```diff
@@ -829,12 +829,110 @@ def app_makedefault(auth, app, domain=None):
...
+
+    Keyword argument:
+        app
+        domain
+
+    """
+    from yunohost.domain import domain_list
+
+    app_settings = _get_app_settings(app)
+    app_domain = app_settings['domain']
```
Psycojoker: Same comment than before.

src/yunohost/app.py
```diff
@@ -829,12 +829,110 @@ def app_makedefault(auth, app, domain=None):
...
+    app_path = app_settings['path']
+
+    if domain is None:
+        domain = app_domain
+    elif domain not in domain_list(auth)['domains']:
+        raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
+
+    if '/' in app_map(raw=True)[domain]:
```
Psycojoker: Idem

Psycojoker:
Okay, I should have read the global principle before doing code review. I think that using `conf.json.persistant` is not the way we should do that since this file only intent is to be edited by the user, we should never touch it.

It looks like it's a bug in the (re)generation system for `ssoconf` so this current approach is not correct :/

likeitneverwentaway:
Thanks for the review! Handling errors is precisely what I need to improve on, that's good. I'll also check if I can fix the ssoconf. 

> this API changes imply we'll also need to modify the admin interface, right?

Yes, that's why I created a check function, so that we can either display a green "Make default" button or a red/orange "Remove default" on a installed app info.

PR #208, opened 2016-12-11
[fix] Domain default app

a1ex4:
This fixes [125](https://dev.yunohost.org/issues/125) and [40](https://dev.yunohost.org/issues/40). Default domain redirection are now set in the app settings, with `is_domain_default: 'Yes'`. Then this setting is handled by ssowatconf and add the redirection in conf.json, if the setting is not there then nothing is added. When removing a default app the redirection will also be deleted as it is now handled with the app setting and ssowatconf.

### DIFF
--- a/locales/fr.json
+++ b/locales/fr.json
@@ -15,6 +15,7 @@
     "app_location_already_used": "Une application est déjà installée à cet emplacement",
     "app_location_install_failed": "Impossible d'installer l'application à cet emplacement",
     "app_manifest_invalid": "Manifeste d'application incorrect",
+	"app_no_domain": "Cette application n'a pas de domaine",
     "app_no_upgrade": "Aucune application à mettre à jour",
     "app_not_correctly_installed": "{app:s} semble être mal installé",
     "app_not_installed": "{app:s} n'est pas installé",


locales/en.json
```diff
@@ -14,6 +14,7 @@
     "app_location_already_used": "An app is already installed in this location",
     "app_location_install_failed": "Unable to install the app in this location",
     "app_manifest_invalid": "Invalid app manifest",
+	"app_no_domain": "This app doesn't have a domain",
```
julienmalik: TAB here

locales/fr.json
```diff
@@ -15,6 +15,7 @@
     "app_location_already_used": "Une application est déjà installée à cet emplacement",
     "app_location_install_failed": "Impossible d'installer l'application à cet emplacement",
     "app_manifest_invalid": "Manifeste d'application incorrect",
+	"app_no_domain": "Cette application n'a pas de domaine",
```
julienmalik: do not modify this file, it must be handled in weblate

alexAubin:
This looks like an important PR/fix, but I'm worried about the actionmap.

IMHO, this being managed in the "app" category is not semantic. It's much more related to a given domain than to a given app. It'd make more sense to have : 

```bash
# What's the default app for domain ilike.coffee.com?
yunohost domain defaultapp ilike.coffee.com
> e.g. would return "None" or the default app...

# Set the default app as wordpress for this domain
yunohost domain defaultapp ilike.coffee.com --set wordpress

# Remove the default app
yunohost domain defaultapp ilike.coffee.com --clear
```

We could keep the current "app makedefault" for backward compatibility (and maybe depreciate it). Same kind of remark goes for "app ssowatconf" which should be in tools imho, but that's another topic.

What do you guys think ?

Psycojoker:
For reference, since this branch has been delete, I've pushed the relevant commits here https://github.com/YunoHost/yunohost/tree/likeitneverwentaway_fix_domain_default_app

alexAubin:
[Semi-automatic message] 

PR #209, opened 2016-12-11
[fix] Missing 'python-openssl' dependency for Let's Encrypt integration.

M5oul:
[#662](https://dev.yunohost.org/issues/662): missing `python-openssl` dependency for Let's Encrypt integration.


Psycojoker:
I think that you should have added this on the next line for style but this is really a trivial detail.

PR #215, opened 2016-12-16
[fix] Wrong fail2ban regex for yunohost

zimo2001:
In my nginx_error.log, the authentication errors are "helpers.lua", not "access.lua".

PR #216, opened 2016-12-17
[enh] Increase fail2ban maxretry on user login, narrow nginx log files

opi:
We're intended to a general public, after all ;)

### DIFF
--- a/data/templates/fail2ban/jail.conf
+++ b/data/templates/fail2ban/jail.conf
@@ -581,4 +581,5 @@ enabled  = true
 port     = http,https
 protocol = tcp
 filter   = yunohost
-logpath  = /var/log/nginx/*.log
+logpath  = /var/log/nginx*/*error.log
+maxretry = 6


Psycojoker:
Is this link to #215?

opi:
#215 only makes me have a look into fail2ban configuration ;)

Psycojoker:
Can we display this on the user interface? After a failed login

opi:
I don't know; From my tests, It looks like a timeout after 3 failed login. I got to investigate fail2ban configuration

data/templates/fail2ban/jail.conf
```diff
@@ -581,4 +581,5 @@ enabled  = true
 port     = http,https
 protocol = tcp
 filter   = yunohost
-logpath  = /var/log/nginx/*.log
+logpath  = /var/log/nginx*/*error.log
```
zamentur: I checked on 2 distinct yunohost, I have one `nginx` directory but no need `nginx*`. Are you sure it's needed ?

opi: nope, it's probably not needed; It's only copy/pasta from others nginx sections in jail.conf (https://github.com/YunoHost/yunohost/blob/unstable/data/templates/fail2ban/jail.conf#L554-L576)

opi:
should we replace every `nginx*` occurences with `nginx` ? I don't really care, but I want consistent configuration across the whole file !

opi:
up!

as #215 is fixed and released (aka "fail2ban will work on user login"), I really like this improvment to be merged as well, because 2 or 3 wrong password attemp seems low to me.

Psycojoker:
@opi would like this PR to be merged in 2.5 before the stable release. Do we have any opposition with that? I'm ok with merging it.

alexAubin:
Yup I'm okay with it

julienmalik:
ok let's merge this one

Psycojoker:
That's enough positive opinions :)

PR #217, opened 2016-12-17
[enh] Add a script to reset ladp admin password

Psycojoker:
Hello,

I'm opening this PR both to start the work and both to start the discussion.

* do we want an utility to reset ldap admin password?
* if so, how do we want to do it?

This is a common request but making it super available doesn't necessary seems like the best idea in the world.

If we do it I would recommend that we have a separate script only executable by root (which I think that you still need to be root to do all the operations).

### DIFF
--- a/debian/install
+++ b/debian/install
@@ -1,4 +1,5 @@
 bin/* /usr/bin/
+sbin/* /usr/sbin/
 data/bash-completion.d/yunohost /etc/bash_completion.d/
 data/actionsmap/* /usr/share/moulinette/actionsmap/
 data/hooks/* /usr/share/yunohost/hooks/
--- a/sbin/yunohost-reset-ldap-password
+++ b/sbin/yunohost-reset-ldap-password
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+################################
+#   Set a temporary password   #
+################################
+
+# Generate a random temporary password (won't be valid after this script ends !)
+# and hash it
+TMP_LDAPROOT_PASSWORD=`slappasswd -g`
+TMP_LDAPROOT_PASSWORD_HASH=`slappasswd -h {SSHA} -s ${TMP_LDAPROOT_PASSWORD}`
+
+# Stop slapd service...
+service slapd stop
+
+# Backup slapd.conf (to be restored at the end of script)
+cp /etc/ldap/slapd.conf /root/slapd.conf.bkp
+
+# Append lines to slapd.conf to manually define root password hash
+echo 'rootdn "cn=admin,dc=yunohost,dc=org"' >> /etc/ldap/slapd.conf
+echo "rootpw $TMP_LDAPROOT_PASSWORD_HASH" >> /etc/ldap/slapd.conf
+
+# Test conf (might not be entirely necessary though :P)
+slaptest -Q -u -f /etc/ldap/slapd.conf
+
+# Regenerate slapd.d directory
+rm -Rf /etc/ldap/slapd.d
+mkdir /etc/ldap/slapd.d
+slaptest -f /etc/ldap/slapd.conf -F /etc/ldap/slapd.d/ 2>&1
+
+# Set permissions to slapd.d
+chown -R openldap:openldap /etc/ldap/slapd.d/
+
+# Restore slapd.conf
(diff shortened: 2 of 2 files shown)


Psycojoker:
Oh and to do so, I've included @alexAubin script for that (and done a commit on his name) since it's currently the working way to do so.

Of course this PR isn't ready for merge.

julienmalik:
This rootdn/rootpw seemed to me the way to go last time I looked into it, for having a sort of permanent "root backdoor" into ldap.

Currently even root is not able to easily modify the ldap content, and this is a pain. I don't think we will be able to implement correctly migrations if we don't have this. We need something like this for the mail alias feature for example.

So instead of having a script which modifies the rootdn in slapd.conf, I would prefer to have this set permanently, with the password stored carefully in a root-only file.
Anyway, with this script, you prove that root actually can change the admin password, so can modify the ldap content if needed. So better be more officially supported.

alexAubin:
I think we should provide such a script, as it is a recurring issue and it's frustrating not being able to just change the admin password despite being root on the machine.

After a few tests, I improved the script, i.e. now it : 
- generates a random tmp password which is temporarily set as root password for the ldap base
- calls `yunohost tools adminpw --admin-password $TMP_PASSWORD` to ask the user the new password
- make ldap forgets about the tmp root password.

I believe that there's a potential security risk during step 2 : the command (including the password given in clear as an argument) could end up in, for instance, `ps -ef`. A malicious user on the machine could make a script waiting for this line to appear, grab the root password and automatically do nasty stuff on the LDAP base while user input the new password... (This is typically why `sudo` or `su` don't allow to give the password as an argument). 

Alternatively, we could print the temporary password in the terminal, and let the user manually type it.

opi:
My 2 cents about this script : good idea, we need an easy way to reset "admin" password when logged-in as root. However, I would like this script to start with "yunohost-" string to avoid confusion on which package provides this utility. I should test it though ;)

zamentur:
@alexAubin it could be possible to forge a bash script (with root owner and permission) with your command:
`#!/bin/bash
yunohost tools adminpw --admin-password $TMP_PASSWORD`
Like that the password don't appear on a `ps -ef` command (need to be checked).

@julienmalik if you spoke to store the admin password of the user in a clear format, I am against (even in a root file). The machine could be accessed physically, this password could be used on other places. If it's an other password, it's different, but I prefer the script method to temporarily add a root password on the ldap

zamentur:
I have tested the bash script idea, it doesn't work...

bin/reset-ldap-password
```diff
@@ -0,0 +1,62 @@
...
+
+################################
+#   Set a temporary password   #
+################################
+
+# Generate a random temporary password (won't be valid after this script ends !)
+# and hash it
+TMP_LDAPROOT_PASSWORD=`slappasswd -g`
+TMP_LDAPROOT_PASSWORD_HASH=`slappasswd -h {SSHA} -s ${TMP_LDAPROOT_PASSWORD}`
+
```
zamentur: I think we should use a SHA512 password ( -h {CRYPT} -c '$6$%.8s'  )

alexAubin: Even for the temporary password ?

julienmalik:
@zamentur  Yes I was thinking of another password, reserved for root unattended modifications to ldap. not the actual admin password that users set at postinstall time.
But I understand this is actually the subject for another issue/discussion/PR. Let's not pollute the discussion here, I'll try to revive this idea later, hopefully with a PR prototype.

alexAubin:
Bumping for glory - Any opinion on this @zamentur @opi @Psycojoker @julienmalik ?

Psycojoker:
Do we need to modify something in the debian package to be able to ship this script in the .deb? I never know how to thing works.

Psycojoker:
According to opi on xmpp, we don't have to do anything, thing are going to be copied thanks to https://github.com/YunoHost/yunohost/blob/unstable/debian/install#L1

opi:
As this script will be available to every user (/usr/bin), we should probably check for "sudo" at first, otherwise any user could launch this script, and stop "slapd" before any permission is checked.
Or maybe copy this script to /usr/sbin ? (I'm not that much into debian to know the differences)

PR #219, opened 2016-12-19
[enh] Set main domain as hostname

alexAubin:
Attempt to address https://dev.yunohost.org/issues/341, based on https://github.com/labriqueinternet/doctorcube_ynh/pull/9

Still need to think about this and do some tests ... Not entirely sure what to do with /etc/hosts and implications.

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -162,6 +162,45 @@ def tools_maindomain(auth, new_domain=None):
         logger.warning("%s" % e, exc_info=1)
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
+    # Set hostname
+    
+    commands = [     
+        "sudo hostnamectl --static set-hostname %s" % new_domain,
+        "sudo hostnamectl --transient set-hostname %s" % new_domain,
+        "sudo hostnamectl --pretty set-hostname (YunoHost/%s)" % new_domain
+    ]
+
+    for command in commands:
+        p = subprocess.Popen(command.split(),
+                             stdout = subprocess.PIPE, 
+                             stderr = subprocess.STDOUT)
+
+        out, _ = p.communicate()
+
+        if p.returncode != 0:
+            logger.warning(command)
+            logger.warning(out)
+            raise MoulinetteError(errno.EIO, m18n.n('domain_hostname_failed'))          # FIXME - To be added in locales/en.json
+        else:
+            logger.info(out) 
+
+    # Add domain to /etc/hosts
+
+    ipv4line = False
+    ipv6line = False
+    with open("/etc/hosts", "r") as f :
+        for line in f.readlines() :
+            if (line.startswith("127.0.0.1 %s" % new_domain)) : 
+                ipv4line = True
+            if (line.startswith("::1 %s" % new_domain)) : 
+                ipv6line = True
+
+    with open("/etc/hosts", "a") as f :
+        if (not ipv4line) :
+            f.write("127.0.0.1 %s\n" % new_domain)
+        if (not ipv6line) :
+            f.write("::1 %s\n" % new_domain)
+
     # Regen configurations
     try:
         with open('/etc/yunohost/installed', 'r') as f:
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -163,17 +163,17 @@ def tools_maindomain(auth, new_domain=None):
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
     # Set hostname
-    
-    commands = [     
-        "sudo hostnamectl --static set-hostname %s" % new_domain,
-        "sudo hostnamectl --transient set-hostname %s" % new_domain,
-        "sudo hostnamectl --pretty set-hostname (YunoHost/%s)" % new_domain
+    pretty_hostname = "(YunoHost/%s)" % new_domain
+    commands = [
+        "sudo hostnamectl --static    set-hostname".split() + [new_domain],
+        "sudo hostnamectl --transient set-hostname".split() + [new_domain],
+        "sudo hostnamectl --pretty    set-hostname".split() + [pretty_hostname]
     ]
 
     for command in commands:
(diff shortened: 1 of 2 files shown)


src/yunohost/tools.py
```diff
@@ -162,6 +162,45 @@ def tools_maindomain(auth, new_domain=None):
         logger.warning("%s" % e, exc_info=1)
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
+    # Set hostname
+    
+    commands = [     
+        "sudo hostnamectl --static set-hostname %s" % new_domain,
+        "sudo hostnamectl --transient set-hostname %s" % new_domain,
+        "sudo hostnamectl --pretty set-hostname (YunoHost/%s)" % new_domain
```
jvaubourg: I think you should add quotes (for bash) around the parenthesis of the pretty hostname.

alexAubin: I tried before committing, and somehow the quotes ended up in the name itself (in hostnamectl status). This probably has to do with the use of subprocess. Subprocess takes as input a list and each element kind of act like an argument. Not like bash where you put quotes specifically to say "this is to be understood as a single argument" (AFAIK).

Psycojoker: You probably want to use the list based approach of subprocess then https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module

Psycojoker:
```
19:43 @<Bram> Aleks: je sais pas s'il faut faire ça dans /etc/hosts
19:43 @<Bram> par contre je vois que ton code est pas assez solide :/
19:43 @<Bram> faudrait utiliser un re
19:44 @<Bram> re.match("127\.0\.0\.1.+ %s"
19:44 @<Bram> un truc du genre
19:44 @<Bram> tu peux avoir plusieurs domains sur une ligne
19:44 @<Bram> et plusieurs espaces/tabs
19:44 @<Bram> c'est assez le merdique
19:45  <Aleks> Oep, c'est un peu la merde de faire un vrai test :/
19:45  <Aleks> Pourrait aussi tenter de resolv le domaine et voir.si ca renvooe 127.0.0.1
```

Psycojoker:
https://github.com/jonhadfield/python-hosts

opi:
is "put the (main) domain(s) in /etc/hosts file" another issue that we need to address in another PR ?

Psycojoker:
Having 2 differents PR would simplify things, indeed.

alexAubin:
Okay, let's stick to only doing this for the main domain for now then.

I'll have a look at providing a better way to parse /etc/hosts as discussed with @Psycojoker.

Psycojoker:
> I'll have a look at providing a better way to parse /etc/hosts as discussed with @Psycojoker.

Regarding this, I think we can actually do something simplier : in neutrinet.sh we only check if the domain is present, then add it and I think that it's totally enough https://github.com/labriqueinternet/configuration_scripts/blob/master/neutrinet.sh#L120

If the domain is present we don't want to add another line that would overwrite the previous one and if it's present it's that the user already did something. We just need a better check than in neutrinet.sh to avoid having false positive result (for example: look for domain.com and being happy because sub.domain.com is already present)

alexAubin:
We could grep on " domain.tld" (with a space before, and a space or end of line after, maybe)

Psycojoker:
Something like `^ *domain.tld` would be better I think (regex)

alexAubin:
We discussed this on the chat with @Psycojoker today, he suggested [this snippet](https://gist.github.com/Psycojoker/a71144e6ea991a88bd66b327a290da0f) to parse /etc/hosts

alexAubin:
Let's postpone this for 2.6, as this is more an improvement than a bug.

Psycojoker:
The `/etc/hosts` modifications should be removed from this PR.

src/yunohost/tools.py
alexAubin: Reviving this discussion : not sure what to do here. The current code works, and I don't see any benefits of using directly a list (I already do a .split() when using subprocess right after - the only limitation is that we cannot have spaces in the pretty hostname atm, but we don't really care imho)

Psycojoker: Looks cleaner and that would avoid stupid bug like "yes you can have a domain name in unicode".

src/yunohost/tools.py
```diff
@@ -162,6 +162,28 @@ def tools_maindomain(auth, new_domain=None):
         logger.warning("%s" % e, exc_info=1)
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
+    # Set hostname
+    pretty_hostname = "(YunoHost/%s)" % new_domain
```
opi: why does "Pretty hostname" prefixed by "Yunohost" ?

opi: otherwise everything works fine

alexAubin: No particular reason, it's just to have a "pretty" hostname, like not just the FQDN alone but something describing the hostname and machine, if I understand correctly

opi:
please squash before merging this PR to avoir log clutter

opi:
please squash before merging this PR to avoir log clutter

src/yunohost/tools.py
```diff
@@ -162,6 +162,28 @@ def tools_maindomain(auth, new_domain=None):
...
+    commands = [
+        "sudo hostnamectl --static    set-hostname".split() + [new_domain],
+        "sudo hostnamectl --transient set-hostname".split() + [new_domain],
+        "sudo hostnamectl --pretty    set-hostname".split() + [pretty_hostname]
```
zamentur: I 'm not sure this kind of commands are persistent. And you ?

alexAubin: I think hostnamectl is precisely designed to make change persitent (unlike hostname and other ways). Also this is already what Doctor Cube does : https://github.com/labriqueinternet/doctorcube_ynh/blob/6d8580c0216c925cdd2ea435a23a9f0408f6a471/scripts/install#L18-L20

PR #220, opened 2016-12-19
[mod] Cleaner _get_conf_hashes logs

alexAubin:
This is intended to clean some post-installation logs when runned with `--debug`. Currently, since there's no "conffiles" entries in services.yml during post-install, we get some messages lilke :
```
45152 unable to retrieve conf hashes for glances
  File "/usr/lib/moulinette/yunohost/service.py", line 640, in _get_conf_hashes
    return _get_services()[service]['conffiles']
KeyError: 'conffiles'
```
This is a problem for my continous integration stuff which alwats report these errors (see post-installations logs [here](http://bicyclette.netlib.re/))

### DIFF
--- a/src/yunohost/service.py
+++ b/src/yunohost/service.py
@@ -640,16 +640,16 @@ def _get_pending_conf(services=[]):
 def _get_conf_hashes(service):
     """Get the registered conf hashes for a service"""
 
-    d = _get_services()
+    services = _get_services()
 
-    if service not in d.keys():
+    if service not in services:
         logger.debug("Service %s is not in services.yml yet.", service)
         return {}
-    elif 'conffiles' not in d[service].keys():
+    elif 'conffiles' not in services[service]:
         logger.debug("No configuration files for service %s.", service)
         return {}
     else:
-        return d[service]['conffiles']
+        return services[service]['conffiles']
 
 
 def _update_conf_hashes(service, hashes):
--- a/src/yunohost/service.py
+++ b/src/yunohost/service.py
@@ -639,12 +639,17 @@ def _get_pending_conf(services=[]):
 
 def _get_conf_hashes(service):
     """Get the registered conf hashes for a service"""
-    try:
-        return _get_services()[service]['conffiles']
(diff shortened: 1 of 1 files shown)


src/yunohost/service.py
```diff
@@ -639,12 +639,17 @@ def _get_pending_conf(services=[]):
 
 def _get_conf_hashes(service):
     """Get the registered conf hashes for a service"""
-    try:
-        return _get_services()[service]['conffiles']
-    except:
-        logger.debug("unable to retrieve conf hashes for %s",
-                     service, exc_info=1)
+
+    d = _get_services()
```
opi: Maybe I'm old-fashion, but I would prefer an explicit variable name, like `services`, because `d` does not mean anything to me ;)

Psycojoker: I'd prefer that too

src/yunohost/service.py
```diff
@@ -639,12 +639,17 @@ def _get_pending_conf(services=[]):
...
+
+    if service not in d.keys():
```
Psycojoker: The `.keys()` is useless here since this is way is called by default on a for loop.

src/yunohost/service.py
```diff
@@ -639,12 +639,17 @@ def _get_pending_conf(services=[]):
...
+        logger.debug("Service %s is not in services.yml yet.", service)
+        return {}
+    elif 'conffiles' not in d[service].keys():
```
Psycojoker: Same.

alexAubin:
Fixed, thanks for the feedback :smile:

Psycojoker:
I know this is annoying to do but I think that we need to move those strings to en.json to make them translatable :/

alexAubin:
On the one hand, yes. On the other hand, these are only intended to show up in --debug, where most of the strings ain't translated anyway :/ ...

Psycojoker:
Ah right, sound acceptable to me then.

opi:
thanks, code is much clearer now.

PR #222, opened 2016-12-22
[enh] Use ssl-cert group instead of metronome for certificates

alexAubin:
See https://dev.yunohost.org/issues/673

I think this should be included in stable instead of migrating it later (though it shouldn't be a big deal anyway)

- metronome/prosody
- mumble
- matrix

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -223,8 +223,8 @@ def _certificate_install_selfsigned(domain_list, force=False):
 
         # Set appropriate permissions
         _set_permissions(new_cert_folder, "root", "root", 0755)
-        _set_permissions(key_file, "root", "metronome", 0640)
-        _set_permissions(crt_file, "root", "metronome", 0640)
+        _set_permissions(key_file, "root", "ssl-cert", 0640)
+        _set_permissions(crt_file, "root", "ssl-cert", 0640)
         _set_permissions(conf_file, "root", "root", 0600)
 
         # Actually enable the certificate we created
@@ -506,7 +506,7 @@ def _fetch_and_enable_new_certificate(domain, staging=False):
 
     domain_key_file = "%s/%s.pem" % (TMP_FOLDER, domain)
     _generate_key(domain_key_file)
-    _set_permissions(domain_key_file, "root", "metronome", 0640)
+    _set_permissions(domain_key_file, "root", "ssl-cert", 0640)
 
     _prepare_certificate_signing_request(domain, domain_key_file, TMP_FOLDER)
 
@@ -571,7 +571,7 @@ def _fetch_and_enable_new_certificate(domain, staging=False):
         f.write(signed_certificate)
         f.write(intermediate_certificate)
 
-    _set_permissions(domain_cert_file, "root", "metronome", 0640)
+    _set_permissions(domain_cert_file, "root", "ssl-cert", 0640)
 
     if staging:
         return


Psycojoker:
I think this PR should be combined with a migration of all existing files. I think that this could be delay for the next release since this isn't a blocked problem.

julienmalik:
Agreed with Bram, this PR should have code to handle the migration of permissions of existing files, for consistency of the whole system (and accross installations)

Psycojoker:
Should we merge this one now or should we wait to have the migration framework?

alexAubin:
If the migration framework is there in a few (less than 4~6) weeks, we can wait for the migration framework I'd say...

alexAubin:
Note that Matrix seems to need access to certificates and need to tweak permissions : https://forum.yunohost.org/t/matrix-messaging/2275/45

zamentur:
The mumble app too

alexAubin:
Okay guys,

I'm reviving this PR because the migration framework is still far from being ready. We've delayed this PR (4 months old now) with the assumption that it would arrive soon. In the meantime, people working on the Matrix and Mumble apps need to do some dirty tweaking of the certs permissions in order for their app to work...

I'd be really in favor of merging this now. To me, it doesn't bring much inconsistencies on installs due to the fact that the migration will actually happen "by itself" due to the renewal of LE certs every three months. The only small inconsistency that would remain would be people with an old self-signed certificate. But then a) it doesn't really matter imho because you can easily get a LE certificate, b) you can easily regenerate your self-signed certificate.

opi:
note that a dirty migration can be done at the Debian level, in the `postinst` script.

I agree on merging soon too

M5oul:
I think there is an issue.
On unstable, I manually renew my certs and only the cert changed to `root:ssl-cert`.
The key did not changed, and stayed with `root:metronome` permissions:
```bash
-rw-r----- 1 root ssl-cert  3619 May 22 16:52 /etc/yunohost/certs/${domain.tld}/crt.pem
-rw-r----- 1 root metronome 2488 May 22 16:52 /etc/yunohost/certs/${domain.tld}/key.pem
```

On my PR for Prosody migration, I modified more line than on this PR: https://github.com/YunoHost/yunohost/pull/240/commits/871850a4fbbfce68368589405cbfc5a676c52454

alexAubin:
Thanks for reporting this

It's a bit unexpected (because the perms of the key is set [here](https://github.com/YunoHost/yunohost/blob/c21acd06d650b3090580e57c2f52e525f687ab41/src/yunohost/certificate.py#L509) (and then the file is moved to the final place [here](https://github.com/YunoHost/yunohost/blob/c21acd06d650b3090580e57c2f52e525f687ab41/src/yunohost/certificate.py#L565))

But I think I saw a similar issue a few monthes ago... An easy fix is to re-force the permissions of the key at the end though. I'll try that and open the corresponding PR

alexAubin:
I'm wondering if this behavior could come from the `/tmp/` being on another filesystem than /etc/. In that case, `shutil.move()` does not just perform a simple rename, it copy the file, and might not be able to keep the permissions in the process...

M5oul:
I tried to use certificates with Mumble.
From Mumble logs, certificates can't be read:
```bash
<C>2017-05-23 08:45:03.465 Failed to read /etc/yunohost/certs/${domain}/crt.pem
<C>2017-05-23 08:45:03.466 Failed to read /etc/yunohost/certs/${domain}/key.pem
```
With `root:ssl-cert` permissions on the certificate and the key:
```bash
-rw-r----- 1 root ssl-cert 3595 May 22 18:04 /etc/yunohost/certs/${domain}/crt.pem
-rw-r----- 1 root ssl-cert 2484 May 22 18:04 /etc/yunohost/certs/${domain}/key.pem
```

That is because `ssl-cert` group does not exist:
```bash
groups ssl-cert
groups: ssl-cert: no such user
```

opi:
`ssl-cert` is not a user, but a group : `grep ssl-cert /etc/group`

opi:
> Does this PR should add to ssl-cert group permissions for metronome/prosody, mumble, matrix users?

metronome only, because mumble or matrix users are created by the relevant apps. Btw, metronome user is already on the `ssl-cert` group (`groups metronome`)

M5oul:
Ok, so each app needs to create a user on `ssl-cert` group.

M5oul:
Ok, I finally succeed to use certificates with Mumble server.

PR #224, opened 2017-01-02
[fix] Fix 693 invalid domain dyndns postinstall

zamentur:
This morning I have got 2 issues on unstable branches when I have tried to postinstall.
### First issue : moulinette error about logging.SUCCESS
One was due to a non built commit on moulinette (it should be done this night).

The 2 others are fixed by this PR.
### Second issue : postinstall on domain (with just one . ) fails due to --ignore-dyndns option

> yunohost tools postinstall test.local --verbose --debug
fails due to the missing option --ignore-dyndns. But it shouldn't fail because it's a short domain so it's not a dyndns one...

Note: I am not sure about the good behaviour to have with 3 parts domain ( eg: toto.example.org ). We don't support for the moment dyndns on it but in future it could be the case...

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -178,7 +178,8 @@ def tools_postinstall(domain, password, ignore_dyndns=False):
 
     Keyword argument:
         domain -- YunoHost main domain
-        ignore_dyndns -- Do not subscribe domain to a DynDNS service
+        ignore_dyndns -- Do not subscribe domain to a DynDNS service (only
+        needed for nohost.me, noho.st domains)
         password -- YunoHost admin password
 
     """
@@ -203,6 +204,10 @@ def tools_postinstall(domain, password, ignore_dyndns=False):
                 else:
                     raise MoulinetteError(errno.EEXIST,
                                           m18n.n('dyndns_unavailable'))
+            else:
+                dyndns = False
+    else:
+        dyndns = False
 
     logger.info(m18n.n('yunohost_installing'))
 
@@ -296,7 +301,6 @@ def tools_postinstall(domain, password, ignore_dyndns=False):
     os.system('service yunohost-firewall start')
 
     service_regen_conf(force=True)
-
     logger.success(m18n.n('yunohost_configured'))
 
 


src/yunohost/tools.py
```diff
@@ -296,7 +301,7 @@ def tools_postinstall(domain, password, ignore_dyndns=False):
     os.system('service yunohost-firewall start')
 
     service_regen_conf(force=True)
-
+    logger.success(m18n.n('yunohost_configured'))
```
alexAubin: Duplicated line ? Is this a typo ?

zamentur:
It seems the erreor on ssl is the same on other regenconf hooks!
Very strange! but my fix is not enough general. 
I have checked the diff on yunohost/yunohost since 15 days without success ! Maybe something in moulinette ?

alexAubin:
I'm proposing the previous commit (2c5d55a) to fix this issue. This refactoring some part of the dyndns handling, so i understand if you guys want to revert/remove this commit. But to me there's too much madness in the current code (c.f. my comment on https://dev.yunohost.org/issues/693).

So the commit : 
- adds `_dyndns_provides` to check if a provider (e.g. dyndns.yunohost.org) can provide/handle a given domain (e.g. toto.nohost.me). This is meant to replace the current check about `len(domain.split('.')) >= 3`, and makes us able to differentiate between `toto.nohost.me` and `toto.someother.tld`. For the first one, we (apriori) implicitly want to have dyndns for it. For the second one, we don't want dyndns stuff (or at least not supported atm)
- adds `_dyndns_available` to check if a given domain is available. This avoid current duplicated code doing requests to https://dyndns.yunohost.org/test/<thedomain> and interpreting the results their own ways.

With this, I refactorized the begining of the postinstall : if the user asks of a domain ending in `.nohost.me`, the postinstall understands that we want dyndns for it. If the user asks for `ynh.local.stuff` or `toto.netlib.re`, the postinstall understands that we don't want dyndns for it.

And the new meaning of `--ignore-dyndns` is « I want toto.nohost.me, but I know it's not available, so don't try to configure the dyndns »

alexAubin:
As discussed on the chat, we could also : 
- have the few lines fix right now
- post-poned the refactorizing (already implemented) for the very beginning of 2.6.x

Psycojoker:
> As discussed on the chat, we could also : 
> - have the few lines fix right now
> - post-poned the refactorizing (already implemented) for the very beginning of 2.6.x

I agree.

Thanks for tackling this bug <3

alexAubin:
I removed 2c5d55a from this branch by force-pushing, and rebased it to a new branch (see #228 ).

PR #225, opened 2017-01-03
[fix] Adding check that domain is resolved locally for cert management

alexAubin:
This should fix https://dev.yunohost.org/issues/651


locales/en.json
```diff
@@ -253,6 +253,7 @@
     "certmanager_domain_http_not_working": "It seems that the domain {domain:s} cannot be accessed through HTTP. Please check your DNS and nginx configuration is okay",
     "certmanager_error_no_A_record": "No DNS 'A' record found for {domain:s}. You need to make your domain name point to your machine to be able to install a Let's Encrypt certificate! (If you know what you are doing, use --no-checks to disable those checks.)",
     "certmanager_domain_dns_ip_differs_from_public_ip": "The DNS 'A' record for domain {domain:s} is different from this server IP. If you recently modified your A record, please wait for it to propagate (some DNS propagation checkers are available online). (If you know what you are doing, use --no-checks to disable those checks.)",
+    "certmanager_domain_not_resolved_locally": "The domain {domain:s} cannot be resolved locally. This might happen if you recently modified your DNS record. If so, please wait a few hours for it to propagate. If the issue persists, consider adding {domain:s} to /etc/hosts. (If you know what you are doing, use --no-checks to disable those checks.)",
```
Psycojoker: s/locally/from inside your YunoHost server

"This will happen if your server isn't configured to use YunoHost DNS server and that [you have recently ...]"

"You can also switch your server DNS server to uses YunoHost DNS server by specifying localhost in /etc/resolv.conf"

src/yunohost/certificate.py
```diff
@@ -809,6 +816,15 @@ def _domain_is_accessible_through_HTTP(ip, domain):
     return True
 
 
+def _domain_is_resolved_locally(domain):
+    try:
+        requests.head("http://%s/" % domain)
```
Psycojoker: We should try to resolve the server using a DNS query and checking that this is exactly this problem, not by doing a HTTP request (this can trigger other problems that aren't related to DNS and the catchall exception can hide other errors).

locales/en.json
alexAubin: Okay for "from inside your Yunohost server".

But to me, the parts about "use Yunohost DNS server" are totally confusing. I would interpret it as : I should use a specific service like OpenDNS or something, but provided by Yunohost's team ? 

Also I don't agree with the interpretation of /etc/resolv.conf. To me there's no right or wrong resolv.conf, and even if you use "nameserver 127.0.0.1" (such that it uses the local dnsmasq service ?), "toto.nohost.me" won't resolve if it's not in /etc/hosts... (or if we tweak dnsmasq in some other ways, but to me /etc/hosts is the most reliable :/)

src/yunohost/certificate.py
alexAubin: Fixed in 86f97d6

NB : I used `socket.gethostbyname()` because it's seems closer to what `requests` uses than `dns.resolve`... `dns.resolve` will perform a 'full' DNS request and return the public ip, but if the domain is in `/etc/hosts`, `requests` will actually use what's in `/etc/hosts`... `socket.gethostbyname()` takes `/etc/hosts` into account.

locales/en.json
Psycojoker: > Also I don't agree with the interpretation of /etc/resolv.conf. To me there's no right or wrong resolv.conf, and even if you use "nameserver 127.0.0.1" (such that it uses the local dnsmasq service ?), "toto.nohost.me" won't resolve if it's not in /etc/hosts... (or if we tweak dnsmasq in some other ways, but to me /etc/hosts is the most reliable :/)

It will resolve, try doing a `yunohost domain add pouet.pouet` then "ping pouet.pouet" and you'll get an ip for `pouet.pouet`.

Interestingly it doesn't resolve other domains and I don't know why.

src/yunohost/certificate.py
```diff
@@ -809,6 +817,15 @@ def _domain_is_accessible_through_HTTP(ip, domain):
     return True
 
 
+def _domain_is_resolved_locally(public_ip, domain):
+    try:
+        ip = socket.gethostbyname(domain)
+    except socket.error:
```
Psycojoker: I still think that we need to log the exception.

src/yunohost/certificate.py
```diff
@@ -809,6 +817,15 @@ def _domain_is_accessible_through_HTTP(ip, domain):
...
+        return False
+
+    return (ip in ["127.0.0.1", public_ip])
```
Psycojoker: No (). I think that we also need to handle the wonderful ipv6 case here :/

alexAubin: Agreed we should ideally support IPv6, but that's gonna take a few more days since it's not trivial from [the documentation of `gethostbyname()`](https://docs.python.org/2/library/socket.html#socket.gethostbyname) ... We already know that the whole cert management (and probably Yunohost) does not support ipv6-only hosts, so ? :/

locales/en.json
alexAubin: Meh, indeed that works, but for some reason I had bind9 up and dnsmasq down (but thought I was using dnsmasq) ... but that just kinda prove that making sure we use dnsmasq and have it running is a complicated problem (some servers have resolvconf generating /etc/resolv.conf automatically... + Internet Cube does some other stuff). We know that `/etc/hosts` just works.

src/yunohost/certificate.py
zamentur: If the user set 127.0.0.2 mydomain.tld in his host it could break things.

PR #226, opened 2017-01-03
[fix] new rspamd version replace rspamd.socket with rspamd.service

zamentur:
https://dev.yunohost.org/issues/645

cf https://rspamd.com/announce/2016/08/01/rspamd-1.3.1.html

>Systemd activation has been removed from Rspamd and Rmilter
>
>Over the recent years, we have experienced constant issues reported by users about Systemd socket activation. This feature seems to be completely broken in systems with both IPv4/IPv6 enabled. Moreover, it seems to be harmful as Rspamd setup time is quite significant whilst socket activation is mostly intended for interactive on-demand daemons. Socket activation has been added under the pressure of Debian packaging rules, though Rspamd is completely rotten in Debian official repos and we strongly discourage Debian users from using the official Debian packages. While we are still looking for a new Debian maintainer for Rspamd, I can declare that there will be no sockets activation enabled by default in Rspamd nor Rmilter.
>
> The switch to standalone mode should be transparent for users. The only significant difference is use of rspamd.service instead of rspamd.socket in service management commands. In some cases you might need to restart Rmilter after upgrade (for example, in Debian Jessie):
>
>systemctl restart rmilter.service

So I think we should consider to give the socket activation idea up, for exemple all internetcube with a FFDN vpn have an ipv4 and ipv6.

So I suggest to simply do a restart. The service should be up after the boot.

May be there will ben an increase in memory ?

(Edit from Alex : using quote markdown instead of code block for better readability)

### DIFF
--- a/data/hooks/conf_regen/28-rmilter
+++ b/data/hooks/conf_regen/28-rmilter
@@ -9,8 +9,6 @@ do_pre_regen() {
 
   install -D -m 644 rmilter.conf \
     "${pending_dir}/etc/rmilter.conf"
-  install -D -m 644 rmilter.socket \
-    "${pending_dir}/etc/systemd/system/rmilter.socket"
 }
 
 do_post_regen() {
@@ -40,14 +38,11 @@ do_post_regen() {
   [ -z "$regen_conf_files" ] && exit 0
 
   # reload systemd daemon
-  [[ "$regen_conf_files" =~ rmilter\.socket ]] && {
-    sudo systemctl -q daemon-reload
-  }
+  sudo systemctl -q daemon-reload
 
-  # ensure that the socket is listening and stop the service - it will be
-  # started again by the socket as needed
-  sudo systemctl -q start rmilter.socket
-  sudo systemctl -q stop rmilter.service 2>&1 || true
+  # Restart rmilter due to the rspamd update
+  # https://rspamd.com/announce/2016/08/01/rspamd-1.3.1.html
+  sudo systemctl -q restart rmilter.service
 }
 
 FORCE=${2:-0}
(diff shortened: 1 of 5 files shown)


data/templates/rmilter/rmilter.conf
```diff
@@ -6,7 +6,7 @@
 pidfile = /run/rmilter/rmilter.pid;
 
 # rmilter is socket-activated under systemd
```
julienmalik: this comment should be changed

zamentur:
I have tested with postinstall, but I hadn't tested the upgrade case.

Psycojoker:
I don't have any knowledge of this field so I'm going to trust you on this one.

Thanks a lot for working on this blocking problem :heart:

data/hooks/conf_regen/28-rmilter
```diff
@@ -9,8 +9,6 @@ do_pre_regen() {
 
   install -D -m 644 rmilter.conf \
     "${pending_dir}/etc/rmilter.conf"
-  install -D -m 644 rmilter.socket \
-    "${pending_dir}/etc/systemd/system/rmilter.socket"
```
julienmalik: `/etc/systemd/system/rmilter.socket` should be removed from the system if it exists.

On my system I still have this file, coming from previous yunohost versions.
I have :

```
admin@ynh:~$ sudo systemctl restart rmilter.socket                                                       
Job for rmilter.socket failed. See 'systemctl status rmilter.socket' and 'journalctl -xn' for details.   
admin@ynh:~$ systemctl status rmilter.socket                                                             
● rmilter.socket                                                                                         
   Loaded: loaded (/etc/systemd/system/rmilter.socket; enabled)                                          
   Active: inactive (dead) since Wed 2017-01-04 14:30:39 CET; 2min 26s ago                               
   Listen: 127.0.0.1:11000 (Stream)        
```

There is probably something more to do to deregister it from systemd but I don't know the proper incantation (deamon-reload does this ? is it supposed to be done in post_regen hook ?)

PR #228, opened 2017-01-08
[enh] Refactorize some parts of dyndns

alexAubin:
Follow up of https://github.com/YunoHost/yunohost/pull/224, to be merged in 2.6.x

TODO : Implement comments from Bram [here](https://github.com/YunoHost/yunohost/commit/2c5d55a7f87f7a7eb5b6e3e59ef7559d8b5f9ebf)

Original description : 

---

- adds `_dyndns_provides` to check if a provider (e.g. dyndns.yunohost.org) can provide/handle a given domain (e.g. toto.nohost.me). This is meant to replace the current check about `len(domain.split('.')) >= 3`, and makes us able to differentiate between `toto.nohost.me` and `toto.someother.tld`. For the first one, we (apriori) implicitly want to have dyndns for it. For the second one, we don't want dyndns stuff (or at least not supported atm)
- adds `_dyndns_available` to check if a given domain is available. This avoid current duplicated code doing requests to https://dyndns.yunohost.org/test/<thedomain> and interpreting the results their own ways.

With this, I refactorized the begining of the postinstall : if the user asks of a domain ending in `.nohost.me`, the postinstall understands that we want dyndns for it. If the user asks for `ynh.local.stuff` or `toto.netlib.re`, the postinstall understands that we don't want dyndns for it.

And the new meaning of `--ignore-dyndns` is « I want toto.nohost.me, but I know it's not available, so don't try to configure the dyndns »

### DIFF
--- a/src/yunohost/dyndns.py
+++ b/src/yunohost/dyndns.py
@@ -66,6 +66,38 @@ re_dyndns_private_key = re.compile(
     r'.*/K(?P<domain>[^\s\+]+)\.\+157.+\.private$'
 )
 
+def _dyndns_provides(provider, domain):
+
+    logger.debug("Checking if %s is managed by %s ..." % (domain, provider))
+
+    try:
+        r = requests.get('https://%s/domains' % provider)
+    except requests.ConnectionError:
+        raise MoulinetteError(errno.ENETUNREACH,
+                m18n.n('dyndns_provider_unreachable', provider=provider))
+
+    dyndomains = json.loads(r.text)
+    dyndomain  = '.'.join(domain.split('.')[1:])
+
+    return (dyndomain in dyndomains)
+
+
+def _dyndns_available(provider, domain):
+
+    logger.debug("Checking if domain %s is available on %s ..." % (domain, provider))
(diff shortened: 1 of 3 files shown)


src/yunohost/dyndns.py
```diff
@@ -66,6 +66,38 @@ def __init__(self, line):
     r'.*/K(?P<domain>[^\s\+]+)\.\+157.+\.private$'
 )
 
+def _dyndns_provides(provider, domain):
+
+    logger.debug("Checking if %s is managed by %s ..." % (domain, provider))
+
+    try:
+        r = requests.get('https://%s/domains' % provider)
```
Psycojoker: We should add timeout here (we hadn't made this decision at the time you've made those commits).

alexAubin:
As discussed on the chat, PR is not ready to be merged, I'm flagging this as inactive until I have some time to work on it (or anybody wants to)

alexAubin:
Closing because this was remade from scratch in https://github.com/YunoHost/yunohost/pull/353

PR #229, opened 2017-01-15
Global settings

Psycojoker:
Hello,

Prototype of global settings module for YunoHost. The usage is quite simple and obvious (like a key/value store basically).

Example usage:

```
root@yunohost:/home/vagrant# yunohost settings list                                                                                                                                                              
root@yunohost:/home/vagrant# yunohost settings set version -v 3 -n ssl
root@yunohost:/home/vagrant# yunohost settings list
ssl: 
  version: 3
root@yunohost:/home/vagrant# yunohost settings list -n ssl
version: 3
root@yunohost:/home/vagrant# yunohost settings get version -n ssl -d 1
3
root@yunohost:/home/vagrant# yunohost settings get version -n ssl --default 1                                                                                                                                    
3
root@yunohost:/home/vagrant# yunohost settings exists -n ssl version                                                                                                                                             
True
root@yunohost:/home/vagrant# yunohost settings exists -n ssl chocapic
False
root@yunohost:/home/vagrant# yunohost settings remove -n ssl version
root@yunohost:/home/vagrant# yunohost settings remove -n ssl version
Error: The key 'version' doesn't exists in the global settings, you can see all the available keys by doing 'yunohost settings list
root@yunohost:/home/vagrant# yunohost settings remove -n ssl version -s
root@yunohost:/home/vagrant# yunohost settings list -n ssl
Warning: The namespace 'ssl' is empty or doesn't exists in the global settings
```

Design decisions:

* you can't get a settings without providing a default value, this is done to force the user to handle the case where the value doesn't exists and avoid bugs
* if the users wants to test the presence of a key he can uses `exists`
* every settings are namespaced to force some organisation and avoid ending up with a big mess where we start to namespace settings key by ourself using "namespace_"

We shouldn't merged this PR if we don't have a need yet for this module (but I know that it was quite expected by @julienmalik).


Psycojoker:
Documentation:

```
16:46  <ju> il manque un truc à mon gout, c'est de pouvoir les documenter
16:47  <ju> voire une short desc + une long doc
16:49 @<Bram> namespace:
16:49 @<Bram>   key:
16:49 @<Bram>     value: value
16:49 @<Bram>     shortdoc: caramba
16:49 @<Bram>     doc: pouet
```

```
16:50  <ju> p-e plus tard on voudra des "type"
16:50  <ju> bool/string/...
16:50 @<Bram> faudra aussi penser à la traduction en fait
16:51 @<Bram> je sais pas si c'est approche de la doc est bonne pour ça
```

Free-for-all ou enforcés ?
```
16:51 @<Bram> c'est pour ça qu'il faudrait réfléchir au free-for-all vs "set de settings limités"
```

Psycojoker:
Référence https://dev.yunohost.org/issues/254

```
17:01 @<Bram> ju: j'ai l'impression qu'un about:config serait effectivement bien mieux que ce que j'ai fait
17:01 @<Bram> là de ce que j'en lis on a
17:01 @<Bram> nom - statut - type - valeur
17:01 @<Bram> on peut donc facilement y joindre une clef pour la short/long doc
17:01 @<Bram> et la traduire
17:01  <ju> ouais le typage ce sera utile pour l'ui
17:01 @<Bram> et on définie nous même la liste des settings disponible
17:02 @<Bram> ça semble moins casse gueule que mon approche
17:02  <ju> ben c'est pas très loin je trouve
17:02 @<Bram> ben, moi c'est du CRUD sur du yaml
17:03  <ju> tu veux fixer en dur la liste exhaustive des settings dispo ?
17:03 @<Bram> et about:config c'est une table sql sur laquelle t'as pas le droit de rajouter d'entrées
17:03 @<Bram> oui
17:03 @<Bram> ça me semble plus solide
17:03  <ju> ça pourrait bien servir aux apps d'avoir un système centralisé de settings
17:03 @<Bram> ben elles ont déjà leurs propres settings non ?
17:03  <ju> genre les apps qui sont liées les unes aux autres
17:03 @<Bram> le fixer en dur permet la traduction
17:03 @<Bram> ah
17:03 @<Bram> hmmm
17:03  <ju> mais j'avoue j'ai pas hyper réfléchi à ça pour les apps
17:03 @<Bram> et c'est vrai qu'on a d'autres trucs variables
17:04 @<Bram> genre les domains
17:04 @<Bram> on va vouloir faire de la config par domaine
17:04  <ju> oui
17:04  <ju> ça c'est sur
17:04 @<Bram> après si c'est par "un truc variable en db"
17:04 @<Bram> ça peut se générer facilement
17:04 @<Bram> au lieu de permettre de rentrer ce qu'on veut
17:04  <ju> mais pour les domaines ya moyen de mettre des trucs dans ldap
17:04 @<Bram> on a d'ailleurs + intérêt à le générer
17:05 @<Bram> ah oui
17:05 @<Bram> c'est mieux tu crois ?
17:05  <ju> c'était l'idée de jerome dans le passé
17:05 @<Bram> on aurait pas intérêt à avoir toutes les settings au même endroit ?
17:05  <ju> ya du pour et du contre
17:06  <ju> on sera + free qu'avec ldap
17:06  <ju> mais va falloir synchronisé
17:07  <ju> rien d'impossible
17:07  <ju> mais source de bug/trucs qui divergent c tout
17:09  <Aleks> ju : heu c'etait quoi deja le truc avec livebox ?
17:09 @<Bram> ça serait pas mal d'avoir un systématisme en fait
17:10 @<Bram> "tous les settings généraux vont dans X"
17:10 @<Bram> "tous les settings particuliers dans Y"
17:10 @<Bram> (X pouvant être égale à Y)
17:10 @<Bram> là c'est un peu le bronx j'ai l'impression
17:10 @<Bram> (ou le vide plus exactement)
```

Je pense que je vais donc adapter cette PR pour plutôt faire quelque chose à la `about:config`

alexAubin:
Many thanks for this work 😮 !

Imho we should flag this as a medium decision, as this feature will be used by many foreseen features and we need to design it right.

Psycojoker:
Make sens :)

Psycojoker:
I've pushed an unfinished work in the direction discussed here. Only minor things needs to be done (like i18n).

I might also add a command "reset" to reset all settings to default and save the old settings in a backup file. 

I'm also considering saving unknown settings key in a settings_unkown.json file in case those could be useful (as they might come from downgrading or other things).

I've also written a very small quick and dirty unittest framework that I haven't committed. I have 17 tests for the settings. I was really annoying to don't have any internal unittest framework for YunoHost. I'll push it in another PR once this one is merged as I don't want to mix things together (and it's hard to work simultanuously on 2 branchs with git). You can find a copy/pasta here: https://gist.github.com/Psycojoker/117a0a81c6c98afc17f5d1e6fb1388b2

We should really modify moulinette to have tests. This things serves its purpose in the meantime.

Psycojoker:
Ok, I now consider this module to be ready for review. I'm updating the gist with the new tests.

data/actionsmap/yunohost.yml
```diff
@@ -925,6 +925,54 @@ monitor:
...
 
 #############################
+#         Settings          #
+#############################
+settings:
+    category_help: Manage YunoHost global settings
+    actions:
+
+        ### settings_list()
+        list:
```
alexAubin: To be discussed / implemented later maybe, but it could be nice to have an option to list only differences with the default settings.

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+        if not isinstance(value, bool):
+            raise MoulinetteError(errno.EINVAL, m18n.n(
+                'global_settings_bad_type_for_setting', setting=key,
+                received_type=type(value).__name__, expected_type=key_type))
+    elif key_type == "int":
+        if not isinstance(value, int) or isinstance(value, bool):
+            raise MoulinetteError(errno.EINVAL, m18n.n(
+                'global_settings_bad_type_for_setting', setting=key,
+                received_type=type(value).__name__, expected_type=key_type))
+    elif key_type == "string":
```
alexAubin: Is there any hope to be able to refactorize this with a dict such as
```
key_type_dict = { "int" : int, "bool": bool }
```
, hopefully including the enum type somehow, or is there no real gain ?

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+                    unknown_settings[key] = value
+    except Exception as e:
+        raise MoulinetteError(errno.EIO, m18n.n('global_settings_cant_open_settings', reason=e),
+                              exc_info=1)
+
+    if unknown_settings:
+        try:
+            _save_settings(unknown_settings, location=unknown_settings_path)
+        except Exception as e:
+            logger.warning("Failed to save uknown settings (because %s), abording." % e)
```
alexAubin: Typos (uknown, aboding) + should be moved to m18n string ?

src/yunohost/tests/conftest.py
```diff
@@ -0,0 +1,16 @@
+import sys
+import moulinette
+
+sys.path.append("..")
+
+old_init = moulinette.core.Moulinette18n.__init__
+
```
alexAubin: Ins't this file meant to be introduced in the unitary test PR ?

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+        return settings
+
+    unknown_settings = {}
+    unknown_settings_path = SETTINGS_PATH_OTHER_LOCATION % "unknown"
+
+    if os.path.exists(unknown_settings_path):
+        try:
+            unknown_settings = json.load(open(unknown_settings_path, "r"))
+        except Exception as e:
+            logger.warning("Error while loading unknown settings %s" % e)
```
alexAubin: To be internationalized

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+# * int
+# * string
+# * enum (in form a python list)
+
+# we don't store the value in default options
+DEFAULTS = OrderedDict([
+    ("example.bool", {"type": "bool", "default": True}),
+    ("example.int", {"type": "int", "default": 42}),
+    ("example.string", {"type": "string", "default": "yolo swag"}),
+    ("example.enum", {"type": "enum", "default": "a", "choices": ["a", "b", "c"]}),
```
alexAubin: Could we define a list of real-life settings we'd like to have ? Imho that would help to have a better feeling about namespaces and how we'll use it in the future. There could be some things such as : 
```
notifications.mail -> "kload@yoloswag.com"
security.allow_ssh_users_by_default -> "True"
security.fail2ban_harshness ->  [ "low", "mid", "high" ]
security.ssl_security -> [ "low", "high" ] # Would impact the key size for certificates and DH parameters
```

data/actionsmap/yunohost.yml
```diff
@@ -925,6 +925,54 @@ monitor:
...
+        ### settings_get()
+        get:
+            action_help: get an entry value in the settings
+            api: GET /settings/<key>
+            arguments:
+                key:
+                    help: Settings key
+
+        ### settings_set()
+        set:
```
alexAubin: If this is meant to be used by end users somehow, shall we have a system of hooks or actions triggered somehow ? E.g. let's imagine there's a setting to define the SSL security policy (bigger/smaller key size), when does the effective change happens ?

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+# the key is "global_settings_setting_%s" % key.replace(".", "_")
+
+# type can be:
+# * bool
+# * int
+# * string
+# * enum (in form a python list)
+
+# we don't store the value in default options
+DEFAULTS = OrderedDict([
```
Psycojoker: We want to enforce a fixed list of settings, putting it directly into the code is a good way to do that.

Furthermore, it allows to use OrderedDirect which is impossible in json (and super annoying to do in yaml and yaml is slow).

I also have some problems with the "one day we might have" which leads to over engineered code for situations where a later modification is trivial, so if it's a problem one day, we'll fix it this day :)

src/yunohost/settings.py
Psycojoker: Sure, I just hadn't have any of those settings so I've put placeholders. I think that those will come with other PRs.

src/yunohost/settings.py
Psycojoker: I don't see this bringing that much value here tbh, I prefer this simplest way because it avoid one indirection (having to look in a dictionary elsewhere thus making the code a little bit harder to read) to the price of not being 100% elegant, but I think that simplicity prevails here.

If those types where serializable in json I would have used them.

src/yunohost/settings.py
Psycojoker: It's a low level debug log, I don't think that most users should look at it.

src/yunohost/settings.py
Psycojoker: Same than before (but fixing typos, thx <3)

Psycojoker: Hm, is "abording" wrong? https://en.wiktionary.org/wiki/abording

data/actionsmap/yunohost.yml
Psycojoker: Ah indeed, good idea :)

src/yunohost/tests/conftest.py
Psycojoker: Well, I needed to run my tests somehow :p

data/actionsmap/yunohost.yml
Psycojoker: Yeah, that's the discussion we had and I don't have strong opinions on that. Right now I would suggest starting with this simple approach, keep the settings internal, and look on how other PR brings new requirements.

And if those requirements a hook based approach, sure, let's do it, but I want to avoid YAGNI for now (you know, like those offsets :p).

src/yunohost/settings.py
Psycojoker: It is.

src/yunohost/settings.py
alexAubin: Okay, agreed for now :).

src/yunohost/settings.py
alexAubin: Okay, agreed !

data/actionsmap/yunohost.yml
alexAubin: After discussing this on the chat with @Psycojoker, we thought about using "decorators" to be able to track which (python) functions should be triggered when changing the value of a setting. This should be relatively easy to implement and add later. Same goes for bash/'external' hooks if we want to go in that direction.

Psycojoker:
Up

src/yunohost/tests/conftest.py
alexAubin: I mean, it probably conflicts with it since it's merged, so probably need to remove it to solve the conflicts, but okay ;)

Psycojoker: No, it's the same commit, git can handle that.

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+from moulinette.core import MoulinetteError
+from moulinette.utils.log import getActionLogger
+
+logger = getActionLogger('yunohost.settings')
+
+SETTINGS_PATH = "/etc/yunohost/settings.json"
+SETTINGS_PATH_OTHER_LOCATION = "/etc/yunohost/settings-%s.json"
+
+# a settings entry is in the form of:
+# name: {type, value, default, description, [possibilities]}
```
opi: use the wording "choices" as this is the one used later in real code

opi: Please describe the concept and usage of "namespace" for `name` key;

Psycojoker:
2017-03-13-221400_452x424_scrot

data/actionsmap/yunohost.yml
```diff
@@ -925,6 +925,54 @@ monitor:
...
+                -n:
+                    full: --namespace
+                    help: 'The namespace in which the settings is stored (eg: "ssl"))'
+                    extra:
+                        required: True
+
+        ### settings_reset()
+        reset:
+            action_help: reset all settings to their default value
+            api: POST /settings/reset
```
opi: REST nitpick : Should it be `DELETE /settings` instead ?

data/actionsmap/yunohost.yml
```diff
@@ -925,6 +925,54 @@ monitor:
...
+            arguments:
+                -y:
+                    full: --yes
+                    help: Yes I'm sure I want to do that
+                    action: store_true
+
```
opi: I think we miss the ability to reset a specific setting ( `DELETE /settings/<key>` )

src/yunohost/settings.py
```diff
@@ -0,0 +1,189 @@
...
+
+    for key, value in DEFAULTS.copy().items():
+        settings[key] = value
+        settings[key]["value"] = value["default"]
+        settings[key]["description"] = m18n.n("global_settings_setting_%s" % key.replace(".", "_"))
+
+    if not os.path.exists(SETTINGS_PATH):
+        return settings
+
+    unknown_settings = {}
```
opi: Please comment usage of `unknown_settings`

Psycojoker:
Implemented opi's feedback.

data/actionsmap/yunohost.yml
```diff
@@ -925,6 +925,57 @@ monitor:
...
+                -v:
+                    full: --value
+                    help: new value
+                    extra:
+                        required: True
+
+        ### settings_reset()
+        reset:
+            action_help: reset all settings to their default value
+            api: DELETE /settings/reset
```
opi: the `/reset` part has no more meaning here, as we use `DELETE` method now.

data/actionsmap/yunohost.yml
```diff
@@ -925,6 +925,57 @@ monitor:
...
+            action_help: reset all settings to their default value
+            api: DELETE /settings
+            arguments:
+                -y:
+                    full: --yes
+                    help: Yes I'm sure I want to do that
+                    action: store_true
+
+        ### settings_default()
+        default:
```
alexAubin: I'm not a huge fan of this design. I would really like having rather something like : 

```
# Reset all settings
yunohost settings reset --all --yes

# Reset a single setting
yunohost settings reset <key>
```

(Not sure about the --yes though if user already has to put --all)

data/actionsmap/yunohost.yml
zamentur: +1

data/actionsmap/yunohost.yml
zamentur: I agreed

zamentur:
I just read it, it seems good. I don't know how to manage settings for domain. Ia ma not sure the ldap solutions is a good idea.
I will test your code later.

alexAubin:
>how to manage settings for domain

You mean how to manage settings specific to a given domain ? I think this point has not been discussed

data/actionsmap/yunohost.yml
Psycojoker: Ok.

On the other hand that makes me thing that "cert-install" and related commands should also have a "--all" flag, when commands doesn't have something like that I often end up writing the command without arguments to have its arguments displayed and end up running it for all possible arguments :x

Psycojoker:
> You mean how to manage settings specific to a given domain ? I think this point has not been discussed

Well, we have discussed that for now those settings are for now, only for static settings, not dynamic ones like one linked to a domain or a user. Both of those are for now stored in ldap and for apps they are stored in app related settings. We can discuss that but that's out of the scope of this PR.

zamentur:
Proposition (from @alexAubin @e-lie):

yunohost settings get example.bool --full

yunohost settings reset example.bool

yunohost settings reset-all

Psycojoker:
> yunohost settings reset example.bool
>
> yunohost settings reset-all

Following a discussion we had on IRC I'm opposed to this API because:

* resetting the whole configuration with a backup and putting back a single option to its default value are 2 different operations, both in its intention and in GUI
* removing "--yes" is removing a protection against doing unvoluntary mistakes, especially for a as critical command at this one where we don't offer a `undo` operation, I'm very very very opposed to this one
* having a command that does destructives/modifications operation without additional arguments is a bad practice because it's a common pattern for a user to just use a command without arguments to get the command help displayed in linux (I still think that `cert-install` and other behavior is bad)

I would rather rename `default` to `todefault` than make those modifications.

zamentur:
About --yes options, there is no other api where we do that. `yunohost app remove` is more destructive and there is no --yes option. And I think to be obliged to do request like this one "DELETE /settings?yes=1" is not a good idea.

About "default" cmd, instinctively, a default command could be misunderstood:
`yunohost settings default example.bool `=> we were thinking it was a command to get the default value of example.bool
to-default is more understandable

What do you think about "reinit" (might be better than "to-default") ?

alexAubin:
>resetting the whole configuration with a backup and putting back a single option to its default value are 2 different operations, both in its intention and in GUI

After thinking about this, I actually disagree :/. Let's have a look at how we describe what we're trying to achieve here to each other : "we want to have one command to reset all settings, or just reset a single one.". (You can replace 'reset' by 'put its value back to default', but that's the definition of reset in this context). The main difference is that in one case, we want to backup stuff, and not in the other case. But it's actually kind of arbritrary (its driven by the assumption that maybe the user doesn't know what it's doing ?) and doesn't make these two things fundamentally different in the intention. Therefore I don't quite understand the necessity of having two different terms, where in fact in both case we mainly want to reset something. To me that'll just add confusion :/

opi:
> Ok guys, we need to move on on this PR.

For people like me, is the first comment up-to-date, or can we have a sum up with command-line instruction & description ? (aka "Le documentation"). Thanks

julienmalik:
+1 @opi I also lost track of this

alexAubin:
>can we have a sum up with command-line instruction & description ? (aka "Le documentation")

Well I know it's not really documentation, but there's the current actionmap [here](https://github.com/YunoHost/yunohost/blob/609daf2ca48e7b5a7316961fd462b6ab0d0e6986/data/actionsmap/yunohost.yml#L927-L974)

The main point remaining was to converge on the CLI/API design for resetting / put a given (or all) settings back to default value.

There's also the question of wether merging this or not if we have no use case implemented for now (c.f. the first post "We shouldn't merged this PR if we don't have a need yet for this module"). But honestly I think it doesn't hurt to have this "framework" merged even if it's not used, so that we can move on to actually implement settings in other PRs...

src/yunohost/settings.py
```diff
@@ -0,0 +1,236 @@
...
+
+    """
+    settings = _get_settings()
+
+    # For now on, we backup the previous settings in case of but we don't have
+    # any mecanism to take advantage of those backups. It could be a nice
+    # addition but we'll see if this is a common need.
+    # Another solution would be to use etckeeper and integrate those
+    # modification inside of it and take advantage of its git history
+    old_settings_backup_path = SETTINGS_PATH_OTHER_LOCATION % datetime.now().strftime("%F_%X")
```
julienmalik: it seems inconsistent to me that the backup is done only with reset-all. Either we support backup, and then we perform a backup at each user modification, either we don't support backup at all for now.

i don't have a strong opinion on whether to perform backup or not. as long as we don't have a huge set of settings, I tend to think we can do without.

julienmalik:
I don't have a very strong opinion on the remaining reset/reset-all/to-default disagreement.

To try to move forward, this is how I see it :

reset/reset-all is crystal clear to me. I agree with @alexAubin about https://github.com/YunoHost/yunohost/pull/229#issuecomment-290695492 they are both the same operation, just applied to a different set. Also I add my previous comment about doing backup consistently in those two cases (then also on the set command). And to move forward, I think we can simply delete this backup for now and keep the backup discussion for later (we already have etckeeper installed, so nothing is really lost?).

since @Psycojoker is very picky about the --yes, and that it is indeed a simple but efficient protection against misunderstanding, I would prefer we add it back.
I don't care about the "beauty" of the API call (which regular user uses that directly? it's done by the web admin)
About consistency with other commands, that's a good point, but I'm on @Psycojoker side on this : better add similar protection on existing commands (app removal, cert modification).




Please don't take my review too seriously, as my opinion above requires yet another layer of modifs... If you guys agree on the current state, I'm ok with this.

alexAubin:
>since @Psycojoker is very picky about the --yes, and that it is indeed a simple but efficient protection against misunderstanding, I would prefer we add it back.
I don't care about the "beauty" of the API call (which regular user uses that directly? it's done by the web admin)
About consistency with other commands, that's a good point, but I'm on @Psycojoker side on this : better add similar protection on existing commands (app removal, cert modification).

About that point, to me the --yes should be for CLI only. We could get some inspiration from [here](https://github.com/YunoHost/yunohost/blob/7718ed60004d1cd34066f8467e375aa71f287adb/src/yunohost/tools.py#L423) to know if we're called from the CLI and ask for explicit confirmation from the user. But this should really be discussed for all "dangerous" commands (app remove, ...) and not just for this one.

As said, my point of view is that using the CLI (in linux in general) **is** (to some extend) dangerous (see rm, cp, mv). Some command do have a --yes like option (like apt-get install), but from my understanding this is because their behavior changes a lot according to the context (depending of the package, some others are likely to be removed / installed) and justify to ask the user for confirmation after he reads everything. But our "app remove" and "settings reset-all" always does the same thing.

In any case, we can still implement this a confirmation mechanism later if the majority is in favor of it.

PR #231, opened 2017-01-16
[mod] start putting timeout in certificate code

Psycojoker:
As discussed with @julienmalik and @alexAubin, I'm starting this PR so we can discuss about how we should handle timeouting requests in the cert management.

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -571,7 +571,7 @@ def _fetch_and_enable_new_certificate(domain, staging=False):
 
     try:
         intermediate_certificate = requests.get(INTERMEDIATE_CERTIFICATE_URL, timeout=30).text
-    except Timeout:
+    except Timeout as e:
         raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_couldnt_fetch_intermediate_cert'))
 
     # Now save the key and signed certificate
@@ -844,7 +844,7 @@ def _domain_is_accessible_through_HTTP(ip, domain):
     try:
         requests.head("http://" + ip, headers={"Host": domain}, timeout=10)
     except Timeout as e:
-        logger.warning(m18n.n('certmanager_http_check_timeout'))
+        logger.warning(m18n.n('certmanager_http_check_timeout', domain=domain, ip=ip))
         return False
     except Exception as e:
         logger.debug("Couldn't reach domain '%s' by requesting this ip '%s' because: %s" % (domain, ip, e))
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -572,8 +572,7 @@ def _fetch_and_enable_new_certificate(domain, staging=False):
     try:
         intermediate_certificate = requests.get(INTERMEDIATE_CERTIFICATE_URL, timeout=30).text
     except Timeout:
-        # XXX what should we do here? retry?
-        pass
+        raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_couldnt_fetch_intermediate_cert'))
 
     # Now save the key and signed certificate
     logger.info("Saving the key and signed certificate...")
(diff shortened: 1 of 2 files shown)


src/yunohost/certificate.py
```diff
@@ -567,7 +569,11 @@ def _fetch_and_enable_new_certificate(domain, staging=False):
         raise MoulinetteError(errno.EINVAL, m18n.n(
             'certmanager_cert_signing_failed'))
 
-    intermediate_certificate = requests.get(INTERMEDIATE_CERTIFICATE_URL).text
+    try:
+        intermediate_certificate = requests.get(INTERMEDIATE_CERTIFICATE_URL, timeout=30).text
+    except Timeout:
+        # XXX what should we do here? retry?
```
alexAubin: If we didn't get the intermediate certificate, the certificate won't work, so I'd say we just do a `raise MoulinetteError(errno.EINVAL, m18n.n('certmanager_couldnt_fetch_intermediate_cert'))`

(Btw, I never know if we do care about which errno to use ?)

src/yunohost/certificate.py
```diff
@@ -837,7 +843,10 @@ def _dns_ip_match_public_ip(public_ip, domain):
 
 def _domain_is_accessible_through_HTTP(ip, domain):
     try:
-        requests.head("http://" + ip, headers={"Host": domain})
+        requests.head("http://" + ip, headers={"Host": domain}, timeout=30)
+    except Timeout as e:
+        # XXX what should we do here? retry?
```
alexAubin: If we really have a 30 sec timeout (which to me is crazy for what we are trying to do here), then we should return False.

But maybe we could have something like : 

- have a 10-15 second timeout first
- if it doesn't work, have something like "Warning : having trouble contacting domain. Retrying..."
- retry with a 30 second timeout
- return False if still timeouting

But I'm having a hard time imagining a situation where the domain was up and everything running properly, but it takes more than 30 seconds for a HEAD request to complete :/...

julienmalik: 30 sec also seems a lot to me. 5 to 10 should be more than enough. Here we are trying to contact ourself, and we just want to detect the case where it does not work at all (due to hairpinning or other weird firewall setups).

src/yunohost/certificate.py
julienmalik: trusting @alexAubin  on this

src/yunohost/certificate.py
julienmalik: I may not see the big picture on this, but I would go for something way more simple : if timeout after 5 to 10 seconds, then just abort with a good error message, like "timeout reached when your server tried to contact itself via its public IP address. you may very well be experiencing hairpinning, or the firewall/router ahead of your server is misconfigured"

alexAubin: @julienmalik Well yep, but @Psycojoker was pushing to have a 30 sec timeout 😛 . But I'd be in favor of this too.

alexAubin:
Haven't tested the real timeout, but I tested the lines inside the exceptions 😕

Psycojoker:
We need one more approval for this one.

src/yunohost/certificate.py
```diff
@@ -567,7 +569,10 @@ def _fetch_and_enable_new_certificate(domain, staging=False):
         raise MoulinetteError(errno.EINVAL, m18n.n(
             'certmanager_cert_signing_failed'))
 
-    intermediate_certificate = requests.get(INTERMEDIATE_CERTIFICATE_URL).text
+    try:
+        intermediate_certificate = requests.get(INTERMEDIATE_CERTIFICATE_URL, timeout=30).text
+    except Timeout:
```
opi: you should add `as e` to be conpliant with yunohost coding standards (tm)

src/yunohost/certificate.py
```diff
@@ -837,7 +842,10 @@ def _dns_ip_match_public_ip(public_ip, domain):
 
 def _domain_is_accessible_through_HTTP(ip, domain):
     try:
-        requests.head("http://" + ip, headers={"Host": domain})
+        requests.head("http://" + ip, headers={"Host": domain}, timeout=10)
+    except Timeout as e:
+        logger.warning(m18n.n('certmanager_http_check_timeout'))
```
opi: error message should include domain and ip information to be more relevant

alexAubin:
Thanks for the feedback !

PR #238, opened 2017-01-23
[enh] Add postgresql helpers

zamentur:
I pull request just for memo and to signal this work (it was in my _common.sh file for packaging...).

The work is unfinished. The api is similar to ynh_mysql helpers.

### DIFF
--- a/data/helpers.d/psql
+++ b/data/helpers.d/psql
@@ -1,12 +1,7 @@
-#!/bin/bash
-
-#=================================================
+# Create a master password and set up global settings
+# Please always call this script in install and restore scripts
 #
-# POSTGRES HELPERS
-#
-# Point of contact : Jean-Baptiste Holcroft <jean-baptiste@holcroft.fr>
-#=================================================
-
+# usage: ynh_psql_test_if_first_run
 ynh_psql_test_if_first_run() {
 	if [ -f /etc/yunohost/psql ];
 	then
@@ -27,8 +22,12 @@ ynh_psql_test_if_first_run() {
 		fi
 
 		systemctl start postgresql
-                su --command="psql -c\"ALTER user postgres WITH PASSWORD '${pgsql}'\"" postgres
-		# we can't use peer since YunoHost create users with nologin
+		sudo --login --user=postgres psql -c"ALTER user postgres WITH PASSWORD '$pgsql'" postgres
+
+		# force all user to connect to local database using passwords
+		# https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html#EXAMPLE-PG-HBA.CONF
+		# Note: we can't use peer since YunoHost create users with nologin
+		#  See: https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/user
 		sed -i '/local\s*all\s*all\s*peer/i \
 		local all all password' "$pg_hba"
 		systemctl enable postgresql
@@ -49,7 +48,7 @@ ynh_psql_connect_as() {
 	user="$1"
 	pwd="$2"
 	db="$3"
-	su --command="PGUSER=\"${user}\" PGPASSWORD=\"${pwd}\" psql \"${db}\"" postgres
+	sudo --login --user=postgres PGUSER="$user" PGPASSWORD="$pwd" psql "$db"
 }
 
 # # Execute a command as root user
@@ -59,7 +58,7 @@ ynh_psql_connect_as() {
 # | arg: db - the database to connect to
 ynh_psql_execute_as_root () {
 	sql="$1"
-	su --command="psql" postgres <<< "$sql"
+	sudo --login --user=postgres psql <<< "$sql"
 }
 
 # Execute a command from a file as root user
@@ -70,7 +69,7 @@ ynh_psql_execute_as_root () {
 ynh_psql_execute_file_as_root() {
 	file="$1"
 	db="$2"
-	su -c "psql $db" postgres < "$file"
+	sudo --login --user=postgres psql "$db" < "$file"
 }
 
 # Create a database, an user and its password. Then store the password in the app's config
@@ -84,7 +83,6 @@ ynh_psql_execute_file_as_root() {
 # | arg: pwd - Password of the database. If not given, a password will be generated
 ynh_psql_setup_db () {
 	db_user="$1"
-	app="$1"
 	db_name="$2"
 	new_db_pwd=$(ynh_string_random)	# Generate a random password
 	# If $3 is not given, use new_db_pwd instead for db_pwd.
@@ -93,7 +91,7 @@ ynh_psql_setup_db () {
 	ynh_app_setting_set "$app" psqlpwd "$db_pwd"	# Store the password in the app's config
 }
 
-# Create a database and grant optionnaly privilegies to a user
+# Create a database and grant privilegies to a user
 #
 # usage: ynh_psql_create_db db [user [pwd]]
 # | arg: db - the database name to create
@@ -104,7 +102,7 @@ ynh_psql_create_db() {
 	user="$2"
 	pwd="$3"
 	ynh_psql_create_user "$user" "$pwd"
-	su --command="createdb --owner=\"${user}\" \"${db}\"" postgres
+	sudo --login --user=postgres createdb --owner="$user" "$db"
 }
 
 # Drop a database
@@ -115,8 +113,8 @@ ynh_psql_create_db() {
 ynh_psql_remove_db() {
 	db="$1"
 	user="$2"
-	su --command="dropdb \"${db}\"" postgres
-	ynh_psql_drop_user "${user}"
+	sudo --login --user=postgres dropdb "$db"
+	ynh_psql_drop_user "$user"
 }
 
 # Dump a database
@@ -128,7 +126,7 @@ ynh_psql_remove_db() {
 # | ret: the psqldump output
 ynh_psql_dump_db() {
 	db="$1"
-	su --command="pg_dump \"${db}\"" postgres
+	sudo --login --user=postgres pg_dump "$db"
 }
 
 
@@ -139,7 +137,7 @@ ynh_psql_dump_db() {
 ynh_psql_create_user() {
 	user="$1"
 	pwd="$2"
-        su --command="psql -c\"CREATE USER ${user} WITH PASSWORD '${pwd}'\"" postgres
+        sudo --login --user=postgres psql -c"CREATE USER $user WITH PASSWORD '$pwd'" postgres
 }
 
 # Drop a user
@@ -148,5 +146,5 @@ ynh_psql_create_user() {
 # | arg: user - the user name to drop
 ynh_psql_drop_user() {
 	user="$1"
-	su --command="dropuser \"${user}\"" postgres
+	sudo --login --user=postgres dropuser "$user"
 }
--- a/data/helpers.d/psql
+++ b/data/helpers.d/psql
@@ -1,3 +1,40 @@
+#!/bin/bash
+
+#=================================================
+#
+# POSTGRES HELPERS
+#
+# Point of contact : Jean-Baptiste Holcroft <jean-baptiste@holcroft.fr>
+#=================================================
+
+ynh_psql_test_if_first_run() {
+	if [ -f /etc/yunohost/psql ];
+	then
+		echo "PostgreSQL is already installed, no need to create master password"
+	else
+		pgsql=$(ynh_string_random)
+		pg_hba=""
+		echo "$pgsql" >> /etc/yunohost/psql
+
+		if [ -e /etc/postgresql/9.4/ ]
+		then
+			pg_hba=/etc/postgresql/9.4/main/pg_hba.conf
+		elif [ -e /etc/postgresql/9.6/ ]
+		then
+			pg_hba=/etc/postgresql/9.6/main/pg_hba.conf
+		else
+			ynh_die "postgresql shoud be 9.4 or 9.6"
+		fi
+
+		systemctl start postgresql
+                su --command="psql -c\"ALTER user postgres WITH PASSWORD '${pgsql}'\"" postgres
+		# we can't use peer since YunoHost create users with nologin
+		sed -i '/local\s*all\s*all\s*peer/i \
+		local all all password' "$pg_hba"
+		systemctl enable postgresql
+		systemctl reload postgresql
+	fi
+}
 
 # Open a connection as a user
 #
@@ -7,9 +44,12 @@
 # usage: ynh_psql_connect_as user pwd [db]
 # | arg: user - the user name to connect as
 # | arg: pwd - the user password
-# | arg: db - the database to connect to
+# | arg: db - the database to connect to
 ynh_psql_connect_as() {
-    ynh_die "ynh_psql_connect_as is not yet implemented"
+	user="$1"
+	pwd="$2"
+	db="$3"
+	su --command="PGUSER=\"${user}\" PGPASSWORD=\"${pwd}\" psql \"${db}\"" postgres
 }
 
 # # Execute a command as root user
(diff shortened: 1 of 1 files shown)


data/helpers.d/psql
```diff
@@ -0,0 +1,88 @@
...
+
+
+# Create a user
+#
+# usage: ynh_psql_create_user user pwd [host]
+# | arg: user - the user name to create
+# | arg: pwd - the password to identify user by
+ynh_psql_create_user() {
+        ynh_psql_execute_as_root \
+        "CREATE USER ${1} WITH PASSWORD '${2}';"
```
Psycojoker: Postgresql est livré avec une commande buildin `createuser` qui permet justement de ne pas avoir à gérer de mot de passe et utiliser l'authentification unix. On devrait utiliser ça à la place.

Le seul cas où on peut avoir envie d'utiliser un mot de passe c'est pour php qui est trop débile pour se connecter à postgresql sans password (et là on peut avoir un helper spécial "php est trop idiot").

alexAubin:
Flagging this as work needed because it looks like it's not done yet

Psycojoker:
Up ?

alexAubin:
Recent work on the COIN app could be useful here I guess : https://github.com/YunoHost-Apps/coin_ynh/commit/b3cecb64cd89fdd4e4306734e602a895f96613ee

data/helpers.d/psql
Jibec: While your request to use createuser is valid, to use Unix auth, we need to allow user connection, here is what our helper does (nologin):
https://github.com/YunoHost/yunohost/blob/83dca8e7c6ec4efb206140c234f51dfa5b3f3bf7/data/helpers.d/user#L55

Trying `su - user` will lead to "This account is currently not available."
So, default Postgres configuration (local all all peer => local connection for all database and all users uses unix users) is not compatible with our current user creation helper.

I understood this from: https://askubuntu.com/a/486661 (I hope I understood it well ;))

So, we'll have to use PGPASSWORD, but there is a warning:
PGPASSWORD behaves the same as the password connection parameter. Use of this environment variable is not recommended for security reasons, as some operating systems allow non-root users to see process environment variables via ps; instead consider using the ~/.pgpass file (see Section 31.15).

Jibec:
For your information, I rewrote part of these helpers:
https://github.com/YunoHost-Apps/weblate_ynh/blob/master/scripts/_common.sh

The most important one is the global checker: ynh_psql_test_if_first_run
https://github.com/YunoHost-Apps/weblate_ynh/blob/master/scripts/_common.sh#L5

I also renamed some helpers to match the MySQL ones, and finished the TODO ones.
It still needs some testing, because I'm nor dev nor admin.
I have no idea how to contribute directly to this PR...

alexAubin:
To be updated with most recent postgres header from the experimental header repo ... https://github.com/YunoHost-Apps/Experimental_helpers/blob/master/postgres/postgres

alexAubin:
Apparently Maniack updated those a few weeks ago, and looking at the history these helpers looks pretty stable (and used by many apps)

So re-activating this PR ;)

Jibec:
Glad to see it evaluated as stable enough for integration!

There is still some warning messages I'd like to fix before making it core. See experimental helpers issues. Note: anyone can fix the issue, there is no embargo.

About delay, is it to be fixed for Debian 9 migration?

I need direction about how to contribute this PR.

Psycojoker:
Shouldn't we close this then and open a new PR once it's ready to be moved into core?

alexAubin:
Soooo, updated the helpers according to what's been merged in https://github.com/YunoHost-Apps/Experimental_helpers/issues/13

Does that looks okay to everybody now ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+	then
+		echo "PostgreSQL is already installed, no need to create master password"
+	else
+		pgsql=$(ynh_string_random)
+		pg_hba=""
+		echo "$pgsql" >> /etc/yunohost/psql
+
+		if [ -e /etc/postgresql/9.4/ ]
+		then
+			pg_hba=/etc/postgresql/9.4/main/pg_hba.conf
```
JimboJoe: Is the `/etc/postgresql/9.4/main/pg_hba.conf` file removed when migrating YunoHost along with PostgreSQL 9.6...?

alexAubin: Yes I think so, I think I recall I checked this and the whole /etc/postgresql/9.4 folder gets removed so the current code should be fine

alexAubin:
(Changed the author of previous commit to please Jibec-sama :wink: )

Jibec:
```
   ***     ***    
 **   ** **   **  
*       *       *    
*               *   
 *    THANKS   *      
  **         **   
    **     **
      ** ** 
        *  
          
```

alexAubin:
Okay guys, I'm merging this since I guess we all agree to include postgresql helpers in general, and this is currently the stable version ... Feel free to amend them later if needed

maniackcrudelis:
Huh, do am I the only to see all this variables with really common name ("user", "pwd", "file", ...) which are not local !?
Also the comments explaining how to use the helpers are full of typo.

Either we revert this PR to fix it or I can add some comments, but is there someone who know how this helper works ? I'll probably have some technical questions.

Jibec:
OK, let's get rid of this PR.

maniackcrudelis:
> **Either** we **revert** this PR to fix it **or** I can **add some comments**, but is there someone who know how this helper works ? I'll probably have some technical questions.

-_-

alexAubin:
Well it's still not officially release so it doesn't seem too critical ? I think we can just make a new PR that fixes what needs to be fixed ?

alexAubin:
Why not just go to https://github.com/YunoHost/yunohost/edit/unstable/data/helpers.d/psql ? :stuck_out_tongue:

maniackcrudelis:
Because I can't make comments here (can I ?), and I need some question about some part of the code.
This helper isn't from me, I don't know how it's suppose to work. For some part at least.

maniackcrudelis:
...well... right now I'm not sure that anyone who has code that is about to read my comments...
Anyway, at least, there're going to be somewhere...

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
+# Create a master password and set up global settings
+# Please always call this script in install and restore scripts
+#
+# usage: ynh_psql_test_if_first_run
+ynh_psql_test_if_first_run() {
+	if [ -f /etc/yunohost/psql ];
+	then
+		echo "PostgreSQL is already installed, no need to create master password"
+	else
+		pgsql=$(ynh_string_random)
```
maniackcrudelis: What is $pgsql ?
Isn't this variable should be locale ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+		pg_hba=""
```
maniackcrudelis: What is it ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+		echo "$pgsql" >> /etc/yunohost/psql
```
maniackcrudelis: Why >> instead of > ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+
+		if [ -e /etc/postgresql/9.4/ ]
+		then
+			pg_hba=/etc/postgresql/9.4/main/pg_hba.conf
+		elif [ -e /etc/postgresql/9.6/ ]
+		then
+			pg_hba=/etc/postgresql/9.6/main/pg_hba.conf
+		else
+			ynh_die "postgresql shoud be 9.4 or 9.6"
```
maniackcrudelis: Why ?
What's going to happen with the future versions ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+		fi
+
+		systemctl start postgresql
+		sudo --login --user=postgres psql -c"ALTER user postgres WITH PASSWORD '$pgsql'" postgres
```
maniackcrudelis: Why not just root ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+
+		# force all user to connect to local database using passwords
+		# https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html#EXAMPLE-PG-HBA.CONF
+		# Note: we can't use peer since YunoHost create users with nologin
+		#  See: https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/user
+		sed -i '/local\s*all\s*all\s*peer/i \
+		local all all password' "$pg_hba"
```
maniackcrudelis: Indent here

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+# example: ynh_psql_connect_as 'user' 'pass' < /path/to/file.sql
+#
+# usage: ynh_psql_connect_as user pwd [db]
+# | arg: user - the user name to connect as
+# | arg: pwd - the user password
+# | arg: db - the database to connect to
+ynh_psql_connect_as() {
+	user="$1"
+	pwd="$2"
+	db="$3"
```
maniackcrudelis: All variables should be locale, unless there's a real reason to do otherwise. Particularly with such common names !
So we need to know if there's any variables which could/should be use outside the scope of these functions.

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+		local all all password' "$pg_hba"
+		systemctl enable postgresql
+		systemctl reload postgresql
+	fi
+}
+
+# Open a connection as a user
+#
+# example: ynh_psql_connect_as 'user' 'pass' <<< "UPDATE ...;"
+# example: ynh_psql_connect_as 'user' 'pass' < /path/to/file.sql
```
maniackcrudelis: Input redirection shouldn't be used without real descriptor management.
That kind of redirection could break loops.

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+#
+# usage: ynh_psql_connect_as user pwd [db]
+# | arg: user - the user name to connect as
+# | arg: pwd - the user password
+# | arg: db - the database to connect to
+ynh_psql_connect_as() {
+	user="$1"
+	pwd="$2"
+	db="$3"
+	sudo --login --user=postgres PGUSER="$user" PGPASSWORD="$pwd" psql "$db"
```
maniackcrudelis: Same as previous, why not using root ?

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+}
+
+# # Execute a command as root user
+#
+# usage: ynh_psql_execute_as_root sql [db]
+# | arg: sql - the SQL command to execute
+# | arg: db - the database to connect to
```
maniackcrudelis: $db isn't used in this helper.

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+ynh_psql_execute_as_root () {
+	sql="$1"
+	sudo --login --user=postgres psql <<< "$sql"
```
maniackcrudelis: Does not seems to be executed by "root" as mentioned in the helper name.

data/helpers.d/psql
```diff
@@ -0,0 +1,150 @@
...
+# Drop a database
+#
+# usage: ynh_psql_drop_db db
+# | arg: db - the database name to drop
+# | arg: user - the user to drop
+ynh_psql_remove_db() {
+	db="$1"
+	user="$2"
+	sudo --login --user=postgres dropdb "$db"
+	ynh_psql_drop_user "$user"
```
maniackcrudelis: The user will be also deleted, not only the database !

data/helpers.d/psql
```diff
@@ -0,0 +1,152 @@
...
+	db="$1"
+	su --command="pg_dump \"${db}\"" postgres
+}
+
+
+# Create a user
+#
+# usage: ynh_psql_create_user user pwd [host]
+# | arg: user - the user name to create
+ynh_psql_create_user() {
```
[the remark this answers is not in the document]
maniackcrudelis: Same for some other helpers here.

data/helpers.d/psql
Psycojoker: Postgresql is a bit weird and except you to do everything using the "postgres" user which is the ~root~ of postgresql while the root user won't have any permission at all to work on postgresql by default.

So that's the postgresql way of doing things (don't ask why).

data/helpers.d/psql
Josue-T: We could also get the version by something like : 
`dpkg -l | grep -P -o "postgresql\s*\d\.\d" | grep -P -o "\d\.\d"`

data/helpers.d/psql
Psycojoker: Yes it should.

data/helpers.d/psql
Psycojoker: This is for compatibility between stretch and jessie.

data/helpers.d/psql
Psycojoker: > can be used too according to the code.

data/helpers.d/psql
Psycojoker: I guess we'll need to update the helpers because we can't predict how things are going to be yet.

data/helpers.d/psql
Psycojoker: Apparently, from the spirit of this helper, it's because the user it closely linked to the database, that's not a user for other things.

data/helpers.d/psql
maniackcrudelis: Still, what $pg_hba is?
Either a comment should explain what is it, or a concrete variable name.
If someone has to fix anything around here, he should know what's going on and how without trying to decrypt the code.

data/helpers.d/psql
maniackcrudelis: ?

Using >> instead of > isn't without consequences ! That's why I asked if there's a reason or if it's just a mistake.

data/helpers.d/psql
maniackcrudelis: Still, why 9.4 or 9.6 only?

data/helpers.d/psql
maniackcrudelis: But the helper name is "ynh_psql_remove_db" and the comment says "Drop a database".
Comment as well as helper name have to be consistent.
So here there's a double matter, either the name/comment should be fixed, or the helper isn't supposed to delete the user.

That's why I said before that I need someone who have write this helper to answer my questions.
But clearly, some are fleeing any confrontation with my questions...

data/helpers.d/psql
JimboJoe: It's used to store the pgsql password for `postgres` user. A comment must be added.

data/helpers.d/psql
JimboJoe: It's used to store the path the the `pg_hba.conf` file ([client authentication control file](https://www.postgresql.org/docs/9.6/static/auth-pg-hba-conf.html)). A comment must be added.

data/helpers.d/psql
JimboJoe: As the file is used to store the unique password (analogous to mysql handling), it should be a `>`.

data/helpers.d/psql
JimboJoe: 9.4 was the Jessie version, 9.6 is the Stretch version. Still, a comment must be added, and the helpers will need to be upgraded at next Debian migration (analogous to php-fpm helpers).

data/helpers.d/psql
JimboJoe: For this I think you need to be more explicit, this is the current official way to use mysql helpers... Waht would you recommend?

data/helpers.d/psql
JimboJoe: For what it's worth, this behavior is analogous to the mysql helpers' behavior (see [here](https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/mysql#L155)).

JimboJoe:
To help move this forward, I've pushed a work I did a couple of weeks ago: migrate Nextcloud to PostgreSQL. The most difficult part is missing (migration...), but you'll clearly see the API differences between MySQL and PostgreSQL.

https://github.com/YunoHost-Apps/nextcloud_ynh/pull/125

I would say that the most striking differences are:
- the need to use `ynh_psql_test_if_first_run`
- the way the password is (not) handled, as compared to mysql helpers.

The PostgreSQL helpers were originally developed by @zamentur , and moved forward by @jibec. Maybe they could step into that conversation... :wink:

Jibec:
With postgres helpers being included in the next release, don't you wish to wait for it?
I'm glad to see a mysql to postgres migration. Maybe we can ask core team to help us (and start with a less critical application :p). I'm pretty sure migrate core to postgres would be appreciated.
-- 
Jean-Baptiste Holcroft

Jibec:
Oops, I thought I answered the nextcloud PR and not the core one ;)

Psycojoker:
I've pushed quick fixes implementing most of the feedback here before the release that we've integrated with Aleks to help this issue move forward: https://github.com/YunoHost/moulinette/commit/fe22b1024c6929caf4dfc911f39623076e8acb39

But we also have chosen not to advertise those helpers in the releases note and to implement ASAP additional feedbacks to get to a good point while discouraging their usage before reaching that point.

data/helpers.d/psql
maniackcrudelis: No, the corresponding helper for mysql is [here](https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/mysql#L59-L70) and "drop" only the database, as its name let's imagine.
Instead, the helper you pointed out, ynh_mysql_remove_db, says clearly "[Remove a database if it exists, and the associated user](https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/mysql#L139)"

PR #245, opened 2017-02-08
[enh] New helper ynh_abort_if_errors

maniackcrudelis:
Helpers to implemente trap and set -eu.

Simply to use, you only add `ynh_check_error` in your script.

### DIFF
--- a/data/helpers.d/system
+++ b/data/helpers.d/system
@@ -36,8 +36,8 @@ ynh_exit_properly () {
 # Stop immediatly the execution if an error occured or if a empty variable is used.
 # The execution of the script is derivate to ynh_exit_properly function before exit.
 #
-# Usage: ynh_check_error
-ynh_check_error () {
+# Usage: ynh_abort_if_errors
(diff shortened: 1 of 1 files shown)


data/helpers.d/system
```diff
@@ -0,0 +1,43 @@
...
+	ynh_die	# Exit with error status
+}
+
+# Exit if an error occurs during the execution of the script.
+#
+# Stop immediatly the execution if an error occured or if a empty variable is used.
+# The execution of the script is derivate to ynh_exit_properly function before exit.
+#
+# Usage: ynh_check_error
+ynh_check_error () {
```
alexAubin: Imho a name like `ynh_trap_errors` or `ynh_abort_if_errors` would be more meaningful, but if people are happy with the current name it's okay.

Psycojoker: Agree.

maniackcrudelis:
ynh_abort_if_errors seems nice.

I use this helper with all my packages. It works fine.

PR #246, opened 2017-02-08
[fix] Can't use common.sh on restore operation

zamentur:
- [x] Need https://github.com/YunoHost/moulinette/pull/134

Redmine ticket https://dev.yunohost.org/issues/621

This pull request allow to use _common.sh in restore script.

I have deleted the copy in a tmp dir, because we are already in a tmp dir...

To import common.sh in a restore script

source ../settings/scripts/_common.sh

And not 

source _common.sh


Psycojoker:
This one looks rather critical. Should we make a bugfix release for it?

Psycojoker:
Why were we doing restore in a tmp dir instead?

opi:
YunoHost/moulinette#134 is now merged and released in [2.5.3 testing release](https://github.com/YunoHost/moulinette/releases/tag/debian/2.5.3)

zamentur:
@Psycojoker to untar files before to run the restore script.

An improvement could be to implement a way to extract files directly from  the archives (with different archives format: tar, dir or borg archives)

zamentur:
One week was needed and we have ou 2 positive views (ljf and Aleks). So I merge

PR #247, opened 2017-02-08
[enh] New helpers for equivs use

maniackcrudelis:
- `ynh_package_remove` and `ynh_package_autoremove` were just moved up.
- Small modifications on `ynh_package_install_from_equivs`. Just added some comments and used popd instead of cd.
- Added `ynh_app_dependencies` to manage easily installation of dependencies with equivs
- And `ynh_remove_app_dependencies` to remove them.

### DIFF
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -104,11 +104,12 @@ ynh_package_install_from_equivs () {
     ynh_package_is_installed "$pkgname"
 }
 
-# Install dependencies with a equivs control file
+# Define and install dependencies with a equivs control file
+# This helper can/should only be called once per app
 #
-# usage: ynh_app_dependencies dep [dep [...]]
+# usage: ynh_install_app_dependencies dep [dep [...]]
 # | arg: dep - the package name to install in dependence
-ynh_app_dependencies () {
+ynh_install_app_dependencies () {
     dependencies=$@
     manifest_path="../manifest.json"
     if [ ! -e "$manifest_path" ]; then
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -89,19 +89,19 @@ ynh_package_install_from_equivs () {
 
     # Build and install the package
     TMPDIR=$(ynh_mkdir_tmp)
-    cp "$controlfile" "${TMPDIR}/control" && pushd "$TMPDIR"	# pushd is like a cd, but it stores the previous directory
+    # Note that the cd executes into a sub shell
     # Create a fake deb package with equivs-build and the given control file
     # Install the fake package without its dependencies with dpkg
     # Install missing dependencies with ynh_package_install
-    equivs-build ./control 1>/dev/null \
+    (cp "$controlfile" "${TMPDIR}/control" && cd "$TMPDIR" \
+     && equivs-build ./control 1>/dev/null \
      && sudo dpkg --force-depends \
           -i "./${pkgname}_${pkgversion}_all.deb" 2>&1 \
-     && ynh_package_install -f
+     && ynh_package_install -f)
     [[ -n "$TMPDIR" ]] && rm -rf $TMPDIR	# Remove the temp dir.
 
     # check if the package is actually installed
     ynh_package_is_installed "$pkgname"
-    popd	# Like a cd on the directory stored by pushd
 }
 
 # Install dependencies with a equivs control file
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -116,7 +116,11 @@ ynh_app_dependencies () {
     fi
     version=$(sudo python3 -c "import sys, json;print(json.load(open(\"$manifest_path\"))['version'])")	# Retrieve the version number in the manifest file.
     dep_app=${app//_/-}	# Replace all '_' by '-'
-    cat > ./${dep_app}-ynh-deps.control << EOF	# Make a control file for equivs-build
+
+    if ynh_package_is_installed "${dep_app}-ynh-deps"; then
+		echo "A package named ${dep_app}-ynh-deps is already installed" >&2
+    else
+        cat > ./${dep_app}-ynh-deps.control << EOF	# Make a control file for equivs-build
 Section: misc
 Priority: optional
 Package: ${dep_app}-ynh-deps
@@ -126,9 +130,10 @@ Architecture: all
 Description: Fake package for ${app} (YunoHost app) dependencies
  This meta-package is only responsible of installing its dependencies.
 EOF
-    ynh_package_install_from_equivs ./${dep_app}-ynh-deps.control \
-        || ynh_die "Unable to install dependencies"	# Install the fake package and its dependencies
-    ynh_app_setting_set $app apt_dependencies $dependencies
+        ynh_package_install_from_equivs ./${dep_app}-ynh-deps.control \
+            || ynh_die "Unable to install dependencies"	# Install the fake package and its dependencies
+        ynh_app_setting_set $app apt_dependencies $dependencies
+    fi
 }
 
 # Remove fake package and its dependencies
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -109,8 +109,12 @@ ynh_package_install_from_equivs () {
 # usage: ynh_app_dependencies dep [dep [...]]
 # | arg: dep - the package name to install in dependence
 ynh_app_dependencies () {
-	dependencies=$1
-	version=$(sudo python3 -c "import sys, json;print(json.load(open('../manifest.json'))['version'])")	# Retrieve the version number in the manifest file.
+    dependencies=$@
+    manifest_path="../manifest.json"
+    if [ ! -e "$manifest_path" ]; then
+    	manifest_path="../settings/manifest.json"	# Into the restore script, the manifest is not at the same place
+    fi
+    version=$(sudo python3 -c "import sys, json;print(json.load(open(\"$manifest_path\"))['version'])")	# Retrieve the version number in the manifest file.
     dep_app=${app/_/-}	# Replace all '_' by '-'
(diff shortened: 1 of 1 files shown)


data/helpers.d/package
```diff
@@ -55,48 +71,67 @@ ynh_package_install() {
...
 
-# Remove package(s)
+# Install dependencies with a equivs control file
 #
-# usage: ynh_package_remove name [name [...]]
-# | arg: name - the package name to remove
-ynh_package_remove() {
-    ynh_apt remove $@
+# usage: ynh_app_dependencies dep [dep [...]]
+# | arg: dep - the package name to install in dependence
```
zamentur: May be we could read dependencies from manifest.json in "requirements".

Exemple of "Depends" statements:

`Depends: libc6 (>= 2.2.1), exim | mail-transport-agent`

So we should offer this possibilities too in doc of this helpers
ynh_app_dependencies "libc6 (>= 2.2.1), exim | mail-transport-agent`"

data/helpers.d/package
```diff
@@ -55,48 +71,67 @@ ynh_package_install() {
...
+ynh_app_dependencies () {
+	dependencies=$1
```
zamentur: $@ no?

data/helpers.d/package
```diff
@@ -55,48 +71,67 @@ ynh_package_install() {
...
+	version=$(sudo python3 -c "import sys, json;print(json.load(open('../manifest.json'))['version'])")	# Retrieve the version number in the manifest file.
+    dep_app=${app/_/-}	# Replace all '_' by '-'
+    cat > ./${dep_app}-ynh-deps.control << EOF	# Make a control file for equivs-build
+Section: misc
+Priority: optional
+Package: ${dep_app}-ynh-deps
+Version: ${version}
+Depends: ${dependencies}
+Architecture: all
+Description: Fake package for ${app} (YunoHost app) dependencies
```
zamentur: Why "Fake", it's not fake. It is a minimalist meta package.

data/helpers.d/package
maniackcrudelis: Is really more simple to read them from the manifest?

data/helpers.d/package
maniackcrudelis: Yes ! Of courses !!

data/helpers.d/package
```diff
@@ -55,48 +71,67 @@ ynh_package_install() {
...
 #
-# usage: ynh_package_remove name [name [...]]
-# | arg: name - the package name to remove
-ynh_package_remove() {
-    ynh_apt remove $@
+# usage: ynh_app_dependencies dep [dep [...]]
+# | arg: dep - the package name to install in dependence
+ynh_app_dependencies () {
+	dependencies=$1
+	version=$(sudo python3 -c "import sys, json;print(json.load(open('../manifest.json'))['version'])")	# Retrieve the version number in the manifest file.
```
maniackcrudelis: So maybe we should add a check for this file before. And change the path if needed.

data/helpers.d/package
```diff
@@ -55,48 +71,67 @@ ynh_package_install() {
...
+    dep_app=${app/_/-}	# Replace all '_' by '-'
```
maniackcrudelis: Why not?

data/helpers.d/package
maniackcrudelis: It's an empty package. So it's not really a package ;)
But it's just a word, we can change it

data/helpers.d/package
```diff
@@ -55,48 +71,67 @@ ynh_package_install() {
...
      && sudo dpkg --force-depends \
           -i "./${pkgname}_${pkgversion}_all.deb" 2>&1 \
-     && ynh_package_install -f) \
-        && ([[ -n "$TMPDIR" ]] && rm -rf $TMPDIR)
+     && ynh_package_install -f
+    [[ -n "$TMPDIR" ]] && rm -rf $TMPDIR	# Remove the temp dir.
 
     # check if the package is actually installed
     ynh_package_is_installed "$pkgname"
+    popd	# Like a cd on the directory stored by pushd
```
zamentur: This popd could break retro compatibility.

maniackcrudelis:
I corrected the manifest's path and took all args for the dependencies.
But I not tested this modifications...

data/helpers.d/package
maniackcrudelis: So...
after some tests on nextcloud package. With `cd`, the directory still the same after this helper. I don't know why...
And with `pushd` and `popd`, popd fail because the stack is empty ! So, I missing something...

I'll correct this helper to remove pushd and replace cd.

maniackcrudelis: I'm so stupid...
What have missing is simply that the `cd` executes in a sub shell !
So I restore this behavior and remove these useless popd and pushd.

And add a small comment to prevent a another miss-correction :)

data/helpers.d/package
zamentur: More simple for the packager, because the data will be deduplicated.

maniackcrudelis: Yes, the name of dependencies packages will be duplicated into the restore script.
But, it's seems me less clear to store them into the manifest instead of install script.

Because it's still a install's action, and we keep the command into this scripts to install them.
So I'm not really convinced that it's better.

polytan02: From my own perspective, I don't really mind if we put the dependencies in each script or in the manifest, as long as it get automated and that the system is clean after the remove script. :)

In the manifest it sounds cool meaning we don't need to write it each time, however, it is very easy to add one line in install, upgrade and restore such as `ynh_app_install_dependencies dep1 dep2 dep3` and for the remove script `ynh_app_remove_dependencies`

data/helpers.d/package
polytan02: In Gentoo, they use these fake packages too. They call them "META" packages. We could reuse the terminology.

maniackcrudelis:
Tested with my packages, that works fine

data/helpers.d/package
alexAubin: We talked about this on the chat quickly with @maniackcrudelis : 
- the main issue is that there are sometimes optionnal dependencies. @maniackcrudelis cited the case of Abiword for etherpad (Abiword is optionnal and weights 300 Mo)
- therefore simply having a "requirements" field in the app manifest is not flexible enough
- however, having these in the app manifest had the advantage of having a clean and simple way to know what an app depends on for usages outside the app's scripts (e.g. from the yunohost core, for some use case we don't know yet)
- the solution is to add a new app setting with the list of dependencies *effectively used*. So for instance, if etherpad installs abiword, it should be listed in that setting. If not, then it's not listed. Simply calling `ynh_app_dependencies` should append the new dependencies to this setting.

By the way, here we're talking about debian packages, but the dependency/requirement problematic is much larger if we start talking about apps which include sources (e.g. .tar.gz) of fetch sources (wget, git clone, ...). For example with Wekan I had to install meteor, which can be considered as an "intermediate" dependency, with a wget. What if another app needs meteor ? It's pretty complicated to keep track and handle these dependencies properly :/ ...

maniackcrudelis: Done https://github.com/YunoHost/yunohost/pull/247/commits/2bb32c26b64f274380adafaefbf877667ac3e092

data/helpers.d/package
```diff
@@ -55,48 +71,72 @@ ynh_package_install() {
...
 }
 
-# Remove package(s)
+# Install dependencies with a equivs control file
 #
-# usage: ynh_package_remove name [name [...]]
-# | arg: name - the package name to remove
-ynh_package_remove() {
-    ynh_apt remove $@
+# usage: ynh_app_dependencies dep [dep [...]]
```
alexAubin: Imho this should be renamed as `ynh_install_app_dependencies` to be the symetric of `ynh_remove_app_dependencies`

data/helpers.d/package
```diff
@@ -55,48 +71,72 @@ ynh_package_install() {
...
+Package: ${dep_app}-ynh-deps
+Version: ${version}
+Depends: ${dependencies// /, }
+Architecture: all
+Description: Fake package for ${app} (YunoHost app) dependencies
+ This meta-package is only responsible of installing its dependencies.
+EOF
+    ynh_package_install_from_equivs ./${dep_app}-ynh-deps.control \
+        || ynh_die "Unable to install dependencies"	# Install the fake package and its dependencies
+    ynh_app_setting_set $app apt_dependencies $dependencies
```
alexAubin: So, do I understand this correctly : this helper (ynh_app_dependencies) should be called once and only once in the install script, right ? Otherwise this helper will overwrite the equivs control file and this line (ynh_app_setting_set) will overwrite the previous apt_dependencies.

Is this expected ? I ask this specifically about the Abiword/etherpad thing you were talking about. Are you preparing a list of dependencies THEN calling this helper with the entire list - or are you calling this helper two times, once with the mandatory depencies and once with the optional dependencies ?

data/helpers.d/package
maniackcrudelis: Yes, why not.

data/helpers.d/package
maniackcrudelis: So, yes, it's exactly that. But as you said, if we call this helper twice, it overwrite the previous control file.
So, I think we should add a security to prevent this case.

For etherpad, a small part of code is better than a long talk.
```
if [ $abiword -eq 1 ]
then
	ynh_app_dependencies npm nodejs-legacy abiword
else
	ynh_app_dependencies npm nodejs-legacy
fi
```

maniackcrudelis: > But as you said, if we call this helper twice, it overwrite the previous control file.

Fixed and tested.

alexAubin: Awesome ! Thanks a lot !

PR #248, opened 2017-02-08
[enh] New helpers for logrotate

maniackcrudelis:
Add and remove logrotate config

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -1,0 +1,53 @@
+# Use logrotate to manage the logfile
+#
+# usage: ynh_use_logrotate [logfile]
+# | arg: logfile - absolute path of logfile
+#
+# If no argument provided, a standard directory will be use. /var/log/${app}
+# You can provide a path with the directory only or with the logfile.
+# /parentdir/logdir/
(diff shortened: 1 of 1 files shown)


zamentur:
What about legal obligation, I think in France we should delete logs after 366 days.

alexAubin:
Imho this is beyond the scope of this PR. At the moment I don't think nginx logs (and others) olders than 1 year are deleted... But that's probably something we should address.

Psycojoker:
Up

alexAubin:
I haven't tested it, but if you guys (@maniackcrudelis and @Psycojoker) did, then let's merge it

JimboJoe:
@maniackcrudelis could you please point us to an application using that helpers?

maniackcrudelis:
Minidlna use it, but the repository is not up to date.
But, we'll don't merge this helper for the moment. We have a small error with the optional argument, that trig the previous set -u

maniackcrudelis:
Tested, it's ok.

alexAubin:
This PR looks mergeable to me. (Not tested / reviewed but I trust you guys 😉 )

PR #250, opened 2017-02-09
[enh] Cleaner postinstall logs during CA creation

alexAubin:
Right now the post-install still displays some ugly logs such as : 

```
Generating a 2048 bit RSA private key
.+++
.......+++
writing new private key to '/usr/share/yunohost/yunohost-config/ssl/yunoCA/ca/cakey.pem'
---
Updating certificates in /etc/ssl/certs... 0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d....done.
```
which pollutes the screen. With this PR and #159 (which removes the ugly wget verbose logs), we can achieve much cleaner postinstall logs.

Edit : the purpose of this PR is not to simply remove information, it is to replace it with more meaningful one.

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -265,21 +265,31 @@ def tools_postinstall(domain, password, ignore_dyndns=False):
     # Create SSL CA
     service_regen_conf(['ssl'], force=True)
     ssl_dir = '/usr/share/yunohost/yunohost-config/ssl/yunoCA'
-    command_list = [
+    commands = [
         'echo "01" > %s/serial' % ssl_dir,
         'rm %s/index.txt'       % ssl_dir,
         'touch %s/index.txt'    % ssl_dir,
         'cp %s/openssl.cnf %s/openssl.ca.cnf' % (ssl_dir, ssl_dir),
-        'sed -i "s/yunohost.org/%s/g" %s/openssl.ca.cnf ' % (domain, ssl_dir),
+        'sed -i s/yunohost.org/%s/g %s/openssl.ca.cnf ' % (domain, ssl_dir),
         'openssl req -x509 -new -config %s/openssl.ca.cnf -days 3650 -out %s/ca/cacert.pem -keyout %s/ca/cakey.pem -nodes -batch' % (ssl_dir, ssl_dir, ssl_dir),
         'cp %s/ca/cacert.pem /etc/ssl/certs/ca-yunohost_crt.pem' % ssl_dir,
         'update-ca-certificates'
     ]
 
-    for command in command_list:
-        if os.system(command) != 0:
+    for command in commands:
+        p = subprocess.Popen(
+            command.split(), stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+        out, _ = p.communicate()
+
+        if p.returncode != 0:
+            logger.warning(out)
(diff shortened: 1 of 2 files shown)


Psycojoker:
I think it's better to show what is happening to the user than to hide things because this will be "cleaner". We are here to empower our users after all and I don't see aesthetic prevailing on that.

alexAubin:
This is not just aesthetic. Dumping 2500 lines of logs does not empower anybody. In fact, important information just get lost if there are too many lines. Specifically printing stuff like
```
Generating a 2048 bit RSA private key
.+++
.......+++
```
does not empower anybody. The lambda user won't know if 2048 is bad, good or overkilled.Most won't even know what's a RSA private key.  What are the dots and + signs ? 

Lot of ugly and unecessary logs have already been moved to the logger.info stream (instead of straight to stdout) already, such as dumping the full details of CA certificate stuff used to self-sign certificates (which does not empower anybody either). I don't see why stuff like this should stay :/ ...

polytan02:
I like the logs, and I like to know that something is happening and that I need to wait, especially when it takes a long time.

alexAubin:
Bump :confused:

I kind of agree with @polytan02 that the cert generation can take 10~30 seconds. On the other hand, just printing "......+++" does not tell what's happening, confuse non-technical people, and just pollute logs.

Right now, we don't have a proper way to tell the user that an operation started and might take some time, because only logger.warning() and logger.success() print something by default.

Anyway, I don't quite understand why people are against this PR. In many other PRs / contexts, we have already cleaned some output which were not meaningful for the average user. For instance, during the Let's Encrypt PR, we removed the dump of the self-signed certificate in stdout. Because it just prints around 30 lines of technical stuff which are meaningless for the average human being. But even this was more meaningful than the ".....+++" I'm willing to 'hide'...

src/yunohost/tools.py
```diff
@@ -265,21 +265,31 @@ def tools_postinstall(domain, password, ignore_dyndns=False):
...
     service_regen_conf(['ssl'], force=True)
     ssl_dir = '/usr/share/yunohost/yunohost-config/ssl/yunoCA'
-    command_list = [
+    commands = [
         'echo "01" > %s/serial' % ssl_dir,
         'rm %s/index.txt'       % ssl_dir,
         'touch %s/index.txt'    % ssl_dir,
         'cp %s/openssl.cnf %s/openssl.ca.cnf' % (ssl_dir, ssl_dir),
-        'sed -i "s/yunohost.org/%s/g" %s/openssl.ca.cnf ' % (domain, ssl_dir),
+        'sed -i s/yunohost.org/%s/g %s/openssl.ca.cnf ' % (domain, ssl_dir),
```
opi: Why do you removed these quotes ? (and why include this change in this PR ?)

alexAubin: It's been a long time, but I think it's because of the difference commands are executed with os.system compared to subprocess. os.system() is closer to what you would really type in a real-life shell, so you can put quotes which are then forgotten, because they are only here to be able to have, for instance, arguments with space in it. 

With subprocess, you directly feed a list and therefore don't need quotes. And in fact , if you put quote, they'll stay there (as if you escaped them, e.g. typing `sed -i \"s/foo/bar/g\"`).

opi: agreed

PR #252, opened 2017-02-12
[fix] handle the case where services[service] is set to null in the s…

ZeHiro:
…ervices.yml. Fix #785

### DIFF
--- a/src/yunohost/service.py
+++ b/src/yunohost/service.py
@@ -658,6 +658,11 @@ def _update_conf_hashes(service, hashes):
                  service, hashes)
     services = _get_services()
     service_conf = services.get(service, {})
+
+    # Handle the case where services[service] is set to null in the yaml
+    if service_conf is None:
+        service_conf = {}
+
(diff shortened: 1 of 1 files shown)


ZeHiro:
The fix will work in this specific case of the detected error, but perhapse it should be better to sanitize the null/None values in a proper helper. load_and_sanitize_yaml (path_to_yaml) : return sanitized Dict

alexAubin:
Don't know much about conf-regen, but I had to tweak also get_conf_hashes to fix the Traceback stuff when trying to reproduce.

Thanks for the PR btw !

ZeHiro:
You're welcome, it was an easy fix :-)

alexAubin:
Bump ! Would be nice to have another feedback, this is relatively easy / small to review.

I tested it by editing services.yml, replacing the whole ssl block by just `ssl : null`, then running `yunohost service regen-conf --force` if I remember correctly.

PR #254, opened 2017-02-14
Add unit testing for global settings

Psycojoker:
TO BE MERGED AFTER GLOBAL SETTINGS.

After years of frustration, I've finally took the time to write tests for YunoHost \o\ /o/

I had to do an ugly hack to "enter the moulinette world". Sadly, those tests only run on a setup environment (like ynh-dev).

To launch the tests:

    apt-get install python-pytest
    cd /usr/lib/moulinette/yunohost
    py.test tests

### DIFF
--- a/src/yunohost/tests/conftest.py
+++ b/src/yunohost/tests/conftest.py
@@ -1,0 +1,16 @@
+import sys
+import moulinette
+
+sys.path.append("..")
+
+old_init = moulinette.core.Moulinette18n.__init__
+
+
+def monkey_path_i18n_init(self, package, default_locale="en"):
+    old_init(self, package, default_locale)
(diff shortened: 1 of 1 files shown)


Psycojoker:
Only consider the last commit, the others ones are from the global_settings.

Psycojoker:
Finally I've only include the base files need to write tests instead of everything from global settings.

I suggest that we merge this branch quickly as it will allow us to write tests in every other branches and that this branch doesn't modify any behavior and thus won't break anything.

To have an example of tests just look at https://github.com/YunoHost/yunohost/blob/1e7fd4251f7fa97d9d03b20a69e937ad9a45bd77/src/yunohost/tests/test_settings.py

Psycojoker:
Up

PR #257, opened 2017-02-15
[fix] Any address in the range 127.0.0.0/8 is a valid loopback address

julienmalik:


Psycojoker:
I fear that the "in the range 127.0.0.0/8" can be very confusing for people not used to this terminology. Appart from that the code looks fine for me.

opi:
@Psycojoker this message is only displayed in the log file, and not in the human admin interface (usage of `logger.debug()` ); So it will be read only by some advanced users

julienmalik:
@Psycojoker it's a debug log anyway, not really on the frontend side of things
Please rephrase if you have something better

PR #259, opened 2017-02-25
[enh] upgrade ciphers suit to more secure ones

Psycojoker:
Following this topic https://forum.yunohost.org/t/ssl-security-disable-diffie-hellman-key-exchange-per-default/2540 I'm opening this PR following his instruction for further discussion.

*Please not that I haven't tested this PR.*

This PR, if good, should be backported to stable.

*Edit by Alex:*
### Status / summary

So right now the proposal is to follow the Modern Compat' recommendation from Mozilla (cf. [this](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility) and this [this](https://mozilla.github.io/server-side-tls/ssl-config-generator/)). 

Pros : 
- Mozilla are good guys and we can trust their recommendations ;
- With modern compat, we don't need DH params anymore (we used non-custom 1024 bit DH params so far, which was bad) ;
- We probably get A+ on SSLLabs by default ;
- Modern compat' is compatible with all ~relatively recent browsers (cf [this](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility))

Cons:
- People with ~old computers/setup/browsers/smartphones (e.g. Safari < 9, Android ~ 2.x) might not have access to yunohost instances, and we have no mechanic yet to allow admin to set "softer" ciphers nginx conf (same discussion as [here](https://github.com/YunoHost/yunohost/pull/161#issuecomment-226692995)). We need to discuss to which extend this is an issue or not.

### Todo

- Propagate the changes to [yunohost_admin.conf](https://github.com/YunoHost/yunohost/blob/unstable/data/templates/nginx/plain/yunohost_admin.conf)
- Test on several different browsers ?

### DIFF
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -24,15 +24,14 @@ server {
     ssl_certificate_key /etc/yunohost/certs/{{ domain }}/key.pem;
     ssl_session_timeout 5m;
     ssl_session_cache shared:SSL:50m;
+
+    # Ciphers with modern configuration
+    # Source : https://mozilla.github.io/server-side-tls/ssl-config-generator/
+    ssl_protocols TLSv1.2;
+    ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256';
     ssl_prefer_server_ciphers on;
-    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-    ssl_ciphers 'EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH';
 
     add_header Strict-Transport-Security "max-age=31536000;";
-
-    # Uncomment the following directive after DH generation
-    # > openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048
-    #ssl_dhparam /etc/ssl/private/dh2048.pem;
 
     access_by_lua_file /usr/share/ssowat/access.lua;
 
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -26,7 +26,7 @@ server {
     ssl_session_cache shared:SSL:50m;
     ssl_prefer_server_ciphers on;
     ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-    ssl_ciphers ALL:!aNULL:!eNULL:!LOW:!EXP:!RC4:!3DES:+HIGH:+MEDIUM;
+    ssl_ciphers 'EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH';
 
     add_header Strict-Transport-Security "max-age=31536000;";
 
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -19,9 +19,12 @@ server {
     ssl_certificate_key /etc/yunohost/certs/yunohost.org/key.pem;
     ssl_session_timeout 5m;
     ssl_session_cache shared:SSL:50m;
+
+    # Ciphers with modern configuration
+    # Source : https://mozilla.github.io/server-side-tls/ssl-config-generator/
+    ssl_protocols TLSv1.2;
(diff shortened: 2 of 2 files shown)


data/templates/nginx/server.tpl.conf
```diff
@@ -26,7 +26,7 @@ server {
     ssl_session_cache shared:SSL:50m;
     ssl_prefer_server_ciphers on;
     ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-    ssl_ciphers ALL:!aNULL:!eNULL:!LOW:!EXP:!RC4:!3DES:+HIGH:+MEDIUM;
+    ssl_ciphers 'EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH';
```
taziden: I'd suggest switching to something like `ssl_ciphers ALL:!aNULL:!eNULL:!LOW:!EXP:!RC4:!3DES:!MD5:!MEDIUM:+HIGH;`
and adding `ssl_ecdh_curve secp384r1;`

julienmalik:
IMHO we should simply use https://mozilla.github.io/server-side-tls/ssl-config-generator/ as our "ssl config authority". Nice thing is that we get the browser compatibility associated with each version (old/intermediate/modern). Also I really like the idea of basing our default ssl config on peer-reviewed up-to-date recommandations.

The intermediate version seems reasonable to choose as default (to keep compatibility with Android 2.3, java 7 and IE). And when the "settings" branch is integrated we will be able to allow end-users to choose other profiles (old, modern, or maybe custom?), so that you can easily become A++++ on ssllabs if that's what you're aiming for....

julienmalik:
Also, you are only taking care of nginx here. what about postfix/dovecot/metronome ? should their cipher list be configured as nginx is an open question.

taziden:
> The intermediate version seems reasonable to choose as default
We're using a  "Modern"-like setup for Ilico.org TLS services for months/years now and so far, we didn't have any complaints from nearly a hundred users (and they are not geeks).
I see no reaso to still support Android 2.3.7, Java 7 or old versions of IE.
I'd rather choose a strong default suite and then let people choose to downgrade security of their setup according to their needs.

> what about postfix/dovecot/metronome ?

Well, I see no reason to have a different setup for dovecot/postfix/metronome. Having one set of ciphers means less work to maintain and update.

alexAubin:
Related : https://github.com/YunoHost/yunohost/pull/161

alexAubin:
I agree with using mozilla as a reference. Also I agree with @ŧaziden : the modern setup looks good enough to me.

Regarding dovecot / postfix / metronome, we should indeed have a look at the ciphers as well, but if nobody is motivated / has time to implement this in the short term, I'd say let's at least implement and merge this even if just for nginx.

alexAubin:
There are some recommendations for dovecot, postfix, openssh and others here : https://cipherli.st/

alexAubin:
So I took the liberty to commit the recommended mozilla with modern compatibility. Didn't test the PR though, but if we agree on the principle, then it should be pretty straightforward to test the configuration of a few devices.

alexAubin:
As a side note : with modern compatibility, we don't need DH parameters anymore (c.f. https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility) so that also solve https://dev.yunohost.org/issues/92 (except if we want to allow users to have intermediate compability)

alexAubin:
>Dhparams = 4096, sans supression et sans commentaire car, sinon, il y a une faille de sécurité.

???

polytan02:
De mémoire, le dhparam est grisé car si on l'activait, c'etait très, très long sur les brique olimex arm à générer. Du coup on doit l'activer manuellement. Je l'ai fait, mais qui d'autre d'autant qu'il n'y a pas de message dans l'interface d'admin.

Ceci étant dit, si les nouveaux ciphers n'utilisent pas dhparam (lien qui prouve?), alors effectivement, plus besoin de se poser la question : ça dégage ! :)

alexAubin:
>De mémoire, le dhparam est grisé car si on l'activait, c'etait très, très long sur les brique olimex arm à générer. Du coup on doit l'activer manuellement. Je l'ai fait, mais qui d'autre d'autant qu'il n'y a pas de message dans l'interface d'admin. 

Oui. c.f. discussion sur https://dev.yunohost.org/issues/92 .. J'ai un proto de job cron qui lance la commande ssl en background avec low cpu priority pour générer les DH params, mais pas eu le temps de paufiner le truc :/

> Ceci étant dit, si les nouveaux ciphers n'utilisent pas dhparam (lien qui prouve?)

https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility , "DH Parameter size: None (disabled entirely) "

polytan02:
> DHE is removed entirely because it is slow in comparison with ECDHE, and all modern clients support elliptic curve key exchanges.


OK pour moi : dh parameters n'est plus utilisé car on utilise ECDHE à la place.

Psycojoker:
Based on Aleks summary (in the first post) I'm ok with this PR based on mozilla's work. I'm in favor that we test it, merge it, then release a testing asking for help from other testers.

Psycojoker:
Oh and fix the admin_conf too.

zamentur:
What about postfix/main.cf ? I think this ciphers list is not adapted, due to old smtp server ...

alexAubin:
Yes, ideally we would also need to put some love into dovecot, postfix, ssh, metronome and maybe others.... I don't know what's the most relevant between doing only nginx here and others later, or all of them in a single PR :/

zamentur:
I Think it's better to do it service by service

Psycojoker:
Ok, let's do those in other PRs then.

PR #260, opened 2017-02-27
[enh] Show fail2ban logs on admin web interface

thardev:


Psycojoker:
Hello,

Thanks for this PR! Sadly, while this is intuitively and logically the way to do it, our regen-conf mecanism modify the `/etc/yunohost/services.yml` once in production and I fear that modifying it our debian package won't be an efficient way to modify it: it will prompt at upgrade, proably won't be applied most of the time and when it will, that might break the regen-conf.

There is a "migration framework" work that has been started and that should be able to handle those situations, sadly it isn't finished yet. I suggest that we keep this PR open for when it will be ready.

Thanks for your contribution!

opi:
@Psycojoker why are you seeing that much complicated ? just tested this services.yml changes in a VM, and after a regen-conf I can see fail2ban log.

Psycojoker:
@opi the regen-conf haven't break all your stored hash and so on? I would have expect it to do that.

opi:
@Psycojoker absolutly not; the regen-conf use the services.yml file like other configuration file in https://github.com/yunohost/yunohost/blob/unstable/data/hooks/conf_regen/01-yunohost ;

Psycojoker:
Tested, works as expected :)

Psycojoker:
Thanks!

M5oul:
Corresponding ticket [dev@#691](https://dev.yunohost.org/issues/691).

PR #261, opened 2017-03-01
Trigger exception if string key aint defined during unit tests

alexAubin:
Feels dirty, but does the trick :s. Feel free to suggest a better way to do it. (I guess the proper way would be to tweak the moulinette directly)

### DIFF
--- a/src/yunohost/tests/conftest.py
+++ b/src/yunohost/tests/conftest.py
@@ -3,14 +3,101 @@ import moulinette
 
 sys.path.append("..")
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Two people, initiating merging in agreement with https://github.com/YunoHost/project-organization#tableau-r%C3%A9capitulatif-du-nombre-davis-n%C3%A9c%C3%A9ssaire-pour-la-prise-de-d%C3%A9cision

PR #262, opened 2017-03-02
Get rid of Sender Rewriting Scheme

taziden:
As implemented today, SRS through sender_canonical_maps gives a strange behavior, rewriting Return-Path to something along senderlocalpart@mydomain which doesn't seem to make much sense.
I'd suggest to remove these two lines from the configuration, getting rid of the unwanted behavior and simplifying postfix configuration.

### DIFF
--- a/data/templates/postfix/main.cf
+++ b/data/templates/postfix/main.cf
@@ -129,10 +129,6 @@ smtpd_recipient_restrictions =
     reject_unauth_destination,
     permit
 
-# SRS
-sender_canonical_maps = regexp:/etc/postfix/sender_canonical
-sender_canonical_classes = envelope_sender
-
 # Ignore some headers
 smtp_header_checks = regexp:/etc/postfix/header_checks
 


alexAubin:
Has anybody the knowledge to review / confirm this ? I can test if I'm given the proper explanation / instructions but I have no clue if this is the right way to handle this ?

julienmalik:
I think it has been introduced here for a reason, so I would prefer to understand that reason before stripping this out.

My memories far far away tells me this was something to do with accounts with email forward or aliases.

Without these lines, what happens when a user has a mail-forward to some gmail address, and receives mail from another gmail account? Also what happens when this same user (with the forward) is part of an alias on another server ?

julienmalik:
hey @abeudin : any idea where this SRS config you introduced 3 years ago comes from :) ?

julienmalik:
According to the old deb changelog : https://github.com/YunoHost/packages_old/commit/fdc14657dc34332eb1ab449d2a1afe6ec5e9e4a7#diff-11d45831eba4b53b0bf8759384f0c26eR3

this is something linked to "mail forwarding"...

julienmalik:
maybe a start of explanation here https://www.mind-it.info/2014/02/22/forward-postfix-spf-srs/ ?

Psycojoker:
```
09:30 @<beudbeud> Bram c'est kload qu'il l'avait fait
09:31 @<beudbeud> car le SRS c'était pour quand tu fesait un foward de mail
09:31 @<beudbeud> car sinon avec le SPF ça merde
09:32  <Bram> beudbeud: donc on a plutôt intérêt à le garder non ?
09:34 @<beudbeud> normalement oui
```

I still don't understand anything about email but I hope this can help. According to beudbeud we should avoid merging this PR.

alexAubin:
Indeed : https://en.wikipedia.org/wiki/Sender_Rewriting_Scheme

« SRS was devised in order to forward email without breaking the Sender Policy Framework (SPF)»

alexAubin:
On the other hand, the original comment from @taziden was : « As implemented today, SRS through sender_canonical_maps gives a strange behavior, rewriting Return-Path to something along senderlocalpart@mydomain which doesn't seem to make much sense. »

So ... even is it an expected / wanted behavior ? Maybe there are other was to configure the SRS if this is the case.

alexAubin:
If there's no more reason to merge this / work on this, I suggest closing this PR in a few days (along with #263)

alexAubin:
Related issue, which also suggest to remove these lines : https://dev.yunohost.org/issues/335
But still doesn't sounds like the right thing to do.

alexAubin:
Other issues, which seem to better describe problems related to Sender being rewritten in strange ways in emails ... : 
- https://dev.yunohost.org/issues/269
- https://dev.yunohost.org/issues/626

alexAubin:
Closing because that doesn't look like the right way to proceed. Feel free to re-open if needed.

PR #263, opened 2017-03-02
Delete sender_canonical

taziden:
It is linked to my previous PR on SRS.


opi:
linked to #262

alexAubin:
Not sure why this is a separate PR, but since https://github.com/YunoHost/yunohost/pull/262 doesn't look like the right thing to do, deleting sender_canonical doesn't look like a good thing to do either ?

alexAubin:
Closing because that doesn't look like the right way to proceed. Feel free to re-open if needed.

PR #264, opened 2017-03-03
[mod] dnsmasq conf: remove deprecated XMPP DNS record line.

M5oul:
With [dnsmasq configured as a local dns resolver](https://yunohost.org/#/dns_resolver_en).

This file which is automatically configured for domain names on `/etc/dnsmasq.d/domain.tld` are used by DNS resolver dnsmasq.

You could check that locally, the dns record is resolved with one of following commands:
```bash
host -t SRV _jabber._tcp.${domain.tld}
dig _jabber._tcp.${domain.tld} SRV
nslookup -querytype=SRV _jabber._tcp.${domain.tld}
```

In the future, if we want to use dnsmasq as an authoritative DNS server, it's better to remove [jabber._tcp SRV record which is deprecated](https://serverfault.com/questions/582106/should-i-specify-the-srv-record-jabber-tcp).


Psycojoker:
Same than for the other PR.

M5oul:
Needs one more approval.

M5oul:
Merged as anticipated decision as there is three approval reviews.

PR #266, opened 2017-03-05
[enh] Adding new port availability checker

alexAubin:
### Problem / context
- Current `app checkport` is not very semantic : it's not specific to apps and 'check' is really vague
- Current implementation raises an error if port doesn't exists, which led to introducting #230 

### Solution
- introduce a `tools portavailable` in the actionmap, which does the same thing as `app checkport` except it simply returns a dictionnary with the form `{ "available" : "Yes" }` (or No). (Even if that's json, it's pretty easy to build a bash helper around it with a grep.)
- add a deprectation warning for `app checkport`.

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1351,6 +1351,16 @@ tools:
                     help: Show private data (domain, IP)
                     action: store_true
 
+        ### tools_portavailable()
+        portavailable:
+            action_help: Check availability of a local port
+            api: GET /tools/portavailable
+            arguments:
+                port:
+                    help: Port to check
+                    extra:
+                        pattern: *pattern_port
+
 
 #############################
 #           Hook            #
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -556,6 +556,7 @@ app:
         checkport:
             action_help: Check availability of a local port
             api: GET /tools/checkport
+            deprecated: true
             arguments:
                 port:
                     help: Port to check
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1352,8 +1352,8 @@ tools:
                     help: Show private data (domain, IP)
                     action: store_true
 
-        ### tools_portavailable()
-        portavailable:
+        ### tools_port_available()
+        port-available:
             action_help: Check availability of a local port
             api: GET /tools/portavailable
             arguments:
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -572,6 +572,6 @@ def tools_portavailable(port):
         s.connect(("localhost", int(port)))
         s.close()
     except socket.error:
-        return { "available" : "Yes" }
+        return { "available" : True }
     else:
-        return { "available" : "No" }
+        return { "available" : False }
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -572,6 +572,6 @@ def tools_port_available(port):
         s.connect(("localhost", int(port)))
         s.close()
     except socket.error:
-        return { "available" : True }
+        return True
     else:
-        return { "available" : False }
+        return False
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -31,6 +31,7 @@ import errno
 import logging
 import subprocess
 import pwd
+import socket
 from collections import OrderedDict
 
 import apt
@@ -555,3 +556,22 @@ def tools_diagnosis(auth, private=False):
         diagnosis['private']['domains'] = domain_list(auth)['domains']
 
     return diagnosis
+
+
+def tools_portavailable(port):
+    """
+    Check availability of a local port
+
+    Keyword argument:
+        port -- Port to check
+
+    """
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(1)
+        s.connect(("localhost", int(port)))
+        s.close()
+    except socket.error:
+        return { "available" : "Yes" }
+    else:
+        return { "available" : "No" }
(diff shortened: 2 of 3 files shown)


maniackcrudelis:
We've already pushed a new helper for this, at least for the apps.
And, this helper doesn't use app checkport anymore.
https://github.com/YunoHost/yunohost/pull/233

src/yunohost/tools.py
```diff
@@ -555,3 +556,22 @@ def tools_diagnosis(auth, private=False):
...
+    """
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(1)
+        s.connect(("localhost", int(port)))
+        s.close()
+    except socket.error:
+        return { "available" : "Yes" }
+    else:
+        return { "available" : "No" }
```
Psycojoker: I think that we should uses boolean values here (`True`/`False`), they are supported by json and will be easier to uses in the admin interface if needed.

Psycojoker:
I don't know if it's not beyond the scope of this PR, but it would be great if an application could register a port for itself, the same way it's done for domains. This way, the check will be cleaner and this will also avoid errors for the situation where an application needs a port but is stoped/offline/only need it from time to time.

alexAubin:
> You should put imports at file's beginning.

Unfortunately in this case (and also for registerurl in the other PR), it creates an infinite import loop. I think this has to do with the way imports are managed in Python 2 :/

alexAubin:
> I don't know if it's not beyond the scope of this PR, but it would be great if an application could register a port for itself, the same way it's done for domains. This way, the check will be cleaner and this will also avoid errors for the situation where an application needs a port but is stoped/offline/only need it from time to time.

Agreed, but I think it's really beyond the scope of this PR :/ ... It would be nice to have this in The Future:tm: though. Though I'm not sure how to handle services like nginx, postfix, mongodb, etc which 'consume' a port but it's not visible if they're disabled/inactive

Psycojoker:
> Agreed, but I think it's really beyond the scope of this PR :/ ... It would be nice to have this in The Future￼ though. Though I'm not sure how to handle services like nginx, postfix, mongodb, etc which 'consume' a port but it's not visible if they're disabled/inactive

A quick fix for the current situation is simply to include a black list of common known used ports for those services, not perfect but will already avoid some errors.

src/yunohost/tools.py
```diff
@@ -555,3 +556,22 @@ def tools_diagnosis(auth, private=False):
...
+    Check availability of a local port
+
+    Keyword argument:
+        port -- Port to check
+
+    """
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(1)
+        s.connect(("localhost", int(port)))
```
zamentur: Should we test to connect to localhost or to the main ip ?
May be adding a host optionnal argument could be an idea.

For the openvpn apps, in some case I need to test it on another ip. I am agree it's not a common case.

What about ipv6 ? And if a service is down ?

data/actionsmap/yunohost.yml
```diff
@@ -1351,6 +1351,16 @@ tools:
                     help: Show private data (domain, IP)
                     action: store_true
 
+        ### tools_portavailable()
+        portavailable:
```
zamentur: portavailable or port-available ?

src/yunohost/tools.py
alexAubin: I think the idea of this helper is mainly to check if a port is available *on the local machine*, not to test if a port is reachable from outside (which is a much more complex issue since anyway you need to ask another machine, which is what is done with ports.yunohost.org). Though it would definitely be interesting to have such a helper, imho it's not the same semantic (should be 'portreachable' or something like that). (Also : that would assume that the port is open in the firewall)

alexAubin: About ipv6 : localhost should resolve correctly even if we're in ipv6 (I assume it depends of the content of /etc/hosts, but by default there's the ipv6 equivalent of 127.0.0.1 on most setups I believe...)

And if a service is down : well... then we go back to the point brought my Bram : the clean way would be that each time a program needs a port, it should 'book' it. But that's quite a lot of work and beyond the scope of this PR which was just meant to fix a semantic issue.

data/actionsmap/yunohost.yml
alexAubin: The most widely used convention seems to imply "portavailable". (Even though I'm the one who introduced cert-install, cert-status and cert-renew :confused: )

alexAubin:
> A quick fix for the current situation is simply to include a black list of common known used ports for those services, not perfect but will already avoid some errors.

Agreed, but imho that's really beyond the scope of this PR, which is defined by the two bullet points at the beginning. We can still implement this in a dedicated PR.

src/yunohost/tools.py
```diff
@@ -555,3 +556,22 @@ def tools_diagnosis(auth, private=False):
...
+        s.close()
+    except socket.error:
+        return { "available" : True }
```
alexAubin: Same discussion as for the url available thing : we should simply return the boolean.

data/actionsmap/yunohost.yml
alexAubin: Following discussions about this (https://dev.yunohost.org/issues/872), let's rename this 'port-available' :wink:

alexAubin:
(Calling for merging approval again)

src/yunohost/tools.py
```diff
@@ -555,3 +556,22 @@ def tools_diagnosis(auth, private=False):
...
+    """
+    Check availability of a local port
+
+    Keyword argument:
+        port -- Port to check
+
+    """
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(1)
```
zamentur: I have tried without and with 0.2 , it takes the same time. (1 second)
It could be cool if it was faster. But it's ok for me like that

PR #267, opened 2017-03-06
[fix] Split checkurl into two functions : availability + booking

alexAubin:
### Status

- Ready for reviews
- <s>This PR needs https://github.com/YunoHost/moulinette/pull/136 to work properly !</s> Merged

### Related to these discussions / tickets
- https://github.com/YunoHost/yunohost/pull/235#discussion_r97322038
-  https://github.com/YunoHost/yunohost/pull/235#discussion_r103211591
- https://dev.yunohost.org/issues/446
- #185

### Problem
- `app checkurl` is not semantic because 'check' is pretty vague, and this function has sideeffects if the app parameter is given
- @maniackcrudelis wants to provide a bash helper around it but many things either can't be done or are ambiguous because of the current implemntation

### Solution
- This PR split this function in two, one is `tools urlavailable` which returns `{ "available" : "Yes" }` (or no) if domain.tld/path can be used. (We can easily build a bash helper around it) 
- The other one is `app bookurl` which set the appropriate settings for an app to be considered installed on a given url.
- `app checkurl` gets a depreciation warning

<s>This is still not fully tested. I need to setup and write some propre unit tests with some dummy app. (The `bookurl` part can only be tested during an app install :/). Writing the tests will be pretty great for future development though ;)</s> Done.

<s>This is also related to #185 because we might want to point user to 'change_url' when user are trying to use 'bookurl' on an already installed app.</s>


data/actionsmap/yunohost.yml
```diff
@@ -578,6 +578,19 @@ app:
                     full: --app
                     help: Write domain & path to app settings for further checks
 
+        ### app_bookurl()
+        bookurl:
+            action_help: Book a web path for a given app
+            api: GET /tools/bookurl
```
Psycojoker: Should probably be a POST/PUT request to create a new entry according to Rest semantic.

src/yunohost/app.py
```diff
@@ -899,6 +899,40 @@ def app_checkport(port):
...
+    url_available = tools_urlavailable(auth, url)["available"] is "Yes"
+    if not url_available:
+        raise MoulinetteError(errno.EINVAL,
+                              m18n.n('app_location_unavailable'))
+
+    app_setting(app, 'domain', value=domain)
+    app_setting(app, 'path', value=path)
+
+
 def app_checkurl(auth, url, app=None):
```
Psycojoker: Shoudln't checkurl now uses those 2 new functions?

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+
+    # Abort if domain is unknown
+    if domain not in domain_list(auth)['domains']:
+        raise MoulinetteError(errno.EINVAL, m18n.n('domain_unknown'))
+
+    # Fetch apps map
+    apps_map = app_map(raw=True)
+
+    # Loop through all apps to check if path is taken by one of them
+    available = "Yes"
```
Psycojoker: Shouldn't we uses a boolean value? This is json serializable and will be easier to uses in the admin interface.

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+    if domain in apps_map:
+        # Loop through apps
+        for p, a in apps_map[domain].items():
+            if path == p:
+                available = "No"
```
Psycojoker: We should return here since we have our answer.

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+            # We also don't want conflicts with other apps starting with
+            # same name
+            elif path.startswith(p) or p.startswith(path):
+                available = "No"
```
Psycojoker: Same regarding return.

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+
+
+    return { "available" : available }
+
+def _parse_app_url(url):
```
Psycojoker: Should probably be named something like `_normalized_url`.

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+    if url.startswith("https://"):
+        url = url[len("https://"):]
+    elif url.startswith("http://"):
+        url = url[len("http://"):]
+
+    # Remove trailing slash
+    url = url.rstrip("/")
+
+    # Check url contains at least a slash
+    if "/" not in url:
```
Psycojoker: This will fails if the url is `domain.com/` because we `rstrip` before.

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+        raise MoulinetteError(errno.EINVAL, m18n.n('invalid_url_format'))
+
+    # Now we extract the domain and uri/path
+    domain = url[:url.index('/')]
```
Psycojoker: I think that `url.split("/")[0]` would be nicer and also safer (since `string.index` will raise in the substring is not present).

Also this won't behave as expected if the domain is `domain.com/` will with `string.split` it will.

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+
+    path = url[url.index('/'):]
```
Psycojoker: Same comment than before.

However, here we will need something like:

```
if "/" in url:
    path = url.split("/", 1)[1]
else:
    path = "/"
```

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+    if not path.endswith('/'):
+        path = path + '/'
+
+    return domain, path
```
Psycojoker: A more general comment for this function is: why the hell do we merge domain+path when calling YunoHost function to split them afterward? This doesn't make any sens and make more work.

I never saw a location where we have domain+path already merged, we always merge them then split them.

I think that we should change the API and simply expect the domain and path and not their combination, this will simplify the code a lot :)

src/yunohost/tools.py
```diff
@@ -555,3 +555,63 @@ def tools_diagnosis(auth, private=False):
...
+def _parse_app_url(url):
+
+    # We want url to be of the format :
+    #  some.domain.tld/foo
+
+    # Remove http/https prefix if it's there
+    if url.startswith("https://"):
+        url = url[len("https://"):]
+    elif url.startswith("http://"):
+        url = url[len("http://"):]
```
Psycojoker: I don't know if using `url.lstrip("http://")` and `url.lstrip("https://")` wouldn't make this code better.

data/actionsmap/yunohost.yml
alexAubin: Indeed, will fix that !

src/yunohost/app.py
alexAubin: Hm, I dunno yet, I don't realize how risky it is to change it.. But maybe if the two new functions are well tested and if we keep an eye on the app CI on testing/unstable it's okay to change it

src/yunohost/tools.py
alexAubin: Yes indeed

src/yunohost/tools.py
alexAubin: Unfortunately a.lstrip(b) removes all characters from b that are at the beginning of a, so for instance "https://hello".lstrip("https://") would return "ello" :confused:

src/yunohost/tools.py
alexAubin: Hm yes indeed, will have a look into that

src/yunohost/tools.py
Psycojoker: Damn, that's stupid, I though it was treating the whole string as a single unit :(

alexAubin:
Added unit tests. You'll need to `git clone https://github.com/YunoHost/test_apps ./apps` inside the tests folder.

polytan02:
Do we use this tester in the samme way as package_linter ?

src/yunohost/app.py
alexAubin: This is a remaining item. Still not sure what to do about this 😕

alexAubin:
>Do we use this tester in the samme way as package_linter ?

Not sure I get the question. What do you mean by "the samme way as package_linter" ?

src/yunohost/app.py
Psycojoker: Not sure either 😕

alexAubin:
Maybe we should address https://dev.yunohost.org/issues/688 also with this PR and adapt the corresponding messages so they are meaningful

Psycojoker:
I really prefer that we avoid putting too much things in a PR and making it too hard to merge. See refactoring of fetchlist as a reference.

alexAubin:
>I really prefer that we avoid putting too much things in a PR and making it too hard to merge. See refactoring of fetchlist as a reference.

Okay, agreed, we can make a small PR later

src/yunohost/app.py
alexAubin: So, I'd rather avoid to change app_checkurl to be sure to avoid breaking things. Using the new functions would not be trivial because a split of the domain/path would still be needed and it's not clear how I should handle the "app" argument.

data/actionsmap/yunohost.yml
```diff
@@ -1351,6 +1368,21 @@ tools:
                     help: Show private data (domain, IP)
                     action: store_true
 
+        ### tools_urlavailable()
+        urlavailable:
```
zamentur: I am for the use of "yunohost tools url-available" . 
Dash is used everywhere but in app cmd. We should fix this and correct app cmd (there is depreciation mechanism in moulinette)

data/actionsmap/yunohost.yml
```diff
@@ -578,6 +579,22 @@ app:
                     full: --app
                     help: Write domain & path to app settings for further checks
 
+        ### app_registerurl()
+        registerurl:
```
zamentur: Is there a way to mask this kind of command from cli help ? 
This command is reserved to package script no ?

yunohost app --help is dreadfull...

src/yunohost/tools.py
```diff
@@ -555,3 +555,56 @@ def tools_diagnosis(auth, private=False):
...
+            if path == p:
+                available = False
+                break
+            # We also don't want conflicts with other apps starting with
+            # same name
+            elif path.startswith(p) or p.startswith(path):
+                available = False
+                break
+
+    return {"available": available}
```
alexAubin: We were discussing this with someone (don't remember who) and maybe simply returning the boolean should do the trick. A single boolean is a valid json and we already have a few functions returning a single value. This'd make parsing the output for helpers painless.

data/actionsmap/yunohost.yml
alexAubin: We discussed this IRL, but I don't think there's a mechanism at the moment in the Moulinette to 'hide' commands.

alexAubin:
(Calling again for merging approval)

PR #268, opened 2017-03-06
[mod] proposition of new way to handle change log

Psycojoker:
Hello,

This is an "in pratice" proposition of a new way to handle the changelog for releases.

Right now everything is done via "yunodump" that gave us a list of the commits since the last release which is cool but end up being quite tedious for the releaser since s·he has to look at every commits and try to make sense out of them which isn't always something easy to do.

Following some discussion on the dev chan, I propose a new way to attempt to do that:

* we have a copy of the `debian/changelog` file in the root of the project
* for every new modification/PR merge, we update it (if possible, in the PR, otherwise after)
* once we need to release, we simply have to copy/paste this file, update the last list (date of release) and *voilà*, we are ready to release

This PR is the changelog ready for the next testing release and also a place to discuss this new behavior proposition.

We'll probably have to update `yunobump` in accordance.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,0 +1,1515 @@
+yunohost (2.6.1) stable; urgency=low
+
+  [ thardev ]
+  * show fail2ban logs on admin web interface #260
+
+  [ ljf ]
+  * [fix] Can't use common.sh on restore operation (#246)
+
+  Thanks to all contributors: thardev, ljf and to the people who are
+  participating to the beta and giving us feedback <3
+
+
+ -- XXX <XXX@XXX>  Day, dd Mon 2017 hh:mm:ss +0100
+
+yunohost (2.6.0) testing; urgency=low
+
+  Important changes
+
+  - [enh] Add unit test mechanism (#254)
+  - [fix] Any address in the range 127.0.0.0/8 is a valid loopback address for localhost
+  - [enh] include script to reset ldap password (#217)
+  - [enh] Set main domain as hostname (#219)
+  - [enh] New bash helpers for app scripts: ynh_system_user_create, ynh_system_user_delete, helper ynh_find_port
+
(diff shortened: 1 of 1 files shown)


julienmalik:
Another way to improve the situation is to always squash PR before merging.

Jibec:
I think git do offer the ability to extract this "Full changes log".  What additionnal information does it brings here?
Who's gonna benefit this information?

in _yunohost (2.6.0)_ you use a major item/minor item but you come back with _yunohost (2.6.1)_ with the per contributor changelog. I prefer the major/minor.

Please note we also have a partial changelog in documentation: https://github.com/YunoHost/doc/blob/master/changelog_2_0.md

I'm not clear about the objective of this changelog:
* to inform about API changes,
* to communicate about what you do,
* to inform endusers about features,
* to be exhaustive,

right now, it is exhaustive.

julienmalik:
without a more automated process, this is going to be a partial changelog.
if we are going this way, then we shall have a PR test (in travis ?) which checks that for each PR, the changelog is updated. Basically I disagree with the "if possible, in the PR, otherwise after".
the copy/paste to the real debian/changelog should be automated too. and then we'll need to handle formatting issues, because the debian/changelog has some strict rules for formatting.

alexAubin:
I don't have a strong opinion about this, but I think I kind of agree with @julienmalik. Also imho we should systematically Squash before/during merging of PRs, otherwise this leads to what's described in the first post (the releaser [...] has to look at every commits and try to make sense out of them which isn't always something easy to do.).

zamentur:
It seems the adopted solution is the "squash and merge" one, so I suggest to close this PR ?

Psycojoker:
Ah? Everyone seems to continue using the classical merge from what I see.

But yeah, we can close this ticket I guess.

PR #269, opened 2017-03-06
Clean services.yml

Psycojoker:
Hello,

This is an attempt to clean `/etc/yunohost/services.yml` from outdated services.

Right we have a series of old/not used anymore/doesn't make sens services and this confused users (can be confirmed from several feedback with this internet cube, this was a very frequent request like "is it normal that those services doesn't run?").

`services.yml` is used for 2 things: saying which services are tracked with which logs files AND keeping track of the regen-conf. This collide sometime because it happens that we handle the regen-conf of stuff that *aren't services* but they end up being listed as inactive services.

This PR attempt to do 3 things to fix that:

* don't list as a service an entry that doesn't have a `status` key, by default the "service %(service name)s" was used when no `status` key was present, we ended up with services like `ssl` which didn't make any sens
* set as null the services `nsswitch` and `udisks2` because I THINK (not sure) that they aren't used anymore (I couldn't find any trace of them and they were always inactives)
* add `bind9` as a null service entry, this will remove bind9 from old YunoHost `services.yml` where it was still present before the dnsmasq migration (this concern quite a lot of Internet Cube actually)

### DIFF
--- a/data/hooks/conf_regen/01-yunohost
+++ b/data/hooks/conf_regen/01-yunohost
@@ -65,7 +65,7 @@ with open('/etc/yunohost/services.yml') as f:
 updated = False
 for service, conf in new_services.items():
     # remove service with empty conf
-    if not conf:
+    if conf is None:
         if service in services:
             print("removing '{0}' from services".format(service))
             del services[service]
--- a/data/templates/yunohost/services.yml
+++ b/data/templates/yunohost/services.yml
@@ -50,7 +50,7 @@ yunohost-firewall:
 nslcd:
    status: service
    log: /var/log/syslog
-nsswitch: null
+nsswitch: {}
 bind9: null
 tahoe-lafs: null
 memcached: null
--- a/data/templates/yunohost/services.yml
+++ b/data/templates/yunohost/services.yml
@@ -50,10 +50,8 @@ yunohost-firewall:
 nslcd:
    status: service
    log: /var/log/syslog
-nsswitch:
-   status: service
-udisks2:
-   status: service
+nsswitch: null
+udisks2: null
 amavis: null
 postgrey: null
 spamassassin: null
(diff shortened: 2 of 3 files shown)


Psycojoker:
Redmine ticket https://dev.yunohost.org/issues/647

julienmalik:
I don't understand why we need those "null" services in this file (amavis, ...) : can you give a hint ?

Psycojoker:
From what I understand (and confirmed by my testing), YunoHost uses them to clean `/etc/yunohost/services.yml` by removing them from it if they are set to `null`. If they aren't present, YunoHost will ignore them and they will stay here.

This is why I've added `bind9: null`, to remove it from old YunoHost instances (for example my Internet Cube) where it is still present.

julienmalik:
For udisk2, ok, there does not seem to be any "service udisk2" available.
For nsswitch, it is still used (https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/conf_regen/46-nsswitch), but from what I understand, there is no daemon running behind. In my `/etc/yunohost/services.yml` I have 
```
nsswitch:
  conffiles:
    /etc/nsswitch.conf: 77056212bae8d35929df0eca7f2e13e7
  status: service
```
I don't know if it's an issue with your PR, but seemed worth mentionning since you say you did not find trace of it.
Also I have some other deprecated services in my services.yml files :
* tahoe-lafs
* memcached
* udisk-glue

And seems like there are some buggy "status" definitions. The following have a correctly running daemon, but the admin interface report them as "Not Running":
* rspamd
* rmilter

Psycojoker:
rmilter and rspamd are running for me, are you on the last version? We've moved from socket activation to services.

Indeed for nsswitch, I've also encountered a strange behavior with my patch for it, I will investigate.

I'm going to add the other services you've pointed to.

Thanks!

Psycojoker:
Oups, thank for noticing :)

M5oul:
Mergeable on the 14 March with two positives reviews.
Mergeable on the 10 March with three positives reviews.

Psycojoker:
I still need to investigate the nsswitch case.

Psycojoker:
Works for nsswitch. The inline python in the regenconf shell script is...

PR #270, opened 2017-03-07
[fix] launch ssowatconf at the end of a broken install to avoid sso bad state

Psycojoker:
Attempt to fix https://dev.yunohost.org/issues/773


M5oul:
Could be merge the 10 March as an anticipated decision as there is three positives reviews.

PR #271, opened 2017-03-14
[fix] Regenerate SSOwat conf during main_domain operation. #672

opi:
After this change, `/etc/ssowat/conf.json` is updated every time the main domain is changed.

Fixes https://dev.yunohost.org/issues/672

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -185,6 +185,9 @@ def tools_maindomain(auth, new_domain=None):
         else:
             logger.info(out)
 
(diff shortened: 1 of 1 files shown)


alexAubin:
At first sight this looks good. We need to check that this does not break postinstall though (for weird reasons).

opi:
postinstall went fine on my vagrant test box

alexAubin:
Accordingly to https://github.com/YunoHost/project-organization, will merge around March 22 if no opposition

PR #273, opened 2017-03-15
Allow underscore in backup name

maniackcrudelis:
Allow underscore in backup name, to allow backup with the name of multi-instance apps.

If there are no reason to avoid this character.

The error message allow this character, so there are a error in one of them.
https://github.com/YunoHost/yunohost/blob/unstable/locales/en.json#L156

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -678,7 +678,7 @@ backup:
                     help: Name of the backup archive
                     extra:
                         pattern: &pattern_backup_archive_name
-                            - !!str ^[\w\-\.]{1,30}(?<!\.)$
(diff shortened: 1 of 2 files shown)


maniackcrudelis:
So, after some tests. With or without underscore, the backup work fine.
But, however, the error message does not tell about the limit of 30 characters!

Psycojoker:
Could you update the error message then plz?

alexAubin:
>So, after some tests. With or without underscore, the backup work fine.

You mean after applying this PR ?

maniackcrudelis:
No before, and that is strange I think. Maybe I missed my search in the code.

So, however, sorry Bram, I didn't receive any mail for your answer. And I didn't see what you said me.

PR #274, opened 2017-03-16
[enh] Helpers for common IO operations

alexAubin:
### Status

- Kinda stable now (with tests). Need feedback !
- Wondering if we should put this in the Moulinette

### Problem

Follow-up of a discussion we had two or three times already : we keep needing to read / write json files, change file permissions or downloading stuff at various places in the code. This lead to either duplicated code, or different pieces of code that don't handle exceptions consistently. The current situation also discourage to handle exceptions correctly because there are simply too many of them to check (e.g. when doing a `requests.get()`)

### Proposed solution

Introduce a `io.py` that contains low-level functions for all common operations such as : 
- read/write files
- read/write json
- download text data
- change file permissions
- execute ~raw bash commands
- ???


Psycojoker:
Haven't dig into it too much but looks like a really good start :+1:

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def _read_json(file_path):
+    """
+    Read a json file
+
+    Keyword argument:
+        file_path -- Path to the json file
+    """
+    assert isinstance(file_path, str)
```
opi: assertion is already done in _read_file()

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Keyword argument:
+        file_path -- Path of the file to remove
+    """
+    # Assumptions
+    assert isinstance(file_path, str)
+
+    if os.path.exists(file_path):
+        try:
+            os.remove(file_path)
```
opi: `os.remove` will probably raise an IOError or OSError exception that could be catched.

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Keyword argument:
+        url -- The url to download the data from
+    """
+    # Assumptions
+    assert isinstance(file_path, str)
+
+    # Download file
+    try:
+        r = requests.get(url, timeout=30)
```
opi: Timeout could be passed as an optional argument to  _download_text_data().

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+        raise MoulinetteError(errno.ECONNRESET,
+                              m18n.n('io_download_unknown_error',
+                                     url=url))
+    # Assume error if status code is not 200 (OK)
+    if r.status_code != 200:
+        raise MoulinetteError(errno.EBADE,
+                              m18n.n('io_download_bad_status_code',
+                                     url=url, code=r.status_code))
+
+    return r.text
```
opi: Maybe we could provide a wrapper around `_download_text_data()` that download, parse and return valid JSON ?

src/yunohost/io.py
Psycojoker: According to requests documentation we should **always** have a timeout, so only the value should be modifiable http://docs.python-requests.org/en/latest/user/quickstart/#timeouts

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+    assert isinstance(user, str)
+    assert isinstance(group, str)
+    assert isinstance(permissions, int)
+
+    # Check file exist
+    if not os.path.isfile(file_path):
+        raise MoulinetteError(errno.EROENT,
+                              m18n.n('io_no_such_file', file=file_path))
+
+    uid = pwd.getpwnam(user).pw_uid
```
julienmalik: what's happening when the user or group does not exists ?

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+        permissions -- Permissions to set, preferrably in octal format
+    """
+    # Assumptions
+    assert isinstance(file_path, str)
+    assert isinstance(user, str)
+    assert isinstance(group, str)
+    assert isinstance(permissions, int)
+
+    # Check file exist
+    if not os.path.isfile(file_path):
```
julienmalik: imho this function should support directories too, transparently

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program; if not, see http://www.gnu.org/licenses
+
+    yunohost_certificate.py
```
julienmalik: copypasta

src/yunohost/io.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Manage certificates, in particular Let's encrypt
```
julienmalik: copypasta

zamentur:
May be this code should be move into moulinette repository ?

alexAubin:
>May be this code should be move into moulinette repository ?

This is an interesting idea considering that this is pretty independent from yunohost

alexAubin:
I added tests for all cases I can think about and fixed / stabilized the functions accordingly. Code is ready for reviews. Feel free to suggest additional functions/helpers or different API

alexAubin:
Ok I just realized there's a "moulinette/utils/filesystem.py" which implements part of this stuff..

alexAubin:
Closing because moved to https://github.com/YunoHost/moulinette/pull/141

PR #275, opened 2017-03-17
[enh] Refactor backup management to pave the way to borg

zamentur:
I have succeded to conserve retro-compatibility :)

It implements a part of this ticket : https://dev.yunohost.org/issues/334

This commit let you define a post_backup_create hooks which can get a csv file containing a dlist of sourceTABdestination

The file is located in $1/backup.csv .

With that it is possible to connect manually a tool like borg backup.

This commit refactor the way the data are put in tar.gz. Before the backup create operation needing more freespace (twice) to be able to execute.

I think I will push an example of borg_backup post_backup_create hooks

TODO
- [x] Use a CSV to register what backup and where it should be in the BackupArchive 
- [x] Use archivemount to avoid to untar big tar.gz
- [x] Better space management (force to keep a margin space of 100Mo due to dump decompression and apt missing dependencies)
- [x] Refactoring in several methods to be able to get a macro understandable backup process
- [x] Refactoring to separate different backup method (copy, tar, borg, hook)
- [x] Apply the mount bind idea to file with system hardlink
- [x] Add a "method_backup_create" hooks to be able to run custom backup action
- [x] Fix --no-compress option
- [x] If an app backup fail, remove added path from CSV L336
- [x] Write tests files
- [x] Rebase the code with more recent code
- [x] Find a way to support restore script that try to rm inside the mount readonly point => already supported because those legacy restore script don't use set -eu 
- [x] Fix ynh_restore hook (or delete it)
- [x] Support test_backup_sys_stuff_that_does_not_exists
- [x] Set correct IO exception (in _mount function and in other places)
- [x] Add comment on RefactorManager class with a "usage:" clause
- [x] Add comment on helpers (and other part of the main code)
- [x] Try to extract info file from tar if the info json is not in the dir L1037
- [x] Check it's a good idea to restore with the new restore hooks and not with the backup one's ! https://github.com/YunoHost/yunohost/compare/23ca90d47f7577f2f3e76fef611d02641acaf9e2...dc82f40bae5f#diff-640fb4337babed213e6b0d3ef83ab433R894 -> It's a good idea, but these restore scripts should be able to restore old archive. Restore script are saved in the archice in the case it is a user restore script, in this case it's executed. So we should never remove a system restore script (but it can do anything)
- [x] Test un compress backup
- [x] Code something on system/user backup hooks failure https://github.com/YunoHost/yunohost/compare/23ca90d47f7577f2f3e76fef611d02641acaf9e2...dc82f40bae5f#diff-640fb4337babed213e6b0d3ef83ab433R404
- [x] Fix the case where fuse is not available (so archivemount fail and we need to ask the user to make an (partial?) extraction)
- [x] Split set_target in 2
- [x] Merge code between BackupManager.set_targets and RestoreManager.set_targets (if possible)
- [x] Search all FIXME and TODO to check nothing is critical
- [x] Rename "hook" in system
- [x] Check all m18n are in locale file
- [x] Implements 2 missing tests
- [x] tests on vps that the archive mount feature is ok

?
- [ ] Copy app hooks dir in backup and restore it L730 L768

Out of scope
- [ ] Use app_list in place of os.listdir
- [ ] Avoid code injection by strange naming some files
- [ ] Be able to call a specific method with --method option andd test custom backup method
- [ ] on automatic defining backup name, check this name is not already used L163
- [ ] Estimate backup size per hooks and apps (dump database load, dependencies ...) L361 L607
- [ ] Run borg create command L475
- [ ] Remove the app if we are sure it is the same app that will be restored (may be ask user ?) L705
- [ ] Don't ask user if system is installed and it is partial restore (like app) L927
- [ ] Partial restore hooks/apps dependencies (some hook need to be present in some cases) L950
- [ ] Repair mysql if broken and it is a complete restore L951


src/yunohost/backup.py
```diff
@@ -152,6 +152,21 @@ def _clean_tmp_dir(retcode=0):
...
+    try:
+        backup_csv = open(backup_csv_path, 'a')
+    except:
+        logger.error(m18n.n('backup_csv_creation_failed'))
+
+    def add_into_list(source,dest):
+        try:
+            if dest.endswith("/"):
+                dest=dest+os.path.basename(source)
+            backup_csv.write("{0}\t{1}\n".format(source,dest))
```
zamentur: I use the TAB character as separator for the CSV. I am not usre it's a good idea (what if a path contain a tab ?)

src/yunohost/backup.py
```diff
@@ -152,6 +152,21 @@ def _clean_tmp_dir(retcode=0):
         'hooks': {},
     }
 
+    # Initialize backup list
+    backup_csv_path = os.path.join(tmp_dir, 'backup.csv')
+    try:
+        backup_csv = open(backup_csv_path, 'a')
+    except:
```
zamentur: Exception type need to be specified

src/yunohost/backup.py
```diff
@@ -152,6 +152,21 @@ def _clean_tmp_dir(retcode=0):
...
+        logger.error(m18n.n('backup_csv_creation_failed'))
+
+    def add_into_list(source,dest):
+        try:
+            if dest.endswith("/"):
+                dest=dest+os.path.basename(source)
+            backup_csv.write("{0}\t{1}\n".format(source,dest))
+        except:
```
zamentur: Exception type need to be specified

src/yunohost/backup.py
JocelynDelalande: use the python csv writer from [csv module](https://docs.python.org/2/library/csv.html), instead of making it by hand.Iit will take care of required escaping and obey the thumb rule « filter your output » :-)

src/yunohost/backup.py
```diff
@@ -152,6 +152,21 @@ def _clean_tmp_dir(retcode=0):
...
+    backup_csv_path = os.path.join(tmp_dir, 'backup.csv')
+    try:
+        backup_csv = open(backup_csv_path, 'a')
+    except:
+        logger.error(m18n.n('backup_csv_creation_failed'))
+
+    def add_into_list(source,dest):
+        try:
+            if dest.endswith("/"):
+                dest=dest+os.path.basename(source)
```
JocelynDelalande: use os.path.join

src/yunohost/backup.py
```diff
@@ -152,6 +152,21 @@ def _clean_tmp_dir(retcode=0):
...
+    # Initialize backup list
+    backup_csv_path = os.path.join(tmp_dir, 'backup.csv')
+    try:
+        backup_csv = open(backup_csv_path, 'a')
+    except:
+        logger.error(m18n.n('backup_csv_creation_failed'))
+
+    def add_into_list(source,dest):
+        try:
+            if dest.endswith("/"):
```
JocelynDelalande: I think I'm lacking background to fully understand, but what do you want to test exactly ?

src/yunohost/backup.py
```diff
@@ -255,29 +274,66 @@ def _clean_tmp_dir(retcode=0):
...
     # Calculate total size
-    backup_size = int(subprocess.check_output(
-        ['du', '-sb', tmp_dir]).split()[0].decode('utf-8'))
+    backup_size = 0
+    with open(backup_csv_path, "r") as backup_csv:
+        for line in backup_csv:
+            line = line.rstrip('\n')
+            source, dest = line.split("\t", 1)
+            if source != os.path.join(tmp_dir,'info.json'):
+                backup_size += int(subprocess.check_output(
```
JocelynDelalande: We may want to use [https://docs.python.org/2/library/os.html#os.stat](os.stat) to get it a more pythonic and simple way :-)

src/yunohost/backup.py
```diff
@@ -255,29 +274,66 @@ def _clean_tmp_dir(retcode=0):
...
+                    output_directory, avail_output[1], backup_size)
+        _clean_tmp_dir(3)
+        raise MoulinetteError(errno.EIO, m18n.n(
+            'not_enough_disk_space', path=output_directory))
+
+    # Copy files (retro-compatibility)
+    if tmp_dir == output_directory:
+
+        with open(backup_csv_path, "r") as backup_csv:
+            for line in backup_csv:
```
JocelynDelalande: Same remark as before for the writer, it's better to use the csv reader from the csv module.

src/yunohost/backup.py
```diff
@@ -291,18 +347,18 @@ def _clean_tmp_dir(retcode=0):
 
         # Add files to the archive
         try:
-            tar.add(tmp_dir, arcname='')
-            tar.close()
+            with open(backup_csv_path, "r") as backup_csv:
```
JocelynDelalande: same again :-) (seems like slight refacto may be done on that part anyway)

src/yunohost/backup.py
zamentur: I have not change the way to run command, it was in the code before, I don't know why.
I think jerome had good reason for that before, may be @Psycojoker or @opi know why we use subprocess here ?

src/yunohost/backup.py
zamentur: We need a way to read this csv from bash post_backup_create hooks. May be a CSV with delimiter could be more difficult to read...

JocelynDelalande: I understand, but the code as is very fragile (a path with a tab and everything is broken: possibly overwriting by accident or attack arbitrary files). So I'd see two solutions:
- find (or write) a robust function to parse CSV in bash
- reject explicitly any argument with a separator (in our case, a tab) in it

src/yunohost/backup.py
opi: no idea :/

src/yunohost/backup.py
```diff
@@ -70,6 +70,9 @@ def backup_create(name=None, description=None, output_directory=None,
         ignore_apps -- Do not backup apps
 
     """
+
+    import csv
```
M5oul: Should be put on file beginning with other imports.

src/yunohost/backup.py
```diff
@@ -152,6 +155,24 @@ def _clean_tmp_dir(retcode=0):
...
+        field_names = ['source', 'dest']
+        backup_csv = csv.DictWriter(backup_csv, fieldnames=field_names, \
+                                    quoting=csv.QUOTE_ALL)
+    except (IOError, OSError, csv.Error):
+        logger.error(m18n.n('backup_csv_creation_failed'))
+
+    def add_into_list(source,dest):
+        try:
+            if dest.endswith("/"):
+                dest=os.path.join(dest,os.path.basename(source))
```
M5oul: `pep8` please.

zamentur:
As discuss on xmpp with opi and Maniack C, I have decided to keep the mount bind concept to be able to keep a retro compatibility with backup/restore script.

Another advantage is the archive will be more readable by a human.

So, this pr won't fix the false result of "du -sb" on /home/yunohost.backup/tmp .

zamentur:
TODO
- [x] Use a CSV to register what backup and where it should be in the BackupArchive 
- [x] Use archivemount to avoid to untar big tar.gz
- [x] Better space management (force to keep a margin space of 100Mo due to dump decompression and apt missing dependencies)
- [x] Refactoring in several methods to be able to get a macro understandable backup process
- [x] Refactoring to separate different backup method (copy, tar, borg, hook)
- [x] Apply the mount bind idea to file with system hardlink
- [x] Add a "method_backup_create" hooks to be able to run custom backup action
- [ ] Be able to call a specific method with --method option
- [ ] Fix --no-compress option
- [ ] If an app backup fail, remove added path from CSV L336
- [ ] Copy app hooks dir in backup and restore it L730 L768
- [ ] Code something on system/user backup hooks failure L267
- [ ] Set correct IO exception (in _mount function and in other places)
- [ ] Write tests files
- [ ] Check it's a good idea to restore with the new restore hooks and not with the backup one's ! L649


Out of scope
- [ ] on automatic defining backup name, check this name is not already used L163
- [ ] Estimate backup size per hooks and apps (dump database load, dependencies ...) L361 L607
- [ ] Run borg create command L475
- [ ] Remove the app if we are sure it is the same app that will be restored (may be ask user ?) L705
- [ ] Don't ask user if system is installed and it is partial restore (like app) L927
- [ ] Partial restore hooks/apps dependencies (some hook need to be present in some cases) L950
- [ ] Repair mysql if broken and it is a complete restore L951
- [ ] Try to extract info file from tar if the info json is not in the dir L1037

zamentur:
I think you can begin to review this code.
Note I plan to fix the 6 points listed above

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+        _, tmp_csv = tempfile.mkstemp(prefix='backupcsv_')
+        env_var['YNH_BACKUP_DIR'] = self.collect_dir
+        env_var['YNH_BACKUP_CSV'] = tmp_csv
+        return env_var
+
+    def _commit_path_in_csv(self, tmp_csv):
+        """ Commit collected path from system or app hooks """
+        with open(tmp_csv, 'r') as input:
+            while True:
+                data = input.read(100000)
```
alexAubin: This looks weird ? What are you trying to achieve here ? If you want to append the content of `tmp_csv` to `self.csv_file`, why not just do it line by line ? Or everything at once (i.e. something like `self.csv_file.write(input.read()))` ?

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
 
-    # Initialize backup info
-    info = {
-        'description': description or '',
-        'created_at': timestamp,
-        'apps': {},
-        'hooks': {},
-    }
+    def _open_csv(self):
+        """ Initialize backup list """
```
alexAubin: Maybe reclarify here what this CSV will correspond to (e.g. the fields)

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+        self.csv_path = os.path.join(self.collect_dir, 'backup.csv')
+        try:
+            self.csv_file = open(self.csv_path, 'a')
+            self.fieldnames = ['source', 'dest']
+            self.csv = csv.DictWriter(self.csv_file, fieldnames=self.fieldnames,
+                                      quoting=csv.QUOTE_ALL)
+        except (IOError, OSError, csv.Error):
+            logger.error(m18n.n('backup_csv_creation_failed'))
+
+    def _get_env_var(self):
```
alexAubin: If this is only for hooks, I'd call this explicitly something like `_get_env_for_backup_hooks`

src/yunohost/backup.py
```diff
@@ -165,28 +255,39 @@ def _clean_tmp_dir(retcode=0):
...
 
-    # Backup apps
-    if not ignore_apps:
+    def _collect_apps_files(self, apps=[]):
+        """ Prepare backup for each selected apps """
+
+        if apps is None:
+            return
         # Filter applications to backup
         apps_list = set(os.listdir('/etc/yunohost/apps'))
```
alexAubin: Don't know if that's in the scope, but ideally this should be changed to something like `app_list(installed=True)`

src/yunohost/backup.py
```diff
@@ -199,257 +300,374 @@ def _clean_tmp_dir(retcode=0):
...
+            subprocess.call(['install', '-Dm555', app_script, tmp_script])
+
+            # Prepare env. var. to pass to script
+            app_id, app_instance_nb = _parse_app_instance_name(
+                app_instance_name)
+            env_dict = self._get_env_var()
+            env_dict["YNH_APP_ID"] = app_id
+            env_dict["YNH_APP_INSTANCE_NAME"] = app_instance_name
+            env_dict["YNH_APP_INSTANCE_NUMBER"] = str(app_instance_nb)
+            env_dict["YNH_APP_BACKUP_DIR"] = tmp_app_bkp_dir
```
alexAubin: I don't know if that help and is relevant, but it might be cleaner/useful to have a method that does all that job of "preparing the backup environment for an app" (the env variables, but also the mkdir and the "install -Dm555" (whatever that is)), from a given app_instance_name, and returns the dictionary corresponding to env vars.

src/yunohost/backup.py
```diff
@@ -199,257 +300,374 @@ def _clean_tmp_dir(retcode=0):
...
+
+            hook_exec(tmp_script, args=[tmp_app_bkp_dir, app_instance_name],
```
alexAubin: Imho this should be the only line in the `try:` block. The other lines shouldn't be risky (except maybe the mkdir and instal -Dm555, but that should be an exception about failing to prepare the backup env, not the app backup in itself).

src/yunohost/backup.py
```diff
@@ -199,257 +300,374 @@ def _clean_tmp_dir(retcode=0):
...
+        else:
+            # Add app info
+            i = app_info(app_instance_name)
+            self.info['apps'][app_instance_name] = {
+                'version': i['version'],
+                'name': i['name'],
+                'description': i['description'],
+            }
+        finally:
+            filesystem.rm(tmp_script, force=True)
```
alexAubin: Here we could have a `_clean_app_backup_env(env_vars)` (method to write) maybe

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
 ARCHIVES_PATH = '%s/archives' % BACKUP_PATH
+MARGIN_SPACE_SIZE = 100
 
 logger = getActionLogger('yunohost.backup')
 
 
-def backup_create(name=None, description=None, output_directory=None,
-                  no_compress=False, ignore_hooks=False, hooks=[],
-                  ignore_apps=False, apps=[]):
+class Archive:
```
alexAubin: There's a class "Archive" and later a class "BackupArchive". The terminology is confusing ... isn't "Archive" also meant to handle .. backup archives too ?

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
-        'description': description or '',
-        'created_at': timestamp,
-        'apps': {},
-        'hooks': {},
-    }
+    def _open_csv(self):
+        """ Initialize backup list """
+        self.csv_path = os.path.join(self.collect_dir, 'backup.csv')
+        try:
+            self.csv_file = open(self.csv_path, 'a')
```
alexAubin: It's weird to open this file without closing it :/

But then I imagine it's related to the way csv.DictWriter works a few line later ? 

Couldn't we store the information with dicts and lists and then, once it's fully ready, export it as a CSV file instead or something ?

src/yunohost/backup.py
```diff
@@ -199,257 +300,374 @@ def _clean_tmp_dir(retcode=0):
...
             logger.debug("unable to open '%s' for writing",
                          archive_file, exc_info=1)
-            _clean_tmp_dir(2)
+            self.clean(2)
             raise MoulinetteError(errno.EIO,
                                   m18n.n('backup_archive_open_failed'))
 
         # Add files to the archive
+        def _tar_path(self, source, dest):
+            tar.add(source, arcname=dest)
```
alexAubin: I think the use of "arcname" should be explained with a comment ?

JocelynDelalande: or just being explicit like  `archive_name` :-)

JocelynDelalande:
Won't have time to review soon, sadly.

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
-            tmp_dir = output_directory
-            env_var['CAN_BIND'] = 0
-    else:
-        output_directory = ARCHIVES_PATH
-
-    # Create archives directory if it does not exists
-    if not os.path.isdir(ARCHIVES_PATH):
-        os.mkdir(ARCHIVES_PATH, 0750)
+    def __init__(self, name=None, description='', collect_dir=None):
+        self.info = {
```
zamentur: You right, historically it was in this dict but it could be a good idea to change that.

src/yunohost/backup.py
zamentur: This file is closed on L120. But I am agree to open a file in a function and close it in another is crazy.
I will implement your suggestion

src/yunohost/backup.py
zamentur: It's for hook and app_backup_script

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+    def _get_env_var(self):
+        """ Define environment variable for hooks call """
+        env_var = {}
+
+        _, tmp_csv = tempfile.mkstemp(prefix='backupcsv_')
+        env_var['YNH_BACKUP_DIR'] = self.collect_dir
+        env_var['YNH_BACKUP_CSV'] = tmp_csv
+        return env_var
+
+    def _commit_path_in_csv(self, tmp_csv):
```
zamentur: I have implemented this after some fail tests.
The goal of this function is to recopy the csv produce by a hook or a backup script of an app into the true CSV. Like that if the script fail, we don't backup the listed files.

But may be it's not the good behaviour and we should register path to backup even if the script fail...

zamentur: _mark_for_backup is used to collect and list into the CSV some files or directory the python algorythm want to add by default in the archive.

src/yunohost/backup.py
zamentur: Copy pasta from http://stackoverflow.com/questions/36875258/copying-one-files-contents-to-another-in-python
So it's in case the file is too big or if there isn't a lot of memory available. But you're right, it shouldn't be a problem because the file produce by a script should contains about 5 lines ...
So your proposition is a good idea.

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+                source = os.path.join(self.collect_dir, source)
+            if dest.endswith("/"):
+                dest = os.path.join(dest, os.path.basename(source))
+            self.csv.writerow({'source': source, 'dest': dest})
+        except csv.Error:
+            logger.error(m18n.n('backup_csv_addition_failed'))
+
+    def _collect_hooks_files(self, hooks=[]):
+        """
+        Prepare backup for each selected system part
```
zamentur: Almost. It concerns hooks in data/hooks/backup/ (system hooks) and in /etc/yunohost/hooks.d/backup/ (user hooks)

src/yunohost/backup.py
```diff
@@ -165,28 +255,39 @@ def _clean_tmp_dir(retcode=0):
...
-                        for f in i['hooks']:
-                            shutil.copy(f['path'], tmp_hooks_dir)
+        logger.info(m18n.n('backup_running_hooks'))
+
+        # Execute hooks
+        env_dict = self._get_env_var()
+        ret = hook_callback('backup', hooks_filtered, args=[self.collect_dir],
+                            env=env_dict, chdir=self.collect_dir)
+
+        if ret['succeed']:
```
zamentur: I don't know, it's an old piece of code. We should check the return of hook_calback

src/yunohost/backup.py
zamentur: I agree

src/yunohost/backup.py
zamentur: @JocelynDelalande  It can't be explicit, tar.add is define by the tarfile python library, not by yunohost.
Or may be you want to rename dest into archive_name ?

zamentur:
Need https://github.com/YunoHost/yunohost/pull/293 so i merge in it

src/yunohost/backup.py
zamentur: Done with cd1f69c47403c882ab6d571850d0d1ed1c0db67b

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+            'apps': {},
+            'hooks': {},
+        }
+        # Define backup name
+        if not name:
+            name = self._define_backup_name()
+
+        self.name = name
+        self.collect_dir = collect_dir
+        if self.collect_dir is None:
```
zamentur: Not really http://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none
collect_dir should be None or string

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+        self._init_collect_dir()
+        self._open_csv()
+
+    def collect_files(self, hooks=[], apps=[]):
+        """
+        Collect all files to backup
+        hooks: list of backup hooks to execute, if hooks is an empty list,
+        backup all hooks. If it's None, backup nothing
+
+        apps: list of apps to backup. Backup all apps, if apps is an empty list.
```
zamentur: Done with fae50ff

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
+    def _define_backup_name(self):
+        """ Define backup name """
+        # FIXME: case where this name already exist
+        return time.strftime('%Y%m%d-%H%M%S')
+
+    def _init_collect_dir(self):
+        """ Initialize preparation directory """
+
+        if not self.collect_dir:
+            self.collect_dir = "%s/tmp/%s" % (BACKUP_PATH, name)
```
zamentur: It was dead code, but there was an error on the path the good one is BACKUP_PATH/tmp/name
Fixed with 27f6a979f8ba6596b66d2f7299ed600836c15ae1

src/yunohost/backup.py
```diff
@@ -483,83 +709,286 @@ def _is_hook_in_backup(h):
...
+            hook_exec(app_script, args=[tmp_app_bkp_dir, app_instance_name],
+                      raise_on_error=True, chdir=tmp_app_bkp_dir, env=env_dict)
+        except:
+            logger.exception(m18n.n('restore_app_failed',
+                                    app=app_instance_name))
 
-            # Parse app instance name and id
-            # TODO: Use app_id to check if app is installed?
-            app_id, app_instance_nb = _parse_app_instance_name(app_instance_name)
+            app_script = os.path.join(tmp_script_dir, 'settings/scripts/remove')
```
alexAubin: tmp_script_dir is not defined in this scope :s

src/yunohost/backup.py
```diff
@@ -483,83 +709,286 @@ def _is_hook_in_backup(h):
...
+            filesystem.chmod(app_setting_path, 0400, 0400, True)
+            filesystem.chown(os.path.join(app_setting_path, 'scripts'),
+                             'admin', None, True)
+
+            # Prepare env. var. to pass to script
+            env_dict = {}
+            env_dict["YNH_APP_ID"] = app_id
+            env_dict["YNH_APP_INSTANCE_NAME"] = app_instance_name
+            env_dict["YNH_APP_INSTANCE_NUMBER"] = str(app_instance_nb)
+            env_dict["YNH_APP_BACKUP_DIR"] = tmp_app_bkp_dir
```
alexAubin: I think we need YNH_BACKUP_CSV also (ynh_restore_file complains about it missing)

src/yunohost/backup.py
```diff
@@ -483,83 +891,299 @@ def _is_hook_in_backup(h):
...
+        env_var['YNH_BACKUP_CSV'] = os.path.join(self.work_dir, "backup.csv")
+        return env_var
 
-                # Setup environment for remove script
-                env_dict_remove = {}
-                env_dict_remove["YNH_APP_ID"] = app_id
-                env_dict_remove["YNH_APP_INSTANCE_NAME"] = app_instance_name
-                env_dict_remove["YNH_APP_INSTANCE_NUMBER"] = str(app_instance_nb)
+def backup_create(name=None, description=None, output_directory=None,
+                  no_compress=False, ignore_hooks=False, hooks=[],
```
alexAubin: I'd really like if we could rename this `ignore_hooks` and `hooks` terminology into something like "system". The "hook" notion is really ambiguous and obscure for random users, and even as a dev it's confusing because "app scripts" are sometimes referred to as "hooks" (or "app hooks"). 

Also, seeing something like 

> yunohost backup create --hooks data --apps owncloud 
> yunohost backup create --hooks conf --apps owncloud

I would interpret it as "this is probably to backup only owncloud's data or owncloud's conf", whereas

> yunohost backup create --system data --apps owncloud 
> yunohost backup create --system conf --apps owncloud

is much clearer imho.

src/yunohost/backup.py
```diff
@@ -49,111 +51,257 @@
...
+            'hooks': self.hooks_return
+        }
+
+    def __repr__(self):
+        return json.dumps(self.info)
+
+    def add(self, method):
+        self.methods.append(method)
+
+    def collect_files(self, hooks=[], apps=[]):
```
alexAubin: Same remark here [about the "hook" terminology](https://github.com/YunoHost/yunohost/pull/275/files#r114081796)

src/yunohost/backup.py
```diff
@@ -199,257 +358,506 @@ def _clean_tmp_dir(retcode=0):
...
+            # We should compute size for each apps and for each hooks
+            # how to know the dependencies size ?
+
+            # Different behaviour if it's only conf hook restoration
+            data_hooks = [hook for hook in hooks if hooks.startswith('data')]
+            if apps is not None or data_hooks != []:
+                logger.warning(m18n.n('may_be_not_enough_disk_space',
+                                      free_space=free_space,
+                                      needed_space=needed_space))
+            elif free_space < CONF_MARGIN_SPACE_SIZE * 1024 * 1024:
```
alexAubin: CONF_MARGIN_SPACE_SIZE is not defined here ? Is this a typo for MARGIN_SPACE_SIZE ?

src/yunohost/backup.py
```diff
@@ -199,257 +358,506 @@ def _clean_tmp_dir(retcode=0):
...
+                raise MoulinetteError(
+                    errno.EIO, m18n.n('not_enough_disk_space',
+                                      path=BACKUP_PATH))
+
+            # TODO partial restore don't need all backup size
+            # We should compute size for each apps and for each hooks
+            # how to know the dependencies size ?
+
+            # Different behaviour if it's only conf hook restoration
+            data_hooks = [hook for hook in hooks if hooks.startswith('data')]
```
alexAubin: This line fails of "hooks" is None

src/yunohost/backup.py
```diff
@@ -199,257 +358,506 @@ def _clean_tmp_dir(retcode=0):
...
+        self.manager = restore_manager
+
+    def _check_is_enough_free_space(self):
+        """ Check free space in output directory at first """
+        backup_size = self.manager.size
+        cmd = ['df', '--block-size=1', '--output=avail', self.repo]
+        avail_output = subprocess.check_output(cmd).split()
         if len(avail_output) < 2 or int(avail_output[1]) < backup_size:
             logger.debug('not enough space at %s (free: %s / needed: %d)',
                          output_directory, avail_output[1], backup_size)
```
alexAubin: This line will fail if `len(avail_output) < 2`, because then `avail_output[1]` doesn't exists :confused: 

Also it'd be nice for tests, if we could factorize out the part where we fetch the available space (or not use subprocess with df) in a separate function. (So that we can mock the output to be 0 for instance) :D

alexAubin: (`output_directory` is not defined either :s)

src/yunohost/backup.py
alexAubin: Also I think this specific line is intended to check the case "there's more free space than the raw estimated backup size BUT not if you consider the margin". Therefore the right comparison should be `if free_space > self.info['size']` and should be placed before the `if hooks=[] and apps=[]` I think.

data/helpers.d/filesystem
```diff
@@ -1,57 +1,126 @@
...
-        eval $SUDO_CMD rm -rf "${DESTPATH}"
-    fi
+# Restore all files linked to the restore hook or to the restore app script
+#
+# usage: ynh_restore
+#
+ynh_restore () {
+    local REL_DIR="${YNH_CWD#$YNH_BACKUP_DIR/}"
+    REL_DIR="${REL_DIR%/}/"
+    sudo grep -ohP $'^\".*\",\"$REL_DIR.*\"$' ${YNH_BACKUP_CSV} | \
```
alexAubin: I guess you want to use `_get_archive_path` also here like in `ynh_restore_file` ?

zamentur: I wish, but _get_archive_path isn't currently adapted for that.

src/yunohost/backup.py
zamentur: _get_env_var() contains YNH_BACKUP_CSV

src/yunohost/backup.py
zamentur: fixed

src/yunohost/backup.py
alexAubin: Alrighty

src/yunohost/backup.py
zamentur: Done with e3e3e00

maniackcrudelis:
huh
In fact I don't know what I can say. I can't say if the helper part (not read the other part) is good or not for me.

This helper has a real lack of comments, and it's too difficult to read and understand it.
And, what is YNH_CWD ?

src/yunohost/backup.py
zamentur: Done with 25f60c9

src/yunohost/backup.py
zamentur: Done with 302a1aa

src/yunohost/backup.py
zamentur: It has been replaced by a dictionnary. Now  the csv file is created at the end of the collect process

src/yunohost/backup.py
zamentur: Renamed by
_commit_path_in_csv => _import_to_list_to_backup
_mark_for_backup => _add_to_list_to_backup

src/yunohost/backup.py
zamentur: Closed

src/yunohost/backup.py
zamentur: Done

src/yunohost/backup.py
zamentur: done with d02b71d

src/yunohost/backup.py
zamentur: Closed

src/yunohost/backup.py
```diff
@@ -49,111 +51,199 @@
...
-        description -- Short description of the backup
-        output_directory -- Output directory for the backup
-        no_compress -- Do not create an archive file
-        hooks -- List of backup hooks names to execute
-        ignore_hooks -- Do not execute backup hooks
-        apps -- List of application names to backup
-        ignore_apps -- Do not backup apps
-
+    This class prepares files to backup and apply one or several backup method
+    on it.
```
zamentur: Done

src/yunohost/backup.py
zamentur: Done
Rename by 
Archive -> BackupManager
BackupArchive -> RestoreManager

src/yunohost/backup.py
zamentur: app_list(installed=True, raw=True) should be better but the code behind app_list is very big...

src/yunohost/backup.py
zamentur: I think install -Dm555 should be an option of hook_exec.

src/yunohost/backup.py
zamentur: tmp_scripts is not in env_vars. So there is no advantages

src/yunohost/backup.py
zamentur: Done with e3e3e00

src/yunohost/backup.py
zamentur: Fixed

src/yunohost/backup.py
zamentur: fixed by 0dc2488

src/yunohost/backup.py
zamentur: I change my mind, finally for me it's ok like in 0dc2488

src/yunohost/backup.py
alexAubin: Hm well yeah, agreed. Forget about my comment. I think we should write a small helper "_list_installed_app_ids()" in app.py some day. Probably when we implement the "app status" (installed, installing, broken..) thing.

src/yunohost/backup.py
alexAubin: Agreed, I'll open a ticket about that.

src/yunohost/backup.py
zamentur: This code has been completely rewrited

zamentur:
@maniackcrudelis to know what is YNH_CWD see #293 

> This helper has a real lack of comments, and it's too difficult to read and understand it.
It happens soon :)

zamentur:
@maniackcrudelis I have added some comments on the bash helpers part

alexAubin:
Hi guys,

we now think this PR is stable and ready for reviews and comments if you have time. It ended up being a heavy refactoring / rewriting of the way the backup and restore work internally, but we really deem it to be necessary for The Future™. The diff is therefore quite big, which also comes from the fact that backup/restore are complicated and high-level features. Note that all the changes are expected to be backward-compatible, so everything should be transparent for both users and for app packagers.

We summarize below the main items that we think are important, hopefully to help you understand this PR.

### A full rework of the backup/restore architecture has been done, using classes and methods to handle and clarify the various parts of the workflow

#### Backup workflow

1. **Backup method selection** : tar, copy, custom, (and soon borg)
2. **Targets selection** : which apps and system parts to backup
3. **File collection** : running the various hooks and backup scripts for each targets, the files which are meant to be backuped are listed
4. **Free space available check** : verify that there are enough free space available to create the archive at its destination
5. **Archive creation** : put all the files collected into the archive, using the chosen backup method
6. **Cleaning**

#### Restore workflow

1. **Targets selection** : which apps and system apps to restore from the archive
2. **Free space available check** : verify that there are enough free space available to restore the targets
3. **Archive mounting** : mount the archive (e.g. mount the .tar.gz archive using archivemount/fuse instead of untarring it) to make the files available
4. **Restore of each target** : use the restore script / restore hook for each target to restore files from the mounted archive
5. **Cleaning / archive unmount**

### In parallel (or because) of this refactoring, this PR brings the following improvements

#### **Minimize space overhead during backup**

- During backup, there's no need anymore to copy the files to a temporary location before creating the archive. The files meant to be backuped are listed, then added to the archive directly. This change is transparent for the app packager as long as they used `ynh_backup` in their backup script. (But `cp` still works anyway)
- The "list files to be backuped" mechanic is actually needed to later add Borg, since you cannot cp files directly in a backup archive. Instead, you must feed it with a list of files to backup.

#### **Minimize space overhead during restore**

During restore, the archive is not uncompressed, and therefore the space needed for restoration is minimal. Also the "free space available" check is now made per target. So you can restore a single small app from a huge archive containing a 10 Go owncloud, even if you don't have 10 Go available.

#### **Backup/restore app helpers simplification**

- The destination of the backup can be automatically determined in `ynh_backup`. So you don't need to `ynh_backup /etc/foo ${backup_dir}/conf/foo` anymore, you can simply `ynh_backup /etc/foo`
- The PR introduces `ynh_restore_file` and `ynh_restore`, which allows you to restore a single file, or all files backuped with `ynh_backup` :
    - `ynh_restore_file /etc/foo` will restore the file backuped with `ynh_backup /etc/foo`
    - `ynh_restore_file /etc/foo` will restore the file backuped with `ynh_backup /etc/foo conf/foo`
    - `ynh_restore` will restore all files backuped by the app

#### **Backup method modularity**

Backup methods are now implemented in separate classes, which allow us to easily extend the backup methods available. In particular, this will allow us to easily implement Borg as just another backup method. In the future, we plan to have a --method option in the CLI.

#### **Functionnal tests**

Using the new test mechanism, we implemented several (20+ !) tests of backup/restore scenarios, including some typical failure cases, and restore of archive from 2.4 to ensure backward compatibility.

#### **Support missing archive info.json** 

'yunohost backup info' is now able to extract the Info.jsonfrom the archive to give information. It allows to restore tar.gz archive without the associated info.json

opi:
After a `git checkout -b enh-334-ynh-backup-helpers-change origin/enh-334-ynh-backup-helpers-change`, I got the following errors : https://paste.yunohost.org/pujeyificu.vbs
( note that I can't `git rebase unstable` because of conflict )

alexAubin:
Eh yeah, I don't know how ljf developped, but I had to include this small bit in hook.py : https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/hook.py#L329-L332

opi:
crap. You can't expect people to test this PR if it's not straight forward :/

opi:
It works very mieux with this `ỲNH_CWD``https://paste.yunohost.org/gemasagiqe.sm \o/

maniackcrudelis:
OK, so I've read the helpers, ynh_backup, ynh_restore and ynh_restore_file. And all seems good for me.

I guess you've already test somes cases.
So for me, the next part is to check all apps via our CI.

For this part of the PR, I agree :)

alexAubin:
Thanks for the feedback guys !

zamentur:
@opi @Aleks: I think the change you describe is in another PR already merged https://github.com/YunoHost/yunohost/pull/305
May be we should rebase the pr or merge 305 into it.

zamentur:
[ ] can't call "source ../settings/scripts/_common.sh" in app backup script
[ ] ynh_install_app_dependencies is not compatible with readonly mount

alexAubin:
Merging soon™️

NicolasPetton:
> I think I will push an example of borg_backup post_backup_create hooks

I'd like to see such examples 😄

PR #276, opened 2017-03-19
Fix for missing YunoHost tiles

JimboJoe:
This PR:
- brings back [this PR](https://github.com/YunoHost/yunohost-config-nginx/pull/3/files) which seems to have stayed behind when migrating from `yunohost-config-nginx`. It fixes the missing tile for Jirafeau app (and maybe others).
- adds a fix to serve `ynhpanel.*` files whatever the app ningx rules are (some apps nginx conf files block them, like duniter or nextcloud).

This work is linked to [this issue](https://dev.yunohost.org/issues/285), as well as [this PR](https://github.com/YunoHost/SSOwat/pull/71) on SSOwat.

### DIFF
--- a/data/templates/nginx/plain/yunohost_panel.conf.inc
+++ b/data/templates/nginx/plain/yunohost_panel.conf.inc
@@ -1,2 +1,8 @@
+# Insert YunoHost panel
 sub_filter </head> '<script type="text/javascript" src="/ynhpanel.js"></script></head>';
 sub_filter_once on;
+# Apply to other mime types than text/html
+sub_filter_types application/xhtml+xml;
+# Prevent YunoHost panel files from being blocked by specific app rules
+location ~ ynhpanel\.(js|json|css) {
+}


alexAubin:
Wow, that sounds great !

Do you know if this will be propagated automatically on existing instance ? (i.e. is yunohost_panel.conf.inc managed by the regen-conf ?)

JimboJoe:
@alexAubin unfortunately I have no idea.

alexAubin:
Alright :P I'll see if how it goes when applying the PR on my dev instance. This question is also related to the cypher PR

M5oul:
Tested. Works with Jirafeau.

opi:
files in `data/templates/nginx/plain` are overridden every time regen-conf is applied (=  during Debian package update). See https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/conf_regen/15-nginx#L23-L24

PR #277, opened 2017-03-21
Exclude media mount bind

maniackcrudelis:


M5oul:
Could you explain behind the motivation of this modification?

maniackcrudelis:
In each user home folder, there is a ~~mount bind~~ symbolic link to /media.
But, usually, /media is used for mount other hard disks. In this case, the backup process will try to backup all other hard disk with each user home folder.

But... I tried just now `find /home/* -type d -prune | awk -F/ '{print $NF}'` and this command list only the first level directory in /home.
So, my proposal si not functionnal...

alexAubin:
After discussing on the chat about this with Maniack, we decided to close this for now. Need to think about the proper fix.

PR #278, opened 2017-03-22
[enh] ip6.yunohost is now served through HTTPS.

opi:


opi:
It's been a while since https://letsencrypt.org/2016/07/26/full-ipv6-support.html

PR #280, opened 2017-03-24
New helper ynh_replace_string

maniackcrudelis:
Sed abstraction, and a way to use something other than sed.
I added also an escape of the delimiter used by sed, to prevent a fail in this case.

### DIFF
--- a/data/helpers.d/string
+++ b/data/helpers.d/string
@@ -9,3 +9,18 @@ ynh_string_random() {
       | tr -c -d 'A-Za-z0-9' \
       | sed -n 's/\(.\{'"${1:-24}"'\}\).*/\1/p'
 }
+
+# Substitute a string by another in a file
+#
+# usage: ynh_substitute_char string_to_find replace_string file_to_analyse
+# | arg: string_to_find - String to replace in the file
+# | arg: replace_string - New string that will replace
+# | arg: file_to_analyse - File where the string will be replaced.
+ynh_substitute_char () {
+	delimit=@
+	match_char=${1//${delimit}/"\\${delimit}"}	# Escape the delimiter if it's in the string.
+	replace_char=${2//${delimit}/"\\${delimit}"}
+	workfile=$3
+
+	sudo sed --in-place "s${delimit}${match_char}${delimit}${replace_char}${delimit}g" "$workfile"
+}
--- a/data/helpers.d/string
+++ b/data/helpers.d/string
@@ -10,13 +10,13 @@ ynh_string_random() {
       | sed -n 's/\(.\{'"${1:-24}"'\}\).*/\1/p'
 }
 
-# Substitute a string by another in a file
+# Substitute/replace a string by another in a file
 #
-# usage: ynh_substitute_char string_to_find replace_string file_to_analyse
-# | arg: string_to_find - String to replace in the file
-# | arg: replace_string - New string that will replace
-# | arg: file_to_analyse - File where the string will be replaced.
-ynh_substitute_char () {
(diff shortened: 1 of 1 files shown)


zamentur:
As explained personnaly I prefer a template method, which is compatible with all variable defined by the user
To see the difference, here is a use case of the ynh_substitute_char helper :
https://github.com/YunoHost/yunohost/pull/285/files#diff-7a709fc6cf57996dbbb12ffbe3d80b19R66

And here the same with a ynh_configure helper (template style)
https://github.com/YunoHost-Apps/limesurvey_ynh/blob/master/scripts/_common.sh#L382
https://github.com/YunoHost-Apps/limesurvey_ynh/blob/master/scripts/_common.sh#L373

I think templates are better because we can defined var outside by putting them into environment variables of the script with the "export" bash instruction.

Example of use:
ynh_configure CONF_FILE  DESTINATION
```
export path_url="/"
export domain="example.com"
export port="3128"
export app="$YNH_APP_NAME"
export final_path="/var/www/$YNH_APP_ID"
ynh_configure nginx.conf /etc/nginx/conf.d/$domain.d/$app.conf
```

maniackcrudelis:
Don't do too much abstraction ! ™

Yes, jinja2 is more secure than sed, though in this helper I've fixed the problem with the delimiter.

Use something other than sed is possibely a good idea.
But use one template for a whole config file will break our flexibility ™, and even more if you want to use $path_url (or something like that) in different place in your config file, but under condition.

And that break the readability of the code, because you can't see which variables will be changed when. (™)

zamentur:
May be we could let the app developper the choice between j2 method or substitute method by testing if the file is a j2 or a "classic" file ?

If helpers built on this ynh_subsitute_char don't offer the possibility to declare variables to replace outside of the helper, I think we will be unable to use is in some cases.

alexAubin:
My two cents on this, but I think the cost of adding this new helper (which I suggest with 4507b59 to be renamed as ynh_replace_string) is relatively low and we already gain in readability / semantic in a lot of places.

Migrating all apps to jinja or whatever kind of templates will take much more time than simply replacing sed calls in existing install scripts with this helper. Actually I'm not sure to understand right now what the template method really brings. Except, ok, a simple and unified way to manage configuration files, but I doubt that this is the biggest problem at the moment :confused: (compared to, for instance, properly managing nodejs dependencies :D)

maniackcrudelis:
> But I disagree with the limited usage in other ynh_add_XXX_config.

For all of our ynh_add_XXX_config helpers, I think we should keep helpers that just create a default config. But our default config.

If some packagers need to add another specific configs in this files, they shall do that in their scripts.
Because, when we read their code, we should read that they use a standart config and some other specifics adding. And for them, we need to keep a clear sight on what they changed.

data/helpers.d/string
```diff
@@ -9,3 +9,18 @@ ynh_string_random() {
...
+
+# Substitute/replace a string by another in a file
+#
+# usage: ynh_replace_string string_to_find replace_string target_file
+# | arg: string_to_find - String to be searched and replaced in the file
+# | arg: replace_string - String that will replace matches
+# | arg: target_file - File in which the string will be replaced.
+ynh_replace_string () {
+	delimit=@
+	match_char=${1//${delimit}/"\\${delimit}"}	# Escape the delimiter if it's in the string.
```
JimboJoe: to be fully consistent with the function renaming (with which I agree), you could rename these two variables as well.

alexAubin: Indeed, done :wink:

PR #281, opened 2017-03-24
New helper ynh_secure_remove

maniackcrudelis:
A secure way to remove a file or directory.
Prevent to knew issues.

Tested with this paths:
- / -> Not removed
- /var -> Not removed
- /var/www -> Not removed
- /var/www/file -> Removed
- /opt -> Not removed
- /opt/file -> Removed
- /home/yunohost.app -> Not removed
- /home -> Not removed
- /home/ -> Not removed
- // -> Not removed
- /etc/cron.d/ -> Not removed
- /etc -> Not removed
- /etc/ -> Not removed
- /etc/X11 -> Removed
- /etc/X11/$var -> Removed (if $var is not empty)

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
     chmod 755 $TMP_DIR
     echo $TMP_DIR
 }
+
+# Remove a file or a directory securely
+#
+# usage: ynh_secure_remove path_to_remove
+# | arg: path_to_remove - File or directory to remove
+ynh_secure_remove () {
+	path_to_remove=$1
+	forbidden_path=" \
+	/var/www \
+	/home/yunohost.app"
+
+	if [[ "$forbidden_path" =~ "$path_to_remove" \
+		# Match all path or subpath in $forbidden_path
+		|| "$path_to_remove" =~ ^/[[:alnum:]]+$ \
+		# Match all first level path from / (Like /var, /root, etc...)
+		|| "${path_to_remove:${#path_to_remove}-1}" = "/" ]]
+		# Match if the path finish by /. Because it's seems there is an empty variable
+	then
+		echo "Avoid deleting of $path_to_remove." >&2
+	else
+		if [ -e "$path_to_remove" ]
+		then
+			sudo rm -R "$path_to_remove"
+		else
+			echo "$path_to_remove doesn't deleted because it's not exist." >&2
+		fi
+	fi
+}
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -92,19 +92,19 @@ ynh_secure_remove () {
 	/home/yunohost.app"
 
 	if [[ "$forbidden_path" =~ "$path_to_remove" \
-		# Match all path or subpath in $forbidden_path
+		# Match all paths or subpaths in $forbidden_path
 		|| "$path_to_remove" =~ ^/[[:alnum:]]+$ \
-		# Match all first level path from / (Like /var, /root, etc...)
+		# Match all first level paths from / (Like /var, /root, etc...)
 		|| "${path_to_remove:${#path_to_remove}-1}" = "/" ]]
-		# Match if the path finish by /. Because it's seems there is an empty variable
+		# Match if the path finishes by /. Because it seems there is an empty variable
 	then
-		echo "Avoid deleting of $path_to_remove." >&2
+		echo "Avoid deleting $path_to_remove." >&2
 	else
 		if [ -e "$path_to_remove" ]
 		then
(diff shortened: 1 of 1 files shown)


data/helpers.d/filesystem
```diff
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
...
+# usage: ynh_secure_remove path_to_remove
+# | arg: path_to_remove - File or directory to remove
+ynh_secure_remove () {
+	path_to_remove=$1
+	forbidden_path=" \
+	/var/www \
+	/home/yunohost.app"
+
+	if [[ "$forbidden_path" =~ "$path_to_remove" \
+		# Match all path or subpath in $forbidden_path
```
JimboJoe: path**s** or subpath**s**

data/helpers.d/filesystem
```diff
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
...
+		|| "$path_to_remove" =~ ^/[[:alnum:]]+$ \
+		# Match all first level path from / (Like /var, /root, etc...)
```
JimboJoe: path**s**

data/helpers.d/filesystem
```diff
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
...
+		|| "${path_to_remove:${#path_to_remove}-1}" = "/" ]]
+		# Match if the path finish by /. Because it's seems there is an empty variable
```
JimboJoe: Match if the path finish**es** by /. because it ~~'s~~ seems there is an empty variable

data/helpers.d/filesystem
```diff
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
...
+	then
+		echo "Avoid deleting of $path_to_remove." >&2
```
JimboJoe: Avoid deleting ~~of~~ $path_to_remove
or: avoid deleti**on** of $path_to_remove

data/helpers.d/filesystem
```diff
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
...
+	else
+		if [ -e "$path_to_remove" ]
+		then
+			sudo rm -R "$path_to_remove"
+		else
+			echo "$path_to_remove doesn't deleted because it's not exist." >&2
```
JimboJoe: **wasn't** deleted because it **doesn't** exist

maniackcrudelis:
JimboJoe, commit, especially for that kind of corrections.

JimboJoe:
As I wrote, I would have committed changes if I had commit rights... not only organizationally, but technically :-)

maniackcrudelis:
Oh yes, sorry, you said that.
I think you should ask to someone (not me) to have this rights.

So, I make this commit ;)

data/helpers.d/filesystem
```diff
@@ -80,3 +80,31 @@ properly with chmod/chown." >&2
     chmod 755 $TMP_DIR
     echo $TMP_DIR
 }
+
+# Remove a file or a directory securely
+#
+# usage: ynh_secure_remove path_to_remove
+# | arg: path_to_remove - File or directory to remove
+ynh_secure_remove () {
+	path_to_remove=$1
```
alexAubin: Following the discussion we had a few weeks ago in the meeting, I think we should add a check here to refuse to do anything if we have a second argument

maniackcrudelis: I'm sorry but I don't any remains about what we had said about that during our previous meeting. Can you extend your idea?

alexAubin: We were discussing the usage of wildcards (`*`) with ynh_secure_remove, for instance : `ynh_secure_remove /var/www/$OOPS/*` with `$OOPS` and empty string. This will be expanded into `ynh_secure_remove /var/www/app1 /var/www/app2 /var/www/app3` which for now is a legit request from the point of view of `ynh_secure_remove`.

So the proposal was to throw an error if $2 is not empty (we cannot check that the helpers was called with a * in the path or not :s), and also to check in the App linter if `*` is used on the same line than ynh_secure_remove. Does that sounds reasonable ? We could also merge the PR as it is right now, and discuss this particular point in a separate thread/PR if you guys prefer.

maniackcrudelis: I think, at a time or another we will have some critics about this limitation. Because, someone will want to remove several directories at one time. Or even the content of a directory with a regex.
But, however, it can be a real problem if a wildcard it badly uses.

So, I think using a such limitation can be the better way to keep the point of this helper.

polytan02: I am ok with forbidding * and it is fine to me to use the helper once for each file (or folder) to remove.
And we can still do a "for" loop if required to compact code a bit.

PR #282, opened 2017-03-24
New helper ynh_setup_source

maniackcrudelis:
THE trollest helper :D

An helper to handle download, checksum, copy, patches, extra files and even a way to install without internet connection.
Really usefull to simplify this part of the packaging.


maniackcrudelis:
Need https://github.com/YunoHost/yunohost/pull/281

data/helpers.d/utils
```diff
@@ -24,3 +24,86 @@ ynh_get_plain_key() {
...
+            && for p in ../sources/patches/${source_id}-*.patch; do \
+                sudo patch -p1 < $p; done) \
+            || ynh_die "Unable to apply patches"
+    fi
+
+    # Add supplementary files
+    if test -e "../sources/extra_files"; then
+        sudo cp -a ../sources/extra_files/. "$dest_dir"
+    fi
+
```
zamentur: I suggest to add a mechanism to restore data and persistent module. Like that this helper could be used in upgrade script too.
https://github.com/YunoHost-Apps/limesurvey_ynh/blob/master/scripts/_common.sh#L189

data/helpers.d/utils
```diff
@@ -29,55 +29,81 @@ ynh_get_plain_key() {
...
+    local src_id=${2:-app} # If the argument is not given, source_id equal "app"
+	# Default value
+	local src_format="tar.gz"
+    local src_sumprg="sha256sum"
+    local src_filename="${src_id}.${src_format}"
+    local src_in_subdir=true #false if source are directly in the archive root   
+
+    # Load value from configuration file (see above for a small doc about this file
+    # format)
+    source ../conf/${src_id}.src
```
maniackcrudelis: As we discussed yesterday, I disagree with this source of the variables !
That's really not clear which variables are load from this file.

And with that, I can load whatever from this file !
Type src_url= sudo rm -r --no-preserve-root / in your source.file and the helper will execute this command without any intervention. And even if the variable will just echoed in the helper !
Just leave a space in a value, and your whole line doesn't work. And the right part will be executed by bash. (A space implies a command to execute)

Maybe we can modify the line with something else than cat and grep. But we should keep a clear load of each variable.
Just keep this code readable and secure !

data/helpers.d/utils
```diff
@@ -29,55 +29,81 @@ ynh_get_plain_key() {
...
+    src_format=$(echo "$src_format" | tr '[:upper:]' '[:lower:]')
+    if [ src_filename = "" ] ; then
```
maniackcrudelis: That will failed on a set -u.
And this variable is already set at line 48

data/helpers.d/utils
```diff
@@ -29,55 +29,81 @@ ynh_get_plain_key() {
...
 	else	# If not, download the source
-		wget -nv -O source.$arch_format $src_url
+		wget -nv -O $src_filename $src_url
     fi
 
 	# Check the control sum
-	echo "$src_checksum source.$arch_format" \
-		| md5sum -c --status || ynh_die "Corrupt source"
+	echo "$src_checksum $src_filename" \
+		| ${src_sumprg} -c --status || ynh_die "Corrupt source"
```
maniackcrudelis: Here again, it's really not clear than $src_sumprg can be overwritten by the source file !

data/helpers.d/utils
```diff
@@ -29,55 +29,81 @@ ynh_get_plain_key() {
...
+# usage: ynh_setup_source dest_dir [source_id]
+# | arg: dest_dir  - Directory where to setup sources
 # | arg: source_id - Name of the app, if the package contains more than one app
 ynh_setup_source () {
-	source_id=${1:-app} # If the argument is not given, source_id equal "app"
-	src_url=$(cat ../conf/${source_id}.src | grep SOURCE_URL | cut -d= -f2-)
-	src_checksum=$(cat ../conf/${source_id}.src | grep SOURCE_SUM | cut -d= -f2-)
-	arch_format=$(cat ../conf/${source_id}.src | grep ARCH_FORMAT | cut -d= -f2-)
-	local_source="/opt/yunohost-apps-src/$YNH_APP_ID/source.$arch_format"
+	local dest_dir=$1
```
maniackcrudelis: dest_dir should be set by default at $final_path.

data/helpers.d/utils
maniackcrudelis: By default, all extra files will not affected by a uncompress in the directory.
If you have some extra files, like wp-content for wordpress, you can simply apply the upgrade by a uncompress or a copy. Your files files will not affected.

So is it really necessary?

zamentur: I think about modules or data the user has added after a standart setup

zamentur: > If you have some extra files, like wp-content for wordpress, you can simply apply the upgrade by a uncompress or a copy. Your files files will not affected.
Manual upgrade of several apps advice to clean the "final_path" before to unpack the tar, and next copy persistent data (modules etc...). It's cleaner, some files may need to be deleted.
https://docs.nextcloud.com/server/11/admin_manual/maintenance/manual_upgrade.html

zamentur:
- [ ] Find a way to read the .src file quickly and securely
- [ ] Add a ynh_restore persistent feature (for upgrade script usage)

data/helpers.d/utils
zamentur: ok we need another solutions that support optionnal params.

data/helpers.d/utils
```diff
@@ -24,3 +24,86 @@ ynh_get_plain_key() {
...
+# src_sum="Control sum"
+# src_format="tar.gz" # (Optional) default value: tar.gz
+# src_in_subdir=false # (Optional) Put false if source are directly in the archive root
+# src_filename="example.tar.gz" (Optionnal) default value: ${src_id}.${src_format}
+#
+#
+# usage: ynh_setup_source dest_dir [source_id]
+# | arg: dest_dir  - Directory where to setup sources
+# | arg: source_id - Name of the app, if the package contains more than one app
+ynh_setup_source () {
```
alexAubin: Sorry to be pushy about helper names, but I'd really like the names to be intuitive about what exactly it is they're doing. For this one I'd suggest something like `ynh_download_and_unpack_sources()`

maniackcrudelis: It's not really just a download and an unpack.
This helper, as it is now, unpack directly in the directory.
So it's more like a whole setup for a webapp.

data/helpers.d/utils
maniackcrudelis: Ok, yes I get it.

But, I really prefer that this 2 helpers was call in the upgrade script and not in this helper.
So like that, for the pakage that need this, we can type simply

```
ynh_save_persistent data "data1" "data2"
ynh_setup_source
ynh_restore_persistent data
```

That's stayed clear, and more explicit.

data/helpers.d/utils
maniackcrudelis: With my original proposition, we can use optional parameters.

Simply like that:
```
src_url=$(grep SOURCE_URL "../conf/${source_id}.src" | cut -d= -f2-)
src_url=${src_url:-default_value}
```

data/helpers.d/utils
zamentur: ok

data/helpers.d/utils
zamentur: We need more documentation about what this helpers can do

data/helpers.d/utils
```diff
@@ -96,15 +97,16 @@ ynh_setup_source () {
...
         if $src_in_subdir ; then
-            local tmp_dir=$(mktemp -d)
-            unzip -quo $src_filename -d "$tmp_dir"
+            local tmp_dir=$(sudo mktemp -d)
+            sudo unzip -quo $src_filename -d "$tmp_dir"
             sudo cp -a $tmp_dir/*/. "$dest_dir"
             ynh_secure_remove "$tmp_dir"
         else
-            unzip -quo $src_filename -d "$dest_dir"
+            sudo nzip -quo $src_filename -d "$dest_dir"
```
M5oul: typo

zamentur:
PR needed: https://github.com/YunoHost/yunohost/pull/293

data/helpers.d/utils
```diff
@@ -24,3 +24,115 @@ ynh_get_plain_key() {
...
+    local src_sum=$(grep 'SOURCE_SUM=' "$YNH_EXECUTION_DIR/../conf/${src_id}.src" | cut -d= -f2-)
+    local src_sumprg=$(grep 'SOURCE_SUM_PRG=' "$YNH_EXECUTION_DIR/../conf/${src_id}.src" | cut -d= -f2-)
+    local src_format=$(grep 'SOURCE_FORMAT=' "$YNH_EXECUTION_DIR/../conf/${src_id}.src" | cut -d= -f2-)
+    local src_in_subdir=$(grep 'SOURCE_IN_SUBDIR=' "$YNH_EXECUTION_DIR/../conf/${src_id}.src" | cut -d= -f2-)
+    local src_filename=$(grep 'SOURCE_FILENAME=' "$YNH_EXECUTION_DIR/../conf/${src_id}.src" | cut -d= -f2-)
+    
+    # Default value
+    src_sumprg=${src_sumprg:-sha256sum}
+    src_in_subdir=${src_in_subdir:-true}
+    src_format=$(echo "$src_format" | tr '[:upper:]' '[:lower:]')
```
maniackcrudelis: Missing a default value, as you said in the help before.

data/helpers.d/utils
```diff
@@ -24,3 +24,115 @@ ynh_get_plain_key() {
...
+    fi
+
+    # Apply patches
+    if (( $(find $YNH_EXECUTION_DIR/../sources/patches/ -type f -name "${src_id}-*.patch" 2> /dev/null | wc -l) > "0" )); then
+        local old_dir=$(pwd)
+        (cd "$dest_dir" \
+            && for p in $YNH_EXECUTION_DIR/../sources/patches/${src_id}-*.patch; do \
+                patch -p1 < $p; done) \
+            || ynh_die "Unable to apply patches"
+        cd $old_dir
```
maniackcrudelis: Need to be tested, but because you use a sub shell with the parenthesis. I think it isn't needed to put a cd after this.

maniackcrudelis:
As it is now, it's ok for me.

alexAubin:
Should be mergeable once #293 and #281 are merged.

alexAubin:
#281 got merged

data/helpers.d/utils
JimboJoe: I think you can avoid using `cd` and backup/restore the current path by using `patch -d` option.

zamentur:
YNH_EXECUTION_DIR has been renamed as YNH_CWD.

There is a PR for that(maybe already merged).

alexAubin:
>There is a PR for that(maybe already merged).

Yup, it got merged

JimboJoe:
You've introduced  this in https://github.com/YunoHost/yunohost/commit/afe28470c0c7776afe4824c9ec81abd6eab52d40.
So if I get it right, it should be in next testing, and this helper has to change on this point, doesn't it?

PR #283, opened 2017-03-24
New helper ynh_restore_file

maniackcrudelis:
A way to simplify the restoring of the files and directories.
And a little abstraction to use something other than cp.

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -52,6 +52,9 @@ ynh_backup() {
     # ... or just copy the directory
     eval $SUDO_CMD mkdir -p $(dirname "${DESTPATH}")
     eval $SUDO_CMD cp -a "${SRCPATH}" "${DESTPATH}"
+
+    # Store the source path
+    echo "${DESTPATH} ${SRCPATH}" | tee -a backup_list > /dev/null
 }
 
 # Deprecated helper since it's a dangerous one!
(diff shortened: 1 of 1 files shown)


data/helpers.d/filesystem
```diff
@@ -80,3 +83,23 @@ properly with chmod/chown." >&2
...
     echo $TMP_DIR
 }
+
+# Restore a file or a directory to the stored path
+#
+# Use the registered path in backup_list by ynh_backup to restore the file at the good place.
+#
+# usage: ynh_restore_file file_to_restore
+# | arg: file_to_restore - File or directory to restore
+ynh_restore_file () {
```
zamentur: I am ok for adding this hook but I think we should use a "ynh_restore" that restore all backuped files directly .

ynh_restore_file should be reserved for particular cases.

data/helpers.d/filesystem
```diff
@@ -52,6 +52,9 @@ ynh_backup() {
     # ... or just copy the directory
     eval $SUDO_CMD mkdir -p $(dirname "${DESTPATH}")
     eval $SUDO_CMD cp -a "${SRCPATH}" "${DESTPATH}"
+
+    # Store the source path
+    echo "${DESTPATH} ${SRCPATH}" | tee -a backup_list > /dev/null
```
maniackcrudelis: Ok, so if you work on a new backup system. I think this helper should be follow your work.

How can we use this list with your PR?

data/helpers.d/filesystem
maniackcrudelis: Once helper to restore all files is a bad idea I think.
First, you lose all your flexibility because you lost all possibility to restore like you want with some actions between each restore.

And second, you can't see what is restored and when. For debugging, I think it's really a bad idea.

Too much abstraction kill abstraction !
We lose readability, we lose flexibility, and we lose what it's really happened in the script.

Don't do too much abstraction !

zamentur:
I close this PR, an equivalent helper is implemented in https://github.com/YunoHost/yunohost/pull/275 .

PR #284, opened 2017-03-24
New helpers ynh_add_fpm_config and ynh_remove_fpm_config

maniackcrudelis:
Standard configuration of php-fpm.
Use local files stored in conf/, so it's still possible to use a specific config

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -57,13 +57,13 @@ ynh_remove_logrotate () {
 # usage: ynh_add_fpm_config
 ynh_add_fpm_config () {
 	finalphpconf="/etc/php5/fpm/pool.d/$app.conf"
-	ynh_compare_checksum_config "$finalphpconf" 1
+	ynh_backup_if_checksum_is_different "$finalphpconf" 1
 	sudo cp ../conf/php-fpm.conf "$finalphpconf"
 	ynh_replace_string "__NAMETOCHANGE__" "$app" "$finalphpconf"
 	ynh_replace_string "__FINALPATH__" "$final_path" "$finalphpconf"
 	ynh_replace_string "__USER__" "$app" "$finalphpconf"
 	sudo chown root: "$finalphpconf"
-	ynh_store_checksum_config "$finalphpconf"
+	ynh_store_file_checksum "$finalphpconf"
 
 	if [ -e "../conf/php-fpm.ini" ]
 	then
@@ -82,6 +82,6 @@ ynh_add_fpm_config () {
 # usage: ynh_remove_fpm_config
 ynh_remove_fpm_config () {
 	ynh_secure_remove "/etc/php5/fpm/pool.d/$app.conf"
-	ynh_secure_remove "/etc/php5/fpm/conf.d/20-$app.ini"
+	ynh_secure_remove "/etc/php5/fpm/conf.d/20-$app.ini" 2>&1
 	sudo systemctl reload php5-fpm
 }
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Need https://github.com/YunoHost/yunohost/pull/280 https://github.com/YunoHost/yunohost/pull/281 and https://github.com/YunoHost/yunohost/pull/286

alexAubin:
#280 and #281 got merged, still need #286

data/helpers.d/backend
```diff
@@ -51,3 +51,37 @@ ynh_remove_logrotate () {
...
+
+	sudo systemctl reload php5-fpm
+}
+
+# Remove the dedicated php-fpm config
+#
+# usage: ynh_remove_fpm_config
+ynh_remove_fpm_config () {
+	ynh_secure_remove "/etc/php5/fpm/pool.d/$app.conf"
+	ynh_secure_remove "/etc/php5/fpm/conf.d/20-$app.ini"
```
JimboJoe: Maybe make this silent, as the ini file is optional and when it doesn't exist, the remove script logs:
```
/etc/php5/fpm/conf.d/20-app.ini wasn't deleted because it doesn't exist.
```

maniackcrudelis: Done, not totally silently, but like that, we can keep a warning in the log in case of error.

data/helpers.d/backend
```diff
@@ -51,3 +51,37 @@ ynh_remove_logrotate () {
...
 }
+
+# Create a dedicated php-fpm config
+#
+# usage: ynh_add_fpm_config
+ynh_add_fpm_config () {
+	finalphpconf="/etc/php5/fpm/pool.d/$app.conf"
+	ynh_backup_if_checksum_is_different "$finalphpconf" 1
+	sudo cp ../conf/php-fpm.conf "$finalphpconf"
+	ynh_replace_string "__NAMETOCHANGE__" "$app" "$finalphpconf"
```
JimboJoe: Finally after testing this helper together with the [nginx helper](https://github.com/YunoHost/yunohost/pull/285), I think using `__NAME__` would be more consistent with the other helper.

maniackcrudelis: Hum, as you said just before, I think too that nobody change the fpm config file.
And, if I agree with you for the fact that `__NAME__` will be more consistent than `__NAMETOCHANGE__` with the nginx helper.
I think also that will affect all currently app which use php-fpm.


So, maybe it's better to break thoses apps now, before more apps uses this helper...

And, for a default php-fpm file, I think it's a good idea, but not like that. I prefer a default config in the exemple app. And each app copy this file in in own conf directory.
More control, and more aware that you use it.

alexAubin:
#286 has been merged

data/helpers.d/backend
JimboJoe: These are very minor points, so I'll follow your common sense ;-)

JimboJoe:
This PR needs to be updated accordingly with #286 last minute helper name changes (and no I can't update it myself as I still don't have rights ;-)

alexAubin:
>This PR needs to be updated accordingly with #286 last minute helper name changes (and no I can't update it myself as I still don't have rights ;-)

Uh, what exactly ? It looks to me like all the "checksum" helpers have the right name ? Except maybe `ynh_compare_checksum_config` ?

JimboJoe:
Tested and LGTM 👍

Rafi594:
This new helper work fine with nextcloud

frju365:
work well !

PR #285, opened 2017-03-24
New helpers ynh_add_nginx_config and ynh_remove_nginx_config

maniackcrudelis:
Standard configuration of nginx
Use local files stored in conf/, so it's still possible to use a specific config


maniackcrudelis:
Need https://github.com/YunoHost/yunohost/pull/280 https://github.com/YunoHost/yunohost/pull/281 and https://github.com/YunoHost/yunohost/pull/286

data/helpers.d/backend
```diff
@@ -51,3 +51,41 @@ ynh_remove_logrotate () {
 		sudo rm "/etc/logrotate.d/$app"
 	fi
 }
+
+# Create a dedicated nginx config
+#
+# usage: ynh_add_nginx_config
+ynh_add_nginx_config () {
+	finalnginxconf="/etc/nginx/conf.d/$domain.d/$app.conf"
+	ynh_compare_checksum_config "$finalnginxconf" 1
```
alexAubin: Uh, it's not clear for me what this line is intended to do ? Is it meant to backup the file for, like, if you runned ynh_add_nginx_config twice for instance ?

alexAubin:
#281 got merged

data/helpers.d/backend
maniackcrudelis: No simply to prevent to trashed a modified config by the user.
In fact, this line is designed to be used in the upgrade script.

JimboJoe: So with the new naming `ynh_backup_if_checksum_is_different`, it is clearer: we backup the file if it was modified, and then proceed to overwrite it, don't we?

data/helpers.d/backend
```diff
@@ -51,3 +51,41 @@ ynh_remove_logrotate () {
...
+	sudo cp ../conf/nginx.conf "$finalnginxconf"
```
JimboJoe: Could it be possible to pass the "source" filename as a parameter?
For example I have several apps for which I had to come up with 2 nginx conf files depending of whether the app is install at root or on a subdir (for example [here](https://github.com/YunoHost-Apps/wallabag2_ynh/tree/code_refactoring/conf)).

data/helpers.d/backend
maniackcrudelis: Yes that's it

data/helpers.d/backend
maniackcrudelis: We can possibly add the source file as an optional argument.
But, I think the better way is to use one config file instead of 2. Even is you need to configure differently if your app is installed on root or on a subpath.

Like I did with searx.
https://github.com/YunoHost-Apps/searx_ynh/blob/a12c4b6e925023d30e04bf044f5580077a7a62e1/conf/nginx.conf
https://github.com/YunoHost-Apps/searx_ynh/blob/a12c4b6e925023d30e04bf044f5580077a7a62e1/scripts/install#L59-L66

alexAubin:
#286 has been merged

PR #286, opened 2017-03-24
New helpers ynh_store_file_checksum and ynh_backup_if_checksum_is_different

maniackcrudelis:
Helpers for avoid destruction of personalised config files.
If the config file was manually modified, make a backup of it.

The name of this backup is returned, so the packager can choose which of this both file will used by default.


data/helpers.d/filesystem
```diff
@@ -80,3 +80,43 @@ properly with chmod/chown." >&2
...
+	config_file_checksum=checksum_${1//[\/ ]/_}	# Replace all '/' and ' ' by '_'
+	ynh_app_setting_set $app $config_file_checksum $(sudo md5sum "$1" | cut -d' ' -f1)
+}
+
+# Verify the checksum and backup the file if it's different
+#
+# usage: ynh_compare_checksum_config current_config_file [compress]
+# | arg: current_config_file - The file on which the checksum will be perfomed.
+# | arg: compress - 1 to compress the backup instead of a simple copy
+# A compression is needed for a file which will be analysed even if its name is different.
```
JimboJoe: I'm not sure to understand the idea here...

alexAubin: +1

Naively I thought the compress option was meant for big files..? I don't understand the "will be analyzed" part.

alexAubin:
I implemented @JimboJoe's comment but I didn't test the helper :s But for me it sounds like a nice idea and I'm okay to merge as is (if tests are successful)

data/helpers.d/filesystem
```diff
@@ -80,3 +80,49 @@ properly with chmod/chown." >&2
     chmod 755 $TMP_DIR
     echo $TMP_DIR
 }
+
+# Calculate and store a file checksum into the app settings
+#
+# $app should be defined globally when calling this helper
```
maniackcrudelis: No it's not needed.
You should consider to a variable can't be limited to the parent script.
A variable, if it set to `local` can't be used out of the function which create it. But, it can be used in any function beneath itself.

So, in your script, if $app is set to `local`, you still can use it in any functions you call. And if you not set it at `local` nor `global`, you can use it the same way.

The matter is only if you set a variable at `local` and try to use it outside this function.

data/helpers.d/filesystem
```diff
@@ -80,3 +80,49 @@ properly with chmod/chown." >&2
...
+# Verify the checksum and backup the file if it's different
+# This helper is primarily meant to allow to easily backup personalised/manually 
+# modified config files.
+#
+# $app should be defined when calling this helper
+#
+# usage: ynh_backup_if_checksum_is_different file [compress]
+# | arg: file - The file on which the checksum test will be perfomed.
+# | arg: compress - 1 to compress the backup instead of a simple copy
+# A compression is needed for a file which will be analyzed even if its name is different.
```
JimboJoe: I think the question is still pending concerning the meaning of this... Could you please give a precise use case?

maniackcrudelis: A precise case, and even the reason for this option is a nginx config file.
If you copy a nginx config file and put it just next to the original file, both files will be analysed by nginx. With a conflict in sight...

JimboJoe: Indeed, thanks! I think we can afford an additional line of comment to add this explanation, because without it it was really not clear for me.

zamentur:
The behaviour of this code is almost the same than the regen-conf...
May be there is a way to unify or prepare a unification of these 2 parts of the code

data/helpers.d/filesystem
alexAubin: >If you copy a nginx config file and put it just next to the original file, both files will be analysed by nginx.

I tried this but this is not the case ? In `my.domain.tld.conf`, there's an include of `my.domain.tld.d/*.conf`. So if your file is called `wordpress.conf.backup.20170101-133700` (or something alike) it's not red by nginx. Or are you thinking about another case ?

alexAubin:
Will merge on Sunday (21th) if no more comments

data/helpers.d/filesystem
```diff
@@ -98,29 +98,20 @@ ynh_store_file_checksum () {
...
 		then	# If the checksum is now different
-			backup_file="$file.backup.$(date '+%d.%m.%y_%Hh%M,%Ss')"
-			if [ $compress_backup -eq 1 ]
-			then
-				sudo tar --create --gzip --file "$backup_file.tar.gz" "$file"	# Backup the current file and compress
-				backup_file="$backup_file.tar.gz"
-			else
-				sudo cp -a "$file" "$backup_file"	# Backup the current file
-			fi
+			backup_file="/home/yunohost.conf/backup$file.backup.$(date '+%d.%m.%y_%Hh%M,%Ss')"
```
JimboJoe: Here shouldn't there be `/` between `backup` and `$file`...?

maniackcrudelis: I don't think so, because $file should be an absolute path.
But, in fact, I didn't test this code... (shame on me, I know...)

But, to prevent a error, if a double slash it's not a problem. It's better than nothing.

PR #288, opened 2017-03-24
New helper ynh_local_curl

maniackcrudelis:
Just an abstraction around curl command.
To prevent forgot arguments and simplify this difficult command.

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -27,6 +27,8 @@ ynh_get_plain_key() {
 
 # Curl abstraction to help with POST requests to local pages (such as installation forms)
 #
+# $domain and $path_url should be defined externally (and correspond to the domain.tld and the /path (of the app?))
+#
 # example: ynh_local_curl "/install.php?installButton" "foo=$var1" "bar=$var2"
 # 
 # usage: ynh_local_curl "page_uri" "key1=value1" "key2=value2" ...
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -25,16 +25,16 @@ ynh_get_plain_key() {
     done
 }
 
-# Curl abstraction to simplify its call
+# Curl abstraction to simplify its call on a local url
 #
 # Helpfully mostly for curl arguments
 #
-# example: ynh_curl_abstract "/install.php?installButton" "value1=$var1&value2=$var2"
+# example: ynh_local_curl "/install.php?installButton" "value1=$var1&value2=$var2"
 # 
-# usage: ynh_curl_abstract "url_access" "data_post"
+# usage: ynh_local_curl "url_access" "data_post"
 # | arg: url_access - Relative path to pointed for POST data sent. Relative to $path_url.
 # | arg: data_post - POST request to send
-ynh_curl_abstract () {
+ynh_local_curl () {
 	url_access=$1
 	data_post=${2:-}
 	curl -kL -H "Host: $domain" --resolve $domain:443:127.0.0.1 --data "$data_post" "https://localhost$path_url$url_access" 2>&1
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -24,3 +24,19 @@ ynh_get_plain_key() {
         fi
     done
 }
+
+# Curl abstraction to simplify its call
+#
+# Helpfully mostly for curl arguments
+#
+# example: ynh_curl_abstract "value1=$var1&value2=$var2" "/install.php?installButton"
+# 
+# usage: ynh_curl_abstract "data_post" "url_access"
+# | arg: data_post - POST request to send
+# | arg: url_access - Relative path to pointed for POST data sent. Relative to $path_url.
+ynh_curl_abstract () {
+	data_post=$1
+	url_access=$2
+	sleep 1
+	curl -kL -H "Host: $domain" --resolve $domain:443:127.0.0.1 --data "$data_post" "https://localhost$path_url$url_access" 2>&1
+}
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -25,17 +25,27 @@ ynh_get_plain_key() {
     done
 }
 
-# Curl abstraction to simplify its call on a local url
+# Curl abstraction to help with POST requests to local pages (such as installation forms)
 #
-# Helpfully mostly for curl arguments
-#
-# example: ynh_local_curl "/install.php?installButton" "value1=$var1&value2=$var2"
+# example: ynh_local_curl "/install.php?installButton" "foo=$var1" "bar=$var2"
 # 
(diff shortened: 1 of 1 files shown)


data/helpers.d/utils
```diff
@@ -24,3 +24,19 @@ ynh_get_plain_key() {
         fi
     done
 }
+
+# Curl abstraction to simplify its call
+#
+# Helpfully mostly for curl arguments
```
JimboJoe: What do you mean?

data/helpers.d/utils
```diff
@@ -24,3 +24,19 @@ ynh_get_plain_key() {
...
+#
+# example: ynh_curl_abstract "value1=$var1&value2=$var2" "/install.php?installButton"
```
JimboJoe: Here one could assume that you have to pass a variable systematically, although it could be a constant.
Maybe you can reformulate: `"key1=value1&key2=value2"` and state `where you can use shell variables as values (e.g. key1=$var1)`

data/helpers.d/utils
```diff
@@ -24,3 +24,19 @@ ynh_get_plain_key() {
...
+# 
+# usage: ynh_curl_abstract "data_post" "url_access"
+# | arg: data_post - POST request to send
```
JimboJoe: "POST **variables** to send"?

data/helpers.d/utils
```diff
@@ -24,3 +24,19 @@ ynh_get_plain_key() {
...
+# | arg: url_access - Relative path to pointed for POST data sent. Relative to $path_url.
```
JimboJoe: Relative path to URL for POST HTTP request (relative to $path_url)?

data/helpers.d/utils
maniackcrudelis: This helper is helpful particularly for the arguments of curl.
-kL -H "Host: $domain" --resolve $domain:443:127.0.0.1

Like that, the packager doesn't be careful about which arguments it's better to use.

data/helpers.d/utils
maniackcrudelis: It's just an example of POST data.
Maybe it needs to be rewritten

data/helpers.d/utils
maniackcrudelis: Is it a alternative proposition?
In this case, yes why not.

JimboJoe: Yes it is :)

data/helpers.d/utils
maniackcrudelis: No, POST data, so a POST request.
That can be a static data.

JimboJoe: I get your point. I was suggesting that one is not passing POST raw data as such, but a string that represents the data to fit into the variables in the POST request. Not a big deal :-)

data/helpers.d/utils
maniackcrudelis: In my mind, developer refer to the upstream dev of the app.
Packager refer only to the dev who work on the scripts for YunoHost.

"fiddle", I don't know. But why not I think it's also clear.

JimboJoe: I think Bram wrote in the chatroom about that. It is simpler and more gratifying for apps packagers. We should decide about that point :-)
You can use use "deal" instead of "fiddle", it is simpler.

alexAubin:
To me it's not clear what you are trying to achieve with this helper. After reading the code / curl call, it's not just a wrapper around curl because (from what I understand ?) you want to access pages hosted locally and do a POST request (or maybe a GET actually ? I dunno).

Also it's not clear how / where `$domain` and `$path_url` should be defined.

I'm trying to find a better name than `ynh_curl_abstract` which imho is too general and mysterious, but so far I only have something like `ynh_curl_local_post`

maniackcrudelis:
It's a curl call to fill POST data in an install form (usually).
A classic way to install lot of application with an install form.

$domain and $path_url are just passed as a global variables.

data/helpers.d/utils
Psycojoker: Yes, I call them "app developers" and the other ones "core developers". I, indeed, think that it's more gratifying and more accurate (and simpler) as this is really development.

data/helpers.d/utils
```diff
@@ -24,3 +24,19 @@ ynh_get_plain_key() {
...
+ynh_curl_abstract () {
+	data_post=$1
+	url_access=$2
+	sleep 1
```
zamentur: Why do we need to wait ?

data/helpers.d/utils
zamentur: I suggest to switch $1 and $2, it allow to write `ynh_curl_abstract "/install.php?installButton"` if there are no POST data

data/helpers.d/utils
maniackcrudelis: Only a empirical way to using curl ;)
We can try without.

data/helpers.d/utils
maniackcrudelis: Yes, you're right! It's better

alexAubin:
In 52f9f3d, I suggest splitting the second argument into several simpler one. That way the user don't have to worry about concatening everything with `&` : 

Example:
```
ynh_local_curl "/install.php?installButton" "foo=$var1" "bar=$var2"
```

data/helpers.d/utils
```diff
@@ -24,3 +24,28 @@ ynh_get_plain_key() {
...
+	# Define url of page to curl
+	full_page_url=https://localhost$path_url$1
+
+	# Concatenate all other arguments with '&' to prepare POST data
+	POST_data=""
+	for arg in "${@:2}"
+	do
+		POST_data="${POST_data}${arg}&"
+	done
+	POST_data=${POST_data::-1}
```
maniackcrudelis: I guess this line remove the last '&'.
Please comment this kind of line.

I think is really not clear for everyone that don't know bash very well.

alexAubin: Yes, sorry about that. It was actually not obvious at all for me when I found it on the internets ;)

zamentur:
We may use $YNH_APP_ARG_DOMAIN and $YNH_APP_ARG_PATH (if exist)

scith:
Hi, this is nice but I found some problems:
- When do POST args are entered, the command fails. Perhaps something like?
``` bash
if [ -z ${@:2} ];
	then
		# Curl the URL with no POST
		curl -kL -H "Host: $domain" --resolve $domain:443:127.0.0.1 "$full_page_url" 2>&1
	else
```
- The results are very verbose. Is there a way to make it execute in a a more quiet way?

I'll continue experimenting with it to see if I manage to solve these

Thanks

maniackcrudelis:
@scith https://github.com/YunoHost/yunohost/pull/302

PR #289, opened 2017-03-24
New helpers ynh_backup_after_failed_upgrade and ynh_backup_before_upg…

maniackcrudelis:
…rade

Create a backup at the beginning of upgrade script.
Then, if the upgrade script failed, remove the app and restore it.

The backup is kept, to provide a way to restore if the upgrade is finally a bad thing.


data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+# Restore a previous backup if the upgrade process failed
+#
+# usage:
+# ynh_backup_before_upgrade
+# ynh_clean_setup () {
+# 	ynh_backup_after_failed_upgrade
+# }
+# ynh_abort_if_errors
+#
+ynh_backup_after_failed_upgrade () {
```
JimboJoe: Naming is unclear for me; why not `ynh_backup_restore_after_failed_upgrade`?

data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+	echo "Upgrade failed." >&2
+	app_bck=${app//_/-}	# Replace all '_' by '-'
+    # Check if a existing backup can be found before remove and restore the application.
```
JimboJoe: Typos :  Check if a**n** existing backup can be found before remov**ing** and restor**ing** the application.

data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+	if sudo yunohost backup list | grep -q $app_bck-pre-upgrade$backup_number
+    then
+    	# Remove the application then restore it
+		sudo yunohost app remove $app
+        # Restore the backup if the upgrade failed
```
JimboJoe: Here "if the upgrade failed" can be removed.

data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+# ynh_clean_setup () {
+# 	ynh_backup_after_failed_upgrade
+# }
+# ynh_abort_if_errors
+#
+ynh_backup_before_upgrade () {
+	backup_number=1
+	old_backup_number=2
+	app_bck=${app//_/-}	# Replace all '_' by '-'
+    # Check if a backup already exist with the prefix 1.
```
JimboJoe: typo: exist**s**
Also remove the final "." as I first thought it was part of the searched prefix.

data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+	if sudo yunohost backup list | grep -q $app_bck-pre-upgrade1
+    then
+        # Prefix become 2 to preserve the previous backup
```
JimboJoe: typo: become**s**

data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+		backup_number=2
+		old_backup_number=1
+	fi
+
+	# Create another backup
```
JimboJoe: not necessarily "another"

data/helpers.d/utils
```diff
@@ -24,3 +24,62 @@ ynh_get_plain_key() {
...
+	sudo yunohost backup create --ignore-hooks --apps $app --name $app_bck-pre-upgrade$backup_number
+	if [ "$?" -eq 0 ]
+    then
+       	# If the backup succedded, remove the previous backup
```
JimboJoe: typo: succ**eed**ed

JimboJoe:
Thanks @alexAubin for being my scribe ;-)
So the only remaining question from my side is the function naming.

data/helpers.d/utils
maniackcrudelis: So, in this case, to follow the real function of this helper.
May we should rename it ynh_restore_backup_if_upgrade_failed

data/helpers.d/utils
```diff
@@ -24,3 +24,64 @@ ynh_get_plain_key() {
...
+	echo "Upgrade failed." >&2
+	app_bck=${app//_/-}	# Replace all '_' by '-'
+	
+        # Check if an existing backup can be found before removing and restoring the application.
+	if sudo yunohost backup list | grep -q $app_bck-pre-upgrade$backup_number
+    	then
+		# Remove the application then restore it
+		sudo yunohost app remove $app
+		# Restore the backup
+		sudo yunohost backup restore --ignore-hooks $app_bck-pre-upgrade$backup_number --apps $app --force
```
JimboJoe: `--ignore-hooks` should now be replaced by `--ignore-system` (see [here](https://github.com/YunoHost/yunohost/blob/unstable/src/yunohost/backup.py#L1965))

alexAubin:
It would be good to merge the remaining helper... but I don't know how to test them :s Does it looks okay to you @JimboJoe ? (or other people in app's group ?)

JimboJoe:
For me it's tested OK (and used in several packages), but the naming issue and my last comment remain. Afterwards, I think we're good to go!

alexAubin:
Alrighty, I think I'm gonna be a dictator and just click on merge right away since it's already 4 months old 😋

JimboJoe:
Hmm... I'm not that happy with that merge, as the helper name clearly doesn't describe well what it does, and we could have discussed that 5 minutes in the chatroom... :confused:

alexAubin:
Sorry I did not see the end of your previous comment :anguished: We can probably do a microdecision after we find a good name on the chat

PR #290, opened 2017-03-28
[fix] Properly manage resolv.conf, dns resolvers and dnsmasq

alexAubin:
### Status

Ready for review and tests (Feel free also to discuss the strategy adopted)

### Problem

(Imho this is a serious bug and should be treated as a release blocker for 2.6.x)

This is an attempt to fix https://dev.yunohost.org/issues/822 and other related issues. Right now, on some setups (such as ynh-dev / vagrant), resolvconf is not present and therefore `/etc/resolv.conf` does not point to 127.0.0.1 / dnsmasq. This leads to several issues since [the main domain is now the hostname](https://github.com/YunoHost/yunohost/pull/219), such as `Warning: sudo: unable to resolve host the.main.domain`. 

As long as at least `/etc/resolv.conf` or `/etc/hosts` is not managed properly, we will also always have people who will get messages such as "Cannot resolve some.domain.tld locally" (when installing Let's Encrypt certificate) and will have to tweak `/etc/hosts` manually.

### Solution

- Add resolvconf as dependency. This automatically (apparently ??) make `/etc/resolv.conf` point to 127.0.0.1 = dnsmasq ;
- This leads to the question of which 'external' DNS are used. Right now my understanding is that by default dnsmasq use google's DNS and caching is disabled ? Therefore I propose tweaking dnsmasq config to use [friendly DNS resolvers](http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver) and enable caching.

### References

Got some inspiration from these pages : 
- http://www.drazzib.com/docs/admin/dnsmasq.html
- https://wiki.archlinux.org/index.php/dnsmasq
- https://wiki.archlinux.org/index.php/Resolv.conf

### DIFF
--- a/data/templates/dnsmasq/plain/resolv.dnsmasq.conf
+++ b/data/templates/dnsmasq/plain/resolv.dnsmasq.conf
@@ -6,21 +6,21 @@
 # http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver
 
 # (FR) FDN
-80.67.169.12
-80.67.169.40
+nameserver 80.67.169.12
+nameserver 80.67.169.40
 # (FR) LDN
-80.67.188.188
+nameserver 80.67.188.188
 # (FR) ARN
-89.234.141.66
+nameserver 89.234.141.66
 # (FR) gozmail / grifon
-89.234.186.18
+nameserver 89.234.186.18
 # (DE) FoeBud / Digital Courage
-85.214.20.141
+nameserver 85.214.20.141
 # (DE) CCC Berlin
-213.73.91.35
+nameserver 213.73.91.35
 # (DE) Ideal-Hosting
-84.200.69.80
-84.200.70.40
+nameserver 84.200.69.80
+nameserver 84.200.70.40
 # (DK) censurfridns
-91.239.100.100
-89.233.43.71
+nameserver 91.239.100.100
+nameserver 89.233.43.71
--- a/data/templates/dnsmasq/plain/resolv.dnsmasq.conf
+++ b/data/templates/dnsmasq/plain/resolv.dnsmasq.conf
@@ -5,9 +5,7 @@
 # List taken from
 # http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver
 
-# (FR) FDN
-nameserver 80.67.169.12
-nameserver 80.67.169.40
+
 # (FR) LDN
 nameserver 80.67.188.188
 # (FR) ARN
@@ -24,3 +22,6 @@ nameserver 84.200.70.40
 # (DK) censurfridns
 nameserver 91.239.100.100
 nameserver 89.233.43.71
+# (FR) FDN
+nameserver 80.67.169.12
+nameserver 80.67.169.40
--- a/data/templates/dnsmasq/plain/resolv.dnsmasq.conf
+++ b/data/templates/dnsmasq/plain/resolv.dnsmasq.conf
@@ -1,0 +1,26 @@
+# According to dnsmasq's doc, dnsmasq will automatically 
+# choose the fastest-to-answer resolver as primary 
+# (if strict-order is disabled)
+
+# List taken from
+# http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver
+
+# (FR) FDN
+80.67.169.12
+80.67.169.40
+# (FR) LDN
+80.67.188.188
+# (FR) ARN
+89.234.141.66
+# (FR) gozmail / grifon
+89.234.186.18
+# (DE) FoeBud / Digital Courage
+85.214.20.141
+# (DE) CCC Berlin
+213.73.91.35
+# (DE) Ideal-Hosting
+84.200.69.80
+84.200.70.40
+# (DK) censurfridns
+91.239.100.100
+89.233.43.71
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -163,7 +163,8 @@ def tools_maindomain(auth, new_domain=None):
         logger.warning("%s" % e, exc_info=1)
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
-    # Clear nsswitch cache for hosts...
+    # Clear nsswitch cache for hosts to make sure hostname is resolved ...
+    service_regen_conf(['nsswitch'], force=True)
     subprocess.call(['nscd', '-i', 'hosts'])
 
     # Set hostname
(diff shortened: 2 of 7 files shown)


data/templates/dnsmasq/plain/resolv.dnsmasq.conf
```diff
@@ -0,0 +1,26 @@
+# According to dnsmasq's doc, dnsmasq will automatically 
+# choose the fastest-to-answer resolver as primary 
+# (if strict-order is disabled)
+
+# List taken from
+# http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver
```
opi: we should probably inform these organisations that their DNS server is listed here

data/templates/dnsmasq/plain/resolv.dnsmasq.conf
```diff
@@ -0,0 +1,26 @@
...
+
+# (FR) FDN
+80.67.169.12
```
opi: Is it the proper format ? I thought it must be like `nameserver 11.22.33.44` according to dnsmasq documentation: "For the format of this file see resolv.conf(5)."

alexAubin:
Okay so, since adminsys/system designers are crazy people who decided that simply changing the hostname should take 10 differents commands, it turns out that even if `host my.host.name` correcly returns an IP (after properly setting up /etc/resolv.conf and dnsmasq), sudo **still** throws a `unable to resolve host`. Because some you *have to* edit `/etc/hosts` to make your hostname point to 127.0.0.1.

Fortunately, a gentlemen created a small library/package so that you don't need to do this. It's called [nss-myhostname](http://0pointer.de/lennart/projects/nss-myhostname/) : 
> [...] Various software relies on an always resolvable local host name. When using dynamic hostnames this is usually achieved by patching /etc/hosts at the same time as changing the host name. This however is not ideal since it requires a writable /etc file system and is fragile because the file might be edited by the administrator at the same time. nss-myhostname simply returns all locally configure public IP addresses [...]. Patching /etc/hosts is thus no longer necessary.

alexAubin:
In 601e343 and b987807, I realized I had to tweak `/etc/nsswitch.conf` and clear the cache of nscd to make sure the use of nss-myhostname is properly taken into account.

Then, in 329814c, I made some arrangements in dnsmasq about the dns resolvers used. This was discussed in parallel with ljf. The problem was : if we indeed use resolvconf/dnsmasq, then (at the moment) we rely on the dns resolvers provided by the DHCP. However, this conflicts with some setups, such as vpnclient_ynh, which want to tweak the /etc/resolv.conf specifically because they don't want to rely on the DHCP dns providers (these are usually the ISP's DNS which won't answer to requests from the VPN).

Therefore we must have a list of DNS provider which we trust. The list chosen it from the FFDN and other 'open'/ethic DNS providers. Since we don't want the load to be focused on the first entry of the list, we decided to shuffle the list during the regen-conf. (Other alternatives where investigated, such as the number of hop in traceroute, the average ping, or the query time, but none of these where found to be really meaningful to optimize). That way the load is spread around all of the resolvers. (Note also that caching is dnsmasq is now explicitly enabled.)

Admin / apps can still specify their own DNS resolvers they want to use by replacing the content of `/etc/resolv.dnsmasq.conf` (instead of `/etc/resolv.conf`), which won't be overriden by the regen-conf of dnsmasq since it'll be detected as manually modified.

data/templates/dnsmasq/plain/resolv.dnsmasq.conf
```diff
@@ -0,0 +1,27 @@
...
+nameserver 213.73.91.35
+# (DE) Ideal-Hosting
+nameserver 84.200.69.80
+nameserver 84.200.70.40
+# (DK) censurfridns
+nameserver 91.239.100.100
+nameserver 89.233.43.71
+# (FR) FDN
+nameserver 80.67.169.12
+nameserver 80.67.169.40
```
zamentur: J'ai envoyé un mail à la FFDN pour demander leurs avis.

sachaz: DNS Aquilenet récursifs ouverts (gaia-dns.aquilenet.fr 141.255.128.100 2a01:474::100 et hades-dns.aquilenet.fr 141.255.128.101 2a01:474::101)

alexAubin: @sachaz Thanks, I added it to the list (because it sounds legit, anyone feel free to speak up if you're against it)

Any reason why you're not on [this page](http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver) yet ? :P

sachaz: AlexAubin: refesh the page ;)
There is too another ressource: https://www.ffdn.org/wiki/doku.php?id=formations:dns

alexAubin: Nice ! Thanks a lot !

src/yunohost/tools.py
```diff
@@ -163,6 +163,10 @@ def tools_maindomain(auth, new_domain=None):
         logger.warning("%s" % e, exc_info=1)
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
+    # Clear nsswitch cache for hosts to make sure hostname is resolved ...
+    service_regen_conf(['nsswitch'], force=True)
```
alexAubin: Moved it back to the postinstall, it wasnt clear to me why I put in the maindomain function 😕

alexAubin:
> We need a backup/restore hooks for these new configuration files

Do we ? Isn't there a regen-conf call at the end of the restore ? For instance, there's no backup of postfix.

zamentur:
> Do we ?
There are ssh backup/restore hooks, same for mysql ...

> Isn't there a regen-conf call at the end of the restore ?
No there isn't (just a ssowatconf call)

> For instance, there's no backup of postfix
I think it's a missing hook...

I agree a regen-conf fixes a missing backup hook. We might run regen-conf at the end of backup, but the idea of a backup is to restore like it was...

alexAubin:
I understand that the backup should restore like it was, but for dnsmasq (if you forget about the shuffle part), we just want to re-regen the conf, just like postfix, dovecot, fail2ban, and other services which don't need to save any particular data (all can be restored entirely from the regen-conf). This is not the case for ssh or mysql because they need specific data or actions (known keys / host specific to the install, backup/restore the MySQL key, ...). At the moment I don't see which specific data we need to backup.

It starts making sense to backup stuff for dnsmasq if the user edited the resolv.dnsmasq.conf to use custom DNS (and therefore the file is considered as 'managed manually' / 'edited by hand' and thus ignored by regen-conf), but that's also the case for postfix, dovecot, fail2ban, and so on. So to me that's beyond the scope of this PR.

alexAubin:
Merging as proposed in the mail from April 12th to yunoteam.

PR #291, opened 2017-03-31
[fix] Remove version from api

zamentur:
Related to: https://dev.yunohost.org/issues/703

It's for security.

### DIFF
--- a/bin/yunohost-api
+++ b/bin/yunohost-api
@@ -192,12 +192,10 @@ if __name__ == '__main__':
     _init_moulinette(opts.use_websocket, opts.debug, opts.verbose)
 
     # Run the server
-    from yunohost.utils.packages import ynh_packages_version
     ret = moulinette.api(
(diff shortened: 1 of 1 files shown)


opi:
This will break web administration as we display current version in the footer ; We should be able to retreive version from API once logged in

zamentur:
So it will be more difficult to fix this. Currently, additionnal routes don't support authentication !

https://github.com/YunoHost/moulinette/blob/unstable/moulinette/interfaces/api.py#L710

Psycojoker:
On the webadmin the version number is not displayed anymore if the admin isn't logged in, but the API still make it accessible, I don't know if it's what we want https://ynh.local/yunohost/api/version

alexAubin:
[Semi-automatic message] 

alexAubin:
Reviving this because : https://github.com/YunoHost/yunohost-admin/pull/226

PR #292, opened 2017-04-01
[fix] Avoid to remove a apt package accidentally

zamentur:
Related to : https://dev.yunohost.org/issues/876

### DIFF
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -35,7 +35,7 @@ ynh_apt() {
 #
 # usage: ynh_package_update
 ynh_package_update() {
-    ynh_apt --no-remove update
+    ynh_apt update
 }
 
 # Install package(s)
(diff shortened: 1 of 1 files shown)


data/helpers.d/package
```diff
@@ -35,15 +35,15 @@ ynh_apt() {
 #
 # usage: ynh_package_update
 ynh_package_update() {
-    ynh_apt update
+    ynh_apt --no-remove update
```
alexAubin: Does this make sense ? I'm not sure there's any risk to remove a package during an apt-get update (it only refreshes the repository database thing)

opi: but during `install` one can remove a package. If you read apt's man page, you can see that 
> If a hyphen is appended to the package name (with no intervening space), the identified package will be removed if it is installed.
( http://manpages.ubuntu.com/manpages/xenial/en/man8/apt-get.8.html )

Stupid ! but we should take care of everything !

PR #294, opened 2017-04-04
[enh] Tools public key (#831) 

e-lie:
To be able to add ssh access to an other instance and create a remote backup we need to display a public-key.
This new tool function generate a key pair (id_rsa_ynh) inside /root/.ssh if it doesn't exists yet
and return the public key.

Edit by Alex : Addresses https://dev.yunohost.org/issues/831


src/yunohost/tools.py
```diff
@@ -565,3 +565,25 @@ def tools_diagnosis(auth, private=False):
...
+
+    if (not pub) and (not private):
+        os.system('ssh-keygen -P "" -f %s >/dev/null' % key_path)
+
+    with open(key_path+'.pub', 'rt') as f:
+        return f.readline().strip('\n')
+        
+
+
+
```
M5oul: four empty lines.

src/yunohost/tools.py
```diff
@@ -565,3 +565,25 @@ def tools_diagnosis(auth, private=False):
         diagnosis['private']['domains'] = domain_list(auth)['domains']
 
     return diagnosis
+
+def tools_public_key():
+    key_path = '/root/.ssh/id_rsa_ynh'
```
opi: as `admin` user is used for many administration tasks, should we rather create the key in `/home/admin` ?

data/actionsmap/yunohost.yml
```diff
@@ -1345,6 +1345,11 @@ tools:
                     help: Show private data (domain, IP)
                     action: store_true
 
+        ### tools_public_key()
+        public-key:
+            action_help: Get public key or generate a new one
+            api: GET /public-key
+
```
opi: Add auth configuration to allow anonymous requests, something like
```
configuration:
    authenticate: false
    lock: false
```

src/yunohost/tools.py
```diff
@@ -565,3 +565,25 @@ def tools_diagnosis(auth, private=False):
...
+    pub = os.path.exists(key_path+'.pub')
+    private = os.path.exists(key_path)
+
+    if pub and (not private):
+        os.remove(key_path+'.pub')
+        pub = not pub
```
opi: why not `pub = false` or `None` ? imho it will be more readable.

src/yunohost/tools.py
```diff
@@ -565,3 +565,25 @@ def tools_diagnosis(auth, private=False):
...
+    elif (not pub) and private:
+        os.remove(key_path)
+        private = not private
+
+    if (not pub) and (not private):
+        os.system('ssh-keygen -P "" -f %s >/dev/null' % key_path)
```
opi: - Use strong default to `ssh-keygen` command, like `-t rsa -b 4096` or the recent `-t ed25519`
- Add a comment, like `-c "Generated by YunoHost"`.
- Be quiet : `-q`

Some worth reading: https://wiki.mozilla.org/Security/Guidelines/OpenSSH#Key_generation & https://wiki.mozilla.org/Security/Key_Management

src/yunohost/tools.py
alexAubin: I'd vote to have the key stored in something like `/etc/yunohost/keys/` (or whichever folder name sounds right and reflects the idea of 'federated identity').

There were some IRL discussions about how to feed/choose the relevant SSH key with borg backup, and I think the answer is in [here](http://borgbackup.readthedocs.io/en/stable/usage.html) (c.f. `BORG_RSH`)

e-lie: Agreed. I asked @zamentur  who said it would easier to use it with borg in `/root/.ssh` but considering that `BORG_RSH` option I can put it in /etc/yunohost/keys/

e-lie:
I suppose that the key_path should be set as a global setting, but not sure how to do this now...

src/yunohost/tools.py
```diff
@@ -565,3 +565,26 @@ def tools_diagnosis(auth, private=False):
...
+    if pub and (not private):
+        os.remove(key_path+'.pub')
+        pub = False
+    elif (not pub) and private:
+        os.remove(key_path)
+        private = False
+
+    if (not pub) and (not private):
+        if not os.path.exists(keys_folder):
+            os.mkdir(keys_folder)
```
zamentur: What about right permissions for ssh private key ?
I think we need a backup hook to backup this keys

opi: does someone test this PR to use the generated ssh key ? I guess `ssh` command will tell if permissions or not correct

src/yunohost/tools.py
```diff
@@ -565,3 +565,26 @@ def tools_diagnosis(auth, private=False):
...
+    key_path = os.path.join(keys_folder, '/id_{}_ynh'.format(key_type))
+    pub = os.path.exists(key_path + '.pub')
+    private = os.path.exists(key_path)
+
+    if pub and not private:
+        os.remove(key_path + '.pub')
+        pub = False
+    elif not pub and private:
+        os.remove(key_path)
+        private = False
```
Psycojoker: This defensive code looks weird, I'm really wondering in which case we could end up in those situations.

src/yunohost/tools.py
```diff
@@ -565,3 +565,26 @@ def tools_diagnosis(auth, private=False):
         diagnosis['private']['domains'] = domain_list(auth)['domains']
 
     return diagnosis
+
+
+def tools_public_key():
+    key_type = 'ed25519'
```
Psycojoker: We had bad experience in YunoHost infra using this key format, it's apparently not really widly supported yet :/

This also should probably be a CLI argument.

src/yunohost/tools.py
```diff
@@ -565,3 +565,26 @@ def tools_diagnosis(auth, private=False):
...
+        os.remove(key_path + '.pub')
+        pub = False
+    elif not pub and private:
+        os.remove(key_path)
+        private = False
+
+    if not pub and not private:
+        if not os.path.exists(keys_folder):
+            os.mkdir(keys_folder)
+        os.system('ssh-keygen -q -t {0} -N "" -C "Generated by Yunohost" -f {1}'.format(key_type, key_path))
```
Psycojoker: We should probably uses this https://docs.python.org/2/library/subprocess.html#subprocess.check_call to ensure that the call was correctly made.

Psycojoker:
Thinking about it, this remind me a discussion with Moul and I think that I would be interesting to include a dedicated support for ssh keys in yunohost including allowing users to be able to log as ssh.

alexAubin:
[Semi-automatic message] 

PR #296, opened 2017-04-15
Manage SSOwat like other services

alexAubin:
Don't know what I'm doing :P But I'm giving this a try to have some feedback

## Problem

At the moment SSOwat is kinda tightly coupled to Yunohost with the `app ssowatconf` action. Fundamentally there shouldn't be any reason to have SSOwat conf handled in a special way compared to other services like nginx or postfix.

## Solution

Have a regen-conf hook that regenerate /etc/ssowat/conf.json just like it's done for other conf files.

However it's a bit tricky because ssowat needs to know about domains, users and apps - and needs to extensively parse/write json. At first I thought I'd write the regen-conf hook in pure bash, but that was a bit tricky without the bash json tool (jq?). Then I thought I'd write it in python directly. Then after writing and testing, I realized hooks are called explicitly with `/bin/bash (-x) the_hook_file`. So I wrapped everything in a python bash call using stdin as the code :/. This is pretty dirty but can't do much without changing the hook call code :/

What do you guys think ?

### DIFF
--- a/data/hooks/conf_regen/53-ssowat
+++ b/data/hooks/conf_regen/53-ssowat
@@ -1,0 +1,175 @@
+#!/bin/bash
+
+python - <<PYTHONPROGRAM
+# Standard python libs
+import os
+import sys
+import json
+import argparse
+
+# Yunohost specific
+sys.path.append('/usr/lib/moulinette/')
+from moulinette.core import init_authenticator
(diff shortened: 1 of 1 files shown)


alexAubin:
Closing because this PR is too weird and we don't really care /o\

PR #297, opened 2017-04-20
[fix] Add random delay to app fetchlist cron job

alexAubin:
Fixes https://dev.yunohost.org/issues/892 (All yunohost instances update their appslist at midnight precisely)

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1937,7 +1937,8 @@ def _install_appslist_fetch_cron():
     cron_job.append("#!/bin/bash")
     # We add a random delay between 0 and 60 min to avoid every instance fetching
     # the appslist at the same time every night
-    cron_job.append("(sleep $((RANDOM%3600)); yunohost app fetchlist > /dev/null 2>&1) &")
+    cron_job.append("(sleep $((RANDOM%3600));")
+    cron_job.append("yunohost app fetchlist > /dev/null 2>&1) &")
 
     with open(cron_job_file, "w") as f:
         f.write('\n'.join(cron_job))
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1933,8 +1933,14 @@ def _install_appslist_fetch_cron():
 
     logger.debug("Installing appslist fetch cron job")
 
+    cron_job = []
+    cron_job.append("#!/bin/bash")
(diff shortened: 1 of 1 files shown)


M5oul:
Will the cron job script will be updated for current install?
It seems adding a new script update it.

src/yunohost/app.py
```diff
@@ -1933,8 +1933,14 @@ def _install_appslist_fetch_cron():
...
 
+    cron_job = []
+    cron_job.append("#!/bin/bash")
+    # We add a random delay between 0 and 60 min to avoid every instance fetching
+    # the appslist at the same time every night
+    cron_job.append("(sleep $((RANDOM%3600)); yunohost app fetchlist > /dev/null 2>&1) &")
+
     with open(cron_job_file, "w") as f:
-        f.write('#!/bin/bash\n\nyunohost app fetchlist > /dev/null 2>&1\n')
+        f.write('\n'.join(cron_job))
```
M5oul: Why joining with a `\n` at file beginning?

alexAubin:
>Will the cron job script will be updated for current install?
>It seems adding a new script update it.

It will because this is included in the dirty migration during the applist management refactoring here

src/yunohost/app.py
alexAubin: Uh, not sure I get the question ? `char.join(list)` concatenate the element of the list with char in between. So if you have `'\n'.join(["abc", "dev"])`, you get `"abc\ndef"`

M5oul: Ah, ok. I thought it was `\n$cron_job`.

src/yunohost/app.py
```diff
@@ -1933,8 +1933,14 @@ def _install_appslist_fetch_cron():
 
     logger.debug("Installing appslist fetch cron job")
 
+    cron_job = []
+    cron_job.append("#!/bin/bash")
+    # We add a random delay between 0 and 60 min to avoid every instance fetching
+    # the appslist at the same time every night
+    cron_job.append("(sleep $((RANDOM%3600)); yunohost app fetchlist > /dev/null 2>&1) &")
```
opi: please split this line in 2 separate line, one with the delay, and the other with the `fetchlist` command, to improve readability

PR #299, opened 2017-04-21
[fix] Refactor DNS conf management for domains

alexAubin:
### Status

- Ready for reviews
- We need to tweak the dynette [here](https://github.com/YunoHost/dynette/blob/master/dynette.cron.py#L12-L19) to allow all the records of the recommended DNS conf (especially the DKIM/DMARC ones) to be manipulated via the dyndns update stuff. (c.f. [corresponding PR](https://github.com/YunoHost/dynette/pull/6))

### Problem

Right now, the handling of the recommended DNS configuration is messy. There are inconsistencies in yunohost between the "recommended" configuration shown via `yunohost domain dns-conf` and the one used in dyndns update and dnsmasq (c.f. https://dev.yunohost.org/issues/267). For the dydns update, no DKIM key is provided, therefore all nohost.me don't have valid DKIM emails ! (c.f. https://dev.yunohost.org/issues/848)

Also it turns out that `yunohost domain dns-conf` doesn't show the DKIM part...

### Solution

I added a `_build_dns_conf()` helper that build a dictionnary describing the DNS conf for a given domain. This provide a central point in the code to manage the DNS stuff for domains. This is then interpreted in `yunohost domain dns-conf` to have a simple text-display of the recommended conf AND in `yunohost dyndns update` to create the update instructions for the dynette.

This should also help to implement https://dev.yunohost.org/issues/605 and https://dev.yunohost.org/issues/596 (and also a proper sync with dnsmasq also). This might also help #177.

### DIFF
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -198,67 +198,26 @@ def domain_dns_conf(domain, ttl=None):
         ttl -- Time to live
 
     """
+
     ttl = 3600 if ttl is None else ttl
-    ip4 = ip6 = None
 
-    # A/AAAA records
-    ip4 = get_public_ip()
-    result = (
-        "@ {ttl} IN A {ip4}\n"
-        "* {ttl} IN A {ip4}\n"
-    ).format(ttl=ttl, ip4=ip4)
+    dns_conf = _build_dns_conf(domain, ttl)
 
-    try:
-        ip6 = get_public_ip(6)
-    except:
-        pass
-    else:
-        result += (
-            "@ {ttl} IN AAAA {ip6}\n"
-            "* {ttl} IN AAAA {ip6}\n"
-        ).format(ttl=ttl, ip6=ip6)
+    result = ""
 
-    # Jabber/XMPP
-    result += ("\n"
-        "_xmpp-client._tcp {ttl} IN SRV 0 5 5222 {domain}.\n"
-        "_xmpp-server._tcp {ttl} IN SRV 0 5 5269 {domain}.\n"
-        "muc {ttl} IN CNAME @\n"
-        "pubsub {ttl} IN CNAME @\n"
-        "vjud {ttl} IN CNAME @\n"
-               ).format(ttl=ttl, domain=domain)
+    result += "# Basic ipv4/ipv6 records"
+    for record in dns_conf["basic"]:
+        result += "\n{name} {ttl} IN {type} {value}".format(**record)
 
-    # Email
-    result += ('\n'
-        '@ {ttl} IN MX 10 {domain}.\n'
-        '@ {ttl} IN TXT "v=spf1 a mx ip4:{ip4}'
-               ).format(ttl=ttl, domain=domain, ip4=ip4)
-    if ip6 is not None:
-        result += ' ip6:{ip6}'.format(ip6=ip6)
-    result += ' -all"'
+    result += "\n\n"
+    result += "# XMPP"
+    for record in dns_conf["xmpp"]:
+        result += "\n{name} {ttl} IN {type} {value}".format(**record)
 
-    # DKIM
-    try:
-        with open('/etc/dkim/{domain}.mail.txt'.format(domain=domain)) as f:
-            dkim_content = f.read()
-    except IOError:
-        pass
-    else:
-        dkim = re.match((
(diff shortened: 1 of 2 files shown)


src/yunohost/domain.py
```diff
@@ -357,3 +316,92 @@ def _normalize_domain_path(domain, path):
...
+    dnsconf["xmpp"].append(_dns_record("_xmpp-client._tcp", ttl, "SRV", "0 5 5222 %s." % domain))
+    dnsconf["xmpp"].append(_dns_record("_xmpp-server._tcp", ttl, "SRV", "0 5 5269 %s." % domain))
+    dnsconf["xmpp"].append(_dns_record("muc", ttl, "CNAME", "@"))
+    dnsconf["xmpp"].append(_dns_record("pubsub", ttl, "CNAME", "@"))
+    dnsconf["xmpp"].append(_dns_record("vjud", ttl, "CNAME", "@"))
+
+    # Email
+    dnsconf["mail"].append(_dns_record("@", ttl, "MX", "10 %s." % domain))
+
+        # SPF record
```
Psycojoker: detail: unindented comment, same for dkim

src/yunohost/dyndns.py
```diff
@@ -205,41 +205,45 @@ def dyndns_update(dyn_host="dyndns.yunohost.org", domain=None, key=None,
...
-            lines += [
-                'update add %s. 1800 AAAA %s' % (domain, ipv6),
-                'update add pubsub.%s. 1800 AAAA %s' % (domain, ipv6),
-                'update add muc.%s. 1800 AAAA %s' % (domain, ipv6),
-                'update add vjud.%s. 1800 AAAA %s' % (domain, ipv6),
-            ]
+
+        dns_conf = _build_dns_conf(domain)
+        all_records = []
+        for records in dns_conf.values():
```
Psycojoker: Should comment on what this part of the code is doing because people won't have in mind what is coming from `_build_dns_conf` and the result of this loop.

Psycojoker:
I have to admit that I'm not really convinced by this refactoring, before was really explicit (but messy) and gave a very good idea on what was going to be produced, not everything is hidden behind abstractions and implicite data structures and it's way harder to grasp from the code what is happening :confused:

alexAubin:
>I have to admit that I'm not really convinced by this refactoring, before was really explicit (but messy) and gave a very good idea on what was going to be produced, not everything is hidden behind abstractions and implicite data structures and it's way harder to grasp from the code what is happening 😕
>Also the code is not obvious and it took me a bit of time to correctly understood what is happening while it was not the case before :x

Okay, but that's not my fault if DNS is a mess and if sometimes `@` is wanted, sometimes the exact domain is wanted ; some times in the future we'll want to disable mail or xmpp for domains so we want the conf to automatically adapt and be modular ; sometimes records are different if you have ipv6 on the machine, and so on. The previous thing surely had some `print "Hello world"`-like code so invidividual lines were easy to understand, but as a whole it was really hard to grasp and to make it evolve.

So yes, we need abstraction (though it's not like if I added 6 layers of virtual classes or anything), which makes some part less explicit (and some other more explicit), but at least the DNS conf is synced between the different part of the code, and it can easily evolve. But you are free to suggest any better solution to fix the issue 😕

src/yunohost/dyndns.py
```diff
@@ -205,41 +205,45 @@ def dyndns_update(dyn_host="dyndns.yunohost.org", domain=None, key=None,
...
             'show',
             'send'
         ]
+
         with open('/etc/yunohost/dyndns/zone', 'w') as zone:
-            for line in lines:
-                zone.write(line + '\n')
+            zone.write('\n'.join(lines))
+
+        return
```
Psycojoker: wat

Psycojoker:
I've tried to make the internal function a bit more transparent by reducing the number of abstractions. We can revert the commit if people don't like it.

Psycojoker:
Okay, also refactored dyndns-update and remove what looked like a debug `return` statement (that would have prevent dyndns-update to work)

Hope this make the code better.

I didn't touch dns-conf since the code looks pretty obvious.

alexAubin:
Sounds good !

alexAubin:
Made a few tests and fixes with @Psycojoker after deploying the changes in the Dynette. Also made a few tests that the previous code still works and that the transition is transparent for users.

We're good to go !

PR #301, opened 2017-04-24
[fix] Don't attempt to change hostname in LXC

alexAubin:
### Problem

For some reason, `hostnamectl` can't be used to change hostname inside LXC's. This is troublesome for bicyclette because [the postinstall crashes because of this](http://bicyclette.netlib.re) and therefore is quite useless.

### Solution

Use some [black magic](https://stackoverflow.com/questions/20010199/determining-if-a-process-runs-inside-lxc-docker) to check if we're running inside a LXC. I wrapped this in a small  `_is_inside_container()` helper

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -164,14 +164,40 @@ def tools_maindomain(auth, new_domain=None):
         logger.warning("%s" % e, exc_info=1)
         raise MoulinetteError(errno.EPERM, m18n.n('maindomain_change_failed'))
 
-    # Clear nsswitch cache for hosts to make sure hostname is resolved ...
+    _set_hostname(new_domain)
+
+    # Generate SSOwat configuration file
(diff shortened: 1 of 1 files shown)


src/yunohost/tools.py
```diff
@@ -189,17 +215,21 @@ def tools_maindomain(auth, new_domain=None):
...
-        pass
+def _is_inside_container():
+    """
+    Check if we're inside a container (i.e. LXC)
 
-    logger.success(m18n.n('maindomain_changed'))
+    Returns True or False
+    """
+
+    p = subprocess.Popen("sudo grep -qa container=lxc /proc/1/environ".split(),
```
alexAubin: >yunohost can run in docker today

Well, in the last discussions I saw about docker, Yunohost was dropping support of docker...

But I implemented the solution in d67b391 anyway.

PR #302, opened 2017-04-26
[enh] Fix curl helper without POST data

maniackcrudelis:
- Allow to use this helper without any POST data
- Keep curl more quiet


scith:
I still can't get this helper to work... It always result in the localhost not found.
Does it work on your apps?

maniackcrudelis:
As marked at "Ready to merge", and we can assume that's only a minor decision. It's just a small correction of en existing helper.

So I merge it.

PR #303, opened 2017-04-30
[enh] Add a custom --yunodebug option for pytest

alexAubin:
### Problem

Right now it's difficult to have an equivalent to --debug when running pytest. You need to tweak the conftest.py manually.

### Solution

Add a custom --yunodebug option to pytest. (Cannot use --debug because it's already a pytest option)

### DIFF
--- a/src/yunohost/tests/conftest.py
+++ b/src/yunohost/tests/conftest.py
@@ -3,6 +3,10 @@ import moulinette
 
 sys.path.append("..")
 
+
+def pytest_addoption(parser):
+    parser.addoption("--yunodebug", action="store_true", default=False)
+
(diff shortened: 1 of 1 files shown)


src/yunohost/tests/conftest.py
```diff
@@ -99,5 +106,3 @@ def _init_moulinette():
     # Initialize moulinette
     moulinette.init(logging_config=logging, _from_source=False)
 
-
-_init_moulinette()
```
opi: no longer needed ?

alexAubin: Nope, `pytest_cmdline_main` is automatically called by pytest

PR #305, opened 2017-05-07
[fix] ynh_setup_source: use YNH_CWD instead of YNH_EXECUTION_DIR

JimboJoe:
Following this discussion: https://github.com/YunoHost/yunohost/pull/282#issuecomment-299637999
(not tested)


zamentur:
Thanks :)

PR #306, opened 2017-05-08
Fix ynh_local_curl

JimboJoe:
Weirdly, POST arguments aren't given to the request if quotes are used...
I'm puzzled as @maniackcrudelis has certainly tested this helper.
Also @scith wasn't able to make it work, but I don't know if this can be the cause.


maniackcrudelis:
I just tried this helper, with and without double quotes.
It works fine in both cases.

So if you have some problems with this double quotes, kick them out.

PR #308, opened 2017-05-08
[enh] Add a script to test m18n keys usage

alexAubin:
### Problem

Remembering which i18n/m18n keys are defined and their names is a pain in the ass. Also it's easy to forget about testing or defining some keys while focusing on writing actual code.

### Solution

I wrote a draft of script to analyze the code to look for m18n keys which are used. This is not perfect though, as sometimes m18n calls are not simply `m18n.n("foo")`, sometimes the key is computed on the fly (using concatenation, or conditions). Also it doesn't test that the arguments like `{bar:s}` are correctly passed and with the right format (string, integer, ...). But it's a start.

Right now it returns the following : 

```
> Error ! Those keys are used in some files but not defined :
   - ask_path
   - backup_archive_writing_error
> Warning ! Those keys are defined but seems unused :
   - app_change_url_no_script
   - ask_current_admin_password
   - diagnostic_monitor_network_error
   - dnsmasq_isnt_installed
   - domain_zone_exists
   - domain_zone_not_found
   - invalid_url_format
   - new_domain_required
   - no_appslist_found
   - no_restore_script
   - package_not_installed
   - package_unexpected_error
   - path_removal_failed
   - service_conf_would_be_updated
   - service_regenconf_dry_pending_applying
```

### DIFF
--- a/tests/_test_m18nkeys.py
+++ b/tests/_test_m18nkeys.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+
+import re
+import glob
+import json
+import yaml
+
+###############################################################################
+#   Find used keys in python code                                             #
+###############################################################################
+
+# This regex matches « foo » in patterns like « m18n.n(  "foo" »
+p = re.compile(r'm18n\.n\(\s*[\"\']([a-zA-Z1-9_]+)[\"\']')
+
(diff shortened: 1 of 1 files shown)


opi:
7282a170a649fbf375a206cc66ea9d365483df22 is wrong, and out of scope

alexAubin:
I removed 7282a17

PR #311, opened 2017-05-11
Allow regen-conf to manage new files

alexAubin:
### Problem

This is an attempt to fix https://dev.yunohost.org/issues/915 and the more general issue, which is that regen-conf won't automatically manage/"take over" new files. This is a tricky issue so I guess it's easier to understand with an example : 

Right now, the new changes in the way dnsmasq is managed imply that the regen-conf hook of dnsmasq will now try to manage `/etc/default/dnsmasq` (and a few other files) which are already present in the system, but not known by regen-conf (no hash in `/etc/yunohost/services.yml)`.

Therefore, if an user upgrade to 2.6, these files won't be updated/managed as expected and the user ends up with an inconsistent system (e.g. resolv.conf points to dnsmasq, but dnsmasq is not properly configured), unless the user runs a `yunohost service regen-conf dnsmasq --force` himself.

Also, the message 
>The configuration file `/foo/bar` is not managed yet and won't be updated

is quite ambiguous about what's the situation precisely. It can be shown in both cases : 
a) because `/foo/bar` already exists but should be deleted according to the regen-conf hook (e.g. nginx don't like a custom conf `/etc/nginx/conf.d/custom.domain.tld.conf`). But it works anyway so it shouldn't be removed automatically.
b) because `/foo/bar` already exists but the regen-conf wants to "take it over" and manage it now (e.g. the present case with dnsmasq). But it doesn't want to do this automatically.

On some of my instance, I actually have some files such as `/etc/nginx/conf.d/yunohost_admin.inc` (or whatever the actual name) which are not up to date, because somehow they were not managed at the time I installed Yunohost (in 2.2 or 2.3 maybe ?) and therefore falls into the "b" case. Therefore my system is not up to date unless I manually check what's happening and force the regen-conf for nginx. (but at the same time, I might have some file in the 'a' category and therefore the --force will destroy stuff which I want to keep).

### Solution / proposal

The proposal is to assume that it is fine for the regen-conf hooks to take over a configuration file that already exist (if it's not to remove it), c.f. case "b" before. The only inconvenience is that we cannot know if this file was manually edited by the user instead of just being the system default. But that's probably a minority of users, and in any cases the files modified by regen-conf are automatically backuped in `/home/yunohost.conf/backup/`. So if they need to, they can get their files back from there.


src/yunohost/service.py
```diff
@@ -391,15 +391,21 @@ def _pre_call(name, priority, path, args):
...
+                    logger.warning(m18n.n('service_conf_new_managed_file',
+                                          conf=system_path, service=service))
+                    regenerated = _regen(system_path, pending_path)
+                    conf_status = 'new'
+                elif force:
                     regenerated = _regen(system_path)
                     conf_status = 'force-removed'
-                elif force:
-                    regenerated = _regen(system_path, pending_path)
-                    conf_status = 'force-updated'
```
opi: `force-updated` status no more exists ?

alexAubin: Hmyeah, now it's what I called "new". Basically what was "force-updated" before is now done by default (without having to be "forced"), c.f. what I wrote in the "Proposal" of the initial post.

opi:
is it a correct output ? 
```
root@yunohost:/vagrant# yunohost service regen-conf dnsmasq
Warning: The configuration file '/etc/default/dnsmasq' is now managed by the service dnsmasq.
Warning: The configuration file '/etc/dnsmasq.conf' is now managed by the service dnsmasq.
dnsmasq: 
  applied: 
    /etc/default/dnsmasq: 
      status: new
    /etc/dnsmasq.conf: 
      status: new
    /etc/resolv.dnsmasq.conf: 
      status: created
  pending: 
```

alexAubin:
Yup, that's what is expected

alexAubin:
Will merge in a few days if no opposition

PR #315, opened 2017-05-30
[fix] Do not break if disk is not a dict. #919

opi:


opi:
Related issue https://dev.yunohost.org/issues/919

alexAubin:
That looks kinda ok to me, but is it expected to get "simfs: not-available" (c.f. the issue on github) ?

opi:
I don't really know what we can expect from Glances. But it seems legit to have "name: value_as_string" sometimes instead of "name: {foo:bar, baz: buz}"

PR #319, opened 2017-06-06
[enh] Do not change the db_pwd

maniackcrudelis:
For restore script, do not change the password.
Keep the primary password and recreate a new database with it.


Psycojoker:
@opi Yes http://tldp.org/LDP/abs/html/parameter-substitution.html

PR #320, opened 2017-06-13
[Fix] Forgot an asterisk in logrotate helper

maniackcrudelis:


maniackcrudelis:
[Minor decision](https://github.com/YunoHost/project-organization/blob/master/yunohost_project_organization_fr.md#d%C3%A9cision-mineure), close on june 20, or 16 if anticipated.

PR #322, opened 2017-06-20
[fix] Move archivemount to suggested dependency

alexAubin:
### Problem

Archivemount depends on fuse, which cannot be installed on some specific machines likes OpenVZ containers (see https://dev.yunohost.org/issues/942). This make `apt-get install yunohost` crash because fuse can't be configured (and also the update to 2.6 / current testing).

### Proposed solution

- *[This PR]* Move archivemount as a suggested dependency. Therefore it won't be installed automatically.
- *[To be done in another PR]* In the install script, attempt to install fuse (or check fuse can be installed) and install archivemount if this works
- *[To be done later]* For the migration 2.5 -> 2.6, tell people to install archivemount manually if they want to have an optimized 'restore'. (Ideally this could be added as a migration once the migration framework is there)

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -26,7 +26,6 @@ Depends: ${python:Depends}, ${misc:Depends}
  , ssowat, metronome
  , rspamd (>= 1.2.0), rmilter (>=1.7.0), redis-server, opendkim-tools
  , haveged
- , archivemount
 Recommends: yunohost-admin
  , openssh-server, ntp, inetutils-ping | iputils-ping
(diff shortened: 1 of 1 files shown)


zamentur:
After some unconcluant tests to try to simulate a package with archivemount as recommended, I suggest to merge this PR.

Note if the admin as configure the system to setup all suggested packages, it will fail. But by default it's not the case: `APT::Get::Install-Suggests "false";` in /etc/apt.conf.

PR #325, opened 2017-07-01
[fix] "backup restore" crashes when archivemount is not installed

alexAubin:
### Problem

As reported in https://dev.yunohost.org/issues/953, `yunohost backup restore` crashes when archivemount is not installed.

### Solution

Simple quick and dirty fix is to wrap the call in a try/except.

### How to test

```bash
yunohost backup create
yunohost backup list
yunohost backup restore backup_archive_name
```

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -1776,8 +1776,13 @@ class TarBackupMethod(BackupMethod):
         tar.close()
 
         # Mount the tarball
-        ret = subprocess.call(['archivemount', '-o', 'readonly',
-                               self._archive_file, self.work_dir])
+        try:
+            ret = subprocess.call(['archivemount', '-o', 'readonly',
+                                   self._archive_file, self.work_dir])
+        except:
+            ret = -1
+
+        # If archivemount failed, extract the archive
         if ret != 0:
             logger.warning(m18n.n('backup_archive_mount_failed'))
 


src/yunohost/backup.py
```diff
@@ -1776,8 +1776,13 @@ def mount(self, restore_manager):
         tar.close()
 
         # Mount the tarball
-        ret = subprocess.call(['archivemount', '-o', 'readonly',
-                               self._archive_file, self.work_dir])
+        try:
+            ret = subprocess.call(['archivemount', '-o', 'readonly',
```
opi: if archivemount is not there, subprocess.call() does not simply return something != 0 ?

alexAubin: Unfortunately no, and that's where the bug comes from : it raises an exception in that case with a message such as "Command or file not found". C.f. the original Redmine issue

opi: stupid ;)

maniackcrudelis:
Hum, maybe my question is totally foolish.
But, what in case of upgrade ? Because currently, we have a lot of yunohost where the restore simply doesn't work for now.

alexAubin:
>But, what in case of upgrade ? Because currently, we have a lot of yunohost where the restore simply doesn't work for now.

What do you mean in case of upgrade ? Why doesnt the restore work ? Is it because of this bug ? o.O

maniackcrudelis:
Unless I made a mistake, yes, it's this bug with archivemount which break the restore. With a crash of python.
The official CI has this bug, and many of my virtual machine.
But, we don't have this bug on our testing and unstable container. Only on stable.

alexAubin:
Okay, but the purpose of this PR is precisely to fix this bug ?

PR #328, opened 2017-07-13
[fix] Logrotate append

maniackcrudelis:
Add a --non-append option to avoid to add the new config at the end of the existing config file.
In case of upgrade, it's a real problem because it will duplicates the same entry.
https://forum.yunohost.org/t/installation-etherpad-mypads/2890/48

- [ ] Complete test
- [ ] Simple test
- [ ] Code review
- [ ] Approval (LGTM)
- [ ] Approval (LGTM)

[Minor decision](https://github.com/YunoHost/project-organization/blob/master/yunohost_project_organization.md#minor-decision), close on 20 of june, or 16 of june.

I've made a complete test, with or without the arguments. And with the new argument at the first and the second place.

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -2,15 +2,15 @@
 #
 # usage: ynh_use_logrotate [logfile] [--non-append]
 # | arg: logfile - absolute path of logfile
-# | option: --non-append - Replace the config file instead of append this new config.
+# | option: --non-append - Replace the config file instead of appending this new config.
 #
 # If no argument provided, a standard directory will be use. /var/log/${app}
 # You can provide a path with the directory only or with the logfile.
 # /parentdir/logdir/
 # /parentdir/logdir/logfile.log
 #
-# It's possible to use this helper several times, each config will added to same logrotate config file.
-# Unless if you use the option --non-append
+# It's possible to use this helper several times, each config will be added to the same logrotate config file.
+# Unless you use the option --non-append
 ynh_use_logrotate () {
 	local customtee="tee -a"
 	if [ $# -gt 0 ] && [ "$1" == "--non-append" ]; then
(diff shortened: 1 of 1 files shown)


data/helpers.d/backend
```diff
@@ -1,16 +1,26 @@
 # Use logrotate to manage the logfile
 #
-# usage: ynh_use_logrotate [logfile]
+# usage: ynh_use_logrotate [logfile] [--non-append]
 # | arg: logfile - absolute path of logfile
+# | option: --non-append - Replace the config file instead of append this new config.
```
JimboJoe: instead of append**ing**

data/helpers.d/backend
```diff
@@ -1,16 +1,26 @@
...
 #
 # If no argument provided, a standard directory will be use. /var/log/${app}
 # You can provide a path with the directory only or with the logfile.
 # /parentdir/logdir/
 # /parentdir/logdir/logfile.log
 #
 # It's possible to use this helper several times, each config will added to same logrotate config file.
```
JimboJoe: will **be** added to **the**

data/helpers.d/backend
```diff
@@ -1,16 +1,26 @@
...
+# Unless if you use the option --non-append
```
JimboJoe: unless ~~if~~ you use

data/helpers.d/backend
```diff
@@ -40,7 +50,7 @@ $logfile {
 }
 EOF
 	sudo mkdir -p $(dirname "$logfile")	# Create the log directory, if not exist
-	cat ${app}-logrotate | sudo tee -a /etc/logrotate.d/$app > /dev/null	# Append this config to the others for this app. If a config file already exist
+	cat ${app}-logrotate | sudo $customtee /etc/logrotate.d/$app > /dev/null	# Append this config to the others for this app. If a config file already exist
```
JimboJoe: Here you must change the comment : "Append this config to the existing config file, or replace the whole config file (depending on $customtee)"

PR #329, opened 2017-07-14
Run change_url scripts as root as a matter of homogeneity

JimboJoe:
Due to parallel development (I guess) of change-url feature and script execution by root, change_url scripts are not run as root at the moment.
Here is a fix proposal; it can be tested with [this PR](https://github.com/YunoHost/test_apps/pull/1) on the test set.


Psycojoker:
Looks good, I think we can merge this one soon.

PR #330, opened 2017-07-14
[enh] Check that user is legitimate to use an email adress when sending mail

anmol26s:
Mail can only be sent if the sender is owner of the address or have the mail alias linked to his account.

Issue: https://dev.yunohost.org/issues/891#change-3384

### DIFF
--- a/data/templates/postfix/main.cf
+++ b/data/templates/postfix/main.cf
@@ -72,6 +72,7 @@ virtual_alias_domains =
 virtual_minimum_uid = 100 
 virtual_uid_maps = static:vmail 
 virtual_gid_maps = static:mail
+smtpd_sender_login_maps= ldap:/etc/postfix/ldap-accounts.cf
 
 # Dovecot LDA 
 virtual_transport = dovecot 
@@ -113,7 +114,8 @@ smtpd_helo_restrictions =
     permit 
  
 # Requirements for the sender address 
-smtpd_sender_restrictions = 
+smtpd_sender_restrictions =
+    reject_sender_login_mismatch, 
     permit_mynetworks, 
     permit_sasl_authenticated, 
     reject_non_fqdn_sender, 


alexAubin:
Sounds interesting ! :+1: :tada: 

I'll have a look !

alexAubin:
Hi there,

I made the following test so far (with PR activated) : 

- add johnoe the first user (thus has alias to root@the.domain.tld)
- add janedoe as the second user (no alias to root@the.domain.tld)
- login as janedoe in a terminal / ssh
- use `mailx -r root@the.domain.tld recipient@some.otherdomain.tld` to send an email from Jane but as root.
- postfix don't say a word about it :/

Though I understand that using mail/mailx is unauthenticated submissions so maybe this PR only applies to authenticated submission (i.e. the SSL/TLS thing which is what Roundcube/Rainloop/Thundebird/whatever mail client use).

alexAubin:
Tested on Rainloop and I confirm that (using same scenario than in previous comment) : 

Without the patch : 
- I could send a user as root from rainloop from both johndoe and janedoe (i.e. I reproduce the original issue)

With the patch : 
- I can send an email as root from rainloop from johndoe, but not janedoe

To me that's good to be merged 👍

anmol26s:
Its good that the test passes on rainloop but on roundcube the patch was bypassed and I had to disable the option of adding new identities in the roundcube. On all other desktop and android clients I have tested,it was working.

alexAubin:
Do you have an understanding of why it doesnt work out of the box for roundcube ? o.O

anmol26s:
I have no idea , but it don't work with roundcube. Will have to see how the email are authenticated with roundcube. Though it works with other email clients  I have tested. You can go with the pull. The creation of new identities in the roundcube could be disabled as a workaround.

alexAubin:
Will merge this in a few days, feel free to complain if you want to review / test / change this.

anmol26s:
Thanks

PR #331, opened 2017-07-16
[fix] Attempt to fix Sender Rewriting Scheme with postsrsd

alexAubin:
### Problem

Right now, the Sender Rewriting Scheme implementation is a bit weird (see the old regex) and reportedly leaks metadata because the Return Path header is set to sender@main.domain (see https://dev.yunohost.org/issues/626). 

After investigating with fr0g, we realized that even though SRS is usually used to satisfy the SPF when forwarding emails (see 'Crash course' in [this readme](https://github.com/roehling/postsrsd/blob/master/README.md)), the current implementation in postfix does not seem to do what's recommended though (see the same paragraph in README).

### Solution

Several tutorial are poiting to postsrsd to do the job. It also allow to not apply SRS for some domains. So excluding the local domains avoid changing the Return Path to sender@main.domain (e.g. keep the original domain even if it's not the main domain).

So far, I have not been able to actually check how it affects the spaminess of mails though ... (this requires complicated setups because the whole SPF issue arrives in the context of forwards or mailing lists for instance)

### How to test

Pull the branch in. Make sure you have two domains A and B on your instance with (ideally) correct mail DNS records. A is the main domain. Send a mail from B to some external email (e.g. yahoo, gmail, whatever) and check the source. Without the patch, if you inspect the source, you should see `Return-Path: user@domain.A` (i.e. leaking existence of domain A to the recipient) whereas with the patch it's just 'Return-Path: user@domain.B' (i.e. not leaking stuff).

In more complicated setups though (i.e. sending a mail to user@domain.B which is then refowarded to user@gmail.com for instance) you might see something like : `Return-Path: SRS0=RQUm=6T=gmail.com=johndoe@domain.A`. This leaks the existence of 'domain A' to user (of domain B). Which is kinda less dramatic than the previous case imho.

### Reviews

- [ ] Agree in principle (0/2)
- [ ] Simple tests (0/1)
- [ ] Quick review (0/1)
- [ ] Deep tests (0/0)
- [ ] Deep reviews (0/0)

### DIFF
--- a/data/templates/postfix/postsrsd
+++ b/data/templates/postfix/postsrsd
@@ -1,0 +1,43 @@
+# Default settings for postsrsd
+
+# Local domain name.
+# Addresses are rewritten to originate from this domain. The default value
+# is taken from `postconf -h mydomain` and probably okay.
+#
+SRS_DOMAIN={{ main_domain }}
+
+# Exclude additional domains.
+# You may list domains which shall not be subjected to address rewriting.
+# If a domain name starts with a dot, it matches all subdomains, but not
+# the domain itself. Separate multiple domains by space or comma.
+# We have to put some "dummy" stuff at start and end... see this comment :
+# https://github.com/roehling/postsrsd/issues/64#issuecomment-284003762
+SRS_EXCLUDE_DOMAINS=dummy {{ domain_list }} dummy
+
+# First separator character after SRS0 or SRS1.
+# Can be one of: -+=
+SRS_SEPARATOR==
+
+# Secret key to sign rewritten addresses.
+# When postsrsd is installed for the first time, a random secret is generated
+# and stored in /etc/postsrsd.secret. For most installations, that's just fine.
+#
+SRS_SECRET=/etc/postsrsd.secret
+
+# Local ports for TCP list.
+# These ports are used to bind the TCP list for postfix. If you change
+# these, you have to modify the postfix settings accordingly. The ports
+# are bound to the loopback interface, and should never be exposed on
+# the internet.
+#
+SRS_FORWARD_PORT=10001
+SRS_REVERSE_PORT=10002
+
+# Drop root privileges and run as another user after initialization.
+# This is highly recommended as postsrsd handles untrusted input.
+#
+RUN_AS=postsrsd
+
+# Jail daemon in chroot environment
+CHROOT=/var/lib/postsrsd
+
--- a/data/hooks/conf_regen/19-postfix
+++ b/data/hooks/conf_regen/19-postfix
@@ -10,14 +10,24 @@ do_pre_regen() {
   postfix_dir="${pending_dir}/etc/postfix"
   mkdir -p "$postfix_dir"
 
+  default_dir="${pending_dir}/etc/default/"
+  mkdir -p "$default_dir"
+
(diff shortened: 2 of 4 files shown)


alexAubin:
As found by @opi, postsrsd is not available in Jessie. Will need to wait stretch to include this (or else we have to git clone it or something)

lunarthegrey:
Any update on this? Really makes mail funky on YunoHost.

alexAubin:
Well we still need to migrate to Stretch before being able to merge this unfortunately :/

But @frju365 worked on another proposal here : https://github.com/YunoHost/yunohost/pull/374 though I didn't have time yet to test it and don't have much expertise with mail to know if that's the right way to proceed or not.

data/templates/postfix/postsrsd
```diff
@@ -0,0 +1,43 @@
+# Default settings for postsrsd
+
+# Local domain name.
+# Addresses are rewritten to originate from this domain. The default value
+# is taken from `postconf -h mydomain` and probably okay.
+#
+SRS_DOMAIN={{ main_domain }}
```
frju365: could you test again this syntax for the domain : because for me, I get an error : domain1.tld: NOT FOUND

alexAubin: I just retested this tonight and for me it works ... where and when exactly do you see this error popping up ?

alexAubin:
So I retested this tonight and this works for me, i.e. sending an email from domain B doesnt procude some weird Return Path with domain A in it.

I haven't retested the forwarded-mail case yet because that requires a more complex setup.

Would be great to have some more feedback on this

Psycojoker:
There are weird conflicts that needs to be solved :/

frju365:
Can we merge this PR to test ?

zamentur:
Take care there is a dkim issue if domainB is a subdomain of domainA...

alexAubin:
So I retested this again today (hopefully for the last time)

I found that it indeed seems to kinda fix the whole SRS issue (the real fix being to not use email anymore anyway >.>).

For "normal" mail sending : when sending from "secondary" domains, there is no more `Return-Path` weirdly leaking the primary domain.

For "forwarded" mail : I can see SRS is also applied. When forwarding through a secondary domain, I see the following header added : 
```
Return-Path: <SRS0=CxUa=KJ=domain.forwarded.to=forwardeduser@primary.domain.tld>
```
which is what is expected as far as I understand. Modulo the fact that it leaks the primary domain ... but digging into it, there doesn't seem to be a proper "multi domain" implementation of SRS anyway (This is a complicated issue, c.f. 'Multi domain support' and 'Dynamic SRS_DOMAIN' issues in https://github.com/roehling/postsrsd/issues ). Apart from this, it should be closer to "the right way" to handle this, if there's any...

About forwarded mail, I was not however able to check through mail-tester what is the impact of the email score, as they seem to refuse forwarded mail or being "in CC".

I'd be happy to merge this in the coming days/weeks, but I also know that this is a complicated and technical PR é_è. But in case you plan to give some feedback about this, please leave a message so at least we know that we should wait for feedback.

HugoPoi:
Nice work, mail going better and better on yunohost !

PR #332, opened 2017-07-17
[fix] Don't verify SSL during changeurl tests :/

alexAubin:
### Problem

See https://github.com/YunoHost/test_apps/pull/1 : for some reason, when running unit tests for changeurl, python-requests doesnt want to validate the request because of SSL error. Not sure what this is coming from... I think it was working before. Maybe some update in the python-requests version (I've seen some similar weird stuff on tartare where python-requests don't want to validate a requests to dev.yunohost.org because of SSL)

### Proposed solution

Disable the SSL verification for the request 😕


Psycojoker:
That's a bit dirty, but since it's in test I guess it's ok :confused:

PR #334, opened 2017-07-19
[fix] Backup daily cron (Redmine 933)

zamentur:
## Problem
Currently yunohost cron in cron.daily are not backup, only cron.d

## Solution
Backup all cron in cron.* and begining by yunohost

## How to test
```bash
yunohost backup create --ignore-apps --hooks conf_cron -n cronbkp --debug --verbose
rm /etc/cron.*/yunohost*
yunohost backup restore cronbkp --debug --verbose
ls -1B -l /etc/cron.*/yunohost*
```
Check permissions and owner

## PR Status
Untested. 

## Validation

- [ ] **Principle agreement** 1/2 : Bram
- [x] **Quick review** 1/1 : Bram
- [ ] **Simple test** 0/1 : 
- [ ] **Deep review** 0/1: 
- [ ] **Complex test** 0/0

### DIFF
--- a/data/hooks/backup/32-conf_cron
+++ b/data/hooks/backup/32-conf_cron
@@ -10,6 +10,6 @@ source /usr/share/yunohost/helpers.d/filesystem
 backup_dir="${1}/conf/cron"
 
 # Backup the configuration
-for f in $(ls -1B /etc/cron.d/yunohost*); do
-  ynh_backup "$f" "${backup_dir}/${f##*/}"
+for f in $(ls -1B /etc/cron.*/yunohost*); do
+  ynh_backup "$f" "${backup_dir}${f}"
 done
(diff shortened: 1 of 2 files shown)


alexAubin:
I don't know if it's relevant to backup these :confused: 

These cron jobs are not data nor configuration. It was relevant to backup them in the past before we changed the applist system, because at that time, the cron jobs *were* the configuration of which list where fetched. But now they don't contain any configuration info (and cron job should not represent / include configuration info anyway imho).

Therefore, these cron jobs should be added by yunohost itself (which is already done in the postinstall) and imho shouldnt be backuped at all... In the future, we might have more of them, or less of them, or the content might be changed, or they might be managed differently (systemd lel ?). So in fact, by backuping these, we risk situations where in the future some archive will restore cron jobs which are inconsistent with newer versions...

zamentur:
I close this pull request following the meeting

PR #335, opened 2017-07-23
[enh] Add optionnal recovery password when subscribing to nohost.me/noho.st

alexAubin:
### Current status

Work in progress

- [X] Optionnal password when running explicitely with `yunohost dyndns subscribe`
- [ ] Same during CLI postinstall ?
- [ ] Same during web postinstall ?
- [ ] Be able to use the password to actually recover a domain which is not available ?

### Problem

Related issue : https://dev.yunohost.org/issues/6
We want people to be autonomous with their .nohost.me / .noho.st, thus we should have a possibility for users to set a 'recovery password' which they can remember to get back control of their domain if they lost it (e.g. because nobody saves the public/private key when reinstalling an instance)

### Solution

This is the client-side implementation of the new API added in https://github.com/YunoHost/dynette/pull/4. During a dyndns subscribe, the user can define a recovery password for its domain. The password can then later be used to delete the domain from the dynette.


src/yunohost/dyndns.py
```diff
@@ -87,6 +90,29 @@ def dyndns_subscribe(subscribe_host="dyndns.yunohost.org", domain=None, key=None
     except requests.ConnectionError:
         raise MoulinetteError(errno.ENETUNREACH, m18n.n('no_internet_connection'))
 
+    # If no password specified in CLI, ask for one to the user
+    if recovery_password is None and msettings.get('interface') == "cli":
```
Psycojoker: All of those things are actually builtin the moulinette (but sadly! undocumented and not well known :() including password size, we can simplify this code a lot.
https://github.com/YunoHost/yunohost/blob/unstable/data/actionsmap/yunohost.yml#L135-L139

src/yunohost/dyndns.py
```diff
@@ -87,6 +90,29 @@ def dyndns_subscribe(subscribe_host="dyndns.yunohost.org", domain=None, key=None
...
+        msignals.display(m18n.n("dyndns_choose_recovery_password", domain=domain))
+        recovery_password = msignals.prompt(m18n.n("ask_password"), True)
+
+        # Confirm password...
+        if recovery_password != msignals.prompt(m18n.n("confirm_password"), True):
+            raise MoulinetteError(errno.EINVAL, m18n.n('passwords_dont_match'))
+
+    # TODO / FIXME : maybe add constrains on password size?
+
+    if recovery_password == "":
```
Psycojoker: Should do a `recovery_password.strip()`

alexAubin:
[Semi-automatic message] 

PR #345, opened 2017-08-10
Common ssh keys operations support

Psycojoker:
Hello,

The idea of this work is to integrate all common ssh keys operation into yunohost.

2 mains interests from here are:

- `yunohost ssh user enable-ssh` to allow a user to login with ssh
- provide all those operations in the web interface for people that don't know how to do that in cli (or are lazy)

I know that ssh key generation is also needed for federation or backup stuff.

This is a full working prototype, there is probably cleaning that needs to be done in addition to more testing and probably edge cases and some TODOs.

New commands (so we can bikeshed on the command names :yum:):

- `yunohost ssh user list` list all users with ssh related informations (access is enabled for eg)
- `yunohost ssh user allow-ssh $username` allow a user to login in ssh
- `yunohost ssh user disallow-ssh $username` inverse
- `yunohost ssh root-login status` show if root can log in ssh (`PermitRootLogin` stuff)
- `yunohost ssh root-login enable/disable`
- `yunohost ssh key list $username` list ssh keys of the users (only `.pub` files for now, don't know if that should include private files)
- `yunohost ssh key add $username` (`-a` to specify algo, `-n` for a key name) generate a new ssh key pair using `ssh-keygen`
- `yunohost ssh key import $username -u $public_key -i $private_key` import a ssh key pair (from another computer for eg)
- `yunohost ssh key remove $username -n $key_name`
- `yunohost sh authorized-keys list $username` parse and list the content of `~/.ssh/authorized_keys` (assume that a comment before a key is the title of the key)
- `yunohost ssh authorized-keys add $username -k $key` (`-c` to add a comment) add a key (with a comment before) to `~/.ssh/authorized_keys`
- `yunohost ssh authorized-keys remove $username -k $key`

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1418,6 +1418,9 @@ ssh:
                             help: Username of the user
                             extra:
                                 pattern: *pattern_username
+                        -n:
+                            full: --name
+                            help: Key name
                         -a:
                             full: --algo
                             help: Algorithm to generate the ssh key
@@ -1427,6 +1430,7 @@ ssh:
                                 - ed25519
                                 - rsa
                                 - rsa1
+                            default: rsa
 
                 ### ssh_rootlogin_enable()
                 import:
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1448,6 +1448,9 @@ ssh:
                             help: Key as a string
                             extra:
                                 required: True
+                        -c:
+                            full: --comment
+                            help: Key comment
 
                 ### ssh_rootlogin_disable()
                 remove:
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1506,6 +1506,8 @@ ssh:
                         -n:
                             full: --name
                             help: Key name
+                            extra:
+                                required: True
 
                 ### ssh_rootlogin_disable()
                 remove:
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1418,6 +1418,15 @@ ssh:
                             help: Username of the user
                             extra:
                                 pattern: *pattern_username
+                        -a:
+                            full: --algo
+                            help: Algorithm to generate the ssh key
+                            choices:
+                                - dsa
+                                - ecdsa
+                                - ed25519
+                                - rsa
+                                - rsa1
 
                 ### ssh_rootlogin_enable()
                 import:
@@ -1430,6 +1439,11 @@ ssh:
                             help: Username of the user
                             extra:
                                 pattern: *pattern_username
+                        -k:
+                            full: --key
+                            help: Key as a string
+                            extra:
+                                required: True
 
                 ### ssh_rootlogin_disable()
                 remove:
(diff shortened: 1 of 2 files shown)


zamentur:
"yunohost ssh key add $username" => I suggest to rename as "yunohost ssh key create $username" .

Example of usage for "yunohost authorized-keys add":
yunohost authorized-keys add $username -k "ssh-ed25519 AAAAC3N...752CyXG"
yunohost authorized-keys add $username -k 'command="borg serve --restrict-to-path /home/USER/backup",no-pty,no-agent-forwarding,no-port-forwarding,no-X11-forwarding,no-user-rc ssh-ed25519 AAAAC3...J4RQ7ryaUi root@example.com'

Psycojoker:
> "yunohost ssh key add $username" => I suggest to rename as "yunohost ssh key create $username" .

I understand why you want that (and that seems indeed more intuitive) but we already uses `add` nearly everywhere, except for user creation (and backup apparently) and that create an incoherence in the CLI interface.

I would prefer that we either uses `add` or `create` everywhere to keep the same semantic as often as possible (and `add` seems more generic than `create`), I think that this will improve the CLI API experience if only a small set of words need to be remembers and are used everywhere.

alexAubin:
Risking to sound like the annoying guy here :confused: .. But I have these three points of concern about the general design : 

- I'm really not comfortable with the idea that the root/admin would or could manage his users private/public key to *others* servers (i.e. what the `yunohost users key` intends to). I don't really see a use case like "admin of server A wants to add a key to user Alice such that she can connect to unrelated server B" ? Except in very specific cases, imho, a user should never store a private key that gives access to another server .. on a server he doesn't own/administrate himself. And if he does, that should be his own business because there should at least be a passphrase associated to the private key.
- I'd really like if we could avoid to create a new category. Imho it would be more logical to add a `ssh` subcategory to `user`, to have `yunohost user ssh` with actions like `allow`/`disallow`, `add-key`/`remove-key`/`list-keys` (implicitly being authorized keys, because imho that should be the only kind of key the admin manages about his users :/)
- Move the root-login toggle to a global settings.

Psycojoker:
> I'm really not comfortable with the idea that the root/admin would or could manage his users private/public key to others servers (i.e. what the yunohost users key intends to). I don't really see a use case like "admin of server A wants to add a key to user Alice such that she can connect to unrelated server B" ? Except in very specific cases, imho, a user should never store a private key that gives access to another server .. on a server he doesn't own/administrate himself. And if he does, that should be his own business because there should at least be a passphrase associated to the private key.

I understand why you are not comfortable with that because the usages you have in mind are indeed bad, but beside those I think that they are several legitimate usages worth it:

* I want to manage the keys of my own user on my server
* I need to generate a key for a user to give him access to another server and I want to have control on that key (for automated process for example)
* I need keys for automated process like backup
* I use several users on my own for several different activities and I want to easily managed all those keys

And already doing bad operation is so freaking easy that I don't see how integrating ssh management in core would make any difference in that regard and I think that the possibility to do those operations is worth it.

All of that make me think that integrating something like "ssh-copy-id" could be great but that will be for another PR.

> I'd really like if we could avoid to create a new category. Imho it would be more logical to add a ssh subcategory to user, to have yunohost user ssh with actions like allow/disallow, add-key/remove-key/list-keys

I have no strong opinion on that, I think that the idea of not wanting to add new categories is not based on solid argument but more on "what if one day" and I generally think that having several categories is better than merging everything together because you have a way cleaner organisation and way more easier discovery of what possibilities are available.

But if it's important for you why not.

> Move the root-login toggle to a global settings.

Right, I actually wanted to do that but totally forgot because I didn't wrote down what I wanted to do /o\

Psycojoker:
I've started splitting this PR into https://github.com/YunoHost/yunohost/pull/403 and https://github.com/YunoHost/yunohost/pull/404

alexAubin:
[Semi-automatic message] 

PR #346, opened 2017-08-13
[fix] Tell user that domain dns-conf shows a recommendation only

alexAubin:
### Problem

Users who use `yunohost domain dns-conf` sometimes think that this shows some sort of actual configuration. We should tell them explicitly that this is only the recommended configuration which should be configured in the registrar.

Related issue on redmine : https://dev.yunohost.org/issues/559

### Solution

For the CLI, display a warning. For the webadmin, display a warning block (see [related PR](https://github.com/YunoHost/yunohost-admin/pull/168))

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -141,6 +141,7 @@
     "domain_creation_failed": "Unable to create domain",
     "domain_deleted": "The domain has been deleted",
     "domain_deletion_failed": "Unable to delete domain",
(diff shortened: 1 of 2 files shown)


Psycojoker:
It might also be a good idea to rename this command into

    yunohost domain generate-dns-conf

That would make it more obvious that it's not a setted conf.

PR #350, opened 2017-08-15
[fix] Use SHA-512 for passwords in LDAP

Psycojoker:
This started as a "let's clean a bit user.py" until I realized that we were using md5 to store users password. In reaction to that, I went rampage and destroyed every occurrence I could find of using bad hashing algorithm and integrated auto upgrade code ON LOGIN of weakly hashed password to sha-512.

We actually managed to impressive exploit of using: MD5, SSHA and SHA1 for password hashing depending from where it's used.

This PR is **important** and security related, we should merge it soon and probably do a security release only for that.

Since password handling is spread across all YunoHost, this PR also depends on https://github.com/YunoHost/moulinette/pull/152 and https://github.com/YunoHost/SSOwat/pull/87

The information we need to transmit to our users is: you need to ask every account on your YunoHost to successfully relog at least once OR chance their password.

I've personally tested every situation possible EXCEPT a fresh new installation.

I've taken care of splitting every commit according to what it does, IT MIGHT BE EASIER TO READ THIS PR COMMIT BY COMMIT.

As a reference, here is a blogpost that explains how to uses "{CRYPT}" the linux auth system instead of LDAP one that is totally lame and only supports at best SHA1 https://www.redpill-linpro.com/techblog/2016/08/16/ldap-password-hash.html This explains this "$6$" thingy.

Things to test:

* user creation
* user modification
* yunohost tools adminpw
* login using sso
* login on the webadmin

Those are the python commands to read and modify the password of a user:

```python
In [9]: auth.search('uid=neutrinet,ou=users,dc=yunohost,dc=org', attrs=["userPassword"])
Out[9]: [{'userPassword': ['{CRYPT}$6$ULY00E6N7eAID$TEZbi3cZArkWp8oXoFw122YHuoXl4wkxkUouyMZ3z..XnUjHiwA63QTsjwKT7YpxWyvYxSRo/WZGBO5OG5bv51']}]

In [10]: auth.update('uid=neutrinet,ou=users', {"userPassword": "{CRYPT}$1$GQVT7JLZ$hkQN0IxWUp8NKp9ummIrN."})
```

And for admin:

```python
In [4]: auth.search("cn=admin,dc=yunohost,dc=org", attrs=["userPassword"])                                                                                                                                       
Out[4]: [{'userPassword': ['{CRYPT}$6$KIiwe9dnaQYBBd0j$tWK0jR8dU4bsms.lCEiQa23NeYzas1bao5V5.5IlmmwM8lcv9aqrxydK1KdBWXDTorqRIKUsl19LrD4srPlAr0']}]

In [3]: auth.update('cn=admin', {"userPassword": "{SSHA}bJumO0uTJcqXp8N+f2wEcTFVLB7oXFLT"})                                                                                                                      
Out[3]: True
```

Both of those set the password to be "neutrinet" (this is my dev vm password, not a prod password).

Those can be used to set the user/admin password to a crappy hash (`SSHA` or `md5`), then you can test:

* login on sso
* login on admin if
* login on a shell command

And see if those modify the crappy hashed password to uses sha-512 (the one starting with `{CRYPT}$6$` that you can see in the `auth.search`).

### DIFF
--- a/src/yunohost/user.py
+++ b/src/yunohost/user.py
@@ -289,10 +289,7 @@ def user_update(auth, username, firstname=None, lastname=None, mail=None,
         new_attr_dict['cn'] = new_attr_dict['displayName'] = firstname + ' ' + lastname
 
     if change_password:
-        char_set = string.ascii_uppercase + string.digits
-        salt = ''.join(random.sample(char_set, 8))
-        salt = '$1$' + salt + '$'
-        new_attr_dict['userPassword'] = '{CRYPT}' + crypt.crypt(str(change_password), salt)
+        new_attr_dict['userPassword'] = _hash_user_password(change_password)
 
     if mail:
         auth.validate_uniqueness({'mail': mail})
--- a/src/yunohost/user.py
+++ b/src/yunohost/user.py
@@ -80,10 +80,9 @@ def user_list(auth, fields=None):
     for user in result:
         entry = {}
         for attr, values in user.items():
-            try:
+            if values:
                 entry[user_attrs[attr]] = values[0]
-            except:
-                pass
+
         uid = entry[user_attrs['uid']]
         users[uid] = entry
 
--- a/src/yunohost/user.py
+++ b/src/yunohost/user.py
@@ -445,6 +445,16 @@ def _convertSize(num, suffix=''):
 
 def _hash_user_password(password):
     char_set = string.ascii_uppercase + string.ascii_lowercase + string.digits + "./"
+    # This 16 number is chosen according to this documentation stating that
+    # this is the maximum number of salt possible
+    # https://www.safaribooksonline.com/library/view/practical-unix-and/0596003234/ch04s03.html
+    #
+    # SystemRandom is the cryptographically secure random method provided by python stl
+    # You can refer to this https://docs.python.org/2/library/random.html for
+    # confirmation (read the red square), it internally uses /dev/urandom
     salt = ''.join([random.SystemRandom().choice(char_set) for x in range(16)])
+
+    # Using "$6$" means that we uses sha-512 which is the strongest hash available on the system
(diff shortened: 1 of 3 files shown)


alexAubin:
>We actually managed to impressive exploit of using: MD5, SSHA and SHA1 for password hashing depending from where it's used.

src/yunohost/user.py
```diff
@@ -118,33 +115,27 @@ def user_create(auth, username, firstname, lastname, mail, password,
     })
 
     # Validate uniqueness of username in system users
-    try:
-        pwd.getpwnam(username)
-    except KeyError:
-        pass
-    else:
+    all_existing_usernames = {x.pw_name for x in pwd.getpwall()}
```
alexAubin: Didn't know this `{ }` list-comprehension-like thing.. Is this for sets/unordered list ?

Psycojoker: It's a set comprehension yes.

src/yunohost/user.py
```diff
@@ -453,3 +441,10 @@ def _convertSize(num, suffix=''):
             return "%3.1f%s%s" % (num, unit, suffix)
         num /= 1024.0
     return "%.1f%s%s" % (num, 'Yi', suffix)
+
+
+def _hash_user_password(password):
+    char_set = string.ascii_uppercase + string.ascii_lowercase + string.digits + "./"
+    salt = ''.join([random.SystemRandom().choice(char_set) for x in range(16)])
```
alexAubin: Maybe we can add a comment here that this is a crypto-secure(?) source of entropy (if you actually checked it is ;))

Psycojoker: Well, we can point out to python official doc if you want yes.

Red square on the page https://docs.python.org/2/library/random.html or https://docs.python.org/2/library/random.html#random.SystemRandom but red square is better I think

PR #351, opened 2017-08-15
Integrate a command to launch a dev shell from within moulinette context

Psycojoker:
This PR allows to do a

    yunohost tools shell

And be drop into an IPython shell (if installed, else python default shell) in the moulinette context for rapid prototyping and debugging (for example to talk to ldap using the `auth` object)

This also allows

    yunohost tools shell -c "command"

À la

    python -c "command"

I've been using this a lot during the sha-512 thingy for rapid debugging.

This is exactly the same thing that Django's `python manage.py shell`. Once you get addicted to it, it become quite life changer.

I'm tempted to pass this as a micro decision since this is trivial and not that much worth it discussing.

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1478,6 +1478,12 @@ tools:
                     extra:
                         pattern: *pattern_port
 
+        ### tools_shell()
+        shell:
+            configuration:
+                authenticate: all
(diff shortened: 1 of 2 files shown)


opi:
Looks good  (holiday review, from a comfy coach in a crowded friend's house)

PR #354, opened 2017-08-23
[enh] Escape some special character in ynh_replace_string

maniackcrudelis:
## Problem
- *Some characters are interpreted by sed, especially in strong password.  
It sometimes fails the install script*

## Solution
- *Escape backslash and &
Both of them are special characters for sed.*

## Validation
*Medium decision*
- [ ] **Complete test** : 
- ~**Upgrade previous version** :~
- [ ] **Code review** : 
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- ~**CI succeeded** :~
- ~**CI succeeded** :~
~One public package check log is enough instead of two checks.~  
When the PR is mark as ready to merge, you have to wait for 7 days before really merge it.

### DIFF
--- a/data/helpers.d/string
+++ b/data/helpers.d/string
@@ -18,8 +18,18 @@ ynh_string_random() {
 # | arg: target_file - File in which the string will be replaced.
 ynh_replace_string () {
 	delimit=@
-	match_string=${1//${delimit}/"\\${delimit}"}	# Escape the delimiter if it's in the string.
+	# Escape the delimiter if it's in the string.
+	match_string=${1//${delimit}/"\\${delimit}"}
 	replace_string=${2//${delimit}/"\\${delimit}"}
+
+        # Escape any backslash to preserve them as simple backslash.
+        match_string=${match_string//\\/"\\\\"}
+        replace_string=${replace_string//\\/"\\\\"}
+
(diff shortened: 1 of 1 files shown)


data/helpers.d/string
```diff
@@ -18,8 +18,18 @@ ynh_string_random() {
...
 ynh_replace_string () {
 	delimit=@
-	match_string=${1//${delimit}/"\\${delimit}"}	# Escape the delimiter if it's in the string.
+	# Escape the delimiter if it's in the string.
+	match_string=${1//${delimit}/"\\${delimit}"}
 	replace_string=${2//${delimit}/"\\${delimit}"}
+
+        # Escape any backslash to preserve them as simple backslash.
+        match_string=${match_string//\\/"\\\\"}
+        replace_string=${replace_string//\\/"\\\\"}
```
zamentur: I think it creates a problem if the password contains a @ because it will be transformed in \@ and here in \\@ => this will create an error...

maniackcrudelis: You're right, there's a regression here.
It's fixed now.

PR #356, opened 2017-08-26
[enh] enable gzip compression for common text mimetypes in Nginx

ariasuni:
Resolves https://dev.yunohost.org/issues/960

### DIFF
--- a/data/templates/nginx/plain/global.conf
+++ b/data/templates/nginx/plain/global.conf
@@ -1,2 +1,2 @@
 server_tokens off;
-gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
+gzip_types text/plain text/css application/javascript text/xml application/xml application/xml+rss text/javascript;
--- a/data/templates/nginx/plain/global.conf
+++ b/data/templates/nginx/plain/global.conf
@@ -1,1 +1,2 @@
 server_tokens off;
+gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
--- a/data/templates/nginx/plain/global.conf
+++ b/data/templates/nginx/plain/global.conf
@@ -1,2 +1,2 @@
 server_tokens off;
-gzip_types text/plain text/css application/javascript text/xml application/xml application/xml+rss text/javascript;
+gzip_types text/css text/javascript application/javascript;


alexAubin:
Not sure what it's about, but Lapineige on Mastodon heard that they might be some risks with having gzip by default .... I think we should investigate these before merging this

ariasuni:
`gzip on;` is already in `/etc/nginx/nginx.conf` and we cannot put the statement twice.

ariasuni:
Enabling gzip with SSL can make us vulnerable to [BREACH attack](http://breachattack.com/). It works by exploiting cases where some sensitive-data is served (and thus compressed) together with some user input.

The safest approach is probably to not enable gzip compression on HTML (which is already not in the mimetypes list) and maybe JSON if it is used for critical information.

For the rest, gzipping JavaScript and CSS (for example) is completely safe.

Psycojoker:
> and maybe JSON if it is used for critical information

It is used for critical information by the API so I guess we won't enable it for it then.

zamentur:
We might disable xml and xml+rss

Psycojoker:
I've hard discussion with a friend about that, I'm a bit too tired to really make sens out of that but it looks like that:

```
15:12 <psycojoker> t'aurais un avis là dessus niveau sécurité ? https://github.com/YunoHost/yunohost/pull/356
15:58 <friend> ça posait soucis avec IE6
15:58 <friend> et pour la sécu, c'est la compression SSL qui merdoie
15:58 <friend> là, compression + SSL, ça marche
16:03 <psycojoker> ah, c'est 2 choses différentes ?
16:03 <friend> genre, avec stunnel, on pouvait activer la compression
16:03 <friend> ça fait parti du protocole SSL
16:04 <friend> mais c'est aussi une faille
16:05 <psycojoker> https://en.wikipedia.org/wiki/BREACH#Details t'es sûr ? j'ai l'impression que ça parle bien de compresser en http
16:05 <psycojoker> sur du https
16:11 <friend> Therefore, turning off TLS compression makes no difference to BREACH
```

But the wikipedia page is super badly written and the breach website is down. We need to investigate more I think but it might actually be fine.

vollkorn1982:
I was directed here by https://mastodon.social/@yunohost/99305798446836007/ and am trying to explain BEAST in simple terms.
For BEAST to work the following must be true:
* Plaintext ist compressed before encryption
* The attacker can control part of the plaintext
* The attacker can observe the ciphertext
Note, that BEAST is a fundamental (and IMHO unsolvable) problem of compression and is independent of certain implementations or program versions.
## How it works
Let's assume the plaintext is a message and a secret. The message is controllable by the attacker and the secret is what the attacker wants to find out. Let the secret be "bbbb".
Now the attacker will try cleverly selected messages and will for example start with "aaaa". In the simplest case the plaintext will look like this now:
"aaaabbbb". If this is compressed it will yield a text of a certain length, which is minimally necessary to carry this information. For example a simple compression algorithm could compress this as "4a4b", with a length of 4.
If the attacker tries other inputs she might try "bbbb" as well. Then the plaintext will look like this:
"bbbbbbbb". This message has much less information necessary to carry and would be with our simple compression algorithm compressed to the message "8b" of length 2.
The attacker now would compare what happens to the length of the ciphertext when she manipulates the part of the plaintext she has access to. From the length she can tell how similar (as in well compressable together with the secret) her input was to the secret.
So, if you want to turn on compression you should make sure there is no information controllable by some third party being compressed.

Psycojoker:
Thanks for your detailed answer :)

alexAubin:
Hey @vollkorn1982, thanks a lot for the explanation which really gives a bit more sense and intuition to the BREACH attack thing.

One thing I'm still no sure to understand, to be able to converge on which type of files we should authorize, is where/when/what are practical example of situation where an attacker is able to control part of the plaintext...

I'm guessing enabling compression for CSS and Javascript code is okay because those are in 99.9% of the case static content containing no secret. I guess RSS is ~fine too (though you could imagine some setup where the RSS is private). But I don't know about JSON, HTML, XML... For instance I'm guessing that you could imagine a dummy situation where there's an API of an application, and a client regularly does request which returns something like 
```
<xml>
<date>07/01/2017 20:23:45</date>
<supersecretcreditcard>1234 5678 1011 1213</supersecretcreditcard>
</xml>
```
with date varying in time and the credit card number staying constant, and then an attacker could observe the cypher thing for multiple time and be able to extract some info... ? (That's probably an oversimplified scenario from real life stuff).

So as far as I understand, static content with no expected secret info (i.e. CSS, Javascript, RSS?) should be fine, but HTML / XML / JSON is not, in the general case ?

vollkorn1982:
The answer will pretty much never be "HTML is okay" or similar. You should better not think in categories of file types here, but evaluate files individually. Any file could be a problem, if it contains a secret (all do) and attacker controllable content (I won't judge on this). TBH: I would leave compression off.

ariasuni:
We can only enable it for CSS and JavaScript, because every other file type can potentially contain secrets. Agree?

alexAubin:
Ok people, I'm flagging this as "ready to merge" so that we gonna merge it in three days or more, depending of what exactly we want to include in the next release. Feel free to complain if you have comments on this !

PR #357, opened 2017-08-27
[fix] Alow dash at the beginning of app settings

ariasuni:
Fixes https://dev.yunohost.org/issues/485


opi:
nice catch !
IMHO, helpers should use long arguments anytime, to improve readability of code.

Psycojoker:
Thanks :)

PR #359, opened 2017-08-28
[enh] Return log after app upgrade with api

zamentur:
## Problem
I need to display log info after app upgrade in the web admin

## Solution
Do like in yunohost tools upgrade return the log if it's an api call

## How to test
This PR is related to the web admin PR https://github.com/YunoHost/yunohost-admin/pull/172
You can test with by upgrading an app

To upgrade an app, you can change the update_time by editing your apps settings /etc/yunohost/apps/APP/settings.yml

## PR Status
Opinion needed 

## Validation

- [x] **Principle agreement** 1/2 : Bram
- [ ] **Quick review** 0/1 : 
- [ ] **Simple test** 0/1 : 
- [ ] **Deep review** 0/1:

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -38,7 +38,7 @@ import pwd
 import grp
 from collections import OrderedDict
 
-from moulinette import msignals, m18n
+from moulinette import msignals, m18n, msettings
 from moulinette.core import MoulinetteError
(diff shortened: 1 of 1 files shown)


Psycojoker:
I'm pretty sure we can grab the exact log of the upgrade but this one looks good enough for now (and it's an improvement anyway).

PR #360, opened 2017-08-28
Add post hooks for apps operation

zamentur:
## Problem
Some apps are dependent of others apps (vpnclient&hotspot, baikal&agendav, coin&openvpn ...).
There is a lot of scenario where an app could want do an action if another app is installed, removed, upgraded or change url.

## Suggested solution
Add this for hooks

I have set up some env var like recommended in this ticket : https://dev.yunohost.org/issues/818

The var are the same than var sending to the scripts/ACTION

## How to test
/etc/yunohost/hooks.d
```
.
├── post_app_change_url
│   └── youpi
├── post_app_install
│   └── youpi
├── post_app_remove
│   └── youpi
└── post_app_upgrade
    └── youpi
```

youpi
```
#!/bin/bash
printenv | grep YNH >> /root/yolo
```
Like that you can see in the yolo file which variable can be used

## PR Status
Opinion needed 

## Validation

- [ ] **Principle agreement** 0/2 : 
- [ ] **Quick review** 0/1 : 
- [ ] **Simple test** 0/1 : 
- [ ] **Deep review** 0/1:

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -430,7 +430,7 @@ def app_change_url(auth, app, domain, path):
         path -- New path at which the application will be move
 
     """
-    from yunohost.hook import hook_exec
+    from yunohost.hook import hook_exec, hook_callback
 
     installed = _is_installed(app)
     if not installed:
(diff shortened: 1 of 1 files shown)


Psycojoker:
As anyone tested it?

PR #362, opened 2017-08-30
Remove 53 upnp

maniackcrudelis:
## Problem
If the port 53 is open on your ISP router, your dnsmasq can be used by everyone. As I already reported before.  
Rafi59 has report today he has the same issue, seen on the [dashboard of his Pi-hole](https://framapic.org/ZuYcNOuKFBqY/7znqBAajodZo.png).

## Suggested solution
Remove this port of the upnp config.
@zamentur suggest to add also a migration...

## How to test
You can see who used your dnsmasq by using Pi-hole.

## PR Status
Opinion needed

## Validation

- [ ] **Principle agreement** 0/2 : 
- [ ] **Quick review** 0/1 : 
- [ ] **Simple test** 0/1 : 
- [ ] **Deep review** 0/1:


opi:
Does dyndns need this port ?

alexAubin:
>Does dyndns need this port ?

Eh no it doesnt ? Do you mean dnsmasq ? This could be an issue if we wanted people to be able to use their instance as resolver by default, but right now we don't have any doc about this and didnt check I think... But Maniack knows more about this since he tweaked pihole

opi:
I mean the part in YunoHost that trigger our dynette server to update nohost.me/noho.st IP ; Does it use the port 53 ?

alexAubin:
No it doesn't ;) (Otherwise, that would mean that you need to forward port 53 on your ISP box in order to have nohost.me / nohost.st on your box)

opi:
OK, good to me then.

alexAubin:
Is this wanted to not remove port 53 also from the firewall ? (c.f. lines for IPv4 / IPv6)

alexAubin:
After discussion : the goal is to block port 53 for requests from outside the local network, but not from within. Hence the change only for Upnp

PR #364, opened 2017-09-02
[enh] Add fail2ban helpers

JimboJoe:
## Problems
To enhance applications protection against hackers/spammers, etc., we can propose helpers to ease the creating of fail2ban jails.

## Solution
Add `ynh_add_fail2ban_config` and `ynh_remove_fail2ban_config` helpers.
A successful implementation example is the [piwigo app](https://github.com/YunoHost-Apps/piwigo_ynh).

## PR Status
Working, but opinion welcome!
And should be implemented in other applications to validate its principle.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -185,11 +185,18 @@ ynh_remove_fpm_config () {
 }
 
 # Create a dedicated fail2ban config (jail and filter conf files)
+# usage 1: ynh_add_fail2ban_config log_file filter [max_retry [ports]]
+# | arg: -l, --logpath=   - Log file to be checked by fail2ban
+# | arg: -r, --failregex= - Failregex to be looked for by fail2ban
+# | arg: -m, --max_retry= - Maximum number of retries allowed before banning IP address - default: 3
+# | arg: -p, --ports= - Ports blocked for a banned IP address - default: http,https
 #
-# usage: ynh_add_fail2ban_config "list of others variables to replace"
+# -----------------------------------------------------------------------------
 #
-# | arg: list of others variables to replace separeted by a space
-# |      for example : 'var_1 var_2 ...'
+# usage 2: ynh_add_fail2ban_config -t [-v "list of others variables to replace"]
+# | arg: -t, --use_template - Use this helper in template mode
+# | arg: -v, --others_var=  - List of others variables to replace separeted by a space
+# |                           for example : 'var_1 var_2 ...'
 #
 # This will use a template in ../conf/f2b_jail.conf and ../conf/f2b_filter.conf
 #   __APP__      by  $app
@@ -235,7 +242,16 @@ ynh_remove_fpm_config () {
 # To validate your regex you can test with this command:
 # fail2ban-regex /var/log/YOUR_LOG_FILE_PATH /etc/fail2ban/filter.d/YOUR_APP.conf
 ynh_add_fail2ban_config () {
-  local others_var=${1:-}
+  # Declare an array to define the options of this helper.
+  declare -Ar args_array=( [l]=logpath= [r]=failregex= [m]=max_retry= [p]=ports= [t]=use_template [v]=others_var=)
+  local logpath
+  local failregex
+  local max_retry
+  local ports
+  local others_var
+  local use_template
+  # Manage arguments with getopts
+  ynh_handle_getopts_args "$@"
 
   finalfail2banjailconf="/etc/fail2ban/jail.d/$app.conf"
   finalfail2banfilterconf="/etc/fail2ban/filter.d/$app.conf"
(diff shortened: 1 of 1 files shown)


Psycojoker:
As anyone tested those? I have no knowledge of fail2ban.

JimboJoe:
At the moment, only the (nearly official) Piwigo application implemented it. So maybe should we wait for a couple of other successful implementations to merge it...?

alexAubin:
As far as I know, fail2ban's conf format changed in Stretch so that might affect these helpers :s

Psycojoker:
> At the moment, only the (nearly official) Piwigo application implemented it. So maybe should we wait for a couple of other successful implementations to merge it...?

Any update on that?

JimboJoe:
It's currently being implemented in Wordpress, and has evolved [here](https://github.com/YunoHost-Apps/Experimental_helpers/pull/2).

alexAubin:
About this : 

we should be careful that the fail2ban conf format changed a bit in stretch and is backward-incompatible. I don't know if these helpers are affected... but right now, the step in the stretch migration is to remove all existing fail2ban conf and regenerate it from scratch :s

JimboJoe:
I've updated the helpers to be on-par with the ones from Experimental_helpers.
I've tested again piwigo on Stretch and they work as advertised.
I understand you had to delete all existing fail2ban jails when migrating from Jessie to Stretch, so that means existing apps won't be protected by fail2ban until their next upgrade (which I find perfectly acceptable)?

alexAubin:
>I understand you had to delete all existing fail2ban jails when migrating from Jessie to Stretch, so that means existing apps won't be protected by fail2ban until their next upgrade

Yes indeed :s

JimboJoe:
This helper currently [only works on Stretch](https://github.com/YunoHost-Apps/Experimental_helpers/pull/11#issuecomment-390648486)...

zamentur:
Not a problem if it only works on Stretch, we could merged it only in stretch...

Psycojoker:
We are now on stretch, shouldn't we merge this?

data/helpers.d/backend
```diff
@@ -183,3 +183,76 @@ ynh_remove_fpm_config () {
...
+
+  test -n "$logpath" || ynh_die "ynh_add_fail2ban_config expects a logfile path as first argument and received nothing."
+  test -n "$failregex" || ynh_die "ynh_add_fail2ban_config expects a failure regex as second argument and received nothing."
+
+  finalfail2banjailconf="/etc/fail2ban/jail.d/$app.conf"
+  finalfail2banfilterconf="/etc/fail2ban/filter.d/$app.conf"
+  ynh_backup_if_checksum_is_different "$finalfail2banjailconf" 1
+  ynh_backup_if_checksum_is_different "$finalfail2banfilterconf" 1
+
+  sudo tee $finalfail2banjailconf <<EOF
```
Josue-T: As all script all called as `root` why does we continue to use sudo ?

maniackcrudelis:
I've updated the helpers to include the fix for jessie and the last modifications from the experimental repository.
This helper will now need #520

data/helpers.d/backend
maniackcrudelis: Probably because someone have forgotten to remove it.

Josue-T: It could be good to clean this everywhere one day.

JimboJoe:
Yes, but not before #520 gets merged :wink:

alexAubin:
#520 got merged. Any update about this ? :D

alexAubin:
Bumpz D:

Josue-T:
I think it could be useful to have more flexibility about the config file. By example if we need to specify multiple regex or add some ingoreregex or add the "maxlines" option.

alexAubin:
Okay guys, I'm flagging this as inactive :/ ... It'd be really nice if we could have some progress on this one way or another. I don't think we need a 100% perfect solution to deploy this, considering that in the meantime some web apps may remain vulnerable to brute force.

maniackcrudelis:
Let's merge in 3 days if no one disagree until then.

alexAubin:
Following today's meeting, this PR is postponed. Josue will have a look at integrating fail2ban in his apps, which requires to have several regex instead of just one. He'll iterate on the design.

data/helpers.d/backend
```diff
@@ -183,3 +183,76 @@ ynh_remove_fpm_config () {
...
+  max_retry=${max_retry:-3}
+  ports=${ports:-http,https}
+
+  test -n "$logpath" || ynh_die "ynh_add_fail2ban_config expects a logfile path as first argument and received nothing."
+  test -n "$failregex" || ynh_die "ynh_add_fail2ban_config expects a failure regex as second argument and received nothing."
+
+  finalfail2banjailconf="/etc/fail2ban/jail.d/$app.conf"
+  finalfail2banfilterconf="/etc/fail2ban/filter.d/$app.conf"
+  ynh_backup_if_checksum_is_different "$finalfail2banjailconf" 1
+  ynh_backup_if_checksum_is_different "$finalfail2banfilterconf" 1
```
maniackcrudelis: Is anyone know why is there a '1' as a second argument for this helper ?
I can't find any version of `ynh_backup_if_checksum_is_different` with a second argument.

This is crashing getopts, which is another problem to fix

maniackcrudelis:
@YunoHost/apps

alexAubin:
Bump @YunoHost/apps , it'd be nice to have this for 3.5 :s (in particular, @Josue-T , it was noted "Josue will have a look at integrating fail2ban in his apps, which requires to have several regex instead of just one. He'll iterate on the design." :sweat_smile: so if you happen to have time to work on this that'd be great !)

Josue-T:
Yes, I might find some time for that in theses next 2 weeks.
If not I think we should merge this as it is because I might not find the time for the theses next month

Josue-T:
Now I updated the helper.

Tested on [synapse](https://github.com/YunoHost-Apps/synapse_ynh/pull/106)

maniackcrudelis:
@Josue-T that's a completely different helper !
You break getopts usage as well as usage of this helper in one line without template.

Maybe for synapse you need to use a template, but most of the apps don't need that and works perfectly with the previous version of this helper.
So, if you want to ADD a template usage for this helper, why not, but I don't think that to force to use a template is a good idea.

Josue-T:
It's just more like the helper nginx.

I think it's the last time change the way how this helper work. And I think that probably many apps need more flexibility than what this helper provided.

Josue-T:
If you really want to keep first idea. We can make a helper `ynh_add_fail2ban_config` with the old idea. And a other helper `ynh_add_fail2ban_template` with my idea.

maniackcrudelis:
nginx doesn't have a standard config, each app has a different config.
You could have said that it's like php, which is true. And actually, this helper could use a template less architecture. But the helper was built over the way we were handling php by that time.

Anyway, except synapse, no app is using a template for fail2ban. And so far we don't need to add that in apps. So we should keep that helper as simple to use as we can. So, like it was.

Again, if you need a template feature for synapse with fail2ban, I agree to ADD this to the helper. But, this shouldn't break the way the helper was working so far.
Nor breaking getopts...

So, add 2 arguments to the helper, one for using a template, another for adding other variables. That's a good thing for those who could need a specific fail2ban config. But do not break the helper because you need a specific way to handle the config.

Josue-T:
Now there are the 2 possibility

maniackcrudelis:
LGTM, but not tested.

maniackcrudelis:
Bump

@YunoHost/apps

maniackcrudelis:
Let's merge in 3 days

alexAubin:
Thanks for the review folks, finally sounds like we gonna merge this \o/ 

serfefefeveimage

PR #365, opened 2017-09-04
Make tools_ldapinit more verbose

Psycojoker:
Hello,

`tools_ldapinit` is used during post install to initialized the ldap db content/schema (which is stored in some yaml in `data/`).

Problem: everything is being ugly catchall try/except which mask every errors during initialization thus making things very hard to debug when they go wrong because we have strictly no informations.

This PR logs this by showing every errors.

This PR needs more testing to see if it fits the targeted used case.


alexAubin:
Not tested but imho this can be merged as a ~micro decision

PR #368, opened 2017-09-07
'/' was resulting in '//' in change_url

Psycojoker:
Fix https://dev.yunohost.org/issues/999

Tested using "yunohost tools shell" and works as expected.

This seems to be quite blocking for the integration of change_url into apps according to feedback.

To test using "yunohost tools shell" do something like this:

```
$ yunohost tools shell
In [1]: from yunohost.app import normalize_url_path

In [2]: normalize_url_path("/")
Out[2]: '/'

In [3]: normalize_url_path("//")
Out[3]: '/'

In [4]: normalize_url_path("/////")
Out[4]: '/'

In [5]: normalize_url_path("a")                                                                                                                                                                                 
Out[5]: '/a/'

In [6]: normalize_url_path("qsd/qsd/qsd/")                                                                                                                                                                      
Out[6]: '/qsd/qsd/qsd/'
```

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -445,8 +445,9 @@ def app_change_url(auth, app, domain, path):
 
     # Normalize path and domain format
     domain = domain.strip().lower()
-    old_path = '/' + old_path.strip("/").strip() + '/'
-    path = '/' + path.strip("/").strip() + '/'
+
+    old_path = normalize_url_path(old_path)
+    path = normalize_url_path(path)
 
     if (domain, path) == (old_domain, old_path):
         raise MoulinetteError(errno.EINVAL, m18n.n("app_change_url_identical_domains", domain=domain, path=path))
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -2105,3 +2106,10 @@ def random_password(length=8):
 
     char_set = string.ascii_uppercase + string.digits + string.ascii_lowercase
     return ''.join([random.SystemRandom().choice(char_set) for x in range(length)])
+
+
+def normalize_url_path(url_path):
```
opi: probably nitpicking, but this helper could be in a file within the https://github.com/YunoHost/yunohost/tree/unstable/src/yunohost/utils folder ? (maybe out of scope of this PR)

Psycojoker: Probably. It would be ever greater in moulinette utils I think but that's mostly a "quick fix" PR to unblock the app group situation. I don't mind if we do those kind of modifications (as long as it don't block this PR for the next release) but I'm not super available for that now :x

alexAubin: Like this one ? :wink: 

https://github.com/YunoHost/yunohost/blob/3571747718ea47e7a2d522adc8d5b2c03c64d261/data/helpers.d/network#L1-L23

Psycojoker: That doesn't great to call that from python :x

alexAubin: Ah nvm I was thinking @opi wanted to make this available for apps

opi: no, but not scoped in app.py ; But let's merge it now, and we'll cleanup later.

PR #370, opened 2017-09-14
Make MySQL dumps with a single transaction to ensure backup consistency

JimboJoe:
## Problems
Making MySQL dumps while having write accesses permitted may possibly lead to inconsistent backups.

## Solution
As stated in [this thread](https://dba.stackexchange.com/questions/19532/safest-way-to-perform-mysqldump-on-a-live-system-with-active-reads-and-writes/19533#19533), `mysqldump` provides the `--single-transaction` option to produce a checkpoint before creating the dump, thus ensuring a consistent backed-up state.
As showed in this [NextCloud notice](https://nextcloud.com/blog/help-test-nextcloud-12.0.3/), the use of this option seems to be widespread.

The only use of `mysqldump` is in the mysql helper. There's no other use in YunoHost core.

## PR Status
Successfully tested on several apps without side effects.

## Validation

- [x] Principle agreement 2/2 : Aleks, opi
- [x] Quick review 1/1 : Aleks
- [x] Simple test 0/1 :


opi:
MySQL documentation say that

>  To dump large tables, combine the --single-transaction option with the --quick option. 

what about adding this [--quick](https://dev.mysql.com/doc/refman/5.7/en/mysqldump.html#option_mysqldump_quick) option ?

beside that, I approve this PR.

JimboJoe:
From what I see, the [`--opt`](https://dev.mysql.com/doc/refman/5.7/en/mysqldump.html#option_mysqldump_opt) option is enabled by default, and it includes the `--quick` option, so we're fortunately good on this point.

alexAubin:
Hey @JimboJoe , from [this comment](https://github.com/YunoHost/yunohost/pull/371#pullrequestreview-63872891), shall we understand that you tested this PR too ? :P

JimboJoe:
@alexAubin, well I stated "Successfully tested on several apps without side effects." in the PR status 😉 
But I can't check the "Simple test" checkbox myself as the PR initiator, can I?

alexAubin:
Ah, I didnt realize you were the initiator of the PR :P 

Well, imho since it's a small PR, we can trust your test and merge as is (except if someone plans to test it soonish)

PR #372, opened 2017-09-21
Uses hmac-sha512 for dyndns TSIG

Psycojoker:
Hello,

Right now we uses hmac-md5 for TSIG signatures for dyndns, this totally sucks. This PR is here to uses hmac-sha512 instead. The main modifications are:

* when subscribing a new domain, uses hmac-sha512. This is tested and working.
* when updating a dydns entry, check if hmac-md5 is used, if yes, then migrate to hmac-sha512

Missing bits:

* some warning on failure
* <s>sign the new sha512 key using the existing md5 key (this is going to be annoying)</s> useless, tsig is too stupid for that since it's a shared symmetrical key

This is probably the most painful thing I've ever worked on for YunoHost, everything kept failing in every manners possible and we don't have a real dev env for dynette.

Related PR for dynette https://github.com/YunoHost/dynette/pull/7

PS: don't be fooled by the branch name, it's actually sha512 and not sha256


Psycojoker:
Appart from proper logging this PR is ready to be tested/reviewed.

I have personally tested it quite a lot and everything seems to be working as expected.

Psycojoker:
Okay, I declare this PR finished and ready for review.

I have personally (and it's annoying to do but I can explain how) test in the subscription of a new hmac-sha512 key and its update and the automatic migration of an existing hmac-md5 to a hmac-sha512 key and its update. All situation works as expected.

In regard to the complexity of this PR, it's importance (#secutiry) and the lack of people in position to review it I'll probably merge it into some days if no one is available for that :/

In regards to the dynette PR: it's also ready for review and has been running in production since some month now (we don't have any testing location ™) and it's hasn't been causing any bug in production that we are aware of.

Once this PR is merged, to do things correctly, we'll need to do a testing exclusively for it (or maybe with some other minor fixes) and ask people to test the migration.

The migration has a fallback mechanism in case it failed so we should be safe but I prefer not to risk the devil.

alexAubin:
Tested several cases and basically it's working as expected, good job, thanks for that ! 👍 

The 3 minutes delay is a bit odd, but I guess we can live with that...

I might have found an issue where the server returns "Error 500 - Internal Sever Error" when attempting to migrate a key which has already been migrated ? That shouldn't happen in real life, that's most probably because I tweaked some stuff manually during tests.

Otherwise, as discussed on the chat, I think we should really try to package this using the migration framework such as it's ran during an upgrade / dist-upgrade, and even if the IP didn't change, so as to : 
- ensure that all setups are consistent, i.e. if you are in this yunohost version, then we know that this migration has been done ;
- ensure that possible issues with the migration are spotted right away by the admin when upgrading the server, and not in some unknown future where the IP will have changed and the admin saw a mail relating the error.
- (on the technical side, avoid to have many migration/legacy code all over the files, 'cause we already have a lot of these and we should really get more experience and confidence with the migration framework - and we'll probably need to answer similar question for Stretch)

In the worst case (i.e. an exception occurs), the admin should just have to relaunch the upgrade/dist-upgrade or this is a bug and we should fix it ASAP. 

I can have a look into it.

Psycojoker:
I'm ok adding a migration for that but I really don't want to remove the "ensure we are using sha512" code from the `dyndns update`, I really want to make sure that no matter which situation we kill md5.

alexAubin:
So I moved the migration thing to the 'migration framework', and the migration is still being called from `dyndns update` if MD5 is detected. I tested a few things and still need to test the `apt-get upgrade / dist-upgrade` case ideally, especially to see how things behave when an exception is encountered.

Psycojoker:
I think we'll test it with the package built for unstable.

alexAubin:
So I did some tests of deb package upgrade and everything looks okay :+1: . We had several small discussions about the UX / design of failed migrations (i.e. what should happen when an exception is raised). Tried to summarize it in this ticket : https://dev.yunohost.org/issues/1049

For now, the conclusion seems to be that : because `dyndns update` in the cron job re-triggers the migration if needed, this migration will be re-ran automatically so it's okay to not have a proper user-warning in case it fails during the dist-upgrade I guess.

So for me we're good to go :+1:

PR #373, opened 2017-09-23
Update php-fpm helpers to handle stretch/php7 and a smooth migration

JimboJoe:
## Problems
PHP and php-fpm versions are upgraded from 5 to 7.0 when upgrading from jessie to stretch.
This upgrade aims at:
- ensuring a functional php-fpm configuration depending on the Debian release
- ensuring a smooth upgrade/uninstallation for apps in stretch when they had been installed on jessie

## Solution
Set and store the php-fpm configuration directory AND php-fpm service name in app settings, based on Debian release name.
Use that settings when removing the app.

This PR should be merged in YunoHost 2.8 to prepare for later stretch migration.

## PR Status
Successfully tested on jessie and stretch sides.
Not tested on a stretch migrated from jessie yet.

## Validation

- [x] Principle agreement 1/2 : Maniack C
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [x] Deep review 1/1 : Maniack C

PR #374, opened 2017-09-25
[Fix] Adresse de rebond différente du domaine principal.

frju365:
## Problem

issue: https://forum.yunohost.org/t/probleme-mail-dns-2d-domaine/3460

## Solution

- Change the header
- add anonymisation
- En fait, cette correction est le fruit d'une recherche de près de 10h sur postfix. Finalement, j'ai trouvé la solution dans la conf d'un autre logiciel, par hasard. Par soucis d'honnêteté, le site dont j'ai puisé cette conf est mailinabox : 
https://github.com/mail-in-a-box/mailinabox/blob/master/conf/postfix_outgoing_mail_header_filters

## PR Status

REOPENED
TEST/ REVIEW Highly Needed (RHN)
 
## EXAMPLE (for JimBoJoe and Reviewers)

before with postsrsd : 
```
Return-Path: <srs0=owd3=ao=domain2.tld=user@domainprincipal.tld> # ou 'Return-Path: <user@domainprincipal.tld>' sans Postsrsd 

Received: from mwinf5c54 (mwinf5c54.ANONYME.net [10.23.111.104])
by mwinb1c03 with LMTPA;
Wed, 13 Sep 2017 15:28:52 +0200

X-Sieve: CMU Sieve 2.3

Received: from domainprincipal.tld ([45.065.99.90])
by mwinf5c54 with ME
id 91Ur1w00Q0RFXA5011UrSl; Wed, 13 Sep 2017 15:28:52 +0200

X-bcc: destinataire@DomainExterne.tld

X-ME-bounce-domain: DomainExterne.tld

X-ME-engine: default

X-me-spamcause: (0)(0000)gggruggvucftvghtrhhoucdtuddrfeelledrgeeggdeijecutefuodetggdotefrodftvfcurfhrohhfihhlvgemucfogfdpggftiffpkfenuceurghilhhou
hhtmecugedttdenucenucfjughrpefhvffukffftgggsehttdertddttdejnecuhfhrohhmpefluhhlihgvnhcuifhomhgvshcuffhirghsuceojhhulhhivghnsegrshg
tphgrphhhihhlrghtvghlihgvrdgvuheqnecukfhppedugedurddvheehrddufedtrdduleenucfrrghrrghmpehhvghlohepsheggedrvghupdhinhgvthepudeguddrv
dehhedrudeftddrudelpdhmrghilhhfrhhomhepshhrshdtpehofigufeeprghopegrshgtphgrphhhihhlrghtvghlihgvrdgvuhepjhhulhhivghnsehsgeegrdgvuhd
prhgtphhtthhopegtlhhjuhegjeeisehorhgrnhhgvgdrfhhr

X-me-spamlevel: not-spam

X-ME-Helo: domainprincipal.tld

X-ME-IP: 45.065.99.90  # MY IP adress

X-ME-Entity: ofr

Received: from domainprincipal.tld (localhost [IPv6:::1])
by domainprincipal.tld (Postfix) with ESMTPSA id C22136B5
for <destinataire@DomainExterne.tls>; Wed, 13 Sep 2017 15:28:50 +0200 (CEST)

From: Name of the user <user@domain2.tld>

To: destinataire@DomainExterne.tld

Subject: ceci est un message de test

Message-ID: <20170913132850.Horde.JcbFlT1kgDPDbzDz7ZcP2Ca@mail.server-d-envoi-test-domaine.tld>

Date: Wed, 13 Sep 2017 13:28:50 +0000

...

DKIM-Signature: ANONYME-LK2R
```

After : 
```
Return-Path: <user@domain2.tld>

Received: from mwinf5c56 (mwinf5c56.ANONYME.net [10.23.111.106])
by mwinb1c03 with LMTPA;
Tue, 26 Sep 2017 10:36:54 +0200

X-Sieve: CMU Sieve 2.3

Received: from domainprincipal.tld ([45.065.99.90])
by mwinf5c56 with ME
id E8ct1w0030RFXA5018ct2f; Tue, 26 Sep 2017 10:36:54 +0200

X-bcc: destinataire@DomainExterne.tld

X-ME-bounce-domain: DomainExterne.tld

X-ME-engine: default

X-me-spamcause: (0)(0000)gggruggvucftvghtrhhoucdtuddrfeelledrjedvgddtkecutefuodetggdotefrodftvfcurfhrohhfihhlvgemucfogfdpggftiffpkfenuceurghilhhou
hhtmecugedttdenucenucfjughrpefhvffukffftgggsehttdertddttdejnecuhfhrohhmpefluhhlihgvnhcuifhomhgvshcuffhirghsuceojhhulhhivghnsegrshg
tphgrphhhihhlrghtvghlihgvrdgvuheqnecukfhppedugedurddvheehrddufedtrdduleenucfrrghrrghmpehhvghlohepsheggedrvghupdhinhgvthepudeguddrv
dehhedrudeftddrudelpdhmrghilhhfrhhomhepjhhulhhivghnsegrshgtphgrphhhihhlrghtvghlihgvrdgvuhdprhgtphhtthhopegtlhhjuhegjeeisehorhgrnhh
gvgdrfhhr

X-me-spamlevel: not-spam

X-ME-Helo: domainprincipal.tld

X-ME-IP: 45.065.99.90

X-ME-Entity: ofr

Received: from authenticated-user (PRIMARY_HOSTNAME [PUBLIC_IP])nge.fr>; Tue, 26 Sep 2017 10:36:52 +0200 (CEST)

From: Name Of my User <user@domain2.tld>

To: destinataire@DomainExterne.tld

Subject: ceci est un message de test

Message-ID: <20170926083652.Horde.yPJwWVW9iq7zfLHZDP99fPB@mail.server-d-envoi-test-domaine.tld>

Date: Tue, 26 Sep 2017 08:36:52 +0000
...
DKIM-Signature: ANONYME-LK2R
```

### Remerciement

Donc, un grand merci à eux (sans le vouloir).

### DIFF
--- a/data/templates/postfix/main.cf
+++ b/data/templates/postfix/main.cf
@@ -131,10 +131,6 @@ smtpd_recipient_restrictions =
     reject_unauth_destination,
     permit
 
-# SRS
-sender_canonical_maps = regexp:/etc/postfix/sender_canonical
-sender_canonical_classes = envelope_sender
-
 # Ignore some headers
 smtp_header_checks = regexp:/etc/postfix/header_checks
 


JimboJoe:
Thanks for your work!
Could you please put the original comments as well for more clarity?

frju365:
Indeed. It's better with comments because in fact there are many IGNORE header for anonymisation too.

JimboJoe:
Thanks. Can you also edit your PR to give us more information on what the problem is, and how to test your patch?
You can use the same model as the other PR (by example [that one](https://github.com/YunoHost/yunohost/pull/373)).
Thanks!

frju365:
Fait, sauf pour les vérifications... je ne sais pas combien il faut de reviews pour merger, ni quand.

JimboJoe:
From my understanding, this template shouldn't be applied as-is, because you should replace the  `PRIMARY_HOSTNAME` and `PUBLIC_IP` expressions, as was done in the [original mail-in-a-box commit](https://github.com/mail-in-a-box/mailinabox/commit/07beef3db20d6094a6923049eda0597bad48ca52).

What about the e-mails you send with that conf? Don't they have a `Received: from authenticated-user (PRIMARY_HOSTNAME [PUBLIC_IP])` (withouth being replaced by your own hostname/IP) header?

frju365:
indeed. Sorry... I will make a commit later in the day to fix it.

JimboJoe:
From what I've tested, when you send a mail from user@domain2.com (where domain1.com is the main domain), the headers contains a `Return-Path: <user@domain.com>` which is  (I believe) what mailtester points out as bounce address ("adresse de rebond").
So if this is your problem, I don't see how your PR resolves it...

You should really edit the PR with real examples of mail headers sent before and after your change, so that we can have a real understanding.

frju365:
ok, let's open it even if my server is in a bad state. :-) I will add two example of header (anonyme) before and after the fix in the PR summary.

frju365:
That's it.
frju365

JimboJoe:
So, the logical part is that it replaces the
```
Received: from domainprincipal.tld (localhost [IPv6:::1])
```
with 
```
Received: from authenticated-user (PRIMARY_HOSTNAME [PUBLIC_IP])nge.fr>; Tue, 26 Sep 2017 10:36:52 +0200 (CEST)
```

But:
- it doesn't replace the first `Received: from domainprincipal.tld` statement!!
- and yet it replaces the Return-Path
```
Return-Path: <srs0=owd3=ao=domain2.tld=user@domainprincipal.tld> # ou 'Return-Path: <user@domainprincipal.tld>' sans Postsrsd 
```
by
```
Return-Path: <user@domain2.tld>
```
which I understand as fixing your issue...

So the `Return-Path` statement seems to be "calculated" somehow, but I have absolutely no idea why... 😦

frju365:
Perhaps anonymisation... I don't know too but this change solved my issue.

frju365:
c'est bon. J'avais en partie oublié que j'avais changé le main.cf.
sender_cannonical a, je pense l'intérêt "d'uniformiser" les adresse d'envoi.

frju365:
I keep the header conf I made because there is some anonymisation which are very nice.

frju365:
Can someone have a look at it  ?

LeOSW42:
I just changed the main.cf as mentioned and worked fine for me.
It worked either with my main mail address and with aliases.

frju365:
Thanks for the feedback ! :+1:

alexAubin:
I understand this might work for your use cases guys, but that's not the right solution in the general case :/.

The SRS thing is here for a reason, in particular for when your forward mails (e.g. Alice sends an email to Bob@bob.nohost.me, which is forwarded to Bob@gmail.com). This isn't easy to explain in just 2 sentences so if you really are interested, you can read [the wikipedia page](https://en.wikipedia.org/wiki/Sender_Rewriting_Scheme#Background) as a starter I guess :s.

[Our current implementation of SRS](https://github.com/YunoHost/yunohost/blob/unstable/data/templates/postfix/plain/sender_canonical#L1) is pretty buggy at the moment (my understanding is that this $1 thing refers to the main domain, which is not what we want), though I believe it at least allows forwarded email to not be completely rejected (e.g. by gmail.com in my example).

So the fix proposed here is an acceptable workaround if you don't care about email forwards (or similar stuff like mailing lists I guess), but it's not in the general case... In the general case, we have to implement a method to correctly handle SRS in a multi-domain context, which is not trivial but is what postsrsd does as suggested [here](https://github.com/YunoHost/yunohost/pull/331).

alexAubin:
The header_checks / IGNORE stuff is pretty independent from the SRS thing though, I think ;). It might be relevant though I don't have enough knowledge about these to know which implications that can have.

frju365:
yes, but I think that for instance it's better than nothing or the Yunohost previous conf.

frju365:
For instance the PR for the postsrsd (I have tested too) doesn't work for me and is pretty useless for this case.

data/templates/postfix/main.cf
```diff
@@ -131,10 +131,6 @@ smtpd_recipient_restrictions =
     reject_unauth_destination,
     permit
 
-# SRS
-sender_canonical_maps = regexp:/etc/postfix/sender_canonical
```
alexAubin: We can't remove this without breaking the forward feature of YunoHost (among other things). The proper fix is to use postsrsd to handle the Sender Rewriting Scheme properly...

frju365: ok. I agree... I can't test postsrsd right now (I tested but surely not enough). I trust you. Can we keep the header. I think we will need to test the first one... many doubts about it.

frju365:
ok. Postsrs PR merged. This is not needed. Thanks for your review  anyway ;)

frju365:
I will open perhaps a PR for the header ? :/

alexAubin:
> I will open perhaps a PR for the header ? :/

Yes please :+1:

PR #375, opened 2017-09-25
[fix] fix this UnicodeDecodingError on backup-restore

Psycojoker:
Hello,

This fix the infamous UnicodeDecodeError that we have on backup-restore. Tested on a buggy backup, I can confirm that this is working.

This is worth a small release I think.

The problem was that we were returning unicode strings and trying to merge them against regular strings, I've fixed the problem at the source.

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -166,11 +166,11 @@ class BackupRestoreTargetsManager(object):
             or (exclude and isinstance(exclude, list) and not include)
 
         if include:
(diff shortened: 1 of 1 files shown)


M5oul:
Could be merge soon.
And a release must be done just after.

Psycojoker:
Hello,

In regards to:
* the few number of people in positions to test this
* the really small size of the patch
* the importance (and urgence) of it
* the confirmation that it works

I'm merging this. We can talk about releasing soon.

PR #376, opened 2017-09-28
[fix] Do not modify the settings too soon in change_url

maniackcrudelis:
## Problems
If the domain is stored too soon, the backup script, used before any modification, fail because it loads the new domain instead of the old one.
The domain, and the path should be changed only at the end of the script.

## Solution
Can't we use the new `domain_url_available` function instead ?

## PR Status
Not tested.
Only a proposition.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -451,10 +451,7 @@ def app_change_url(auth, app, domain, path):
     if (domain, path) == (old_domain, old_path):
         raise MoulinetteError(errno.EINVAL, m18n.n("app_change_url_identical_domains", domain=domain, path=path))
 
-    # WARNING / FIXME : checkurl will modify the settings
-    # (this is a non intuitive behavior that should be changed)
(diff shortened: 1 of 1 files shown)


Psycojoker:
I'm wondering if we effectively save the domain after that then, here we just check it.

Otherwise no opposition to that since it's coming from the people best informed about that subject.

alexAubin:
Ah indeed, as pointed by @Psycojoker , it should probably be more like : 

```
if not domain_url_available(...) :
    # Throw 
    raise MoulinetteException(...)

domain_url_register(...)
```
(To be tested / implemented)

alexAubin:
(Flagging as "work needed" for now, because as said we need more than just checking the availability.)

alexAubin:
Closing because superseded by https://github.com/YunoHost/yunohost/pull/608

PR #377, opened 2017-10-02
Create milter_headers.conf

taziden:
This file define a header : X-Spam which will be added to mail considered as spam by rspamd.

This configuration is defined here : https://rspamd.com/doc/modules/milter_headers.html and as of the time I'm creating this PR, this documentation page is not up-to-date because X-Spam was supposed to be added when extended_spam_headers is set to True and it's not the case anymore since 1.6.2.

This header is being used by rspamd.sieve which we push into dovecot in order to put spammy e-mails in Junk folder automatically.


Psycojoker:
This diff looks really weird and big for this PR and is it normal that the target for merging branch is strech?

Psycojoker:
Merged into unstable using cherry-picking because the PR was very weird. Thanks taziden :heart:

PR #381, opened 2017-10-10
Trigger error if app dependency install fails (Redmine 1006) + allow for 'or' in dependencies

Josue-T:
I purpose this change to improve the helper 'ynh_install_app_dependencies'.
Before this change if the dependences are not installable the install didn't fail. By these change the helper generate an error and the install stop.


alexAubin:
Not tested yet but this sounds like a legit fix, thanks :+1: :smile:

alexAubin:
Agreed with @JimboJoe after having a deeper look. (That sounds like a kinda important fix though.)

@Josue-T : can you elaborate on the `--no-remove` please :stuck_out_tongue: ? Is this what's preventing the triggering of the error somehow ?

JimboJoe:
Somme discussion on the matter in the chatroom:
```
[20:51:24] <josue> Pour répondre à ta question de "--no-remove" c'est car si tu met cette option lorsque des dépendances ne sont pas installables la commande ne vas pas générer d'erreur mais rien ne sera installé ni supprimé et du coup le helper crois que les dépendances sont installée mais réellement ca n'est pas le cas.
[20:52:56] <Aleks> hm oké o.O
[20:53:07] <Aleks> "If any packages are to be removed apt-get immediately aborts without prompting."
[20:53:11] <Aleks> pas sur de tout piger xD
[20:55:01] *Aleks dégaine le git blame
[20:56:11] <Aleks> https://github.com/YunoHost/yunohost/pull/292
[20:56:33] <Aleks> imho ca conflicte avec ça josue
[20:56:41] <Aleks> mais du coup je sais pas trop quelle est la bonne solution
[20:57:28] <Aleks> on peut toujours faire un test à coup de "dpkg -l" après, ou un truc du genre, mais ça fait un peu iech que la commande retourne pas directement une erreur quoi
[20:59:16] <josue> Alors peut être que ma solution n'est pas idéale mais en tout cas actuellement c'est pas top. Dans tout les cas j'ai fait des test et actuellement tu peu passer n'importe quelle paquet à installer le helper continuera comme si rien n'était.
[21:00:29] <Aleks> ouaip je comprends :s
[21:02:21] <josue> Et même ton dpkg -l te dira que ton paquet de dépendance est installé. Le seul autre moyen c'est de checker pour tous les paquet qui doivent être installé.
[21:05:05] <Aleks> arf ok
```

Josue-T:
Maybe we could try to see on the option `--fix-missing`. Maybe an error could be generated with this option.

Josue-T:
I have found that if we call ynh_die while the subshell finish with an erron (in line 104) the install fail if the dependence is not installable.

So I think this solution is clearly better than the last.

JimboJoe:
Now making, some tests on stretch, I'm not sure this change is a good idea at the moment...
Rationale : nextcloud app needs the `php-zip` dependency (which automatically installs `php7-zip`) for stretch, but this dependency doesn't exist in Jessie.
So in essence, to prepare stretch migration, we may add dependencies that don't exist in jessie, and we surely don't want that to block installation...

Josue-T:
In this case I think the best way is to put a choice in the control file.
Actually the dependence is something like that : `package1 & package2 & package3 & php-zip`
and you purpose for the stretch compatibility is that : `package1 & package2 & package3 & php-zip & php7-zip` but I think it don't solve the problem.

I think to solve that we need to implement a function with give a choice between some package like that :  `package1 & package2 & package3 & (php-zip | php7-zip)`.

JimboJoe:
I'm sorry if I was unclear: neither `php-zip` nor `php7-zip` exist for jessie.

Josue-T:
I understand, but in all case while you set the dependence witch is not installable (in jessie in your case) after the install the package manager won't accept to do anything as long as this the dependence are not solved.

alexAubin:
>Now making, some tests on stretch, I'm not sure this change is a good idea at the moment...
Rationale : nextcloud app needs the php-zip dependency (which automatically installs php7-zip) for stretch, but this dependency doesn't exist in Jessie.
So in essence, to prepare stretch migration, we may add dependencies that don't exist in jessie, and we surely don't want that to block installation...

In fact, today I'm investigating app regressions and I found that searx has exactly the opposite case : as far as I understand, a dependency available in jessie is not available in stretch, but the install of dependencies fails silently. Later, we end up with a `virtualenv: command not found` because no dependency was installed at all. (c.f. [here](http://212.47.243.98:8080/job/searx%20(Official)/lastBuild/consoleText))

If we have different dependencies depending on the distribution (jessie/stretch), imho we should have an explicit if/else block. But in any case, we should definitely trigger an error if the app dependency install fails, 'cause we can't expect the app to function properly (let alone expect it to install correctly) if it does.

JimboJoe:
I was very naively expecting we could move towards stretch without impacting apps too much, but you've just proven the contrary 👍 
So we'll have to handle these cases in official apps prior to YunoHost 3.0 release, and to communicate effectively to app developers when releasing first testing.

Josue-T:
Hello,

I think now that we might have a problem for the migration if we install a different package witch contain some different dependence.
For example : 
- In debian jessie the app need the package **a**, **b**, and **c**.
- In debian strech the package c is not available but an other package **d** replace **c**.

The first idea is to type that : 
```
if $debian_stretch
then
    ynh_install_app_dependencies a b c
else
    ynh_install_app_dependencies a b d
fi
```
But I think this won't work on the migration (jessie -> stretch) because on the migration apt will say "I need the package a, b, c but c is not installable". It will be only while we do the app upgrade that the dependence package will be replaced and the new dependence (for stetch) will be installed.

I think the only way is to make a package compatible with jessie and stretch. We can create a controle file witch look like that : 
```
...
Depends: a, b, (c | d)
...
```
In this case I think it should run because while we do the migration apt will say "I need c or d so I can find d and I will install it".

alexAubin:
>But I think this won't work on the migration (jessie -> stretch) because on the migration apt will say "I need the package a, b, c but c is not installable". It will be only while we do the app upgrade that the dependence package will be replaced and the new dependence (for stetch) will be installed.

As far as I understand it, it's a bit trickier than that. If you do a migration from jessie to stretch (with the whole change in sources.list and dist-upgrade and so on), the system *will keep package c installed, even if it's not available in stretch*. That's a least what I observed for php5 in particular.

So with that in mind, an app installed with package a, b, c should keep working on stretch, even though c is not available and should be replaced with d. Though that could be a concern, because that means this app doesn't have a single installation scheme (i.e. if d becomes absolutely mandatory in the future and the packager didn't pay attention to this in the upgrade script...)

But the thing you propose about the control file looks pretty cool 👍

Josue-T:
> As far as I understand it, it's a bit trickier than that. If you do a migration from jessie to stretch (with the whole change in sources.list and dist-upgrade and so on), the system will keep package c installed, even if it's not available in stretch. That's a least what I observed for php5 in particular.

Ok, but I also think if we have a conflict maybe c could in conflic with d. I don't know if this case exist but I just try to think of all case.

alexAubin:
>Ok, but I also think if we have a conflict maybe c could in conflic with d. I don't know if this case exist but I just try to think of all case.

Ah yes indeed, so in that case the `(c | d)` in control file should handle the conflict like we want I think 👍

Josue-T:
Hello,
I have updated the helper to support the choice between some package.

data/helpers.d/package
```diff
@@ -117,8 +117,13 @@ ynh_package_install_from_equivs () {
 #
 # usage: ynh_install_app_dependencies dep [dep [...]]
 # | arg: dep - the package name to install in dependence
+#   You can give a choice between some package with this syntax : "dep1|dep2"
+#   Example : ynh_install_app_dependencies dep1 dep2 "dep3|dep4|dep5"
+#   This mean in the dependence tree : dep1 & dep2 & (dep3 | dep4 | dep5)
 ynh_install_app_dependencies () {
     local dependencies=$@
+    local dependencies=${dependencies// /, }
```
zamentur: What about if we put 2 comas ?
see https://github.com/YunoHost-Apps/grafana_ynh/blob/e5a150e077eb9120d9e41a3bad6404c63d33f2fe/scripts/_common.sh#L60

Josue-T: Yes, it's possible but in this case we change the synatax of the actuall helper. My purpose is juste add the choice possibility with the same synatx.

Actually we have this synaxe : `ynh_install_app_dependencies dep1 dep2 dep 3`
And with my purpose this will be that : `ynh_install_app_dependencies dep1 dep2 "dep3|dep4|dep5"` witch mean : dep1 & dep2 & (dep3 | dep4 | dep5)

alexAubin: I think the question here was to know what happens if the users already had `,` in the arguments, like grafana_ynh has : `ynh_install_app_dependencies influxdb, grafana`, which means `$dependencies` will end up containing `influxdb,, grafana`, but thinking about it, that shouldn't be any trouble since this line was actually done later in the script. So I'm guessing those debian control files things are robust against typos like having `,,` :+1:

Josue-T: Ah, ok I understand, but I've just moved the replacement, but it was already here. (see : https://github.com/YunoHost/yunohost/pull/381/files#diff-48ee89702d6221dfe3ea24923b0f8f9eR143). I think the debian script is really exigent about the syntax, so two comma won't probably be accepted.

alexAubin:
(This got merged in branch stretch)

maniackcrudelis:
As discussed here https://github.com/YunoHost-Apps/nextcloud_ynh/pull/96, we encounter a situation where a package is needed, but only available on stretch.
So what about a [Recommends](https://www.debian.org/doc/debian-policy/#s-binarydeps) field in the control file to allow optionnal dependencies ?

maniackcrudelis:
So, it's a false good idea...
Unfortunately, dpkg ignores this field. So we can't use it.

alexAubin:
Soooo to try to move forward with this : 

to me it sounds like the nextcloud issue ends being somewhat unrelated to this : we can't really have a dependency like (be on jessie [or] install php-zip) I think ?

Are there some other items / objections / discussions to have on this PR or are we good to go :stuck_out_tongue: ?

alexAubin:
Merging this as there were already several approvals. We can still iterate later if needed.

Thanks for your work @Josue-T and sorry about the long discussion / delay !

JimboJoe:
:tada:

PR #383, opened 2017-11-04
[enh] Add ynh_info helper for app scripts

alexAubin:
### Problem

Following some tickets on the bugtracker (can't find them back because redmine's broken...) + some threads on the forum, there are some issues with how logging is done in yunohost : 

- By default, only 'Success', 'Warning' and 'Error' messages are visible, but in many cases we want to inform the user of something. Generally we end up either not informing him/her, or putting the message in a 'Warning' which is confusing.
- When an app install fails, it triggers the removal of the app. This is fine, but the logging confuses user, as remove scripts usually show a lot of warnings in these case (e.g. attempt to delete a user/db which was not created yet). The user then usually copy/paste the last few lines to the forum, but these lines don't contain the relevant initial error message... We need to think about ways to make this clearer.
- When installing an app from the webadmin, you see the full bash trace/debug output full of technical stuff, but from the CLI by default you see only Success/Warning/Errors. Imho in both case, you should see Info/Success/Warning/Errors.

The goal of this PR is not to address all of this, but to make a step towards more meaningful app script logs.

### Solution

(See also an attempt of explanation with [this sketch](https://framapic.org/pMoGibRSKka9/15Ko2k3DwlCd.png))

- Allows app scripts to send 'Info' messages via a `ynh_info "Hello world"` helper to describe (briefly) what's going on during execution. 
- This is done via a new option `enable_stdinfo` for `hook_exec` (see https://github.com/YunoHost/moulinette/pull/155), it passes a YNH_STDINFO env variable to the script, containing the name of the FIFO pipe to write to
- Move the bash XTRACE to the 'Debug' level (instead of info). 

N.B. : it does *not* enable the Info level to be visible by default (still need --verbose).

### How to test 

- Checkout this branch
- Clone an app locally, add some `ynh_info "Some message"` in a script
- Run the script (e.g. install) with `--verbose`

## Validation

- [] Principle agreement 1/2 :  ljf
- [] Quick review 0/1 : 
- [x] Simple test 1/1 : ljf
- [x] Deep review 1/1 : ljf


src/yunohost/hook.py
```diff
@@ -297,7 +298,7 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
 
 
 def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
-              chdir=None, env=None, user="admin"):
+              chdir=None, env=None, user="admin", enable_stdinfo=False):
```
zamentur: I suggest to allways enable stdinfo, so this args does not seem so usefull.

PR #384, opened 2017-11-10
Solve issue with ynh_restore_file

Josue-T:
While we use the `ynh_restore_file` if the destination already exist the `mv` fonction don't work correctly. By this commit the old directory is renamed.


JimboJoe:
Thanks for your PR and sorry for the delay in reviewing...
You seem to only consider the case of a directory. In my understanding, `DESTPATH` can be a file, and then the `mv` command won't be problematic (other than overwriting the existing file).

Josue-T:
Hello,
Yes you are true, but I think it could depends of the point of view. My PR is more to show that actually there are a problem and to purpose a solution, but this solution could be improved. But to improve this patch I need some point of view to know what is the best thing to do when we have a conflict.

JimboJoe:
Well, after some thinking, I think we should make a prior backup in any case (that's already what we do with applications configuration files). We should maybe log a message in this case as well, as some backup directories could be huge (for example, a Nextcloud data directory, which is never deleted when removing the application).
But we can wait for other points of view 😉

alexAubin:
We discussed this a while ago in a meeting, and it's a tricky issue...

Apparently the items that were noted where to : 
- at least add a warning message when we encountered that case to tell the user that we moved something
- move the existing file/folder (?) to the same place used by regen-conf : `/home/yunohost.conf/backup` (to be consistent, and also to avoid keeping a file/folder that might create conflicts, e.g. if it's a nginx file)

Josue-T:
And if it is a data directory for example in /var/lib/xxx what should we do ?

data/helpers.d/filesystem
```diff
@@ -180,12 +181,18 @@ ynh_restore_file () {
...
     # If ARCHIVE_PATH doesn't exist, search for a corresponding path in CSV
     if [ ! -d "$ARCHIVE_PATH" ] && [ ! -f "$ARCHIVE_PATH" ] && [ ! -L "$ARCHIVE_PATH" ]; then
         ARCHIVE_PATH="$YNH_BACKUP_DIR/$(_get_archive_path \"$ORIGIN_PATH\")"
     fi
-    
+
+    # Move the old directory if it already exist
+	if [[ -e "${DEST_PATH}" ]]
+	then
+		mv "${DEST_PATH}" "${DEST_PATH}_old_$(date +%Y%m%d%H%M)"
```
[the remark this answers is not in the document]
Josue-T: Ok and if it is big what does we do ? We just remove it ?

Josue-T:
Hello,

I have changed changed the things we have discuss. So what do you think now ?

data/helpers.d/filesystem
```diff
@@ -180,12 +181,26 @@ ynh_restore_file () {
...
     if [ ! -d "$ARCHIVE_PATH" ] && [ ! -f "$ARCHIVE_PATH" ] && [ ! -L "$ARCHIVE_PATH" ]; then
         ARCHIVE_PATH="$YNH_BACKUP_DIR/$(_get_archive_path \"$ORIGIN_PATH\")"
     fi
-    
+
+    # Move the old directory if it already exist
+    if [[ -e "${DEST_PATH}" ]]
+    then
+        # Check if the file/dir size is less than 500 Mo
+        if [[ $(du -sb /var/www/horde | cut -d"/" -f1) -le "500000000" ]]
```
alexAubin: I'm guessing you mean `${DEST_PATH}` here ? 😛

Josue-T: Ah, yes, sorry.

Psycojoker:
Haven't tested (I don't work on the helpers) but there is a conflict that needs to be resolved, it seems pretty trivial but I would prefer that someone that is more used to the helpers than me would do it.

PR #385, opened 2017-11-10
Create helper to get package version from the manifest

Josue-T:


JimboJoe:
Nice idea!
How does it fit with this somewhat ["new rule"](https://github.com/YunoHost/doc/pull/464)?

alexAubin:
Indeed, same question as @JimboJoe 

@Josue-T : basically the convention seems to have a version number for the app which is `<upstreamversion>-<packageversion>`, so e.g. 3.0.2-5 (3.0.2 being the upstream version, and 5 being the package version).

We could easily have a function like `ynh_app_upstream_version` that splits and use cut/awk to get the part before `-` ?

Josue-T:
Yes it's a good idea.

Jibec:
Having a helper to extract version is a good idea!

a few comments:

* I think we should we add a rule to https://github.com/YunoHost/package_linter and enforce than 
  We may wish to store it in the application, so we can make use of it for upgrades 👍
* I'm not found of the helper's name, "app" can be either the YunoHost package or the upstream application. I suggest to be more explicit in the function name. If you have two function, one for the package and one for the upstream app, it should solve this.
* I don't understand how the "src" thing may work, I'm afraid it's going to lead to multiple issues in the future. I would prefer to have it in a different PR with an explanation of what we want to do and why.

Josue-T:
> I don't understand how the "src" thing may work, I'm afraid it's going to lead to multiple issues in the future. I would prefer to have it in a different PR with an explanation of what we want to do and why.

I have implemented this helper in my package here : https://github.com/YunoHost-Apps/riot_ynh
You can see that the helper replace the version automatically in the source file : https://github.com/YunoHost-Apps/riot_ynh/blob/master/conf/app.src#L3

Jibec:
sorry but this really is something dirty, I'm surprised it doesn't break the install as checksum can't be correct...

Josue-T:
For the 'ynh_setup_source' I understand your point of view. My point of view was that it's better to don't have the version written in many places in the package.

About char to seprate the upstream and the packageversion is it really a good idea to use this char `-`  because I think that we could find an upstream version with already the char `-`.
For example, if we use the debian package for mumble we have this : 
- The upstream version (debian version) : ` 1.2.18-1`
- The yunohost review : `5`
- The result would be :  `1.2.18-1-5`

I found this page witch explain that : https://en.wikipedia.org/wiki/Software_versioning#Separating_sequences

Would it be better to use something like : `<upstreamversion>~ynh<packageversion>`
With my last example, in this case it would be : ` 1.2.18-1~ynh5`

maniackcrudelis:
Hi Josue-T

About your modification of ynh_setup_source, I agree with Jibec, I think this adding will lead to a lot of troubles. And more than anything else, I really not sure about the gain comming with this modification.
Of course, I understand the idea of having the version number only in the manifest. But, You have also to change the checksum... So, eventually, you have to modify this file...
In a other hand, keeping the version number here will simplify the verification of a checksum and the readability of this file.

For your helper to get the version, it's really interresting, so I copied it in the new repository for experimental helpers
https://github.com/YunoHost-Apps/Experimental_helpers/blob/master/ynh_app_version/ynh_app_version
Here, we can use it, check it and improve it.

alexAubin:
What do we do with this PR guyz ? Shall we keep the ynh_app_version part ?

maniackcrudelis:
`ynh_app_version` will be replace by `ynh_read_manifest "../manifest" "version"`
https://github.com/YunoHost-Apps/Experimental_helpers/blob/master/ynh_read_manifest/ynh_read_manifest_2

PR #386, opened 2017-11-18
[Fix LE] Update acme-tiny as LE updated its ToS

M5oul:
- fix https://dev.yunohost.org/issues/1007
- Renewing a certificate is not problematic, but creating a certificate for a new domain name may failed according to recent feedbacks.
 ~~Import sources as a subtree to ease future cases like this one with a [simple `git subtree pull`](https://github.com/duniter/duniter/issues/466#issuecomment-243899393).~~

---
~~@Psycojoker, sorry, it remove pep8 fixes. But now, they can be done upstream and retrieved then.~~

### DIFF
--- a/src/yunohost/vendor/acme_tiny/acme_tiny.py
+++ b/src/yunohost/vendor/acme_tiny/acme_tiny.py
@@ -39,7 +39,7 @@ def get_crt(account_key, csr, acme_dir, log=LOGGER, CA=DEFAULT_CA):
         raise IOError("OpenSSL Error: {0}".format(err))
     pub_hex, pub_exp = re.search(
         r"modulus:\n\s+00:([a-f0-9\:\s]+?)\npublicExponent: ([0-9]+)",
-        out.decode('utf8'), re.MULTILINE | re.DOTALL).groups()
+        out.decode('utf8'), re.MULTILINE|re.DOTALL).groups()
     pub_exp = "{0:x}".format(int(pub_exp))
     pub_exp = "0{0}".format(pub_exp) if len(pub_exp) % 2 else pub_exp
     header = {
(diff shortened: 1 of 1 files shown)


src/yunohost/vendor/acme_tiny/acme_tiny.py
```diff
@@ -95,7 +82,7 @@ def _send_signed_request(url, payload):
     log.info("Registering account...")
     code, result = _send_signed_request(CA + "/acme/new-reg", {
         "resource": "new-reg",
-        "agreement": "https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf",
+        "agreement": json.loads(urlopen(CA + "/directory").read().decode('utf8'))['meta']['terms-of-service'],
```
opi: Not sur to be the right person to review this change.

- lot of new files, is this because you decide to use git subtree ?
- the main change (appart from pep8) in acme_tiny.py is here, I don't understand this change but it seems legit

M5oul: > lot of new files, is this because you decide to use git subtree ?

Exactly.

>the main change (appart from pep8) in acme_tiny.py is here, I don't understand this change but it seems legit

This change happens [upstream](https://github.com/diafygi/acme-tiny/commit/c4940d229a296b7643e86d2e9ab31a6e0099ba71) to fix this problem.

alexAubin:
>We have to be sure it really fix the problem.
>As the only feedback we have is negative.

Not sure what's going on with julie's setup, but there was a few positive feedback in other threads and on the support's chat.

Regarding the PR itself, I'm not too comfortable with the subtree thing adding a whole bunch of commits unrelated to YunoHost :/ That sounds a bit overkill for a one-file dependency which should be managed by a proper deb package or something alike in the long term

M5oul:
> Regarding the PR itself, I'm not too comfortable with the subtree thing adding a whole bunch of commits unrelated to YunoHost :/ That sounds a bit overkill for a one-file dependency which should be managed by a proper deb package or something alike in the long term

Ah, you are right there [will be a package for jessie](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=acme-tiny).
Let's update that line for now and use the Debian package when YunoHost will be migrated to Stretch.

Psycojoker:
Still good for me.

src/yunohost/vendor/acme_tiny/acme_tiny.py
opi: LGTM

PR #388, opened 2017-11-22
Fix for old apps without backup script.

maniackcrudelis:
Add only a condition to check if there's a backup script before trying to make a backup.
To prevent a failing of the upgrade script in case of missing backup script.

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -59,30 +59,35 @@ ynh_restore_upgradebackup () {
 # ynh_abort_if_errors
 #
 ynh_backup_before_upgrade () {
-	backup_number=1
-	old_backup_number=2
-	app_bck=${app//_/-}	# Replace all '_' by '-'
-	
-    	# Check if a backup already exists with the prefix 1
-	if sudo yunohost backup list | grep -q $app_bck-pre-upgrade1
-    	then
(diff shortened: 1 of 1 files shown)


JimboJoe:
Small fix, could be merged quickly.

data/helpers.d/utils
```diff
@@ -59,30 +59,35 @@ ynh_restore_upgradebackup () {
...
-	app_bck=${app//_/-}	# Replace all '_' by '-'
-	
-    	# Check if a backup already exists with the prefix 1
-	if sudo yunohost backup list | grep -q $app_bck-pre-upgrade1
-    	then
-        	# Prefix becomes 2 to preserve the previous backup
-		backup_number=2
-		old_backup_number=1
-	fi
+	if [ -e "/etc/yunohost/apps/$app/scripts/backup" ]
```
alexAubin: That's a minor thing, but could we maybe simplify this to save 1 indent level ? e.g. :

```
if [ file does not exists ] 
then
     echo "This app doesn't have any backup script." >&2
     return
fi

...the rest of the function...
```

alexAubin:
Yolocommited my indentation change proposal :D @JimboJoe : do you think you could re-test it like this if that's not too troublesome ? :D

JimboJoe:
Tested OK 👍

PR #392, opened 2017-12-08
Solve ynh_use_logrotate

Josue-T:
Actually when we use the command `ynh_use_logrotate /var/log/abcd/` the result in the logrotate conf file is `/opt/yunohost/abcd/logs//.log` witch is not what we want.


maniackcrudelis:
Merge as a micro decision

PR #393, opened 2017-12-10
[enh] Default version number for ynh_install_app_dependencies

maniackcrudelis:
## Problem
- *ynh_install_app_dependencies fail if there's no version number in the manifest*

## Solution
- *Use a default value if $version is empty.*


data/helpers.d/package
```diff
@@ -124,6 +124,9 @@ ynh_install_app_dependencies () {
     	manifest_path="../settings/manifest.json"	# Into the restore script, the manifest is not at the same place
     fi
     version=$(grep '\"version\": ' "$manifest_path" | cut -d '"' -f 4)	# Retrieve the version number in the manifest file.
+    if [ " ${#version}" -eq 0 ]; then
+        version="1.0"
+    if
```
JimboJoe: I think it would look much nicer with a `fi` here 😉

maniackcrudelis: Are you sure ?
I think bash would like the idea of changing its habits.

Psycojoker:
A conflict needs to be resolved for this PR to be merged and I don't know this part of the code so I can't do it :/

Psycojoker:
Thanks :heart:

PR #394, opened 2017-12-13
Fix broken ynh_replace_string

maniackcrudelis:
## Problem
- *The just released commit https://github.com/YunoHost/yunohost/commit/e3a4b307f7c307820c83afce9489f8a128ffb966 break ynh_replace_string if you want to use it with a some special sed characters.*
- *For example, you can't use &, or sub-expression. If you try `ynh_replace_string "\(string\)" "\1"`, the helper will change that into `s@\\(string\\)@\\1@`, which obviously doesn't work...*

## Solution
- *Remove those escapes from ynh_replace_string to restore a correct behavior.*
- *And add a new helper, `ynh_replace_password_string` which use ynh_replace_string, but escape all the potential problematics characters to allow password with special characters.*

This fix should be quickly merged in stable, because it affects most of the change_url scripts.

### DIFF
--- a/data/helpers.d/string
+++ b/data/helpers.d/string
@@ -26,6 +26,27 @@ ynh_replace_string () {
 	local replace_string=$2
 	local workfile=$3
 
+	# Escape the delimiter if it's in the string.
+	match_string=${match_string//${delimit}/"\\${delimit}"}
+	replace_string=${replace_string//${delimit}/"\\${delimit}"}
+
+	sudo sed --in-place "s${delimit}${match_string}${delimit}${replace_string}${delimit}g" "$workfile"
+}
+
+# Substitute/replace a password by another in a file
+#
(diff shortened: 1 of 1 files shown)


JimboJoe:
LGTM and tested OK.
I think we need to be:
* more clear about limitations (you can't use `\` with `ynh_replace_string`), 
* and more clear in the naming: why not `ynh_replace_special_string` instead of `ynh_replace_password_string` (this is not limited to passwords), for instance?

maniackcrudelis:
> more clear about limitations (you can't use \ with ynh_replace_string)

? We can use a \ in ynh_replace_string, that's a part of the problem, we couldn't before this fix.

JimboJoe:
I was referring to the fact that as sed is used, "\\" has a special effect... But nevermind, we already commented about that.

PR #395, opened 2017-12-17
[fix/enh] Use local variables

maniackcrudelis:
## Problem
- *ynh_install_app_dependencies used and modified MY variable "$version"...*

## Solution
- *Use local variables for all our helpers.*
- *I've not forgotten $finalsystemdconf, $finalnginxconf, $finalphpconf and $finalphpini which can be usefull even after the execution of the helper.*


Jibec:
I could not add my commits to this PR :(

I did some corrects, but we still have some issues I don't want to fix without asking you confirmation.
Adding "local" to variable assignments raises this warning to multiple place: https://github.com/koalaman/shellcheck/wiki/SC2155

It also suggest to add "-r" to read lines: https://github.com/koalaman/shellcheck/wiki/SC2162

And file print raises multiples issues such as: https://github.com/koalaman/shellcheck/wiki/SC2068
I don't understand the code to suggest something.

maniackcrudelis:
> Adding "local" to variable assignments raises this warning to multiple place: https://github.com/koalaman/shellcheck/wiki/SC2155

We use also set -u, so a blank variable will trig trap anyway.

> It also suggest to add "-r" to read lines: https://github.com/koalaman/shellcheck/wiki/SC2162

Change something that works fine now doesn't seem like a good idea... And, escape backslash can be a real problem when read is used to read a file line per line.
Which is the main used of read here.

> And file print raises multiples issues such as: https://github.com/koalaman/shellcheck/wiki/SC2068

What do you mean by "file print" ? All our cat with EOF ?

Jibec:
OK, thanks for taking the time to have a look at it.
I meant, the helper file named "print"
-- 
Jean-Baptiste Holcroft

alexAubin:
So are we good with this or ... ?

maniackcrudelis:
For me is all good.

PR #396, opened 2017-12-19
Ajout de connexion stmp over tls et STARTTLS

kitoy30:


alexAubin:
(c.f. discussion on chat : closing because wrong branch, #397 is the good one 😉)

PR #398, opened 2017-12-22
[enh] Tranfer backup_core_only to BACKUP_CORE_ONLY

maniackcrudelis:
## Problem
- *It can be hard to handle the value of `backup_core_only` when it's in settings.yml, especially with restore script.*

## Solution
- *Do not use `backup_core_only` in settings.yml anymore. But use an environment variable, passed only to the backup script when this helper is used.*


data/helpers.d/utils
```diff
@@ -77,7 +77,7 @@ ynh_backup_before_upgrade () {
 	fi
 
 	# Create backup
-	sudo yunohost backup create --ignore-system --apps $app --name $app_bck-pre-upgrade$backup_number
+	sudo BACKUP_CORE_ONLY=1 yunohost backup create --ignore-system --apps $app --name $app_bck-pre-upgrade$backup_number
```
alexAubin: Did you test this ? I'm not sure that the value of BACKUP_CORE_ONLY is propataged to the actual backup script... :confused:

maniackcrudelis: I remember I already try that. But without all the new commits.
Let's try that right now ;)

maniackcrudelis: Just tried with nextcloud (and some upgrade of my helpers and of nextcloud scripts to follow this PR). It works as expected.
In fact, it works because the variable is set in the same line than the yunohost command. So, the shell launch by YunoHost keep the value of the variable.

The only problem here that to use the argument IS_BIG, you need to use a second argument, so the third one can be IS_BIG.
IS_BIG should not be associated to $3 as a positional parameter, but be an option, like --is-big.

alexAubin:
Yolomerging as it's a small PR and opened since like 4 months

PR #399, opened 2017-12-29
Fix security of nginx headers .

frju365:
# Problem

Security problems due to unsafe header : 
  - check : https://wiki.mozilla.org/Security/Guidelines/Web_Security
  - Test of the header : https://observatory.mozilla.org/ ; https://securityheaders.io/ ; https://hstspreload.org

# Solution

- Add preload to the HSTS header
- Add XSS protection (need CSP too)
- Add Referrer-Policy (need to be tesing for other applications) only in ADMIN conf.

# Improvement in future PR :

- Add OCSP Stapling
- Add HPKP-pining to SSL cert.
- Add referrer to server.tpl
- Improve CSP
- Add Report-URI header with proper PHP file in .well-known/report-uri/index.php 

# Status
Need to be testing.

### DIFF
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -43,8 +43,7 @@ server {
     #ssl_dhparam /etc/ssl/private/dh2048.pem;
 
     add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload"; 
-    add_header 'Referrer-Policy' 'same-origin';
-    add_header Content-Security-Policy "upgrade-insecure-requests; object-src 'none'; script-src https: 'unsafe-eval';report-uri /csp-violation-report-endpoint/";
+    add_header Content-Security-Policy "upgrade-insecure-requests; object-src 'none'; script-src https: 'unsafe-eval';";
     add_header X-Content-Type-Options nosniff;
     add_header X-XSS-Protection "1; mode=block";
     add_header X-Download-Options noopen;
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -42,7 +42,12 @@ server {
     # > openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048
     #ssl_dhparam /etc/ssl/private/dh2048.pem;
 
-    add_header Strict-Transport-Security "max-age=31536000;";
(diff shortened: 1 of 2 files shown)


frju365:
Yes, it is needed to test it with several case and apps, to see if nothing is broken. For the referrer header, I will change it to 'same-origin', because 'origin-when-cross-origin' is not really safe (observatory of mozilla)

Psycojoker:
Yeah, I think we need one full testing only for this patch to check that nothing has break :/

frju365:
In fact the conf of the app is first. So, it won't break it. I need one testing at min for admin interface, and a test with at least 5 random apps(to be sure it won't break the app without conf overwriting).

Psycojoker:
We can wait for a whole CI run on this PR no? Once it's merged, I remember we did that for something else.

frju365:
For me, it's good, if someone test it ;) I have tested that with my app, and it seems to work.

frju365:
I think that is good for old browser... But I will check that.

frju365:
I wanted to improve it for the admin conf and perhaps add a true url for the report-uri CSP. Don't merge for instance.

data/templates/nginx/server.tpl.conf
```diff
@@ -42,7 +42,14 @@ server {
     # > openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048
     #ssl_dhparam /etc/ssl/private/dh2048.pem;
 
-    add_header Strict-Transport-Security "max-age=31536000;";
+    add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload"; 
+    add_header 'Referrer-Policy' 'same-origin';
```
zamentur: Need to be tested on django apps before to apply it.

zamentur: The feature to test on apps is the login feature of apps

frju365: Yes. for better testing I will split the PR in 2.

frju365:
PR is now complete.

frju365:
dans "Improvement in future PR", j'en ai profité pour noter les futures améliorations que je souhaiterais faire dans les header (et que j'ai déjà fait chez moi par la même occasion)

Psycojoker:
Let's put that in a bit long testing then :)

PR #403, opened 2018-01-04
Add commands to allow/disallow a yunohost user to connect in ssh

Psycojoker:
Extracted from https://github.com/YunoHost/yunohost/pull/345 to allow smoother merging since this part made consensus.

PR #404, opened 2018-01-04
[enh] add commands to manage authorized-keys of users

Psycojoker:
Extracted from https://github.com/YunoHost/yunohost/pull/345  to allow smoother merging since this part made consensus.

You need to avoid reading the 2 first commits since they are coming from the previous PR.

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1337,6 +1337,74 @@ dyndns:
 
 
 #############################
+#            SSH            #
+#############################
+ssh:
+    category_help: Manage ssh keys and access
+    actions: {}
+    subcategories:
+        authorized-keys:
+            subcategory_help: Manage user's authorized ssh keys
+
+            actions:
+                ### ssh_authorized_keys_list()
+                list:
+                    action_help: Show user's authorized ssh keys
+                    api: GET /ssh/authorized-keys
+                    configuration:
+                        authenticate: all
+                    arguments:
+                        username:
+                            help: Username of the user
+                            extra:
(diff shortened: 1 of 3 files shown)


Psycojoker:
This PR is ready to be merged for me (with the https://github.com/YunoHost/yunohost/pull/403 which is included inside of this one) and for reviews.

Psycojoker:
Hello,

Calling for a lighter merge on this one because Aleks is busy on stretch (and I prefer him to be on that part of the code than reviewing this) and that it seems that no one else has time for this PR https://github.com/YunoHost/project-organization/blob/master/yunohost_project_organization_fr.md#validation-all%C3%A9g%C3%A9-en-cas-de-manque-de-relecteurs

zamentur:
Why not make the cli like it:

> yunohost user ssh [allow|disallow|add|list|remove] <USER>

?

alexAubin:
Meh... I kinda agreed with @zamentur ... I don't understand why we should split and introduce a whole new category for this ... These are all for SSH, and are all about users, so having something like `yunohost user ssh [allow|disallow|add-key|list-key|remove-key]` seems logical... 

I mean, the technical part is probably fine, but what we can't change once it's release is the CLI/API design so we really be careful about it :/

Psycojoker:
Apparently none of you remember that this is coming from a whole PR introducing several related function for ssh in its own modules that's I've splitted to merged step by step and that this move to a part of the feature to the user module ... is a changed you've asked in the previous PR -_-

Also this proposed API is ambiguous because it doesn't which kind of keys it talks about, private or authorized keys.

PR #406, opened 2018-01-05
[enh] add new api entry point to check for meltdown vulnerability

Psycojoker:
This is for the admin interface, the idea is to show an "alert" to inform about meltdown/spectre and to send to the update/upgrade page and then the reboot page for that.

Debian has yet to release the patch for Jessie and I still need to do the admin UI part.

Still, it would be really great to merge and release this ASAP.


src/yunohost/tools.py
```diff
@@ -624,6 +624,11 @@ def tools_diagnosis(auth, private=False):
         # Domains
         diagnosis['private']['domains'] = domain_list(auth)['domains']
 
+    diagnosis['security'] = {
+        # source https://askubuntu.com/questions/992137/how-to-check-that-kpti-is-enabled-on-my-ubuntu
+        "spectre-meltdown": "cpu_insecure" not in read_file("/proc/cpuinfo")
```
randomstuff: The current Debian kernels do not the patch for shwoing `cpu_inscure` in `/proc/cpuinfo`.

randomstuff: Moreover it would only show up if you *do* have an updated kernel anyway (for meltdown) so you don't have to update (?).

Psycojoker: Our targent is Jessie and debian hasn't released the new kernel on it yet so I don't know how it's going to behave, I'm waiting for the release to see that https://security-tracker.debian.org/tracker/CVE-2017-5754

Are you sure you don't need to reboot? That seems strange, everyone is rebooting everywhere, or do you mean that this will show up in cpuinfo even without a reboot?

randomstuff: 1. The current (updated) kernels in Debian stable/sid don't show `cpu_insecure`.

2. As long as you don't reboot you won't see `cpu_insecure` anyway because you need a new kernel to see `cpu_insecure`: if you are still on an old kernel, the kernel don't know about the bug so he can't tell you cabout it. If you see `cpu_insecure` it means you are on a new kernel which (very probably) knows how to do KPTI (unless it was disabled at compile time) so you don't need to reboot (you already did).

randomstuff: AFAIU

randomstuff: Moreover, the `cpu_insecure` stuff is really only about Metldown, not about Spectre (AFAIU). The Spectre part is not really a CPU bug.

Psycojoker: Ok, thanks for the precisions, we'll need to explore other ways then :)

randomstuff: For Meltdown, maybe something like (expressed in sh):

    if test "$(uname -m)" != x86_64 || grep "/boot/config-$(uname -r)" CONFIG_PAGE_TABLE_ISOLATION=y ; then
      echo OK
    else
      echo "Meltdown (maybe)"
    fi

This would list AMD systems as vulnerable even if they're not.

Or:

    if test "$(uname -m)" != x86_64 || dmesg | grep "Kernel/User page tables isolation:" ; then
      echo OK
    else
      echo "Meltdown (maybe)"
    fi

randomstuff:
[This repo](https://github.com/speed47/spectre-meltdown-checker) is supposed to contain some code to try to detect how your system is wrt Spectre/Metldown.

Psycojoker:
On, interesting, thanks :)

I'll test that on debian once they release the fixed kernel for Jessie.

I'm also considering using the very boring and stupid version of "check if this specific version of the kernel is installed or an higher version" that might also do it.

Psycojoker:
Debian just release its fix for jessie https://security-tracker.debian.org/tracker/CVE-2017-5754

I don't have time to work on it exactly right now but it's going to be ASAP.

Psycojoker:
As expected the current patch doesn't work, I'm looking to adapt/integrate the linked script (which works as expected) but we can only do that work the meltdown case since it's the only fixed CVE for debian right now: https://security-tracker.debian.org/tracker/CVE-2017-5754 (2 other ones https://security-tracker.debian.org/tracker/CVE-2017-5753 https://security-tracker.debian.org/tracker/CVE-2017-5715)

Psycojoker:
Between like and the time it took me to realized that this script had a json output this PR has been quite delayed...

This is now done and ready for review.

src/yunohost/tools.py
```diff
@@ -42,8 +42,16 @@
 from moulinette import msettings, msignals, m18n
 from moulinette.core import MoulinetteError, init_authenticator
 from moulinette.utils.log import getActionLogger
+<<<<<<< b60d8ca822d08c8e3fdf8a17505ff3e285b28164
 from moulinette.utils.process import check_output
 from moulinette.utils.filesystem import read_json, write_to_json
+=======
+from moulinette.utils.filesystem import read_json, write_to_json, read_file
+<<<<<<< a934b3fd19403859cf6a46713b82945524f9ac28
```
M5oul: To be fixed!

data/other/spectre-meltdown-checker.sh
```diff
@@ -0,0 +1,980 @@
+#! /bin/sh
```
M5oul: I would prefer a `git subtree` as this script is going to evolve and is not going to become a Debian package as for `acme-tiny`.

src/yunohost/tools.py
```diff
@@ -42,8 +42,16 @@
...
+>>>>>>> [mod] move spectre-meltdown check to diagnosis function
+=======
+from moulinette.utils.process import check_output
+>>>>>>> [enh] uses speed47 script to check for meltdown vulnerability
```
JimboJoe: Looks like there was a conflict resolution "oops"... ;-)

Psycojoker: weird, I though I had fixed that :x (I had a bug when I've tested that, then commit and pushed but apparently something went wrong >.<)

Psycojoker: okay, find it, I've pushed ... the admin, not the core :sweat_smile: 

Thanks @JimboJoe and @M5oul :heart:

Psycojoker:
Modified to use subtree.

M5oul:
Nice!
It's clean that way 👍

src/yunohost/tools.py
```diff
@@ -632,9 +632,31 @@ def tools_diagnosis(auth, private=False):
...
+def _check_if_vulnerable_to_meltdown():
+    # script taken from https://github.com/speed47/spectre-meltdown-checker
+    # script commit id is store directly in the script
+    SCRIPT_PATH = "/usr/share/yunohost/yunohost-config/moulinette/spectre-meltdown-checker/spectre-meltdown-checker.sh"
+
+    # example output from the script:
+    # [{"NAME":"SPECTRE VARIANT 1","CVE":"CVE-2017-5753","VULNERABLE":true,"INFOS":"only 23 opcodes found, should be >= 70, heuristic to be improved when official patches become available"},{"NAME":"SPECTRE VARIANT 2","CVE":"CVE-2017-5715","VULNERABLE":true,"INFOS":"IBRS hardware + kernel support OR kernel with retpoline are needed to mitigate the vulnerability"},{"NAME":"MELTDOWN","CVE":"CVE-2017-5754","VULNERABLE":false,"INFOS":"PTI mitigates the vulnerability"}]
+    for CVE in json.loads(check_output("bash %s --batch json" % SCRIPT_PATH)):
+        # meltdown https://security-tracker.debian.org/tracker/CVE-2017-5754
+        if CVE["CVE"] == "CVE-2017-5754":
```
alexAubin: Eh why not use `--variant 3` when calling the script instead ? :P

src/yunohost/tools.py
```diff
@@ -632,9 +632,31 @@ def tools_diagnosis(auth, private=False):
...
+        }
+    }
+
     return diagnosis
 
 
+def _check_if_vulnerable_to_meltdown():
+    # script taken from https://github.com/speed47/spectre-meltdown-checker
+    # script commit id is store directly in the script
+    SCRIPT_PATH = "/usr/share/yunohost/yunohost-config/moulinette/spectre-meltdown-checker/spectre-meltdown-checker.sh"
```
alexAubin: Eh.. it ends up being a really weird location ... '`yunohost-config/moulinette`' :/ I don't know why the folder 'other' ends up being 'yunohost-config/moulinette' in the first place.. If we want to keep a clean file tree, we should really consider putting this somewhere else like in `src/yunohost/vendor` or in `data/` directly .. or a `data/misc` or something :confused:.

Psycojoker: We need to modify the `debian/something` then if we have another location.

src/yunohost/tools.py
Psycojoker: Probably because it was 5am when I've managed to do what I wanted :yum:

randomstuff:
Yeah, I still don't really understand the subtree thing.

Github is now all confused and displays all the upstream commits in here ;)

Psycojoker:
Thanks everyone for your help :)

PR #407, opened 2018-01-06
[enh] save the conf/ directory of app during installation and upgrade

Psycojoker:
## The problem 
When install an application (or upgrading it), the "conf" folder is not saved which makes some operations quite hard to do, in particular the `change_url` one. This is a long time requested feature from the app group.

Redmine ticket https://dev.yunohost.org/issues/871

## Solution
Save it #lol

## PR Status
Personally tested, work as expected.

## Validation

- [x] Principle agreement 0/2 : JimboJoe
- [x] Quick review 0/1 : JimboJoe
- [x] Simple test 0/1 : JimboJoe
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -621,9 +621,12 @@ def app_upgrade(auth, app=[], url=None, file=None):
             with open(app_setting_path + '/status.json', 'w+') as f:
                 json.dump(status, f)
 
-            # Replace scripts and manifest
-            os.system('rm -rf "%s/scripts" "%s/manifest.json"' % (app_setting_path, app_setting_path))
+            # Replace scripts and manifest and conf (if exists)
+            os.system('rm -rf "%s/scripts" "%s/manifest.json %/conf"' % (app_setting_path, app_setting_path, app_setting_path))
             os.system('mv "%s/manifest.json" "%s/scripts" %s' % (extracted_app_folder, extracted_app_folder, app_setting_path))
+
+            if os.path.exists(os.path.join(extracted_app_folder, "conf")):
+                os.system('cp -R %s/conf %s' % (extracted_app_folder, app_setting_path))
 
             # So much win
             upgraded_apps.append(app_instance_name)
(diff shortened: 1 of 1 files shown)


Psycojoker:
cc @maniackcrudelis

Psycojoker:
This is related a bit to this pull request : right now applications won't have the `conf` folder saved, I've though a bit about "how to solved this" and once my brain had stop thinking about very complexes solutions (like doing a migration), I've came to a simple one:

- add code in `change_url` that will look if the `conf` folder is present
- if it's not, refuse to do a `change_url` and ask for the user to upgrade the app
- during app upgrade the `conf` folder will be saved
- therefor, `change_url` script only needs to be written for the case where the `conf` folder is present

@JimboJoe @maniackcrudelis and other from the app group who write those kind of script I'd like your opinion on this approach.

Psycojoker:
```
19:53 <Aleks> psycojoker : comprends pas le dernier bullet point, "therefor, change_url script only needs to be written for the case where the conf folder is present" ?
19:53 <@psycojoker> Aleks: aujourd'hui le répertoire "conf" est pas présent donc ils font des hacks pour gérer cette situation
19:53 <@psycojoker> et donc, si on fait ça, y aura plus besoin de gérer la situation où y a conf et la situation où y a pas conf

Psycojoker:
Actually @Josue-T told me that we can simply require the good version of YunoHost for that and it seems that this will solve this issue.

maniackcrudelis:
Wonderful :)

And for your idea about asking for an upgrade, seems ok for me. And we can write a little helper to do that.

JimboJoe:
I totally agree with @Josue-T: when using this feature in change_url script, we'll just have to set the right yunohost version as requirement.
@maniackcrudelis what would you need a helper for?

maniackcrudelis:
For these points:
> - add code in change_url that will look if the conf folder is present
> - if it's not, refuse to do a change_url and ask for the user to upgrade the app

It would be something really simple, but just to avoid to have to rewrite the code each time.

JimboJoe:
Well, if you create/modify your change_url script to depend on that new /conf folder and require the right version of yunohost in the same app version, then... the /conf will be copied, won't it (this PR handles upgrade cases)...?

Psycojoker:
Indeed, like @JimboJoe said, the point is that "if your app is upgraded or installed, you'll have the guaranty to have the conf/ folder and therefor won't have to check if conf/ is present of not because it will be there" :)

PR #408, opened 2018-01-07
[Enh] Logfile for DNSMASQ

frju365:
# Problem

- No log File for DNSMASQ

# Solution

- Add  a Log File

# Status
Finished. Testing/review

### DIFF
--- a/data/templates/dnsmasq/plain/dnsmasq.conf
+++ b/data/templates/dnsmasq/plain/dnsmasq.conf
@@ -1,6 +1,9 @@
 domain-needed
 expand-hosts
 
+log-facility=/var/log/dnsmasq.log
+log-queries
+
 listen-address=127.0.0.1
 resolv-file=/etc/resolv.dnsmasq.conf
 cache-size=256
--- a/data/templates/yunohost/services.yml
+++ b/data/templates/yunohost/services.yml
@@ -6,7 +6,7 @@ avahi-daemon:
    log: /var/log/daemon.log
 dnsmasq:
    status: service
-   log: /var/log/daemon.log
+   log: /var/log/dnsmasq.log
 fail2ban:
     status: service
     log: /var/log/fail2ban.log


JimboJoe:
Not tested but LGTM.

General question for every log; do we try:

-  having a specific log file for every application (here it is easier to retrieve them in the web admin UI)
-  or do we use current features, such as `journalctl -u` (though that needs a core enhancement)?
Here these logs are already available via `journalctl -u dnsmasq`.

I'm voting for the first option right now, but the second one could be interesting.

@frju365 what about log rotation here?

Psycojoker:
> @frju365 what about log rotation here?

I've asked him and he isn't aware of how to do that (and me neither). Are you? If so, would you be able to contribute to this PR?

JimboJoe:
Well, we use [this helper](https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/backend#L14-L54) for the apps.
[This dnsmasq-related procedure](http://gyaresu.org/logging-dnsmasq/) seems to be tested but should be enough to configure dnsmasq log rotation...
Regarding the core organization, would it be better to create a new logrotate configuration file, or append to the [existing one](https://github.com/YunoHost/yunohost/blob/unstable/debian/logrotate) (which was introduced in [this PR](https://github.com/YunoHost/yunohost/commit/8ebfdd217c88c7cb4b0e7f4feb37bfd963757f05))?

Psycojoker:
> Regarding the core organization, would it be better to create a new logrotate configuration file, or append to the existing one (which was introduced in this PR)?

imo it's always better ton have one file per subject but I don't know if this specific situation would require another organization.

data/templates/dnsmasq/plain/dnsmasq.conf
```diff
@@ -1,6 +1,9 @@
 domain-needed
 expand-hosts
 
+log-facility=/var/log/dnsmasq.log
+log-queries
```
maniackcrudelis: To have a log of dnsmasq, you don't need that option.
And this option is going to log every dns request, with the name of the domain requested and the IP of who ask to it.
That means, it's quite tricky regarding the privacy of the users who use the dns.

Also, this option without a logrotate would be a suicide ! I stucked my own server with the 11G log in a few days.

Please have a look here http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

zamentur: Same point of view than Maniack

alexAubin:
Ok guys, it sounds like at least 3 people (counting me) think it's not a good idea to log those DNS queries... so closing it, sowwy :/ (Feel free to reopen if there's more to discuss)

PR #424, opened 2018-01-26
[fix] Simplify get_public_ip

alexAubin:
## The problem

Currently, `get_public_ip(6)` display a dirty traceback and confusing error message for instance when running `dyndns update` with no IPv6 sat up. This repeatedly confused user when debugging some nohost.me/noho.st situations, because it appeared to them that something was broken - when in fact it's just that they did not have IPv6, which is fine.

Also we use `get_public_ip` in many places, but always have to do a `try/except` to keep 'None' (or similar value) in the exception.

## Solution

Just return `None` if no ip was found. Also, move `get_public_ip` to a dedicated 'util' file instead of having it in domain.py leading to weird imports.

## PR Status

Not fully tested, so still kinda WIP

## How to test

Check all occurrences of `get_public_ip` are still working properly.

## Validation

- [x] Principle agreement 0/2 : 
- [x] Quick review 0/1 : 
- [x] Simple test 0/1 : 
- [x] Deep review 0/1 :

### DIFF
--- a/src/yunohost/utils/network.py
+++ b/src/yunohost/utils/network.py
@@ -2,7 +2,7 @@
 
 """ License
 
-    Copyright (C) 2015 YUNOHOST.ORG
+    Copyright (C) 2017 YUNOHOST.ORG
 
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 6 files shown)


src/yunohost/utils/network.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+""" License
+
+    Copyright (C) 2015 YUNOHOST.ORG
```
Psycojoker: > 2015 

#lol

PR #426, opened 2018-01-26
[enh] Make etckeeper stfu

alexAubin:
## The problem

c.f. https://dev.yunohost.org/issues/996

etckeeper regularly dumps a shitload of message, for instance when running `apt-get upgrade` and some new files were added to `/etc/`. Sometimes it's like hundreds of line of /etc/certificate stuff which basically nobody cares about and hide the real `apt-get` log.

## Solution

Add `--quiet` to etckeeper.conf. Since it's not really a service and it's just a static file, I put that in the yunohost regen-conf hook which already handle files like the services.yml. Not sure if that's the right solution in terms of design. I guess we can also just add a 'etckeeper' hook easily if needed.

## PR Status

Should be working

## How to test

Checkout this branch. Try to trigger the etckeeper thing, by creating a file like `/etc/foobar`, then run something like an `apt-get upgrade`.

## Validation

- [x] Principle agreement 2/2 : Bram, ljf
- [x] Quick review 1/1 : ljf
- [ ] Simple test 0/1 : 
- [x] Deep review 1/1 :  ljf

### DIFF
--- a/data/templates/yunohost/etckeeper.conf
+++ b/data/templates/yunohost/etckeeper.conf
@@ -1,0 +1,43 @@
+# The VCS to use.
+#VCS="hg"
+VCS="git"
+#VCS="bzr"
+#VCS="darcs"
+
+# Options passed to git commit when run by etckeeper.
+GIT_COMMIT_OPTIONS="--quiet"
+
+# Options passed to hg commit when run by etckeeper.
+HG_COMMIT_OPTIONS=""
+
+# Options passed to bzr commit when run by etckeeper.
+BZR_COMMIT_OPTIONS=""
+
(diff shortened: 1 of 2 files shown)


data/templates/yunohost/etckeeper.conf
```diff
@@ -0,0 +1,43 @@
+# The VCS to use.
+#VCS="hg"
+VCS="git"
+#VCS="bzr"
+#VCS="darcs"
+
+# Options passed to git commit when run by etckeeper.
+GIT_COMMIT_OPTIONS="--quiet"
```
alexAubin: (Note that this is the part different from the default file in debian)

PR #429, opened 2018-02-02
[enh] Manual migrations

alexAubin:
## The problem

As described / motivated in this issue https://dev.yunohost.org/issues/1067 , we need manual migrations for Stretch (and might need it for some other things in the future, e.g. migration that want to use apt). We also want to integrate this in the admin interface so that it's not mandatory for the user to connect in SSH + use the CLI to perform the migration. 

## Solution

Migrations now have a mode (auto or manual) and `yunohost migrations migrate` can be called with `--auto` that will only run `auto` migrations (to be used in the context of the debian configure thing called after (dist-)upgrade).

Also added a few info (description and disclaimer) that will be useful for UX of the admin interface.

## PR Status

Still WIP, might keep updating it while I work on admin integration

## How to test

(will add this once PR is ready)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Credits goes to @Josue-T for spotting issue fixed by 409c9588acace55d9b620ab5e52ef67d65180446  :wink:

alexAubin:
Merging as agreed during today's meeting...

PR #441, opened 2018-03-15
Add nodejs helpers

maniackcrudelis:
## The problem

NodeJS is a pain in the ass...

## Solution

Add helpers to handle NodeJS installation, remove, upgrade for the version needed by each app.
Allow to use as many version as you need.
Automatic upgrade of each version of node installed.

## PR Status

Already used by many apps, and especially Etherpad, which is going to be official.
Can be review.

## How to test

Try etherpad.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/helpers.d/backend
```diff
@@ -213,3 +213,209 @@ ynh_remove_fpm_config () {
...
+# Install Node version management
+#
+# [internal]
+#
+# usage: ynh_install_n
+ynh_install_n () {
+	echo "Installation of N - Node.js version management" >&2
+	# Build an app.src for n
+	mkdir -p "../conf"
+	echo "SOURCE_URL=https://github.com/tj/n/archive/v2.1.7.tar.gz
```
JimboJoe: When do you think we will have to update that reference version?

data/helpers.d/backend
```diff
@@ -213,3 +213,209 @@ ynh_remove_fpm_config () {
...
+# And 2 alias stored in variables:
+#   - $nodejs_use_version: An alias to load a node version in the current shell. Especially useful for systemd.
+#     NB: $PATH will contain the path to node, it has to be propagated to any other shell which needs to use it.
+#   - $ynh_node_exec: An alias to execute a command with node.
+#
+# usage: ynh_use_nodejs
+ynh_use_nodejs () {
+	nodejs_version=$(ynh_app_setting_get $app nodejs_version)
+
+	load_n_path="[[ :$PATH: == *\":$n_install_dir/bin:\"* ]] || PATH=\"$n_install_dir/bin:$PATH\"; N_PREFIX="$n_install_dir""
```
JimboJoe: Could you please elaborate a little on this line...? :blush:

data/helpers.d/backend
```diff
@@ -213,3 +213,209 @@ ynh_remove_fpm_config () {
...
+# ynh_install_nodejs will install the version of node provided as argument by using n.
+#
+# usage: ynh_install_nodejs [nodejs_version]
+# | arg: nodejs_version - Version of node to install.
+#        If possible, prefer to use major version number (e.g. 8 instead of 8.10.0).
+#        The crontab will handle the update of minor versions when needed.
+ynh_install_nodejs () {
+	# Use n, https://github.com/tj/n to manage the nodejs versions
+	nodejs_version="$1"
+	local n_install_script="https://git.io/n-install"
```
JimboJoe: This variable isn't used anywhere.

data/helpers.d/backend
```diff
@@ -213,3 +213,209 @@ ynh_remove_fpm_config () {
...
+	then
+		n rm $nodejs_version
+	fi
+
+	# If no other app uses n, remove n
+	if [ ! -s "$n_install_dir/ynh_app_version" ]
+	then
+		ynh_secure_remove "$n_install_dir"
+		ynh_secure_remove "/usr/local/n"
+		sed --in-place "/N_PREFIX/d" /root/.bashrc
```
JimboJoe: What case is this supposed to cover?

data/helpers.d/backend
maniackcrudelis: Probably an old line, not used anymore...

data/helpers.d/backend
maniackcrudelis: Unless it's also an old useless line, 'n' will add a line to the .bashrc, to load n.
Maybe, we could remove this line just after the installation.

data/helpers.d/backend
JimboJoe: Well, if we need to maintain these helpers, do we only need to rely on CI results on apps using them to know if there something to change here (link broken, security fix, etc.)?

data/helpers.d/backend
JimboJoe: OK, that would require at least a comment for this :wink:

data/helpers.d/backend
maniackcrudelis: I'm not so sure to understand the point.
We have here the last available release, there's no other release since april 2017.
What else do you want we do ?

data/helpers.d/backend
maniackcrudelis: Let me check if it's still relevant before.
And also see if we can remove it just after the installation.

data/helpers.d/backend
JimboJoe: I'm asking the question for the future maintenance.
Do we need to follow tightly new n version releases?

maniackcrudelis: I don't think so unless there's a big modification of how node can be handle.
But since we used here a specific version, I don't think it would be any problem if there's any new version of n.

Except a big modification of node, or a big security issue about n, I think if it would be any upgrade of n, we can take our time to check it and implement it.

data/helpers.d/backend
maniackcrudelis: So... It's an old useless line...
This helper has been refactored many times...

alexAubin:
Uh sooo can we have this in a dedicated nodejs file pliz ? :stuck_out_tongue:

alexAubin:
So I think we may merge this ?

JimboJoe:
I think alike!

PR #442, opened 2018-03-16
[fix] Don't backup cron files

alexAubin:
## The problem

c.f. https://dev.yunohost.org/issues/933

## Solution

This is an alternative solution to https://github.com/YunoHost/yunohost/pull/334

c.f. my comment here : https://github.com/YunoHost/yunohost/pull/334#issuecomment-316438365 , I personally don't understand the motivation to backup cron jobs... They are neither configuration nor data, in the sense that they should be sat up by the postinstall or other relevant pieces of code... 

For instance, we do not backup/restore the `/etc/nginx/conf.d/domain.tld.conf` files because the information that matters is the existence of the domain.tld domain, not the conf file itself since it is to be regenerated by the regen-conf...

If for instance we decide to change the cron job for dyndns domain for various reasons (e.g. change the frequency or optimize the way it works) then a backup before that moment, followed by a restore, will override the new cron with the old one...

## PR Status

Done / ready for review

## How to test

Create a backup, and check that no cron job is backuped..

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


zamentur:
It's a design question between backup strategy and the regen-conf feature. 

We should backup all /etc (and may be complete mysql databases unreferenced by an app)... User don't know how to add some files to backup, /etc is not so big and there is a lot of configuration (yunohost or user config).

But we are not obliged to restore all part of the system configuration. Some users wants to get a "clean" install after restoring...

About cron, I don't see where is the mechanism to regenerate cron like:
/etc/cron.d/yunohost-applist-*
/etc/cron.daily/yunohost-fetch-appslists
/etc/cron.d/yunohost-dyndns
/etc/cron.d/yunohost-monitor

So remove this restore hook isn't enough.


> For instance, we do not backup/restore the /etc/nginx/conf.d/domain.tld.conf files because the information that matters is the existence of the domain.tld domain, not the conf file itself since it is to be regenerated by the regen-conf...

It's not exact, we backup and restore it...

https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/restore/29-conf_nginx

The regen con mechanism manages conflict with user change. If the user has change some things may be he/she wants to backup it and even restore it.

alexAubin:
So in fact, yes, it appears that we do restore nginx configuration files. But since those are handled by the regen-conf, either : a) the regen-conf is properly reset so that those are re-written with up to date conf automatically, or b) the regen-conf is not properly reset and all restored nginx files are flagged as manually modified, and that might lead to trouble now or later.

Imho, the focus should be on restoring a clean, working setup. Restoring manually-edited configuration files, and/or files which are not handled by the regen-conf, is a hazard in the sense that we cannot ensure that it won't break the destination setups because of backward/forward-incompatible stuff...

But this whole question of backuping / restoring /etc/ as a whole is a different question so I would say it's outside of the scope here (though it's related) ... 

>About cron, I don't see where is the mechanism to regenerate cron

It's the same as the mechanism that initially creates them, which might be different for each cron. But for instance, `/etc/cron.daily/yunohost-fetch-appslists` is created during the postinstall. Crons like `/etc/cron.d/yunohost-applist-*` are outdated / legacy stuff, which is another example of why we shouldn't restore crons imho :/ (though the code handling the automatic migration is still there).

`/etc/cron.d/yunohost-monitor` is created when running `monitor_enable()`... in that case, I doubt that it's something done automatically during postinstall, but anyway I expect that restoring this cron alone doesn't make sense if you don't start and enable glances which is what the rest of `monitor_enable` does...

For `/etc/cron.d/yunohost-dyndns`, it's a bit trickier and needs testing to conclude I think, depending on when / how the other files are restored (in particular the keys) but in any case it's not future-proof to just cp this file imho :s

zamentur:
yunohost-certificate-renew

alexAubin:
So imho the proper solution on this is to : 
- Have https://github.com/YunoHost/yunohost/pull/480
- Integrate a regen-conf script to maintain a single crontab-like file, which contains all the cron managed by yunohost
    - e.g. the regen-conf script shall detect that a dyndns domain is enabled, and include the corresponding cron rule

alexAubin:
https://github.com/YunoHost/yunohost/pull/653 got merged so we could move forward with this

PR #443, opened 2018-03-18
Add wget as dependency

zamentur:
## The problem

It seems wget is not in dependency but we use it with ynh_setup_source ... 

## Solution

Add wget in the dependency.

## PR Status

Untested but very simple

## Validation
I think it's a micro decision so I will just wait if someone want to say something and next merge


Psycojoker:
Can be merged as a microdecision imo.

PR #444, opened 2018-04-03
[add] Http2 protocol

frju365:
## The problem

- http/1 deprecated with the nginx on stretch

## Solution

- upgrade to http/2 protocol

## PR Status

Need test / Finished.

## How to test

big question :P
Requirements : Stretch

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


frju365:
ah yes... In fact, I thought that anyway, the next release would be base on stretch and unstable (I don't know if we will have a release in Jessie before.

frju365:
I changed the base branch... but I don't know if it was what you waited for.

frju365:
Ok. I will do another PR.

PR #447, opened 2018-04-14
display debug information on cert-install/renew failure

Psycojoker:
## The problem

When a `yunohost domain cert-install/renew" failed it's really hard to debug because you just have the information "failed" which sucks and increase our support cost (and we don't want that).

## Solution

Display some debug information to help as shown here:

2018-04-14-134000_469x104_scrot

For now it only displays different ip addresses because that's the case I've encountered. We can add more information later on (or you can add more on this PR if you have some in mind).

## PR Status

Working.

## How to test

Add a domain not well configured for your testing environment and try to do a `cert-install`.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -296,6 +296,7 @@ def _certificate_install_letsencrypt(auth, domain_list, force=False, no_checks=F
                 m18n.n("certmanager_cert_install_success", domain=domain))
 
         except Exception as e:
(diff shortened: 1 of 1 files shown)


frju365:
Very needed. Good work. :)

PR #450, opened 2018-04-21
Lazy load some python imports (perfs improved a lot).

pitchum:
## The problem

One some hardware python imports can be very slow.
For example importing module *requests* can take up to 5 seconds:

    python -m timeit -n 1 -r 1 'import requests'
    1 loops, best of 1: 5.53 sec per loop

## Solution

This PR moves some of these slow imports where they are really needed (this is sometimes called lazy module import).

After that these commands became way faster:
- yunohost app setting ...
- yunohost app list
- yunohost domain list
- yunohost domain dns-conf
- yunohost dyndns installcron/removecron
- ... and maybe others

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -32,7 +32,6 @@ import re
 import urlparse
 import errno
 import subprocess
-import requests
 import glob
 import pwd
 import grp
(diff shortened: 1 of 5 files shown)


pitchum:
I'm a dumb git user, I deleted my own branch by accident...
Don't know how to fix this PR yet but fortunately the commit did not disappear: 723e581

PR #451, opened 2018-04-22
[enh] Lazy-load some module for perf improvements

pitchum:
*Sorry this is a clone of the [previous PR](https://github.com/YunoHost/yunohost/pull/450) that I closed accidentally*

## The problem

On some hardware python imports can be very slow.
For example importing module *requests* can take up to 5 seconds:

    python -m timeit -n 1 -r 1 'import requests'
    1 loops, best of 1: 5.53 sec per loop

## Solution

This PR moves some of these slow imports where they are really needed (this is sometimes called lazy module import).

After that these commands became way faster:
- yunohost app setting ...
- yunohost app list
- yunohost domain list
- yunohost domain dns-conf
- yunohost dyndns installcron/removecron
- ... and maybe others

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -32,7 +32,6 @@ import re
 import urlparse
 import errno
 import subprocess
-import requests
 import glob
 import pwd
 import grp
@@ -129,6 +128,7 @@ def app_fetchlist(url=None, name=None):
     else:
         appslists_to_be_fetched = appslists.keys()
 
+    import requests # lazy loading this module for performance reasons
     # Fetch all appslists to be fetched
     for name in appslists_to_be_fetched:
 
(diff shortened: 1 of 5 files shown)


Psycojoker:
Have been doing some research and apparently they are aware of that https://github.com/requests/requests/issues/3213 and a potential improvement https://github.com/requests/requests/pull/3817 which we don't have in stretch https://packages.debian.org/stretch/python-requests because it's version 2.12.4-1 which was released on 14 Dec 2016 and the PR post-date that.

Psycojoker:
Otherwise PR looks good to me if it's well tested :)

alexAubin:
Proposal to merge this in 3 days if there's no opposition

alexAubin:
(Mistakenly merged into stable, fixed with a force-push and moved it to unstable)

PR #453, opened 2018-04-28
[enh] Local cache when using ynh_setup_source

Jibec:
## The problem

When running a lot of tests (mostly using package_check), downloading takes useless time and resources. As both are limited, I try to cache everything I can.

Debian already has a cache for packages, nothing is to be done for YunoHost: https://github.com/YunoHost/package_check/issues/26

Pip also have one, but including in in core is to be discussed (not here): https://github.com/YunoHost/package_check/issues/25

YunoHost has some mechanics, but:

1. the current system doesn't cache, it only looks into a folder to check if a file is already here
2. the folder used for this cache file looks inconsistent

## Solution

add a simple cache mechanism:
 * if a local file exists: overwrite this cache file if the checksum doesn't match
 * if it doesn't: just download a cache file

I kept it simple and if there is an error, it just tells the user about it (I'm unsure this happens much).

## PR Status

Happily using it while testing application with package_check.

## How to test

You may just copy/paste this function in any application an run package_check

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -150,13 +150,19 @@ ynh_setup_source () {
     fi
     local local_src="/opt/yunohost-apps-src/${YNH_APP_ID}/${src_filename}"
 
+    # simple cache mechanism
+    # if cache file exists and the checksum isn't good, download it again
+    # if not, just download the file
     if test -e "$local_src"
-    then    # Use the local source file if it is present
-        cp $local_src $src_filename
-    else    # If not, download the source
-        wget -nv -O $src_filename $src_url
+    then
+    	echo "${src_sum} ${local_src}" | ${src_sumprg} -c --status \
+        	|| wget -nv -O $local_src $src_url
+    else
+    	mkdir -p "/var/cache/yunohost/ynh_setup_source/${YNH_APP_ID}"
+    	wget -nv -O $local_src $src_url
     fi
-
+    cp $local_src $src_filename
+    
     # Check the control sum
     echo "${src_sum} ${src_filename}" | ${src_sumprg} -c --status \
         || ynh_die "Corrupt source"
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -112,7 +112,7 @@ ynh_backup_before_upgrade () {
 #
 # Details:
 # This helper downloads sources from SOURCE_URL if there is no local source
-# archive in /opt/yunohost-apps-src/APP_ID/SOURCE_FILENAME
+# archive in /var/cache/yunohost/ynh_setup_source/APP_ID/SOURCE_FILENAME
 #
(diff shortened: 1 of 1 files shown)


JimboJoe:
While I think the idea of caching data in our package_check environment is really **great**, I'm more concerned with impacting our helpers here, for two main reasons:
* caching upstream archives shouldn't be useful: I think we more reasonably should work on upgrading upstream apps in our applications only if it has changed (for example by comparing the "upstream" part of the package version), to avoid this unnecessary download, but also unnecessary (and risky) linked operations.
* adding a cache means managing it: how can we ensure the /var partition won't get full?

I'm maybe being too cautious here, but maybe would it be possible to restrict that (very welcome) caching mechanism to package_check environment...?

data/helpers.d/utils
```diff
@@ -112,7 +112,7 @@ ynh_backup_before_upgrade () {
 #
 # Details:
 # This helper downloads sources from SOURCE_URL if there is no local source
-# archive in /opt/yunohost-apps-src/APP_ID/SOURCE_FILENAME
+# archive in /var/cache/yunohost/ynh_setup_source/APP_ID/SOURCE_FILENAME
```
maniackcrudelis: /opt/yunohost-apps-src/ isn't a cache for package.
It's a way to allow users to install a package without an internet connection.

So that you can't just replace it.

alexAubin:
Agreed with @JimboJoe : this is a pretty cool thing, but I would restrict it to package_check for now since that brings many question when it comes to putting this on actual instances.

Maybe we could simply have a flag in a form of a file like `/etc/yunohost/use_caching` which would be enabled by package check ?

Jibec:
I would be fine with not enabling this, but would love to gave a quick (and maybe dirty) alternative.

data/helpers.d/utils
zamentur: Yep, this feature may (might?) be used by some project in Africa/Asia with low/no connectivity.

Jibec: Someone with low/no connectivity will probably want to go into /var/cache/apt/archives/ for debian packages. Many of our packages won't work only without debian packages:https://github.com/search?q=org%3AYunoHost-Apps+ynh_install_app_dependencies&type=Code

I'm unsure what's the issue here, isn't this PR an improvement in this low/no internet scenario?

Do we need to add a readme file telling where to find the new folder?
Btw, this folder is not written in our documentation: https://github.com/YunoHost/doc/search?q=yunohost-apps-src&type=Issues

zamentur: I agree with the idea to change the location. I think, the cache should be manage correctly if the partition behind /var/cache is quite full

Jibec:
I don't have the skill for what's requested here (cache management). I think we can close this pull request.

PR #454, opened 2018-04-28
[enh] ECDH Curves

frju365:
## The problem

- Security problem which affect Diffie-Hellman (DH) key exchange parameters

## Solution

- Add ECDH Curves

## PR Status

- Work Finished/Review needed

## How to test

- Install Yunohost :P

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -19,8 +19,8 @@ server {
     ssl_certificate_key /etc/yunohost/certs/yunohost.org/key.pem;
     ssl_session_timeout 5m;
     ssl_session_cache shared:SSL:50m;
-    # As suggested by Mozilla : https://wiki.mozilla.org/Security/Server_Side_TLS
-    ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
+    # As suggested by Mozilla : https://wiki.mozilla.org/Security/Server_Side_TLS and https://en.wikipedia.org/wiki/Curve25519
+    ssl_ecdh_curve X25519:sect571r1:secp521r1:secp384r1;
 
     ssl_prefer_server_ciphers on;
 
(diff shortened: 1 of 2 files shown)


data/templates/nginx/plain/yunohost_admin.conf
```diff
@@ -19,6 +19,7 @@ server {
     ssl_certificate_key /etc/yunohost/certs/yunohost.org/key.pem;
     ssl_session_timeout 5m;
     ssl_session_cache shared:SSL:50m;
+    ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
```
alexAubin: Uh could you elaborate on where those come from ? For instance for the ciphers we cite the mozilla source which is kinda trustable, since all those things are touchy security stuff...

Also how do you know this related to the DH params ? Does that mean we don't need the ssl_dhparam thingy at all ? Or are those DH params but only for eliptic curve ciphers ?

frju365: Just read this : https://wiki.mozilla.org/Security/Server_Side_TLS ;-)

frju365: It was just to change the description from just "security problem"...

frju365: It's not cyphers btw....

alexAubin: Alrighty, I just moved the instructions right next to the other cipher-related things, since that's the same reference ;)

frju365: euh... but these are the same ecdh crurves for the two... modern or intermediate... so I thought/think it's better to set it at the top. like this the final user will be completely lost and perhaps not enable this.

frju365:
I need test on jessie server for the X25519 curve.

frju365:
I don't think the X25519 works on Jessie. Can someone just add this line on his/her server just to check. It won't break installation for information.

frju365:
ok, it's ok now.

PR #455, opened 2018-04-29
[enh] Control the login shell when creating users in ynh_system_user_create

JimboJoe:
## The problem
Some packages need to create system users with a valid login shell, but the user creation helper only allows nologin shell.

## Solution
Allow this helper to create users using the system default login shell.

## PR Status
Finished and tested.

## How to test
You can possibly test the [Discourse package](https://github.com/YunoHost-Apps/discourse_ynh).

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/user
+++ b/data/helpers.d/user
@@ -41,9 +41,10 @@ ynh_system_user_exists() {
 
 # Create a system user
 #
-# usage: ynh_system_user_create user_name [home_dir]
+# usage: ynh_system_user_create user_name [home_dir [use_shell]]
 # | arg: user_name - Name of the system user that will be create
 # | arg: home_dir - Path of the home dir for the user. Usually the final path of the app. If this argument is omitted, the user will be created without home
+# | arg: use_shell - Create a user using the default login shell if present. If this argument is omitted, the user will be created with /usr/sbin/nologin shell
 ynh_system_user_create () {
 	if ! ynh_system_user_exists "$1"	# Check if the user exists on the system
 	then	# If the user doesn't exist
@@ -52,7 +53,12 @@ ynh_system_user_create () {
 		else
 			local user_home_dir="--no-create-home"
 		fi
-		sudo useradd $user_home_dir --system --user-group $1 --shell /usr/sbin/nologin || ynh_die "Unable to create $1 system account"
+		if [ $# -ge 3 ]; then	# If we want a shell for the user
+			local shell="" # Use default shell
+		else
+			local shell="--shell /usr/sbin/nologin"
+		fi
+		useradd $user_home_dir --system --user-group $1 $shell || ynh_die "Unable to create $1 system account"
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


data/helpers.d/user
```diff
@@ -52,7 +53,12 @@ ynh_system_user_create () {
 		else
 			local user_home_dir="--no-create-home"
 		fi
-		sudo useradd $user_home_dir --system --user-group $1 --shell /usr/sbin/nologin || ynh_die "Unable to create $1 system account"
+		if [ $# -ge 3 ]; then	# If we want a shell for the user
```
alexAubin: Hm I don't know what's the best design for this :sweat_smile: That's a small thing but here you could run `ynh_system_user_create foo /bin/false` and it would actually use the default login shell because the third argument is taken as a boolean argument. Though it's difficult to design an elegant solution ... eitheir explicitly gives the shell to use (e.g. `/bin/bash`), or use literal values like `True / False` or `DEFAULT_SHELL / NO_LOGIN` ?

Or we can keep it like this, but at least an example of usage would be good to clarify this :stuck_out_tongue: 

(Otherwise that looks good to me though not tested)

JimboJoe: ...that's why I was seeking your opinion, because I couldn't find out something completely satisfying myself :sweat_smile:
Providing the shell as parameter seems overkill and prevents from using the default shell (unless you allow using the `default` parameter).
Adding an example is definitely a good start!

maniackcrudelis: Hum, and here using a specific shell means also you have to specify a home directory.
Sometimes, I'm asking myself if we shouldn't switch from positional arguments to getopts arguments for the helpers.

JimboJoe: Provided we keep simple enough helpers (which is a good practice!), I recommend to keep positional arguments.

maniackcrudelis: I found a few hours today, I made a helper for using getopts easily.
https://github.com/YunoHost-Apps/Experimental_helpers/pull/21

@JimboJoe, what do you think about this helper ? Don't you think that using this helper would permit to use "getopts" arguments without the complexity of getopts

JimboJoe: In the case of this helper, what would you propose as parameters definition?
Moreover, any idea as how to manage helper API compatibility with existing apps...?

maniackcrudelis: Anyway, as the helper should currently be tested before be merged in the core. This PR shouldn't wait for it.
So let's use positional arguments here, and we will make a new PR when the new helper getopts will be ready.

alexAubin:
Well, turns out #520 got merged in the meantime ... So we could implement --options right away ? :s

alexAubin:
Bumpz ? :s

alexAubin:
Yolo then !

PR #456, opened 2018-05-01
Tweak the CSP config in nginx template for domains

frju365:
## The problem

- Problem with the CSP

## Solution

- Add CSP by the packagers in their app conf
- More permissive CSP

## PR Status

Not tested / Reviews needed (3)

## How to test

- Install Yunohost and add this line in /etc/nginx/conf.d/domain.conf

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
I'm moving the default-src to the "report-only" header (should be the 'debug' mode of CSP, only display errors in console and does not actually enforce it) for the next release until we figure out what exactly we should do with this :confused:

PR #462, opened 2018-05-09
[enh] Improve ynh_add_nginx_config

Josue-T:
## The problem

- Actually the helper `ynh_add_nginx_config` is not really useful if we have an other variable witch need to replaced.
- After the issue `alias_transversal` we always need uncomment the redirection while we install the app in subdir.

## Solution

Add 2 new features :
- Permit to replace some others variables in the template
- Manage automatically the redirection (alias_transversal issue fix) as we do in each apps.

## PR Status

Tested on two apps (seafile and riot).

## How to test

Test the last version of seafile and riot, or implement this new helper in your app.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -134,7 +134,7 @@ ynh_remove_systemd_config () {
 #   __PORT_2__    by $port_2
 #
 ynh_add_nginx_config () {
-	local finalnginxconf="/etc/nginx/conf.d/$domain.d/$app.conf"
+	finalnginxconf="/etc/nginx/conf.d/$domain.d/$app.conf"
 	local others_var=${1:-}
 	ynh_backup_if_checksum_is_different "$finalnginxconf"
 	sudo cp ../conf/nginx.conf "$finalnginxconf"
@@ -161,14 +161,18 @@ ynh_add_nginx_config () {
 	fi
 
 	# Replace all other variable given as arguments
-	for v in $others_var
+	for var_to_replace in $others_var
 	do
-        ynh_replace_string "__${v^^}__" "${!v}" "$finalnginxconf"
+		# ${var_to_replace^^} make the content of the variable on upper-cases
+		# ${!var_to_replace} get the content of the variable named $var_to_replace 
+		ynh_replace_string "__${var_to_replace^^}__" "${!var_to_replace}" "$finalnginxconf"
 	done
 	
 	if [ "${path_url:-}" != "/" ]
 	then
 		ynh_replace_string "^#sub_path_only" "" "$finalnginxconf"
+	else
+		ynh_replace_string "^#root_path_only" "" "$finalnginxconf"
 	fi
 
 	ynh_store_file_checksum "$finalnginxconf"
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -117,7 +117,10 @@ ynh_remove_systemd_config () {
 
 # Create a dedicated nginx config
 #
-# usage: ynh_add_nginx_config
+# usage: ynh_add_nginx_config "list of others variables to replace"
+#
+# | arg: list of others variables to replace separeted by a space
+# |      for example : 'path_2 port_2 ...'
 #
 # This will use a template in ../conf/nginx.conf
 #   __PATH__      by  $path_url
@@ -126,8 +129,13 @@ ynh_remove_systemd_config () {
 #   __NAME__      by  $app
 #   __FINALPATH__ by  $final_path
 #
+#  And dynamic variables (from the last example) :
+#   __PATH_2__    by $path_2
+#   __PORT_2__    by $port_2
+#
 ynh_add_nginx_config () {
-	finalnginxconf="/etc/nginx/conf.d/$domain.d/$app.conf"
+	local finalnginxconf="/etc/nginx/conf.d/$domain.d/$app.conf"
+	local others_var=${1:-}
 	ynh_backup_if_checksum_is_different "$finalnginxconf"
 	sudo cp ../conf/nginx.conf "$finalnginxconf"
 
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
I haven't check this PR, because I've no time...

But, I'm still really not fond of jinja, which implies source files not easy to read.
So, if it's possible to modify the syntax use in the source file, why not, but if we have to use that kind of syntax: https://github.com/YunoHost-Apps/limesurvey_ynh/blob/master/conf/nginx.conf.j2, it's just not readable...

Josue-T:
> So, I read the code, is it really usefull ?

Yes I think

> I mean, if a packager need to modify other variables than the standard ones, I rather prefer to see a ynh_replace_string before or after this helper to keep it clear.

There are a problem with your idea.

If you can't use `ynh_replace_string` before because you need to keep the original nginx config file clean if you want to use it in the change-url script.

If you try to use `ynh_replace_string` after you might have a problem. By example you need to change an other port in your nginx config, the `nginx reload` command in the helper will fail because nginx won't accept a port value witch look like `__PORT2__`.

An other problem is that you can't just use `ynh_replace_string` after. You also need to reload nginx and after to recalculate the checksum for this file.

maniackcrudelis:
> If you can't use ynh_replace_string before because you need to keep the original nginx config file clean if you want to use it in the change-url script.

Ok, that's a new matter we didn't have so far, and yes, it could be a problem now...
I'm still not so fond of the idea of making this helper more complicated, but I should admit that we could have a problem here.
So, I agree.

But, let's enhance this PR ;)

data/helpers.d/backend
```diff
@@ -151,6 +159,18 @@ ynh_add_nginx_config () {
 	if test -n "${final_path:-}"; then
 		ynh_replace_string "__FINALPATH__" "$final_path" "$finalnginxconf"
 	fi
+
+	# Replace all other variable given as arguments
+	for v in $others_var
```
maniackcrudelis: Please use something clearer than "v" for this variable.

data/helpers.d/backend
```diff
@@ -151,6 +159,18 @@ ynh_add_nginx_config () {
...
+	do
+        ynh_replace_string "__${v^^}__" "${!v}" "$finalnginxconf"
```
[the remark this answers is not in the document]
maniackcrudelis: Thanks for the explanation, could you add this answer as comment in the code please?

data/helpers.d/backend
```diff
@@ -151,6 +159,18 @@ ynh_add_nginx_config () {
...
+	done
+	
+	if [ "${path_url:-}" != "/" ]
+	then
+		ynh_replace_string "^#sub_path_only" "" "$finalnginxconf"
+	fi
```
maniackcrudelis: Good idea to manage that here, that has to be done.
But, as we doing that here, we should also manage `#root_path_only`

data/helpers.d/backend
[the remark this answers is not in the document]
Josue-T: Should be fixed

data/helpers.d/backend
Josue-T: Should be also fixed

alexAubin:
This got merged but I'm really not convinced this is a good thing on the mid/long-term here ...

We need to sit down 5 minutes and look what's happening here. To me this is really a short-term and hackhish fix for a more general issue. We now have ~30 lines just dedicated to do some `ynh_replace_string` just in this helper, and some other like this in the helpers for fpm and systemd... Are we gonna add the same hack for those helpers as well the day we need to include some other variables like in the present case ?

What this shows is that we really need a proper template engine (and we're just implicitly implementing one here). Of course that brings back the question of the jinja templating solution ... And we can probably agree that https://github.com/YunoHost-Apps/limesurvey_ynh/blob/master/conf/nginx.conf.j2 isnt the most readable code in the world, but on the other hand it also seems to be due to the fact that nginx confs are full of regex stuff and so on. If we were talking about a systemd jinja2 template example, that would surely be readable. But if syntax really is a problem, we can also look into changing the delimiters (`{{` and `}}`) to something else ...

I mean, if you take a step back on the way we manage templates right now, it's really weird that `$path_url` becomes `__PATH__` in a template and not `__path_url__` or `__PATH_URL__`, and that it shouldn't be confused with `$final_path` which is `__FINALPATH__` in templates... How are you supposed to guess and understand that as a newcomer (except by reading the doc, yes, but that's provided you found it).

I mean this PR is fine on the short term and I understand the motivation of Josue willing to fix his particular issue, but on the mid/long-term this adds bad API design (which we then won't be able to remove to not break backward compat) without the robustness and powerfulness of a true templating engine...

JimboJoe:
I do share @alexAubin's concerns. I haven't taken time to think about it much, but these arguments make absolute sense. We're just used to our current own "templating system" because we got used to it over the past months/years. It doesn't mean it's the best one.
As we are certainly working on YunoHost on a medium-long term, that subject merits a bit more of thinking/debating.
Considering last weeks exchanges and examples, I do personally find jinja2 implementation more intuitive and more maintainable than our own "templating system"...
I hence also wonder why that PR got merged so quickly :wink:

maniackcrudelis:
Should we create a topic on the forum to discuss about that ?
It could be more efficient than discussing here and there on some PR.

maniackcrudelis:
I let you then create a topic and begin the discussion, because it's more your idea than mine. So you can explain your point of view. And then we could debate about it.

alexAubin:
So I created this topic to try to summarize (though as said, I'm pretty biased) : https://forum.yunohost.org/t/templating-engine-for-app-configuration-helpers/4875

PR #463, opened 2018-05-09
Add ynh_render_template helper to render templates with jinja2

alexAubin:
## The problem

Following some discussion about https://github.com/YunoHost/yunohost/pull/457 and seeing https://github.com/YunoHost/yunohost/pull/462 , and there are some more ideas in the core (e.g. have a global setting to disable https redirection, etc...) we really need a proper templating solution both for the core and for the apps.

Quick-and-dirty templates with search-and-replaces are good if you have simple things to do, but at some point you will find that you need conditional blocks (or even loops maybe) which you can't do easily (you gonna use some hack to comment/uncomment lines).

## Solution

Use a proper template engine : jinja2. @zamentur already used j2cli in some apps, allowing to render templates using (exported) bash variables. For example : https://github.com/YunoHost-Apps/limesurvey_ynh/blob/master/conf/nginx.conf.j2

If you are not familiar with jinja2, look at the example and notice that you can do things like : 
```
{% if path_url != "/" %}
rewrite ^{{ path_url }}$ {{ path_url }}/ permanent;
{% else %}
{% set path_url = "" %}
{% endif %}
```
to handle the infamous "path_url is /" case smoothly.

j2cli is unfortunately only available with pip, so we can't have it as a dependency. But by ~~hacking stuff~~reading stack overflow, I found that we can build a simple helpers which call a few line of python on the fly to do what we want.

Hence with this helper, you can just call something like `ynh_render_template ../conf/my.template /etc/my/output` and voila ! (provided that you exported the variables you need at some point before)

## PR Status

...

## How to test

Source the helper and try it on a test template

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -234,3 +234,20 @@ ynh_local_curl () {
 	# Curl the URL
 	curl --silent --show-error -kL -H "Host: $domain" --resolve $domain:443:127.0.0.1 $POST_data "$full_page_url"
 }
+
+# Render templates with Jinja2
+# 
+# Attention : Variables should be exported before calling this helper to be
+# accessible inside templates.
+#
+# usage: ynh_render_template some_template output_path
+# | arg: some_template - Template file to be rendered
+# | arg: output_path   - The path where the output will be redirected to
+ynh_render_template() {
+   local template_path=$1
+   local output_path=$2
+   # Taken from https://stackoverflow.com/a/35009576
+   python2.7 -c 'import os, sys, jinja2; sys.stdout.write(
+                    jinja2.Template(sys.stdin.read()
+                 ).render(os.environ));' < $template_path > $output_path
+}


maniackcrudelis:
I rather prefer:

```
if [ "$path_url" != "/" ]
then
	ynh_replace_string "^#sub_path_only" "" "../conf/nginx.conf"
fi
```
and
```
#sub_path_only rewrite ^__PATH__$ __PATH__/ permanent;
```
to
```
{% if path_url != "/" %}
rewrite ^{{ path_url }}$ {{ path_url }}/ permanent;
{% else %}
{% set path_url = "" %}
{% endif %}
```

The first version seems far clearer then the second one

maniackcrudelis:
Jinja2 is clearly more efficient than sed, but I don't think that we should use jinja2 for apps, since the idea is to keep apps packages simple and accessible for everyone.
At least if we can't have a simpler syntax than that.

Psycojoker:
{% if path_url != "/" %}rewrite ^{{ path_url }}$ {{ path_url }}/ permanent;{% endif %}

Is really not that complicate you know.

Also the `{%` `%}` `{{` `}}` can be modified if needed.

maniackcrudelis:
If `{% %}` `{{ }}` can be modified, it could be interesting.

Not that complicated ? I'm not sure, I asked to someone else to have a look to the 2 methods, clearly the second one seems complicated to read to him.

I'm not against this helper, just that I don't want to see jinja2 used everywhere in our apps. Except if we can have a clear syntax easy to read.

Psycojoker:
> Not that complicated ? I'm not sure, I asked to someone else to have a look to the 2 methods, clearly the second one seems complicated to read to him.

Your second one or the version I've gave you?

Because yes, your version is clearly more complicated :sweat_smile: but your bash one have a bug because it missed a second sed and needs to looks at 2 different files to understand what is happening :/ (and can spread during the whole bash file so it's really hard to see exactly what is happening and modifying to file) while the other one do everything at the same place/file and is more robust (and more powerful).

maniackcrudelis:
I picked up the version in the first comment.
There's nothing missing in my bash version, this part is already handle by the helper.

Again, I'm not against jinja2, but I'm against this syntax not readable.
Using `{{ path_url }}` instead of `__PATH_URL__` is more difficult to read if you have to find where there's variables in a file to be changed.
Having a line like this one (better than the first proposition though): `{% if path_url != "/" %}rewrite ^{{ path_url }}$ {{ path_url }}/ permanent;{% endif %}` isn't easy to read if you don't know the syntax of nginx. It's becoming hard to know what is jinja2 and what is part of the real config file.
A situation we should really think of with all the other config files we can have, specific to apps, or in other languages.

zamentur:
This helper could help in various situation, I have seen some crazy sed to remove commentary to enable some feature in configuration... This helper could manage it properly.

With this helper it's possible to add several variables, could be very useful in helpers defining a configuration (like nginx or php-fpm helpers).

My point of view is: templates (with a template engine) is the cleaner way to manage the problem of a config with some variables.

maniackcrudelis:
> This helpers d'ont force people to use it.

Yes it will !
Unless you're ready to sign with your blood that you're about to maintain your apps for all their life time, you will force other packagers to use it.
That's all the problem I'm pointing out since the beginning. We're not alone to work on our apps, even if we maintain it.
I already have to work on some apps with specific parts like this helper propose. And I really don't want to do that again, just to have some more abstraction...

alexAubin:
I don't see how that forces people to use it ... One of the motivation for this PR is that we need it for stuff in the core, so from that point of view this is kinda independent for the apps part.

In fact, regarding the apps, from my point of view it's rather the opposite : people are forced to use the current nginx / php / systemd / ... helpers if they want to have helpers that will evolve in the future and therefore increases the robustness of the apps - instead of crafting their own. Therefore, people are forced to use the template system implemented in these helpers...

maniackcrudelis:
> this is kinda independent for the apps part.

Are you kidding me !!!???
> we really need a proper templating solution both for the core and for the apps.
[...]
already used j2cli in some apps

And also the example...

So, you need that for the core, well, it's none of my concern. Put that code somewhere else than the apps helpers and I'll not say anything...

alexAubin:
To clarify : considering that's there's currently no consensus on the app template engine solution, and that we need this for the core, the decision to merge this remains kinda independent of what's decided for the apps, though it would be nice to use the same technology.

>Put that code somewhere else than the apps helpers and I'll not say anything...

`data/helpers.d` isn't the "app helpers", or at least they are already used by the core as well, c.f. : 
https://github.com/YunoHost/yunohost/search?p=1&q=%22yunohost%2Fhelpers%22&unscoped_q=%22yunohost%2Fhelpers%22
so I don't see where else we should put this ...

maniackcrudelis:
For the record, and even if this helper is only for the core.
Any app which will use that template will be immediately and completely out of my scope.
Things are said...

PR #469, opened 2018-05-11
move service_status to sytemd via dbus

Psycojoker:
## The problem

Old code is super hackish and limitated and ugly.

## Solution

Move to systemd via dbus (yes that's ugly code thanks dbus) to have nicer high level code that don't rely on hack.

Bonus:

* uniformisation of how to access status information, we don't need this "status" key anymore
* we get way more useful information: active state, description (!) and since how much time this service is running

## PR Status

Ready to be merged.

## How to test

Go to branch, then:

* yunohost service regen-conf (IMPORTANT)
* yunohost service status

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

Example new output:

2018-05-11-045511_729x809_scrot


Psycojoker:
Oh, good catch :o 

And no, haven't tested on the cube, that's supposed to be generic everywhere since it's only about systemd.

Will investigate in case of.

Psycojoker:
Btw, there is a less totally crapy ugly api python library that exists for dbus https://github.com/LEW21/pydbus

But it's only available in stretch :(

That might be worth it to migrate in the future if we uses dbus more (that might open interesting possibilities apart from the idious API).

Psycojoker:
I chose to keep backward compatibility and tested the admin interface.

I should be good now :) (I'm making other invastigation in the mean time)

Psycojoker:
Also this remove ssl and nsswitch from the admin interface, which I think is good for UX, I don't think it's going to cause a problem but I wanted to point it out.

Psycojoker:
Discussed with Aleks, good to go

PR #471, opened 2018-05-11
[fix] Untrusted TLS connection established to

zamentur:
## The problem

In mail log we can found some "Untrusted TLS connection established" message

## Solution

Indicate where are certs

## PR Status

Tested

## How to test

Apply this patch, reload postfix, and try to send a mail, you should see in log this message:

"Trusted TLS connection established to"

## Validation

- [x] Principle agreement 2/2 : Aleks, Jimbojoe
- [x] Quick review 1/1 : Aleks
- [x] Simple test 1/1 : Jimbojoe
- [x] Deep review 1/1 : Jimbojoe

### DIFF
--- a/data/templates/postfix/main.cf
+++ b/data/templates/postfix/main.cf
@@ -45,7 +45,8 @@ smtp_tls_exclude_ciphers = $smtpd_tls_exclude_ciphers
 smtp_tls_mandatory_ciphers= $smtpd_tls_mandatory_ciphers
 smtp_tls_loglevel=1
 
-# Fix "Untrusted TLS connection established to" message in log
+# Configure Root CA certificates
+# (for example, avoids getting "Untrusted TLS connection established to" messages in logs)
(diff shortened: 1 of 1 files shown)


alexAubin:
@JimboJoe 
> Maybe change the log to refer to what we configure vs what we "fix"...?

Not sure what you mean ? o.O

JimboJoe:
Oops I meant the comment, not the log :relieved:

alexAubin:
Still not sure what you mean xD 

Do you mean it's weird that you don't have these errors on your server so the current comment is incorrect / not precise enough ? Personally I don't understand how it can work (the 'trusting' of an SSL connection) without this ... or maybe there's a default value which works on your setup for some reason :s

JimboJoe:
Sorry, I can eventually write a complete sentence in front of a computer :innocent:
The comment states
> Fix "Untrusted TLS connection established to" message in log

which describes a fix.

Shouldn't we formulate the comment to factually explain this configuration, by instance:
> Configure Root CA certificates

PR #472, opened 2018-05-12
bunch of cleaning in services.py

Psycojoker:
Microdecision, waiting a bit in case of.

If you want to review look for a "per commit" basis, all of those are atomic.

I've added a bunch on debug logging to avoid hiding some errors, we might get reports from users due to that.

### DIFF
--- a/src/yunohost/service.py
+++ b/src/yunohost/service.py
@@ -573,6 +573,7 @@ def _give_lock(action, service, p):
     return son_PID
 
 def _remove_lock(PID_to_remove):
+    # FIXME ironically not concurrency safe because it's not atomic...
 
     PIDs = filesystem.read_file(MOULINETTE_LOCK).split("\n")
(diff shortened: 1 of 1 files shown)


src/yunohost/service.py
```diff
@@ -498,11 +511,11 @@ def _run_service_command(action, service):
...
         raise MoulinetteError(errno.EINVAL, m18n.n('service_unknown', service=service))
 
-    cmd = None
-    if action in ['start', 'stop', 'restart', 'reload', 'enable', 'disable']:
-        cmd = 'systemctl %s %s' % (action, service)
-    else:
-        raise ValueError("Unknown action '%s'" % action)
+    possible_actions = ['start', 'stop', 'restart', 'reload', 'enable', 'disable']
+    if action not in possible_actions:
+        raise ValueError("Unknown action '%s', available actions are: %s" % (action, ", ".join(possible_actions)))
```
alexAubin: Maybe an assert here would be slightly more elegant, but that's already pretty better than before :+1:

PR #474, opened 2018-05-14
[mod] display backup location

Psycojoker:
Otherwise people don't know where to find it (had the case on the dev chan).

We should always give this kind of "obvious when you know it" information.

Microdecision, waiting a bit in case of.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -80,7 +80,7 @@
     "backup_cleaning_failed": "Unable to clean-up the temporary backup directory",
     "backup_copying_to_organize_the_archive": "Copying {size:s}MB to organize the archive",
     "backup_couldnt_bind": "Couldn't bind {src:s} to {dest:s}.",
-    "backup_created": "Backup created",
+    "backup_created": "Backup created and stored in /home/yunohost.backup/archives under the name {}",
     "backup_creating_archive": "Creating the backup archive...",
     "backup_creation_failed": "Backup creation failed",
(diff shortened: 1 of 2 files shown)


locales/en.json
```diff
@@ -80,7 +80,7 @@
     "backup_cleaning_failed": "Unable to clean-up the temporary backup directory",
     "backup_copying_to_organize_the_archive": "Copying {size:s}MB to organize the archive",
     "backup_couldnt_bind": "Couldn't bind {src:s} to {dest:s}.",
-    "backup_created": "Backup created",
+    "backup_created": "Backup created and stored in /home/yunohost.backup/archives under the name {}",
```
JimboJoe: Instead of having this directory in the locale constants, shouldn't we pass `ARCHIVES_PATH` as a parameter to the logging instruction?
One could imagine being able to set this path to somewhere else in the future.

alexAubin: In fact I wonder if the actual location of the backup isn't backup-method dependent ... for instance if you use a remote borg backup method, it won't be at all in this location.

Maybe making this message backup-method independent is complicated, so maybe a simple solution would be something like 

```
if backup_method == "tar":
    ... display this message ...
```

Psycojoker: I guess it's the job of the method to provide this information then because the manager doesn't have it.

The code doesn't look ready for that from here but it's very dense and I haven't took the time to dig in it.

alexAubin:
Closing as of today's meeting / Superseded by the other major PR on backup location

PR #476, opened 2018-05-17
Translate service descriptions

Psycojoker:
## The problem

Services description are very bad (like most of them means nothing for regular users) and untranslatable.

Wonderful example:

    Start/stop fail2ban
    Starts and daemonize Glances server

I've gathered the services description in a pad so anyone can easily contribute and makes them having more sens (because I'm bad at that) here, please contribute ™: https://pad.aquilenet.fr/p/8Cg5Miv6Of

## Solution

Make them translatable and overwrite their description when needed.

## PR Status

Need better wording for descriptions.

## How to test

Launch the service status pages.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
Rebased, now we have a clean diff

Psycojoker:
I've pushed a first dump from the pad.

Except if other people joins to contribute we should be ready to merge in a few days.

Psycojoker:
Worst case we can modify that later, it's cheap to do a PR for that.

Psycojoker:
VEWY PWETTY
2018-05-18-043942_1343x920_scrot

PR #478, opened 2018-05-19
We need to set status:null for yunohost pseudo-service

alexAubin:
## The problem

Following #469, we need to add status: null for "pseudo-services" in services.yml. There's a pseudo-service "yunohost" which was forgotten

## Solution

Add `status: null` for the yunohost key in services.yml

## PR Status

Microdecision / waiting for Bram's feedback

## How to test

If you're on the latest unstable, `yunohost service status` should be broken because it can't fetch the status of the "yunohost" service.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/yunohost/services.yml
+++ b/data/templates/yunohost/services.yml
@@ -37,6 +37,8 @@ nslcd:
   log: /var/log/syslog
 nsswitch:
   status: null
+yunohost:
+  status: null
 bind9: null
 tahoe-lafs: null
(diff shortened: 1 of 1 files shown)


Psycojoker:
That's weird because my setup never bought up that service :|

But let's go forward if that work's for you (didn't broke anything for me)

PR #480, opened 2018-05-21
Decouple the regen-conf mechanism from services

alexAubin:
## The problem

Regen-conf is currently heavily coupled to the notion of services, and this is problematic / annoying for many reasons : 
- we have "pseudo-services" like "ssl", "nsswitch" and "yunohost" which are not actual services but just here to handle some configuration files
- it's annoying to create a whole service, then add all the hack to make it "not a real service" when you in fact just want to manage a single file. This is why I ended up putting the management of etckeeper.conf inside the "yunohost" regen-conf hook
- we have other examples of files we would like to manage independently of the notion of service, for instance the dns resolvers (c.f. #352 ) and cron jobs (dyndns, cert-renewal)

## Solution

Decouple the regen-conf mechanism from the concept of services. To do this, the regen-conf related code was moved in a regenconf.py and file hashes are now stored in a regenconf.yml, independently from services.yml.

Of course that's a delicate operation since the regen-conf is one of the core feature of the core :o and a migration is required to properly handle this ... (which is why I ended up working on stretch-unstable to add a 6th migration)

The API also gets changed from `yunohost service regen-conf` (still supported but deprecated) to `yunohost tools regen-conf`.

Nevertheless, after this we'll be able to safely add simple conf-regen hooks for new files without all the weird "pseudo-service" thing.

## PR Status

Needs some heavy testing to make sure this doesn't break anything

## How to test

Pull this branch ... the first time you run a regen-conf, it should automatically migrate. Though we also need to test postinstall and a "real-life upgrade" (which triggers the postinst script and so on)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/regenconf.py
```diff
@@ -0,0 +1,518 @@
...
+
+    # Legacy code to automatically run the migration
+    # This is required because regen_conf is called before the migration call
+    # in debian's postinst script
+    if os.path.exists("/etc/yunohost/installed") \
+       and ("conffiles" in read_file("/etc/yunohost/services.yml") \
+            or not os.path.exists("/etc/yunohost/regenconf.yml")):
+        from yunohost.tools import _get_migration_by_name
+        migration = _get_migration_by_name("separate_regenconf_from_services")
+        migration.migrate()
```
zamentur: I don't know well the migration system. Is there a risk that a user running the regen-conf migrate accidentally to stretch ?

Psycojoker: Normally no, some migrations needs to be explicitly accepted iirc, @alexAubin will confirm.

alexAubin: This is to be migrated in stretch-unstable, so implicitly people will be already on stretch

alexAubin:
Superseded by https://github.com/YunoHost/yunohost/pull/653

PR #481, opened 2018-05-24
Move avahi-daemon to recommended, fix issues#1097

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1097

## Solution

Move avahi-daemon to recommended in debian control file

## PR Status

Weeeell not tested. It should be okay but maybe some more work is needed to adapt the regen-conf because there are explicit references to avahi-daemon in it ...

## How to test

Find a way to uninstall avahi-daemon on an instance and test the regen conf :s

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -22,7 +22,7 @@ Depends: ${python:Depends}, ${misc:Depends}
  , dovecot-ldap, dovecot-lmtpd, dovecot-managesieved
  , dovecot-antispam, fail2ban
  , nginx-extras (>=1.6.2), php5-fpm, php5-ldap, php5-intl
- , dnsmasq, openssl, avahi-daemon, libnss-mdns, resolvconf, libnss-myhostname
+ , dnsmasq, openssl, libnss-mdns, resolvconf, libnss-myhostname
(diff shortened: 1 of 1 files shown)


Psycojoker:
Isn't that going to break regenconf and service stuff https://github.com/YunoHost/yunohost/blob/1ec8b4d5c3661aedddf8acec4ec16d71679c4d16/data/templates/yunohost/services.yml#L3? https://github.com/YunoHost/yunohost/search?q=avahi&unscoped_q=avahi

alexAubin:
Yes :stuck_out_tongue: 

Hence the : 

>maybe some more work is needed to adapt the regen-conf because there are explicit references to avahi-daemon in it ...

alexAubin:
Closing because not sure it's the right strategy ... several people including me have no particular trouble installing YunoHost on a LXC nowadays ? The issue seems trickier than just "Avahi is incompatible with LXC" ...

PR #482, opened 2018-05-26
backup filename limit set to 50

Psycojoker:
## The problem

Apparently 30 chars is too short https://github.com/YunoHost/issues/issues/962

## Solution

Move to 50 as requested (shouldn't we put more actually? I don't see the point of the limit)

## PR Status

Gud

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
I don't know, I've never touch to backup so maybe there was a good reason? Could we increase it even more?

PR #486, opened 2018-05-31
[enh] POC for app actions

Psycojoker:
## The problem

A lot of actions can or needs to be run on the CLI level for a lot of apps (like all owncloud commands, same for wallabag) and more actions could be interesting for applications (import/export/upgrade/fixs/set admin/enable or disable stuff/etc...).

In general that increase the power of app developpers that are already doing a lot of cool things.

CLI api:

```
yunohost app action list $app_id
yunohost app action run $app_id $action -a $arguments --verbose
```

All the app installation arguments mechanism as been reused to be coherent and because that's already working well.

The application must ship a file named `actions.json` stored at the same level than its `manifest.json`.

Here is an example:

```
[{
        "id": "restart_service",
        "name": "Restart service",
        "command": "echo pouet $YNH_ACTION_SERVICE",
        "user": "root",  # optional
        "cwd": "/", # optional
        "accepted_return_codes": [0, 1, 2, 3],  # optional
        "description": {
            "en": "a dummy stupid exemple or restarting a service"
        },
        "arguments": [
            {
                "name": "service",
                "type": "string",
                "ask": {
                    "en": "service to restart"
                },
                "example": "nginx"
            }
        ]
},
{
    ...  # other action
}]
```

Sample run:

```
root@yunohost:/etc/yunohost/apps/zerobin# yunohost app action run zerobin restart_service -a service=prout --verbose
Executing script '/tmp/tmp3CA5II'...
+ echo pouet prout
pouet prout
Success! Action successed!
```

## PR Status

Working but might probably need more work for more things? That could be merged as an experimental feature the time to get feedback from apps developpers.

Working on the admin interface will probably requires more code, like in `app list --raw` but I don't like HUGE PR that nobody review.

Todo:

- [ ] refactor code to be DRY regarding arguments parsing
- [ ] start a POC for the admin, this will probably raised needed modifications to the API

## How to test

Create a `actions.json` file in `/etc/yunohost/apps/$app_id/actions.json` like in the example then play with the commands.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -785,6 +785,20 @@ app:
                     app_id:
                         help: app id
 
+              ### app_action_run()
+              run:
+                  action_help: Run app action
+                  api: PUT /<app>/actions/<action>
+                  arguments:
+                    app_id:
+                        help: app id
+                    action:
+                        help: action id
+                    -a:
+                        full: --args
+                        help: Serialized arguments for app script (i.e. "domain=domain.tld&path=/path")
+
+
       config:
           subcategory_help: Applications configuration panel
           actions:
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -773,6 +773,18 @@ app:
 
     subcategories:
 
+      action:
+          subcategory_help: Handle apps actions
+          actions:
+
+              ### app_action_list()
+              list:
+                  action_help: List app actions
+                  api: GET /<app>/actions
+                  arguments:
+                    app_id:
+                        help: app id
+
       config:
           subcategory_help: Applications configuration panel
           actions:
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1367,6 +1367,31 @@ def app_change_label(auth, app, new_label):
     app_ssowatconf(auth)
 
 
+# ACTIONS todo list
+# * save actions.json
+# commands:
+# yunohost app action list $app
+# yunohost app action run $app $action -d parameters
+# docstring
+
+def app_action_list(app_id):
+    installed = _is_installed(app_id)
+    if not installed:
+        raise MoulinetteError(errno.ENOPKG,
+                              m18n.n('app_not_installed', app=app_id))
+
+    actions = os.path.join( APPS_SETTING_PATH, app_id, 'actions.json')
+
+    if not os.path.exists(actions):
+        return {
+            "actions": [],
+        }
+
+    return {
+        "actions": read_json(actions),
+    }
+
+
 # Config panel todo list:
 # * docstrings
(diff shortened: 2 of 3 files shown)


zamentur:
`        "command": "echo pouet `$YNH_ACTION_SERVICE",``

We might use bash script like for install/upgrade/restore or hooks no ?

Some idea of specific actions:

- load a .cube (for vpnclient_ynh)
- rescan files (for nextcloud_ynh)
- switch public to private or change a specific settings
- scan wordpress integrity ?
- import a db from an other installation without yunohost ?
- test if the app is correctly working (may be this kind of action should be run without authentication ?)
- ????

Psycojoker:
> We might use bash script like for install/upgrade/restore or hooks no ?

They are run as bash script, you can write whatever you want (like `bash /path/to/script`) but I was expecting a lot of action to be single simple commands instead of having to write a lot of small files for that so I added this possibility to make the already hard life of app devs easier.

Psycojoker:
No exactly related but this will increase the desire for it: we'll need a way to easily upload/download files from the admin interface (and from cli I guess, like by providing a http link to it)

Subject to discuss but not needed for a first version: allow script to return something (in json?) that can be used for the admin interface to provide download link/other.

Example:

* download an export
* grab the result text from a command
* etc...

But I would be, again, more interested in gradual integration of this feature instead of waiting 2 years to have the perfect-do-everything-but-without-feedback version.

maniackcrudelis:
A first real use case of this feature here https://github.com/YunoHost-Apps/wordpress_ynh/pull/47.
Work as expected, big thank to Bram for this wonderful work !

Psycojoker:
Following feedback from Maniack, default cwd is now `/etc/yunohost/apps/$app_id/`

Psycojoker:
After a discussion with Maniack I've switch the format to a list instead of a dict because I'm expecting app developpers to want to be able to control in which order actions appears.

alexAubin:
Just randomly jumping in the discussion /o/ 

Are the motivations to have a separate `actions.json` file rather than having something like an `extra_actions` key in the `manifest.json` ? (I think that could avoid some copypasta in some places like app.py or the list_builder where there are specific behavior related to `manifest.json` ?)

Psycojoker:
Sound nicer to separate those different concepts, also that could make the manifest enormous and I wanted to avoid that.

In my experience splitting things generally makes a better organisation (and "actions" seemed different enough from the content of the manifest to put them somewhere else for me).

But that's really a question of taste I think.

Psycojoker:
Implemented the experimental warning as discussed in the meeting.

Merging it as we are going to move on a short circle dev of feedback from the app team and don't advertise this feature and discourage its usage as it is not finished yet.

This one still needs an admin interface.

PR #487, opened 2018-06-01
[enh] Verbose by default

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/971

## Solution

Set verbose by default and remove the corresponding option (`--verbose`). This also implies to reflag some `INFO` message as `DEBUG`, and some `WARNING` as `INFO`.

## PR Status

Should be okay, but this goes with https://github.com/YunoHost/yunohost/pull/383 and https://github.com/YunoHost/moulinette/pull/155

## How to test

Pull the branch in and run some command which will display some INFO messages ... Note that you need https://github.com/YunoHost/moulinette/pull/155 (the locales/en.json part) to have the nice Info: prefix.

## Validation

- [x] Principle agreement 2/2 : Bram, ljf
- [x] Quick review 1/1 : Bram
- [x] Simple test 1/1 : ljf
- [x] Deep review 0/1 : ljf

### DIFF
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -355,13 +355,13 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
     command.append(cmd.format(script=cmd_script, args=cmd_args))
 
     if logger.isEnabledFor(log.DEBUG):
-        logger.info(m18n.n('executing_command', command=' '.join(command)))
+        logger.debug(m18n.n('executing_command', command=' '.join(command)))
     else:
-        logger.info(m18n.n('executing_script', script=path))
+        logger.debug(m18n.n('executing_script', script=path))
 
     # Define output callbacks and call command
     callbacks = (
-        lambda l: logger.info(l.rstrip()),
+        lambda l: logger.debug(l.rstrip()),
         lambda l: logger.warning(l.rstrip()),
     )
     returncode = call_async_output(
--- a/bin/yunohost
+++ b/bin/yunohost
@@ -9,8 +9,8 @@ import argparse
 IN_DEVEL = False
 
 # Level for which loggers will log
-LOGGERS_LEVEL = 'INFO'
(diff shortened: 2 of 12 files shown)


zamentur:
We might want a quiet option that just display the result without any other messages

alexAubin:
>Seems good to me, during my test an INFO from moulinette has been displayed with --debug option (not without). Is it ok ?

I think yea, I'm not sure to understand it 100%, but that's probably related to the way the loggers are enabled and is fine imho (as long as it only shows up with --debug)

>We might want a quiet option that just display the result without any other messages

Don't we have this already ?

zamentur:
You are right, this quiet option return only the result without log message. I was believing it was a complete quiet option (no result display).

src/yunohost/hook.py
```diff
@@ -355,13 +355,13 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
-        logger.info(m18n.n('executing_command', command=' '.join(command)))
+        logger.debug(m18n.n('executing_command', command=' '.join(command)))
     else:
-        logger.info(m18n.n('executing_script', script=path))
+        logger.debug(m18n.n('executing_script', script=path))
 
     # Define output callbacks and call command
     callbacks = (
-        lambda l: logger.info(l.rstrip()),
+        lambda l: logger.debug(l.rstrip()),
```
alexAubin: @maniackcrudelis Beware of this change : I don't know how that impacts the whole package_check / app CI thing but the whole detail of app installs will now start with 'DEBUG' at the beginning of the line instead of 'INFO' (on stretch). Also `--verbose` won't exist anymore

maniackcrudelis: Actually the CI uses --debug, not --verbose and looks for the lines marked WARNING and ERROR.
So, it shouldn't break anything on our CI.
Anyway, our CI stretch should show us if something doesn't work anymore.

PR #489, opened 2018-06-13
[fix] Redirect corectly php5 to php7 file during restore

zamentur:
## The problem

I tried to restore a backup of nextcloud made in jessie on my new stretch instance.

Was not working due to bad php5/php7 management.
https://github.com/YunoHost/issues/issues/1140

## Solution

I use here the backup.csv fileto redirect correctly the file in the good place.

I also delete a _common.sh file, because it's a bad pactice of packaging, that stores accidentally this file in /backup dir.

## PR Status

Untested, I just test the methodology, but i have done a sed manually and delete the _common.sh with pdb.

## How to test

Try to restore a nextcloud from jessie on a stretch version. Other php app are concerned too.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -103,6 +103,7 @@
     "backup_output_directory_not_empty": "The output directory is not empty",
     "backup_output_directory_required": "You must provide an output directory for the backup",
     "backup_output_symlink_dir_broken": "You have a broken symlink instead of your archives directory '{path:s}'. You may have a specific setup to backup your data on an other filesystem, in this case you probably forgot to remount or plug your hard dirve or usb key.",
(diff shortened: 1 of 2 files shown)


JimboJoe:
Could you please elaborate on why using a `_common.sh` is a bad practice for packaging...? :thinking:

zamentur:
The bad pratice is to copy _common.sh in the backup dir, like here: https://github.com/YunoHost-Apps/nextcloud_ynh/blob/master/scripts/backup#L11

It triggers a bug during the restore. It uses this _common.sh copied instead of the one in /home/yunohost.backup/tmp/apps/APP/settings/scripts/ ...

I think make `source ../settings/scripts/_common.sh` should be better

Psycojoker:
Thanks for the modifications :heart:

PR #492, opened 2018-06-14
OCSP Stapling

frju365:
## The problem

- Every time, the browser needs to send a request to LE to check if the SSL Cert is not revoked.

## Solution

- Add OCSP Stapling

## PR Status

Work finished.
Review required.

## How to test

Install Yunohost + nginx

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/hooks/conf_regen/15-nginx
```diff
@@ -38,6 +38,15 @@ do_pre_regen() {
   for domain in $domain_list; do
     domain_conf_dir="${nginx_conf_dir}/${domain}.d"
     mkdir -p "$domain_conf_dir"
+    domain_ca_type=$(yunohost domain cert-status ${domain} | grep CA_type)
+    if [$domain_ca_type = "CA_type: Let's Encrypt"]; then
```
Psycojoker: typo, there should be a space after the "[" otherwise you'll get a syntax error in bash.

Have you tested this code? Same for the other "]".

Also the variable should be quoted here because of spaces.

frju365: Well indeed I change it in the HPKP PR :) I will correct that.

frju365:
Well, it can be quite good to have it. It will prevent servers from sending request to LE each time to check if cert is revoked or not.

data/hooks/conf_regen/15-nginx
```diff
@@ -38,6 +38,15 @@ do_pre_regen() {
   for domain in $domain_list; do
     domain_conf_dir="${nginx_conf_dir}/${domain}.d"
     mkdir -p "$domain_conf_dir"
+    domain_ca_type=$(yunohost domain cert-status ${domain} | grep CA_type | tr -d " ")
+    if [[ "$domain_ca_type" = "CA_type:Let'sEncrypt" ]]; then
+      if [ ! -f "/etc/yunohost/certs/${domain}/chain.pem" ];then
+        wget -O ${pending_dir}/etc/yunohost/certs/${domain}/chain.pem "https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem"
```
alexAubin: We need to be careful about this step, because `/etc/yunohost/certs/${domain}` is a symlink that will get updated when we a renew is done.
Hence : 
- either we regen the nginx conf each time we enable a new certificate, to run this piece of code
- or we move this wget to the python code (I'd prefer that solution because it's more efficient that regenerating the whole conf)

data/hooks/conf_regen/15-nginx
```diff
@@ -38,6 +38,15 @@ do_pre_regen() {
...
+      fi
+      sed "s/#ssl_stapling/ssl_stapling/g" "${nginx_conf_dir}/${domain}.conf"
+      sed "s/#ssl_trusted_certificate/ssl_trusted_certificate/g" "${nginx_conf_dir}/${domain}.conf"
```
alexAubin: As discussed on the chat, I'd really like if we could rely on `ynh_render_template` here instead of sed's :/

frju365: Well, it's just two sed... not really important (as we discussed in the chat). I don't know how ynh_render_template works... but why not.

frju365:
fixed... but I don't know how I can use ynh_render_template ? and if it's needed for this little operation.

frju365:
In fact the wget command was not needed as the cert is already downloaded as intermediate certificate.

PR #493, opened 2018-06-14
HPKP

frju365:
## The problem

"HTTP Public Key Pinning Extension
An example might be your bank, which always have their certificate from CA Company A. With the current certificate system, CA Company B, CA Company C and the NSA CA can all create a certificate for your bank, which your browser will hapily accept because those companies are also trusted root CA's.

If the bank implements HPKP and pin's their first intermidiate certificate (from CA Company A), browsers will not accept certificates from CA Company B and CA Company C, even if they have a valid trust path. HPKP also allows your browser to report back the failure to the bank, so that they know they are under attack." cf. https://raymii.org/s/articles/HTTP_Public_Key_Pinning_Extension_HPKP.html

## Solution

Implement HPKP for LE certificate

## PR Status

Work finished / Not tested

## How to test

Require OCSP (build code on previous PR to avoid conflicts : https://github.com/YunoHost/yunohost/pull/492)
Install yunohost + nginx

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/conf_regen/15-nginx
+++ b/data/hooks/conf_regen/15-nginx
@@ -40,7 +40,7 @@ do_pre_regen() {
     mkdir -p "$domain_conf_dir"
     domain_ca_type=$(yunohost domain cert-status ${domain} | grep CA_type)
     if [$domain_ca_type = "CA_type: Let's Encrypt"]; then
-      if [ ! -f "${pending_dir}/etc/yunohost/certs/${domain}/chain.pem" ];then
+      if [ ! -f "/etc/yunohost/certs/${domain}/chain.pem" ];then
         wget -O ${pending_dir}/etc/yunohost/certs/${domain}/chain.pem "https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem"
       fi
       sed "s/#ssl_stapling/ssl_stapling/g" "${nginx_conf_dir}/${domain}.conf"
--- a/data/hooks/conf_regen/15-nginx
+++ b/data/hooks/conf_regen/15-nginx
@@ -38,8 +38,8 @@ do_pre_regen() {
   for domain in $domain_list; do
     domain_conf_dir="${nginx_conf_dir}/${domain}.d"
     mkdir -p "$domain_conf_dir"
-    domain_ca_type=$(yunohost domain cert-status ${domain} | grep CA_type)
-    if [[ "$domain_ca_type" = "    CA_type: Let's Encrypt" ]]; then
+    domain_ca_type=$(yunohost domain cert-status ${domain} | grep CA_type | tr -d " ")
+    if [[ "$domain_ca_type" = "CA_type:Let'sEncrypt" ]]; then
       ### OCSP Stapling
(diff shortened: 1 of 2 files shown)


data/hooks/conf_regen/15-nginx
```diff
@@ -38,6 +38,26 @@ do_pre_regen() {
...
+      fi
+      sed "s/#ssl_stapling/ssl_stapling/g" "${nginx_conf_dir}/${domain}.conf"
+      sed "s/#ssl_trusted_certificate/ssl_trusted_certificate/g" "${nginx_conf_dir}/${domain}.conf"
+      sed "s/#resolver/resolver/g" "${nginx_conf_dir}/${domain}.conf"
+      
+      ### HPKP
+      if [ -f "/etc/yunohost/certs/${domain}/chain.pem" ]; then
+        hash1=$(openssl x509 -pubkey -in ${pending_dir}/etc/yunohost/certs/${domain}/chain.pem | openssl pkey -pubin -outform der | openssl dgst -sha256 -binary | base64)
+        hash2=$(openssl x509 -pubkey -in ${pending_dir}/etc/yunohost/certs/${domain}/crt.pem | openssl pkey -pubin -outform der | openssl dgst -sha256 -binary | base64)
+        popd
```
Psycojoker: Is it normal? I don't see any `pushd` anywhere in this PR

data/hooks/conf_regen/15-nginx
```diff
@@ -38,6 +38,26 @@ do_pre_regen() {
   for domain in $domain_list; do
     domain_conf_dir="${nginx_conf_dir}/${domain}.d"
     mkdir -p "$domain_conf_dir"
+    domain_ca_type=$(yunohost domain cert-status ${domain} | grep CA_type)
+    if [ $domain_ca_type = "CA_type: Let's Encrypt" ]; then
```
Psycojoker: Variable needs to be quoted for spaces.

julienmalik:
Hi,
Are we sure we want HPKP in yunohost ? I see it is commented out by default, but as it is said to be dangerous, I was wondering if even letting this just a quick 'uncomment' away was a good idea ?

Reading https://blog.qualys.com/ssllabs/2016/09/06/is-http-public-key-pinning-dead or https://scotthelme.co.uk/im-giving-up-on-hpkp/ is quite scary for example

Seems like it is being taken out in chrome: https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/he9tr7p3rZ8/eNMwKPmUBAAJ

julienmalik:
Oh and I even read the code wrong, it is being enabled automatically for LE. I think it should be off by default

frju365:
Indeed. I didn't read these two sources. Interesting. Perhaps it can be very bad, and even more if we think that our users have perhaps no knowledge. 

We can close this PR.

frju365:
Let's close it. I've read the articles and it's seems it's very controversial. Pls, flag it as "wontfix".

PR #495, opened 2018-06-19
[enh] Add MUA autoconfig.

pitchum:
## The problem

Configuring Thunderbird can be tedious.
You have to select the proper protocols and ports for both incoming and outgoing emails.

## Solution

Fortunately there is a mechanism called [Autoconfiguration](https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration ) we can use to make life easier for users of Thunderbird.
TL;DR, the yunohost server only needs to publish a specific XML file on http://domain.tld/.well-known/autoconfig/mail/config-v1.1.xml, Thunderbird will find it and then automatically configure the proper ports and protocols.

## PR Status

- [x] test the publishing mechanism (tested by pitchum)
- [x] update autoconfigs of existing domains each time regenconf is invoked
- [x] remove autoconfigs for removed domains each time regenconf is invoked
- [x] test a complete scenario in thunderbird with a real domain name (tested by pitchum)

## How to test

Create a new domain like *example.net* and create a user account like john@example.net.
Don't forget to update the DNS configuration for this new domain.

Then open Thunderbird, add a new mail account.
In the popup, fill the 3 fields (name, email address *john@example.net* and the password) and then click next.
After a few seconds Thunderbird should display "Configuration found for mail provider" and no manual configuration should be needed.
...

## Validation

- [ ] Quick review 0/1 :
- [x] Simple test 1/1 : pitchum
- [x] Deep review 1/1 :  ljf

### DIFF
--- a/data/hooks/conf_regen/15-nginx
+++ b/data/hooks/conf_regen/15-nginx
@@ -38,11 +38,18 @@ do_pre_regen() {
   for domain in $domain_list; do
     domain_conf_dir="${nginx_conf_dir}/${domain}.d"
     mkdir -p "$domain_conf_dir"
+    mail_autoconfig_dir="${pending_dir}/var/www/.well-known/${domain}/autoconfig/mail/"
+    mkdir -p "$mail_autoconfig_dir"
 
     # NGINX server configuration
     cat server.tpl.conf \
       | sed "s/{{ domain }}/${domain}/g" \
       > "${nginx_conf_dir}/${domain}.conf"
+
+    cat autoconfig.tpl.xml \
+      | sed "s/{{ domain }}/${domain}/g" \
+      > "${mail_autoconfig_dir}/config-v1.1.xml"
+
 
     [[ $main_domain != $domain ]] \
       && touch "${domain_conf_dir}/yunohost_local.conf" \
@@ -56,6 +63,14 @@ do_pre_regen() {
     domain=${file%.conf}
     [[ $domain_list =~ $domain ]] \
       || touch "${nginx_conf_dir}/${file}"
+  done
+
+  # remove old mail-autoconfig files
+  autoconfig_files=$(ls -1 /var/www/.well-known/*/autoconfig/mail/config-v1.1.xml)
+  for file in $autoconfig_files; do
+    domain=$(basename $(readlink -f $(dirname $file)/../..))
+    [[ $domain_list =~ $domain ]] \
+      || (mkdir -p "$(dirname ${pending_dir}/${file})" && touch "${pending_dir}/${file}")
   done
 
   # disable default site
--- a/data/templates/nginx/autoconfig.tpl.xml
+++ b/data/templates/nginx/autoconfig.tpl.xml
@@ -1,0 +1,19 @@
+<clientConfig version="1.1">
+  <emailProvider id="{{ domain }}">
+    <domain>{{ domain }}</domain>
+    <incomingServer type="imap">
+      <hostname>{{ domain }}</hostname>
+      <port>993</port>
+      <socketType>SSL</socketType>
+      <authentication>password-cleartext</authentication>
+      <username>%EMAILLOCALPART%</username>
+    </incomingServer>
+    <outgoingServer type="smtp">
+      <hostname>{{ domain }}</hostname>
+      <port>587</port>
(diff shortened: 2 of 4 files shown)


data/hooks/conf_regen/15-nginx
```diff
@@ -38,12 +38,18 @@ do_pre_regen() {
   for domain in $domain_list; do
     domain_conf_dir="${nginx_conf_dir}/${domain}.d"
     mkdir -p "$domain_conf_dir"
+    mkdir -p "/var/www/${domain}/.well-known/autoconfig/mail/"
```
alexAubin: Thanks for this great piece of work !

I think it would be nice to avoid creating a folder `${domain}` in `/var/www/` for each domain ideally (sometimes, users have to browse it and it can become quite a mess if for some reason you have 10+ domains)

Instead, I propose that we use `/var/www/.well-known/${domain}/autoconfig/mail` ?

Also if we want this to be managed by the regenconf we should put this folder + files in `${pending_dir}`

so : `${pending_dir}/var/www/.well-known/${domain}/autoconfig/mail/`

What do you think ?

pitchum: I've updated the PR following your suggestion but I don't know how to make use of the magic of ${pending_dir}.

So currently it is broken, the following warning is displayed when adding a new domain: *Warning: The configuration file '/var/www/.well-known/domain.tld/autoconfig/mail/config-v1.1.xml' has been manually removed and will not be created*
The expected file tree is created in `/home/yunohost.conf/pending/nginx/var/www/.well-known/` but I don't know yet how to have it properly copied/moved into `/var/www/.well-known/`.

Can you please tell me how to use the ${pending_dir} correctly?

Note that I also added some cleanup code to *rm -rf* the `.well-known` dir upon domain deletion.
I don't think that it's related though.

alexAubin: Indeed the regenconf is a bit non-trivial ;)

Basically, the regenconf scripts should always put files in $pending_dir and not in the true filesystem (at least during the "pre" step). Because it will then compare to the current file on the disk to check if it was manually edited. (And also because maybe the regen-conf is ran with --dry-run).

And each file you manage with the regen-conf, you should also delete with the regenconf. Otherwise, it will be considered as "manually removed".

To remove a file, you can see a few line below that there is a ` touch "${nginx_conf_dir}/${file}"`, i.e. creating an empty file, which corresponds to saying that it should be deleted. (That's a bit hackish, yup)

So in our case, I guess we want to do something similar, i.e. : 
- do a `ls /var/www/.well-known/*/autoconfig/mail/config-v1.1.xml'`
- for each file, extract the domain
- if the domain is not registered in yunohost anymore,
- put an empty file at the same place in $pending_dir

If we want to go further, we could want to completely remove /var/www/.well-known/$domain ... but I don't know if we have a mechanism to remove entire directories... For instance I suspect that, after you remove a domain, there is still a folder /etc/nginx/conf.d/domain.tld.d/ ... maybe that's no bigdeal.

pitchum:
I updated my PR to make use of the magic of regenconf as suggested by @alexAubin and now it works correctly.
@zamentur suggested to take care of migrating any existing domains. Each time regenconf is invoked, configs for other domains are updated so I guess it's enough.

I could successfully test the auto-discovery from within thunderbird but I couldn't yet use the email account due to other problems (probably related to VPN). Still investigating...

Psycojoker:
> @zamentur suggested to take care of migrating any existing domains. Each time regenconf is invoked, configs for other domains are updated so I guess it's enough.

Yes, it should be good since regenconf is called during .deb upgrade.

pitchum:
I updated the PR to advertise port 993 with TLS instead of port 143 with STARTTLS because default firewall rules in yunohost don't allow port 143 (I don't know if it's intentional though).
Now everything works, I think this PR is ready to be merged.

Psycojoker:
Let's merge it then :)

Thanks a lot for your contribution :heart: (and sorry for the delay, I don't think people are very confident with this part of the code to click on merge, I'm not neither but we can sort that out with a beta phase)

alexAubin:
Adding a screenshot to be used by press articles :P 
2018-11-28-145233_1366x768_scrot

PR #498, opened 2018-06-22
[ux] display human understandable choice for boolean type on installation

Psycojoker:
## The problem

Right now, the CLI help to chose a value for a boolean type during an app installation is horrible from a UX point of view, it looks like this:

    Is it a public Zerobin site ? [1 | 0] (default: 0):

Good luck guessing what is true and what is false.

It's a long know problem and it happen to have pissed me off once too often and I was working on that part of the code.

## Solution

Do something that regular non nerdy humans can understand, like this:

    Is it a public Zerobin site ? [yes | no] (default: no): 

It also now accept: "yes", "y", "no", "n", "1", "0" as valid answers and ignore the case.

## PR Status

Tested for me and ready to merge.

## How to test

Install zerobin.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -2074,7 +2074,7 @@ def _parse_action_args_in_yunohost_format(args, action_args, auth=None):
 
                 if arg_default is not None:
                     if arg_type == 'boolean':
-                        ask_string += ' (default: {0})'.format("yes" if arg_type == 1 else "no")
+                        ask_string += ' (default: {0})'.format("yes" if arg_default == 1 else "no")
                     else:
                         ask_string += ' (default: {0})'.format(arg_default)
 
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -2068,11 +2068,15 @@ def _parse_action_args_in_yunohost_format(args, action_args, auth=None):
 
                 # Append extra strings
                 if arg_type == 'boolean':
-                    ask_string += ' [0 | 1]'
+                    ask_string += ' [yes | no]'
                 elif arg_choices:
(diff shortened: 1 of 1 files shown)


Psycojoker:
I realize that this part of the code is absolutely not translated nor translatable, this is not for this PR but that's annoying and not simple (like "how to translate yes/no in other languages") (I see how to do that but it's a bit more time than this PR).

JimboJoe:
I guess it will also impact the `yunohost app setting` parameter to pass for a boolean?

Psycojoker:
It only changes the way the user input in CLI is parsed, not how its store, so I don't think so.

src/yunohost/app.py
```diff
@@ -2068,11 +2068,15 @@ def _parse_action_args_in_yunohost_format(args, action_args, auth=None):
...
                 if arg_type == 'boolean':
-                    ask_string += ' [0 | 1]'
+                    ask_string += ' [yes | no]'
                 elif arg_choices:
                     ask_string += ' [{0}]'.format(' | '.join(arg_choices))
+
                 if arg_default is not None:
-                    ask_string += ' (default: {0})'.format(arg_default)
+                    if arg_type == 'boolean':
+                        ask_string += ' (default: {0})'.format("yes" if arg_type == 1 else "no")
```
alexAubin: Uuh, here we just checked that `arg_type == 'boolean'` and now we check if `arg_type == 1` ? Am I missing something ?

Psycojoker: Ah no, should be arg_defaut, my bad.

PR #501, opened 2018-06-26
[fix] Patch dnsmasq to read correctly dns-root-data

zamentur:
## The problem
https://forum.yunohost.org/t/jessie-users-please-dont-run-upgrades-right-now/5086

## Solution

Patch dnsmasq

## PR Status

Untested

## How to test

?


debian/postinst
```diff
@@ -14,6 +14,9 @@ do_configure() {
       echo "Regenerating configuration, this might take a while..."
       yunohost service regen-conf --output-as none
 
+      echo "Patch Dnsmasq service"
+      patch -N /etc/init.d/dnsmasq dnsmasq.patch | true
```
zamentur: I am not sur the patch is accessible in the current dir

PR #502, opened 2018-06-30
[fix] timeout on get_public_ip otherwish dyndns update is stucked

Psycojoker:
## The problem

`get_public_ip` is used by `dyndns_update`. In some networking configuration (because network) ipv6 is up but goes nowhere, therefor the request never finish.

Currently, we use urllib (yes, urllib1) which never timeout, therefor `dyndns_update` never complet, get stuck and the cron keep spawning more and more which ends up eating all the RAM of the server and making it unusable, which is a pretty serious problem.

Since this is quite server and pretty easy to fix I think we should also release a fix for jessie.

## Solution

Add modern code that times out.

## PR Status

Ready to merge.

## How to test

Run dyndns_update.

## Validation

Microdecision.

### DIFF
--- a/src/yunohost/utils/network.py
+++ b/src/yunohost/utils/network.py
@@ -21,7 +21,9 @@
 import logging
 import re
 import subprocess
-from urllib import urlopen
+import requests
+
+from requests import ConnectionError
 
 logger = logging.getLogger('yunohost.utils.network')
 
@@ -37,8 +39,8 @@ def get_public_ip(protocol=4):
         raise ValueError("invalid protocol version")
 
     try:
-        return urlopen(url).read().strip()
-    except IOError:
+        return requests.get(url, timeout=30).content.strip()
+    except ConnectionError:
         return None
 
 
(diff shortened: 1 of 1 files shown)


Psycojoker:
The user who have reported the problem confirms that the patch works as expected.

src/yunohost/utils/network.py
```diff
@@ -37,8 +39,8 @@ def get_public_ip(protocol=4):
         raise ValueError("invalid protocol version")
 
     try:
-        return urlopen(url).read().strip()
-    except IOError:
+        return requests.get(url, timeout=30).content.strip()
+    except ConnectionError:
```
alexAubin: Uuuh shouldnt we cover other exception than ConnectionError ? We might want to use the helper `download_text` here : https://github.com/YunoHost/moulinette/blob/unstable/moulinette/utils/network.py#L8

Psycojoker: ah, that seems indeed way better

Psycojoker: Done

Psycojoker:
Works as expected:

```
>>> from yunohost.utils.network import get_public_ip
>>> get_public_ip()
u'x.x.x.121'
```

(I, obviously, put the "x" myself here)

PR #504, opened 2018-07-01
Fix nodejs removal in ynh_remove_nodejs

alexAubin:
## The problem

A few minor things forgotten when actually removing nodejs via the helper ?
c.f. https://github.com/YunoHost-Apps/wekan_ynh/issues/6
and https://github.com/YunoHost-Apps/wekan_ynh/blob/master/scripts/_future.sh#L162

## Solution

Add missing lines

## PR Status

To be validated / reviewed by actual nodejs / helper experts :P I have no idea if this is the right thing to do

## How to test

Uh idk :s 

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -138,6 +138,8 @@ ynh_remove_nodejs () {
 	then
 		ynh_secure_remove "$n_install_dir"
 		ynh_secure_remove "/usr/local/n"
+		sed --in-place "/N_PREFIX/d" /root/.bashrc
+		rm -f /etc/cron.daily/node_update
 	fi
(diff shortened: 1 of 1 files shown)


Psycojoker:
LGTM but I don't handle helpers very well

maniackcrudelis:
That's a `rm -f`, not a `rm -rf`, that makes a huge difference !

frju365:
yes, indeed. I was not completely awake ;) But I still think we have a nice helper which is used in the previous remove case.

PR #507, opened 2018-07-09
App without sso

Psycojoker:
## The problem

Fixes https://github.com/YunoHost/issues/issues/1163

## Solution

Allow an application to optout of ssowat if the key 'no_sso' is defined (I don't think checking for a value makes that much sens here).

The only interesting commit of this PR is https://github.com/YunoHost/yunohost/commit/2bf327970b9f9270d6d5453a68b31fa835d22392

## PR Status

Tested and ready to merge, except if we want to bikesheed on the key name but I really don't care about that, just don't block the PR too long for that plz :/

## How to test

    yunohost app setting app_that_should_appear_in_sso no_sso -v true
    yunohost app ssowatconf
    cat /etc/ssowat/conf.json  # ← here you shouldn't see the app in the conf in the users section

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1284,7 +1284,7 @@ def app_ssowatconf(auth):
     redirected_urls = {}
 
     try:
-        apps_list = app_list()['apps']
+        apps_list = app_list(installed=True)['apps']
     except:
         apps_list = []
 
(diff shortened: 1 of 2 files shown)


Josue-T:
Good idea,

Will be really useful for synapse.

PR #509, opened 2018-07-22
[mod] propagate the no_checks logic to acme-tiny code

Psycojoker:
## The problem

The "no-check" logic is not propagated to acme-tiny, this can cause a bug in the following situation:

- domain DNS is correctly configured and is accessible from outside
- but locally the IP to this domain is wrong for some reason
- acme-tiny will attempt to get the certificate through http to make a check
- this check will fail

It's also important to note that this is pretty hard to debug, the error message is bad and in english and is confusing because it doesn't tell that it's a LOCAL attempt to download the file.

If we agree on modifying acme-tiny I will also modify this error message if I don't forgot to do it.

## Solution

Have no respect and modify acme-tiny code.

## PR Status

Ready to merge.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -289,7 +289,7 @@ def _certificate_install_letsencrypt(auth, domain_list, force=False, no_checks=F
                 _check_domain_is_ready_for_ACME(domain)
 
             _configure_for_acme_challenge(auth, domain)
-            _fetch_and_enable_new_certificate(domain, staging)
+            _fetch_and_enable_new_certificate(domain, staging, no_checks=no_checks)
             _install_cron()
 
             logger.success(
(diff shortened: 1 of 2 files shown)


Psycojoker:
While not being super convinced Aleks was ok with merging this one.

Psycojoker:
For the record:

```
00:28 <@Bram> Alaiks: pas d'avis là dessus ? https://github.com/YunoHost/yunohost/pull/509
00:36 <Alaiks> zbleh ben j'suis pas moulte convaincu é_è pour moi le mieux c'est quand même de faire en sorte que le domaine soit résolvable en local avec la bonne ip, et si ça l'est pas alors de mon point de vue c'est un bug et faut le fixer en amont é_è
00:36 <Alaiks> après c'est pas si pire comme modif mais ça augmente quand meme le coup de maj de acme_tiny.py parce que a chaque fois faudra qu'on réapplique le patch etc
00:38 <Alaiks> enfin après si t'es convaincu que ça aide pourquoi pas é_è
00:39 <@Bram> ben ça réduit le nombre de bugs
00:39 <@Bram> et donc
00:39 <@Bram> le nombre de support
00:39 <@Bram> et donc
00:39 <@Bram> on gagne de l'énergie
00:39 <@Bram> je résonne un peu comme ça :x
00:40 <@Bram> et surtout ça ment le --no-check
00:40 <@Bram> vu que ça check quand même
00:40 <@Bram> et j'aime pas quand ça ment >.>

PR #512, opened 2018-08-05
Better error path conflict

Psycojoker:
## The problem

Right now when we try to install an application but it conflicts with an already installed application at the path level we get a kinda "fuck you" error that just says that there is a conflict.

It sucks for the users and doesn't give any hint on how to fix that.

## Solution

Display all conflicting applications.

Example:

2018-08-06-012424_883x465_scrot

## PR Status

Test in "yunohost tools shell" and should be good.

## How to test

Like on the screenshot.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -22,7 +22,7 @@
     "app_location_already_used": "The app '{app}' is already installed on that location ({path})",
     "app_make_default_location_already_used": "Can't make the app '{app}' the default on the domain {domain} is already used by the other app '{other_app}'",
     "app_location_install_failed": "Unable to install the app in this location because it conflit with the app '{other_app}' already installed on '{other_path}'",
-    "app_location_unavailable": "This url is not available or conflicts with the already installed app '{app_label:s}' ({app_id:s}) on '{domain:s}{path:s}'",
+    "app_location_unavailable": "This url is not available or conflicts with the already installed app(s):\n{apps:s}",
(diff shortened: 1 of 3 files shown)


locales/en.json
```diff
@@ -22,7 +22,7 @@
     "app_location_already_used": "The app '{app}' is already installed on that location ({path})",
     "app_make_default_location_already_used": "Can't make the app '{app}' the default on the domain {domain} is already used by the other app '{other_app}'",
     "app_location_install_failed": "Unable to install the app in this location because it conflit with the app '{other_app}' already installed on '{other_path}'",
-    "app_location_unavailable": "This url is not available or conflicts with an already installed app",
+    "app_location_unavailable": "This url is not available or conflicts with the already installed app(s):\n{apps:s}",
```
alexAubin: Hmmm, I never know what's the case where it creates an issue. Isn't this gonna raise a "not all arguments converted during string formatting" error for other translated strings which don't have `{apps}` in them ?

Psycojoker: Meh, I never remember that neither :/

I'm going to look at the moulinette code, we should fix that there.

Psycojoker: Yeah, it's gonna failed... https://github.com/YunoHost/moulinette/blob/stretch-unstable/moulinette/core.py#L92

PR #513, opened 2018-08-08
More agressive deprecation messages for app checkurl and initdb

alexAubin:
## The problem

`app checkurl` and `app initdb` are deprecated but still used in some ~maintained apps.

## Solution

Propose to put a more agressive message (error instead of warning) so that hopefully one day we can decide to get rid of this legacy code :| ...

Also : 
- I pointed to the helper instead of "use app register-url" since it's probably more relevant
- I moved the entire message to app.py because it doesn't make much sense to bother translators with that message. The message is aimed to app packagers which should understand english ? :s 

## PR Status

Ready to merge if people agree

## How to test

Run `app checkurl` and `app initdb`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -14,7 +14,6 @@
     "app_change_url_identical_domains": "The old and new domain/url_path are identical ('{domain:s}{path:s}'), nothing to do.",
     "app_change_url_no_script": "This application '{app_name:s}' doesn't support url modification yet. Maybe you should upgrade the application.",
     "app_change_url_success": "Successfully changed {app:s} url to {domain:s}{path:s}",
-    "app_checkurl_is_deprecated": "Packagers /!\\ 'app checkurl' is deprecated ! Please use 'app register-url' instead !",
     "app_extraction_failed": "Unable to extract installation files",
     "app_id_invalid": "Invalid app id",
     "app_incompatible": "The app {app} is incompatible with your YunoHost version",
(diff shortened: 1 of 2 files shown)


Psycojoker:
Shouldn't we parser apps do see who is using that then open a ticket/PR there? That sound a bit more efficient communication for me :x 

That doesn't prevent from merging this PR but that should have more impact.

Psycojoker:
Another strategy is to said:

- we warn that in this version X those utils will be removed
- ounce we release this version app maintainers needs to specify explicitly that they support this version in their manifest otherwise the installation is going to be broken

But that will requires a lot of release for apps and will be bad for users who won't be able to install apps for minor details.

alexAubin:
Yes, we should clearly take action on a "per app" basis if we really want to get rid of this legacy code ;P

But it brings complicated questions like "how to handle apps which are unmaintained", e.g. is it okay to forge-merge things to use the recommended new helpers :thinking: (personally I don't see any other way if we want to avoid breaking old apps just because of this)

Psycojoker:
> But it brings complicated questions like "how to handle apps which are unmaintained", e.g. is it okay to forge-merge things to use the recommended new helpers thinking (personally I don't see any other way if we want to avoid breaking old apps just because of this)

That's one question we aren't forced to answer you know :p We don't need to take ALL the cases into account, we don't have the manpower for that (and if the app is unmaintained well... users shouldn't probably install it anyway :/)

Already fixing that on all maintained app would be a great process.

PR #514, opened 2018-08-25
[enh] Add fake-hwclock

zamentur:
## The problem
Sometimes there is no Real Time Clock Battery (arm board, old battery ...). The time is needed to be able to make TLS/openvpn authentication securely.

## Solution

Set fake-hwclock as dependency of yunohost.

Fake-hwclock save the clock each hours and at shutdown. On boot, it will check if the current date is under the registered date and if it's the case load the registered date. This is needed 

## PR Status
Ready, Micro decision

## How to test

I think it's ok

## Validation

- [x] Principle agreement 1/1 : 
- [x] Quick review 1/1 :


Psycojoker:
Haven't tested but LGTM according to xmpp discussion.

PR #517, opened 2018-08-26
[enh] Hooks to allow apps to extend the recommended DNS configuration

Josue-T:
## The problem

We can't add a custom rule in the dns config.

## Solution

Add a hook to make a possibility to add a custom rule

## PR Status

Tested. Mybe we need to do some more tests.

Note that this PR depends of #526.

## How to test

- Checkout the branch
- Rebase on the branch `hook_return` (PR #526 )
- Add a dns hook in `/etc/yunohost/hooks.d/custom_dns_rules` with by example that : 
```bash
if [[ $1 = "example.com" ]]
then
echo '[{"type": "SRV", "name": "_matrix._tcp.example.com.", "value": "10 domain.tld.", "ttl": 3600}]' > $YNH_STDRETURN
fi
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/hook.py
```diff
@@ -294,6 +295,13 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
                 result[state][name].append(path)
             except KeyError:
                 result[state][name] = [path]
+
+            #print(hook_return)
+            #for r in hook_return.:
+            result['stdreturn'].extend(hook_return) #for r in hook_return
```
alexAubin: I would rather have `result[name]['stdreturn']` (in anticipation of the new diagnosis system ;P)
And I would propagate this hierarchy also for the state a few line before to have `result[name][state]`  ... which in turns need to be propagated where this `result` value is used (I found backup and service)

Josue-T:
> The docstring of `_build_dns_conf` needs to be updated to inform about this new behavior and potentially new addition to the datastructure.

Should be fixed

src/yunohost/domain.py
```diff
@@ -4,7 +4,8 @@
 
     Copyright (C) 2013 YunoHost
 
-    This program is free software; you can redistribute it and/or modify
+    This progra
+    m is free software; you can redistribute it and/or modify
```
Psycojoker: Looks like a mistake.

src/yunohost/domain.py
```diff
@@ -397,13 +408,23 @@ def _build_dns_conf(domain, ttl=3600):
...
+    res = {
         "basic": [{"name": name, "ttl": ttl, "type": type_, "value": value} for name, ttl, type_, value in basic],
         "xmpp": [{"name": name, "ttl": ttl, "type": type_, "value": value} for name, ttl, type_, value in xmpp],
         "mail": [{"name": name, "ttl": ttl, "type": type_, "value": value} for name, ttl, type_, value in mail],
         "extra": [{"name": name, "ttl": ttl, "type": type_, "value": value} for name, ttl, type_, value in extra],
     }
 
+    # Custom record
+    hookres = hook_callback('custom_dns_rules', args=[domain])
+    for n, val in hookres.items() :
```
Psycojoker: Can you please use full name variable? This makes the code harder to read, especially on the for/list comprehension line (and code is more read and wrote etc...)

Psycojoker:
LGTM once the code use full variable name (and the hook return PR is merged)

alexAubin:
Ran some tests + implemented the renaming suggested by Bram

Planning to merge in the beginning of next week

Josue-T:
Thanks @alexAubin for your help

src/yunohost/domain.py
```diff
@@ -408,21 +408,38 @@ def _build_dns_conf(domain, ttl=3600):
...
+        #       { 'state': 'succeed',
+        #         'stdreturn': [{'type': 'SRV',
+        #                        'name': 'stuff.foo.bar.',
+        #                        'value': 'yoloswag',
+        #                        'ttl': 3600}]
+        #       },
+        #  '/some/path/to/hook2':
+        #       { ... },
+        #  [...]
+        #
```
Psycojoker: Such good commit, much gud code, very wow :scream: :heart:

alexAubin: xD

PR #518, opened 2018-08-26
[fix] Standardize sshd config

zamentur:
## The problem

Some instance don't use the sshd conf of YunoHost.

## Solution

- Remove the from_script file
- make a migration to add a include instruction and ensure the ssh conf is managed
- make a manual migration to ask user what he want do if the conf is different
- implement in postinstall a way to keep the old sshd conf if the user ask for it during the install_script

## PR Status

Ready
Related PR https://github.com/YunoHost/install_script/pull/50

## How to test

Run migrations in different case:

- [ ] With /etc/yunohost/from_script
- [ ] With different port or permitrootlogin at yes
- [ ] ?

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/data_migrations/0007_reset_sshd_config.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if self.mode == "auto":
+            return None
+
+        # Detect major risk to migrate to the new configuration
+        dsa = False
+        ports = []
+        root_login = []
+        port_rgx = r'^[ \t]*Port[ \t]+(\d+)[ \t]*(?:#.*)?$'
+        root_rgx = r'^[ \t]*PermitRootLogin[ \t]([^# \t]*)[ \t]*(?:#.*)?$'
```
zamentur: This regex conflict with https://github.com/YunoHost-Apps/my_webapp_ynh/pull/21

zamentur: This regex conflict with YunoHost-Apps/my_webapp_ynh#21

src/yunohost/data_migrations/0006_manage_sshd_config.py
```diff
@@ -0,0 +1,76 @@
...
+        # If custom conf, add 'Include' instruction
+        ynh_hash = _get_conf_hashes('ssh')[SSHD_CONF]
+        current_hash = _calculate_hash(SSHD_CONF)
+        include_rgx = r'^[ \t]*Include[ \t]+sshd_config\.d/\*[ \t]*(?:#.*)?$'
+        if ynh_hash != current_hash:
+
+            add_include = False
+            for line in open(SSHD_CONF):
+                if re.match(include_rgx, line) is not None:
+                    add_include = True
```
alexAubin: Maybe I'm mistaking, but shouldn't it be the opposite ? If you find a match, don't you want to *NOT* add the include later ?

src/yunohost/data_migrations/0006_manage_sshd_config.py
```diff
@@ -0,0 +1,76 @@
...
+                    break
+
+            if add_include:
+                with open(SSHD_CONF, "a") as conf:
+                    conf.write('Include sshd_config.d/*')
+
+        # Restart ssh and backward if it fail
+        if not _run_service_command('restart', 'ssh'):
```
alexAubin: Is a restart really needed ? Wouldnt a reload be enough ?

src/yunohost/tools.py
```diff
@@ -413,6 +413,12 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False):
     service_start("yunohost-firewall")
 
     service_regen_conf(force=True)
+
+    # Restore specific ssh conf
```
alexAubin: Needs a more extensive comment about what's going on here...

alexAubin:
To be investigated : some people report to not be able to connect with the admin account. Someone ended up installing libpam-ldapd on and OVH VPS.

https://forum.yunohost.org/t/impossible-connection-sur-le-compte-admin-en-ssh-ssh-could-not-resolve-hostname-service/6129/12

alexAubin:
Note that we should really wonder if disabling root login is really what we want, considering that login as admin assumes LDAP is up and running.

Reading users encountering messages like `mdap-sasl-bind-s-():can’t contact ldap server` (c.f. https://forum.yunohost.org/t/probleme-de-mise-a-jour-yunohost/6371/21 ) makes this PR a bit scary ... (Though it has always been the case theoretically that people wouldnt be able to log as root, it's just not the case in practice so far)

Psycojoker:
> Note that we should really wonder if disabling root login is really what we want, considering that login as admin assumes LDAP is up and running.

I've had to debug so much internet cube because of a broken ldap (nslcd was buggy on debian) that without root it would have been a total nightmare, I'm really /not/ trilled by the idea to disable it by default :[

Psycojoker:
> I'm not 100% opposed to simply keeping root access allowed in any cases, we should just be aware that it definitely increases the attack surface so it's a trade off between security and comfort in debugging even in the case where LDAP is down :/

Indeed, but with the new password policy we should be a little bit safe no? At least this mitigate this quite a lot (and also "admin" is a very common user name that is directly tried so removing root is a bit limited on that regards :/)

I like the "local network only" idea but I'm not sure on how much setup it would be realistic.

Josue-T:
My point of view that we should allow root login only with public key.

alexAubin:
Hmokay but then we have a new problem which is to define an SSH key, which the admin should then keep once it'll be needed, and that needs to be generated somehow. Some people don't even have Linux to run a keygen on :/...

Psycojoker:
Indeed, that looks like a really complicated side-solution :x

alexAubin:
So 90e542a re-enable root login on local network ... (192.168.*.* and 10.0.*.*, dunno if that's enough)

data/templates/ssh/sshd_config
```diff
@@ -94,3 +87,11 @@ Match User sftpusers
         AllowTcpForwarding no
         GatewayPorts no
         X11Forwarding no
+
+# root login is allowed on local networks
+# It's meant to be a backup solution in case LDAP is down and
+# user admin can't be used...
+# If the server is a VPS, it's expected that the owner of the
+# server has access to a web console through which to log in.
+Match Address 192.168.0.0/16,10.0.0.0/16
```
Josue-T: Is it not 10.0.0.0/8 and are there not also 172.16.0.0/12
And for IPv6 ? Didn't test but maybe fe80::/16 could be also useful :smile:

zamentur: done, i added link local too

zamentur:
A nice to have could ba a settings to specify which ip can connect as root, but i think it's not completely necessary for this PR.

SylvainCecchetto:
Hi, I don't if the SSH migration from YHN 3.4 come from this merge but anyway.
The migration 8 `ssh_conf_managed_by_yunohost_step2` want to modify the SSH port to 22 if the user (like me) use another custom port.
I have no problem with that but I have a question: Is this migration script also check that the fail2ban SSHD config is also configuring to check on the 22 port.
Because the guide https://yunohost.org/#/security_fr says to modify the fail2ban configuration if we modify the SSH port.

Thank you ;-)

alexAubin:
Nope it doesn't .. if you changed the SSH port, then right now that's up to you to reconfigure things properly :/

SylvainCecchetto:
Ok, thank you for the confirmation ;-)

PR #519, opened 2018-08-27
[fix] 'dyndns update' should check the upstream dns record

irina11y:
fix https://github.com/YunoHost/issues/issues/1063

### DIFF
--- a/src/yunohost/dyndns.py
+++ b/src/yunohost/dyndns.py
@@ -37,14 +37,14 @@ from moulinette.core import MoulinetteError
 from moulinette.utils.log import getActionLogger
 from moulinette.utils.filesystem import read_file, write_to_file, rm
 from moulinette.utils.network import download_json
+from moulinette.utils.process import check_output
+
 
 from yunohost.domain import _get_maindomain, _build_dns_conf
 from yunohost.utils.network import get_public_ip
 
 logger = getActionLogger('yunohost.dyndns')
 
-OLD_IPV4_FILE = '/etc/yunohost/dyndns/old_ip'
-OLD_IPV6_FILE = '/etc/yunohost/dyndns/old_ipv6'
 DYNDNS_ZONE = '/etc/yunohost/dyndns/zone'
 
 RE_DYNDNS_PRIVATE_KEY_MD5 = re.compile(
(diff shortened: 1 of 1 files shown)


src/yunohost/dyndns.py
```diff
@@ -252,6 +226,30 @@ def dyndns_update(dyn_host="dyndns.yunohost.org", domain=None, key=None,
         'zone %s' % host,
     ]
 
+
+    old_ipv4 = check_output("dig @91.224.148.92 +short %s" % domain).strip()
+    old_ipv6 = check_output("dig @91.224.148.92 +short aaaa %s" % domain).strip()
```
Psycojoker: It looks like every YunoHost is going to make a request to a DNS server every 2min, isn't that a lot for those DNS server? I could be like ~3000 requests every 2min (if everyone has updated), I'm not sure it's good.

Psycojoker: After an IRC discussion that ~should~ be alright, we are hitting our own DNS servers.

Psycojoker: I'm still wondering if we shouldn't switch a DNS resolution here, this IP is own by tetaneutral and if we lose it for wathever reason all the dyndns domains are going to be in a shitty situation while if we bind that to the domain we own, it would be way more safe.

alexAubin: +1

src/yunohost/dyndns.py
```diff
@@ -252,6 +226,30 @@ def dyndns_update(dyn_host="dyndns.yunohost.org", domain=None, key=None,
...
+        ipv4 = ipv4_
+
+    if ipv6 is None:
+        ipv6 = ipv6_
+
+    logger.debug("Old IPv4/v6 are (%s, %s)" % (old_ipv4, old_ipv6))
+    logger.debug("Requested IPv4/v6 are (%s, %s)" % (ipv4, ipv6))
+
+    # no need to update
+    if old_ipv4 == ipv4 and old_ipv6 == ipv6:
```
alexAubin: Need to check what happens here when, for instance, there's no IPv6. The previous `check_ouput()` probably returns empty string that is gonna be compared to `None`

alexAubin:
Fixed a few things, proposing to merge soon

PR #521, opened 2018-08-27
Add ynh_check_app_version_changed

maniackcrudelis:
## The problem

During the upgrade of an app, the app itself and the package are always upgraded, even if no upgrade is really needed.

## Solution

- Add the new helper ynh_check_app_version_changed to check the version before an upgrade
- Add also the helpers ynh_read_manifest, ynh_app_upstream_version and ynh_app_package_version.
- These previous helper have been modified (from the experimental version) to support getopts
- ynh_check_app_version_changed has been modified to use ynh_app_upstream_version

## PR Status

Ready to be reviewed

## How to test

Use `ynh_check_app_version_changed` in a package.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/system
+++ b/data/helpers.d/system
@@ -126,7 +126,7 @@ ynh_app_package_version () {
 # To force an upgrade, even if the package is up to date,
 # you have to set the variable YNH_FORCE_UPGRADE before.
 # example: sudo YNH_FORCE_UPGRADE=1 yunohost app upgrade MyApp
-
+#
 # usage: ynh_check_app_version_changed
 ynh_check_app_version_changed () {
   local force_upgrade=${YNH_FORCE_UPGRADE:-0}
--- a/data/helpers.d/system
+++ b/data/helpers.d/system
@@ -67,6 +67,11 @@ ynh_read_manifest () {
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
 
+	if [ ! -e "$manifest" ]; then
+		# If the manifest isn't found, try the common place for backup and restore script.
+		manifest="../settings/manifest.json"
+	fi
+
 	jq ".$manifest_key" "$manifest" --raw-output
 }
 
@@ -85,9 +90,6 @@ ynh_app_upstream_version () {
     ynh_handle_getopts_args "$@"
 
     manifest="${manifest:-../manifest.json}"
-    if [ ! -e "$manifest" ]; then
-        manifest="../settings/manifest.json"	# Into the restore script, the manifest is not at the same place
-    fi
     version_key=$(ynh_read_manifest --manifest="$manifest" --manifest_key="version")
     echo "${version_key/~ynh*/}"
 }
@@ -107,9 +109,6 @@ ynh_app_package_version () {
     ynh_handle_getopts_args "$@"
 
     manifest="${manifest:-../manifest.json}"
-    if [ ! -e "$manifest" ]; then
-        manifest="../settings/manifest.json"	# Into the restore script, the manifest is not at the same place
-    fi
     version_key=$(ynh_read_manifest --manifest="$manifest" --manifest_key="version")
     echo "${version_key/*~ynh/}"
 }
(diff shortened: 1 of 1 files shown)


data/helpers.d/system
```diff
@@ -53,3 +53,108 @@ ynh_abort_if_errors () {
...
+# | arg: -k, --key= - Name of the key to find
+ynh_read_manifest () {
+	# Declare an array to define the options of this helper.
+        declare -Ar args_array=( [m]=manifest= [k]=manifest_key= )
+        local manifest
+        local manifest_key
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+
+	python3 -c "import sys, json;print(json.load(open('$manifest', encoding='utf-8'))['$manifest_key'])"
```
Psycojoker: We should probably just bundle [jq](https://stedolan.github.io/jq/) with YunoHost, that would make things easier (quite minor difference I know).

It's packaged into debian.

Psycojoker: Other comment: do we have a guarantee that python THREE is installed? We don't use it in YunoHost I think :thinking:

maniackcrudelis: I've no idea...
This helper isn't from me. I'd rather prefer a good old "grep | cut" which is faster than that.
But, this helper is already used by some apps, and it seems to work fine so far.

Josue-T: Mybe jq could also fix definitely any problem about encoding. I rememberer I had some problem about encoding in synapse (this problem is solved by `encoding='utf-8'`) and it's why I decided to use python3 which has a better support of encoding.

Psycojoker: jq is now a dependacy of yunohost if you wish to uses it https://github.com/YunoHost/yunohost/pull/532

maniackcrudelis:
@YunoHost/apps

maniackcrudelis:
So... that's wonderful that we now have jq. But this wonderful app crash with a laconic `jq: error: manifest/0 is not defined at <top-level>, line 1`
The doc is helpful when you want to understand how this is supposed to be used.
And trying to find any explanations elsewhere is equally useless...

So ! What do we do with that ?
Should we try to have a helpful helper working and try to find a way later to optimize it ?
Or should we wait several more month before finally take the easy and working way of a simple `grep | cut` ?

alexAubin:
Eh o.O

Not sure to understand what's happening : are you referring to an app in particular ?

maniackcrudelis:
No, the situation is that we have many apps that use this helper, or should.
But this helper isn't merged because of this python line. jq is not well documented nor easy to use.

So I'm quite upset that we have to find a complicated work around to a simple solution.

alexAubin:
Sooo, idk, did some search (not so obvious to find clear doc, agreed)

Anyway, I think just `jq ".version" manifest.json` does the job ?

maniackcrudelis:
So... now this helper use jq instead of python.
Don't know if it's really going to work for any key in the manifest, but it works for version. So far that's all we need.

maniackcrudelis:
Bump

@YunoHost/apps

data/helpers.d/system
```diff
@@ -53,3 +53,108 @@ ynh_abort_if_errors () {
...
+ynh_app_package_version () {
+    declare -Ar args_array=( [m]=manifest= )
+    local manifest
+    # Manage arguments with getopts
+    ynh_handle_getopts_args "$@"
+
+    manifest="${manifest:-../manifest.json}"
+    if [ ! -e "$manifest" ]; then
+        manifest="../settings/manifest.json"	# Into the restore script, the manifest is not at the same place
+    fi
```
alexAubin: I guess the whole thing works, but couldn't we factorize those 4 lines in the `ynh_read_manifest` helper ?

maniackcrudelis: Yep, indeed, I guess we could consider that this will be the same problem for every helper that would use that helper.

PR #520, opened 2018-08-27
Add the internal helper ynh_handle_getopts_args

maniackcrudelis:
## The problem

As said here, https://github.com/YunoHost/yunohost/pull/455, using positional arguments could be a problem when we're adding new options to a helper.

## Solution

Use a helper to allow to use easily getopts to manage arguments.

## PR Status

Ready to be reviewed.

## How to test

Have a look to [experimental helpers](https://github.com/YunoHost-Apps/Experimental_helpers), which already use getopts.

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #524, opened 2018-08-28
Add ynh_delete_file_checksum

maniackcrudelis:
## The problem

During a change_url script, the checksum of the nginx conf file has to be removed.

## Solution

Provide a simple helper to remove a checksum corresponding to a file.

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -299,6 +299,23 @@ ynh_backup_if_checksum_is_different () {
 	fi
 }
 
+# Delete a file checksum from the app settings
+#
+# $app should be defined when calling this helper
+#
+# usage: ynh_remove_file_checksum file
+# | arg: -f, --file= - The file for which the checksum will be deleted
+ynh_delete_file_checksum () {
+	# Declare an array to define the options of this helper.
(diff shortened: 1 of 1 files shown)


JimboJoe:
How will that play with apps that already use this helper with positional arguments (e.g. [wallabag2](https://github.com/YunoHost-Apps/wallabag2_ynh/blob/master/scripts/change_url#L94))?

maniackcrudelis:
Any application which is not using getopts arguments will trig the "Legacy mode" https://github.com/YunoHost/yunohost/pull/520/files#diff-a1f538c2eae234f6ba78e0bed0ae54dbR164

maniackcrudelis:
If you're brave enough, could be nice to try quickly this helper with wallabag2 ;)
My VM is already in use with YunoRunner.

JimboJoe:
I was brave enough, and it's working in legacy mode and in "standard" mode after changing the call syntax :+1:

maniackcrudelis:
Wonderful, thanks for your try

PR #525, opened 2018-08-28
Add yunohost.multimedia helpers

maniackcrudelis:
Add helpers to handle yunohost.multimedia

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Josue-T:
Hello,

Is it not better to implement this directly in the core ?

Psycojoker:
@zamentur there is a question for you ^ (also please justify when you add this tag, it's not the first time I see you doing this without any explanation and that can be quite aggressive)

maniackcrudelis:
Well, looks like explanation aren't needed... So I remove this flag (without any explanation...)

data/helpers.d/multimedia
```diff
@@ -0,0 +1,78 @@
...
+
+# Install or update the main directory yunohost.multimedia
+#
+# usage: ynh_multimedia_build_main_dir
+ynh_multimedia_build_main_dir () {
+        local ynh_media_release="v1.2"
+        local checksum="806a827ba1902d6911095602a9221181"
+
+        # Download yunohost.multimedia scripts
+        wget -nv https://github.com/YunoHost-Apps/yunohost.multimedia/archive/${ynh_media_release}.tar.gz
```
alexAubin: I too am wondering if this shouldn't be included in the core ?... Creating a new repo + all this release and checksum stuff looks like a lot of logistic for only 2 shell scripts (also what about the updates later).

Can't we just add those in the core somehow, and add a call to `ynh_media_build.sh` during the postinstall if there isn't any significant overhead in running it ?

maniackcrudelis:
@YunoHost/apps

data/helpers.d/multimedia
maniackcrudelis: @alexAubin, as said, I can integrate that helper into the core, if you or someone else show me where I should do that.

Gofannon:
I was thinking about adding "multimedia folders" to Airsonic and found this discussion by doing reserches before beginning. I don't have much knowledge about the pro or cons about having the functionality as a "helper" or as "core".
As a packager, it would be nice to know which path is going to be taken and take a decision about it. Also what kind of work is going to be needed and apps are going to "break" if it is included into the core.

----

From a quick search, helper `ynh_multimedia_build_main_dir` is used in 32 scripts :
https://github.com/search?q=org%3AYunoHost-Apps+ynh_multimedia_build_main_dir&type=Code

And "multimedia apps" are used by 40% of people who answered this poll : https://forum.yunohost.org/t/poll-what-do-you-use-yunohost-for/6955

Hope it can help :)

maniackcrudelis:
@Gofannon, no decision's taken but in the meantime you can use the helper.
If an integration into the core is made at one time, it won't break anything.
Furthermore, it's more about including directories and permissions into all YunoHost instances, the others helpers will still be used, as they're part of apps interaction with those directories.

alexAubin:
Still planning to work on it someday to integrate this in the core as it seems something used by several apps, but there's so many things to do ... :/

alexAubin:
Superseded by https://github.com/YunoHost/yunohost/pull/1129

PR #526, opened 2018-08-29
[enh] Allow hooks to return data

Josue-T:
## The problem

- There are no posibility to return a value from a hook.

## Solution

Add a possibility to return a value by creating a Json file in the $YNH_STDRETURN path.

To return a value the hook should look like something like that : 
```
echo '["small test"]' > $YNH_STDRETURN
```

## PR Status

Tested for #517 but need some other test because the value returned by `hook_callback` has been changed.

## How to test

- Checkout the branch
- Test the #517 witch need this feature.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/hook.py
```diff
@@ -388,11 +392,22 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
     elif raise_on_error and returncode != 0:
         raise MoulinetteError(
             errno.EIO, m18n.n('hook_exec_failed', path=path))
-    return returncode
+
+    try:
+        returnjson = read_json(stdreturn)
+    except Exception as e:
+        returnjson = {}
+        errno.EIO, m18n.n('hook_json_return_error', path=path, msg=str(e))
```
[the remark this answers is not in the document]
Josue-T: Yes ;-)

src/yunohost/hook.py
```diff
@@ -388,11 +392,22 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
             logger.error(m18n.n('hook_exec_not_terminated', path=path))
-            return 1
+            return 1, ''
     elif raise_on_error and returncode != 0:
         raise MoulinetteError(
             errno.EIO, m18n.n('hook_exec_failed', path=path))
-    return returncode
+
+    try:
+        returnjson = read_json(stdreturn)
```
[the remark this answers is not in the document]
Josue-T: Should be normally fixed

Psycojoker:
There are conflict to resolve :/

Josue-T:
> There are conflict to resolve :/

Should be fixed

src/yunohost/hook.py
```diff
@@ -388,11 +392,27 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
+            else:
+                returnjson = {}
+    except Exception as e:
+        returnjson = {}
+        raise MoulinetteError(
+            errno.EIO, m18n.n('hook_json_return_error', path=path, msg=str(e)))
+
+    stdreturndir = os.path.split(stdreturn)[0]
+    os.remove(stdreturn)
+    os.rmdir(stdreturndir)
```
Psycojoker: Those operation won't be executed in case there is a raise line 408 thus letting the FS dirty.

src/yunohost/hook.py
```diff
@@ -388,11 +392,27 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
-    return returncode
+
+    try:
+        with open(stdreturn, 'r') as f:
+            if f.read() != '':
+                returnjson = read_json(stdreturn)
+            else:
+                returnjson = {}
+    except Exception as e:
+        returnjson = {}
```
Psycojoker: This line seems useless since there is a raise just after that :o

Josue-T: Is it ok if I replace this line by this : 
```
os.remove(stdreturn)
os.rmdir(stdreturndir)
```
To fix the next comment too.

src/yunohost/hook.py
Josue-T: Should be fixed

src/yunohost/hook.py
Josue-T: Should be fixed

src/yunohost/hook.py
```diff
@@ -288,10 +289,8 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
             else:
                 post_callback(name=name, priority=priority, path=path,
                               succeed=True)
-            try:
-                result[state][name].append(path)
-            except KeyError:
-                result[state][name] = [path]
+
+            result[name] = {'path' : path, 'state' : state, 'stdreturn' : hook_return }
```
zamentur: This PR doesn't change a lot of code using the 'state' key, like here https://github.com/YunoHost/yunohost/blob/115b557b67893111d5b7708f6b3b3226615f3092/src/yunohost/backup.py#L613

The easiest is to keep the result[state][name] like that so something like:
```
result[state][name] = {'path' : path, 'state' : state, 'stdreturn' : hook_return }
```

src/yunohost/hook.py
```diff
@@ -288,10 +289,8 @@ def hook_callback(action, hooks=[], args=None, no_trace=False, chdir=None,
             else:
                 post_callback(name=name, priority=priority, path=path,
                               succeed=True)
-            try:
-                result[state][name].append(path)
```
zamentur: Are you sure there are no ways to get several path for a same name ? For example with custom hook ?

Josue-T: I think I changed that because it was a request from @alexAubin . But not sure.

Can @alexAubin  confirm this ?

src/yunohost/hook.py
Josue-T: I think I changed that because it was a request from @alexAubin . But not sure.

Can @alexAubin  confirm this ?

alexAubin: Uh yeah I don't understand the trouble here ?

To me it doesn't make sense to just put things in either the "succeed" or "failed" box instead of just having directly a dictionnary of hook name (`result[name]`) ?

Dunno why that troublesome to do this instead of keeping the existing `result[state][name]` ?

zamentur: To do this, you need to change a lot of things and you can't just make:
```
ret = hook_callback(...)
if ret['failed']:
    # Some error management
```

This kind of pattern is used in a lot of piece of code. We can't change the format of the return without check all hook_callback call...

zamentur: In more, previously there was possibility to get several hook path for the same name. I think it could be several path on setup using custom hook. I believe mailman_ynh use this feature.

Josue-T: Well,

So what should I do for that now ?

alexAubin: I think we did check all the `hook_callback` call, though I would need to double-check ... but it's not that much a big deal imho :/

About the "several path" thing, gotta check what is the exact behavior but that doesn't look like a blocker thing to me :/

Josue-T: Now it should be fixed

PR #527, opened 2018-08-29
Synchronize root and admin password

zamentur:
## The problem

We suspect some instance have 'yunohost' as password. More generally, when people set or change the admin password, it does not affect the root password, which leads to : 1) more confusion between 'what's the admin password' and 'what's the root password', and 2) leaves hole open if the user forgets to change the default root password ...

## Solution

Add a migration that check if the root password is 'yunohost' : if so, the root password is automatically replaced by the admin password. Otherwise, the migration is manual and will sync root and admin password. Also, using `yunohost tools adminpw` will now also impacts the root password.

## PR Status
Ready

## How to test

```
# Put yunohost as password
passwd
yunohost  tools migrations migrate --debug
# Check password has changed
su
yunohost tools adminpw --debug
# Check password has changed
su
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
I have to admit that this migration looks super tricky and breaking prone to me, I don't feel really confident about having it :fearful: 

I know this is frustrating to re-do some existing work but I think it would be way better if we had a mechanism on authentication to check if the password in a blacklist and require a modification on the password at that moment. That would be way more flexible and futur proof because we could extend this blacklist very easily.

Also this PR seems to contain a totally very different modification that what is announce :neutral_face: https://github.com/YunoHost/yunohost/pull/527/commits/2d5077e2eadc5ec18273fe73d6b96ac94219a7fb That should be in another PR, definitively.

zamentur:
> I know this is frustrating to re-do some existing work but I think it would be way better if we had a mechanism on authentication to check if the password in a blacklist and require a modification on the password at that moment.

So the main problem for you is the way I add a second password on ldap ?
Note : This process is already used in a lot of case with the `yunohost-reset-ldap-password` command.

Until today, i never saw someone on forum/irc saying this manipulation break something. I know it's quite tricky.

> Also this PR seems to contain a totally very different modification that what is announce neutral_face 2d5077e That should be in another PR, definitively.

`yunohost tools adminpw` was announced in this PR, see the solution description. The title of this PR is voluntary not very clear, because i didn't want to publish all details in a title...

To explain a little bit, we need during postinstall to change root password with the admin's one. It's the easiest way i found to be sure root password will not be a default password from our images.

In more, having different password between root/admin users is not so simple to understand for users. So it's a solution pretty well to fix that too.

To understand completely the issue, you should read this PR too: https://github.com/YunoHost/yunohost/pull/518

> That would be way more flexible and futur proof because we could extend this blacklist very easily.

That's not the goal of this PR. To accept good password or not, there is already an other PR https://github.com/YunoHost/yunohost/pull/196 . 

I already reduce the list of password tested in this PR to the strict minimal, because other contributors here say it's a different case if the user put a weak password voluntarily... 

https://github.com/YunoHost/yunohost/pull/527/commits/6aef80cd2502aee71bcbf555d3c4b5669ffe0b5e

zamentur:
7 days, this fix correct a big issue...

src/yunohost/data_migrations/0006_migrate_pwd.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def backward(self):
+
+        pass
+
+    def _get_admin_hash(self):
+        """
+        Ask for admin hash the ldap db
+        Note: to do that like we don't know the admin password we add a second
+        password
```
alexAubin: I'm testing this and : is this method really needed to get the admin hash ?

I just realized that, as root, you can run `slapcat` from which you may extract the relevant info. Here's a quick and dirty command to do that (we might want to have a more robust way to do that though) : 

```
slapcat | grep "dn: cn=admin,dc=yunohost,dc=org" -A20 | grep userPassword -A2 | tr -d '\n ' | tr ':' ' ' | awk '{print $2}' | base64 -d
```

src/yunohost/data_migrations/0006_migrate_pwd.py
```diff
@@ -0,0 +1,60 @@
...
+SMALL_PWD_LIST = ["yunohost", "olinux"]
+
+class MyMigration(Migration):
+    "Migrate password"
+
+    def migrate(self):
+
+        if self._is_root_pwd_listed(SMALL_PWD_LIST):
+            new_hash = self._get_admin_hash()
+            self._replace_root_hash(new_hash)
```
alexAubin: Rethinking about this, I think we need to clarify the purpose of this migration ? (Or maybe I'm kinda lost about it)

It seems like this PR does two different things : 
- check if the root password is "bad", and if so, resync it with the admin password
- change tools_adminpw such that it also changes the root password each time the admin password is modified...

But for existing instance on which the root password ain't "bad", after this migration, we still end up with the root password and admin password being desynchronized. Shouldn't we sync the root and admin password in any case so that we end up with consistent setups in all cases ?

alexAubin:
Note that since we are gonna have a few manual migrations now, we should check how this behaves exactly. I'm thinking that if you have several pending migrations (manual or auto), running `yunohost tools migrations migrate --accept-disclaimer` might run all of them whereas we might want them to be ran one-by-one.

Josue-T:
Will we be able to keep a independent root password after this migration ?

alexAubin:
Yes and no : the spirit of this PR is that root and admin password shall be identical and shall be resynchronized each time `yunohost tools adminpw` is ran. You could still have a different root password by running "passwd" as root, but that's not really the spirit.

alexAubin:
Proposing to merge in a few days ... any more feedback on this ?

PR #528, opened 2018-08-30
[enh] integrate CAA DNS entry into "yunohost domain dns-conf"

irina11y:
enh https://github.com/YunoHost/issues/issues/1153

### DIFF
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -346,6 +346,10 @@ def _build_dns_conf(domain, ttl=3600):
             ["*", ttl, "AAAA", ipv6],
         ]
 
+    basic += [ 
+        ["@", ttl, "CAA", "128 issue 'letsencrypt.org'"]
+    ]
+
     # XMPP
     xmpp = [
         ["_xmpp-client._tcp", ttl, "SRV", "0 5 5222 %s." % domain],
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -330,6 +330,9 @@ def _build_dns_conf(domain, ttl=3600):
             {"type": "TXT", "name": "mail._domainkey", "value": "\"v=DKIM1; k=rsa; p=some-super-long-key\"", "ttl": 3600},
             {"type": "TXT", "name": "_dmarc", "value": "\"v=DMARC1; p=none\"", "ttl": 3600}
         ],
+        "extra": [
+            {"type": "CAA", "name": "@", "value": "128 issue 'letsencrypt.org", "ttl": 3600},
+        ],
     }
     """
 
--- a/src/yunohost/domain.py
+++ b/src/yunohost/domain.py
@@ -200,6 +200,11 @@ def domain_dns_conf(domain, ttl=None):
     for record in dns_conf["mail"]:
         result += "\n{name} {ttl} IN {type} {value}".format(**record)
 
+    result += "\n\n"
+    result += "; Extra"
+    for record in dns_conf["extra"]:
+        result += "\n{name} {ttl} IN {type} {value}".format(**record)
+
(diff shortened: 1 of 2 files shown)


frju365:
I don't understand this PR... because this will return dns-conf. I don't understand why it's set LE whereas the users has perhaps not built/validated a LE certificate. Perhaps he hasn't certificate at all or perhaps have a SSL certs provided by another SSl authority. (surely me who don"t understand the PR... and perhaps it's not finished ;) ) :-)

alexAubin:
>why it's set LE whereas the users has perhaps not built/validated a LE certificate. Perhaps he hasn't certificate at all or perhaps have a SSL certs provided by another SSl authority.

The purpose of the CAA record is to declare which CA can emit a certificate for this domain. So with this record, only Lets Encrypt shall emit a certificate for the domain. That doesn't conflict with the fact that there is no certificate emitted for this domain already.

This does, however, creates a conflict in case a user wants to ask another CA to emit a certificate for this domain. After talking about this during the brique camp, the conclusion was basically that YunoHost's recommendation is to use Lets Encrypt - and we do not anyway officially support any sort of integration with other CA. Using another CA is an advanced use case which requires some tweaking, and this PR just adds the fact that the advanced user shall remove/change this record before requesting a certificate, which imho is not such a burden.

src/yunohost/domain.py
```diff
@@ -346,6 +346,10 @@ def _build_dns_conf(domain, ttl=3600):
             ["*", ttl, "AAAA", ipv6],
         ]
 
+    basic += [
```
alexAubin: Rethinking about this today : I definitely think we should add a new category called "extra" since this ... extra stuff which is not required for things to work. Also this should simplify the integration with `dyndns update` (the dynette might probably refuse to add this record right now, or we need to edit it) and simplify other things I anticipate for the auto-diagnosis

frju365:
Ok. Thanks for the answer. Well, not supporting other CA will be very complicated if we want to open the software to companies or administrations (not the moment today, but perhaps the future of Yunohost).

Psycojoker:
> Ok. Thanks for the answer. Well, not supporting other CA will be very complicated if we want to open the software to companies or administrations (not the moment today, but perhaps the future of Yunohost).

This part of the code is not complicated at all, I don't think we'll have that much problem introducing other CA if needed.

alexAubin:
>I'm really wondering, won't this change breaks the self-signed certificate mechanism that we have? I don't see any condition used for that here.

It won't : this record is only used by CA's before emitting a certificate. It's not analyzed by browers whatsoever

Psycojoker:
> >  I'm really wondering, won't this change breaks the self-signed certificate mechanism that we have? I don't see any condition used for that here.
> 
> It won't : this record is only used by CA's before emitting a certificate. It's not analyzed by browers whatsoever

Oh ok, I should probably rtfm this then ... especially since it's me who have open the issue :sweat_smile:

alexAubin:
I implemented the changes discussed and that should now be alright, at least from my point of view ...

Proposing to merge this in a few days. Is there any more feedback on this ?

src/yunohost/domain.py
```diff
@@ -325,6 +330,9 @@ def _build_dns_conf(domain, ttl=3600):
             {"type": "TXT", "name": "mail._domainkey", "value": "\"v=DKIM1; k=rsa; p=some-super-long-key\"", "ttl": 3600},
             {"type": "TXT", "name": "_dmarc", "value": "\"v=DMARC1; p=none\"", "ttl": 3600}
         ],
+        "extra": [
+            {"type": "CAA", "name": "@", "value": "128 issue 'letsencrypt.org", "ttl": 3600},
```
zamentur: Missing single quote

PR #529, opened 2018-08-30
Allow to add a service description on "yunohost service add"

irina11y:
enh  #1133


src/yunohost/service.py
```diff
@@ -50,15 +50,16 @@
 logger = log.getActionLogger('yunohost.service')
 
 
-def service_add(name, status=None, log=None, runlevel=None):
-    """
+def service_add(name, status=None, log=None, runlevel=None, description=None):
+    """
```
Psycojoker: You might want to remove this additionnal blank space but that's a minor useless detail.

PR #532, opened 2018-08-31
[enh] add jq as a dependancy

Psycojoker:
## The problem

We often need to query json (and maybe modify) files in shell scripts. Right now we uses python or grep/other to do that, that's both not practical (and slow for python) and not very reliable.

## Solution

Add jq as a dependancy so scripts can rely on that, it's a tool meant for that: https://stedolan.github.io/jq/

## PR Status

Ready to merge.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Yolomerging as microdecision :+1:

PR #534, opened 2018-09-01
[enh] New diagnosis system

alexAubin:
## The problem

Setting up and maintaining a server is a hard task, and even though we have documentation regarding port forwarding, DNS and other things, it is not entirely easy to diagnose that things are working smoothly or at their best.

Also, some issues might appear such as lacking disk space not having swap, or one critical service going down for some reason - stuff that you might notice only when somethings breaks for real (ie app failure because of lack of space).

## Solution

This PR proposes a new, high-level diagnosis system, meant to be ran regularly to check for common issues, such as DNS configuration mismatch with the expected conf, having a reasonnable amount of disk space remaining, all critical services being up, and so on.

The diagnosis system shall be ran from time to time with a cron job that should alert the admin in case of problems found.

Features would include : 
- "high-level" tests/reports
- modularity (one diagnoser per file)
- extendability (app could add custom diagnosers)
- cached value (e.g. don't rediagnose the DNS every 30 second, instead keep it in cache for something like an hour or a day, and be able to force a rediagnosing if needed)
- be able to ignore warning/error (sometimes it's expected that the DNS does not match the recommended conf for instance)

More complete list of examples of diagnoser : 
- Internet connectivity / what are the IPv4 and IPv6
- Ports forwarded / reachable
- DNS configuration
- Having some swap available
- Having some RAM available
- Having some disk space available (check the different partitions, e.g. /tmp and /var)
- Check certificates are valid
- Check mail score is reasonable
- Check all critical services are running
- Check the last upgrade of the system and apps happened not so long ago
- Check the last backup of the system and apps happened not so long ago
- Check spectre/meltdown vulnerability
- Misc security checks
- ???

## PR Status

Requires #535
Work in progress

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
We are going to have a confusing CLI api with "yunohost diagnosis" and "yunohost tools diagnosis" :sweat:

frju365:
Question : about the confusing command : why not doing yunohost only one command : yunohost diagnosis ? For a general diagnosis why not using --general option or --all for the diagnosis with everything ?

alexAubin:
Not sure what you mean ? If you refer to the `'yunohost diagnosis'` vs `'yunohost tools diagnosis'`, my intention was that `'yunohost tools diagnosis'` would become obsolete and therefore should be replaced by `'yunohost diagnosis'`

In terms of usage, the basic command would be `'yunohost diagnosis run'` which would run all diagnosis by default. Otherwise you can optionnaly specify categories of diagnosis to be ran, if you have something more precise in mind.

frju365:
ok. It's what I meant. ;-) (run is different from option but wille lay the same role here)

It was in fact to give an idea of solution of what @Psycojoker says.

alexAubin:
Sooo made quite some progress on this. There still are some funky details that might need to be reworked or polished but that's starting to look like something usable and that would be helpful in pinpointing the basic issues that people regularly encounter.

Obviously there are many many additional checks that could be added, but for now it's pretty much in the situation where it should be reviewed and tested on the battlefield / actual servers and iterate on it...

The major remaining feature to implement is the ability to ignore specific warnings ('cause for example you don't want your server complaining everyday about the lack of IPv6 if there's no way to add IPv6 at all because of your ISP, or complaining about XMPP DNS records but you don't care about XMPP)

Here's a screenshot of the current webadmin view : 

2019-08-01-220747_1366x768_scrot

alexAubin:
So now that we are pretty much done with the 3.7 (I guess we'll only need to push some relatively small hotfixes), let's start the 3.8.

This first iteration with the diagnosis is pretty much done, so flagging as ready to merge. Still some stuff to be adressed in follow-up PRs :
- [ ] Having some sort of cron job like at least once a day to run the diagnosis ?
- [ ] Implement some remaining modules, like email stuff
- [ ] Deploy the remote diagnoser service on our infra
- [ ] General polishing and tweaking of stuff
- [ ] ???

PR #535, opened 2018-09-01
Python hooks

alexAubin:
## The problem

Python hooks are required for the new diagnosis system modularity / extendability, and more generally if we want to implement some hooks in python for some reason. This could be the case for example of the famous `app ssowatconf` command which should be a classical regen-conf thing but can't be so far because it has too many python calls and json stuff to be implemented in bash.

## Solution

Split the behavior of the `hook_exec` function depending on the nature of the hook. Python hooks are to be modules loaded by yunohost and expected to have a `main` function.

## PR Status

Almost done but might still move depending on #534 

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -25,9 +25,11 @@
 """
 import os
 import re
+import sys
 import errno
 import tempfile
 from glob import iglob
+from importlib import import_module
 
 from moulinette import m18n
 from moulinette.core import MoulinetteError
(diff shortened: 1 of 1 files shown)


alexAubin:
Superseded by https://github.com/YunoHost/yunohost/pull/747

PR #536, opened 2018-09-03
[enh] Don't backp user home with .nobackup file

zamentur:
## The problem

Sometimes i don't want to backup data of a user (typically user dedicated to backup data from others server).

## Solution
A file in the home dir of the user

## PR Status

Ready

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Jaxom99:
Hi ! just found about this and it's perfectly my usecase. 😄 so thanks, one more time !

Is this publicly  documented somewhere ? I think I should make a PR in https://yunohost.org/#/backup_fr...

PR #543, opened 2018-09-15
[enh] Display only used fingerprint in bootprompt

zamentur:
## The problem

Some fingerprints are display but can't be used because SSH conf don't allow to use those fingerprint.

## Solution

Just display fingerprint used by SSH

## PR Status

Ready, this PR is derivated from an other one which change SSH config...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/bin/yunoprompt
+++ b/bin/yunoprompt
@@ -5,7 +5,7 @@ ip=$(hostname --all-ip-address)
 
 # Fetch SSH fingerprints
 i=0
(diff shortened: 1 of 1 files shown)


alexAubin:
Yolomergin' (n.b. : this goes into another PR, not in stretch-unstable)

PR #546, opened 2018-09-28
mod_auth_ldap: reflect SASL API changes in latest Metronome.

maranda:
# The problem

Users traces with 3.10.13+

## Solution

The session state should always be passed in the profile as util.sasl functioning now relies on it

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/lib/metronome/modules/mod_auth_ldap2.lua
+++ b/lib/metronome/modules/mod_auth_ldap2.lua
@@ -56,8 +56,9 @@ function new_default_provider(host)
      return nil, "Account creation/modification not available with LDAP.";
  end
 
- function provider.get_sasl_handler()
+ function provider.get_sasl_handler(session)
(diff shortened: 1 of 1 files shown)


alexAubin:
Apparently this fixes a critical issue that a lot of issue have with the new metronome version.

We should release a quick hotfix for this. To me this looks like a small, legit change. Any opinion on this, @Psycojoker , @zamentur , @JimboJoe  , @Josue-T  ?

Josue-T:
Didn't tested, but I think too that it's important to fix this as quickly as possible.

Josue-T:
Tested and it work on my side.

Psycojoker:
If it's tested and working you can go for it for me.

PR #548, opened 2018-10-01
Do not use a separate ini file for php pools

maniackcrudelis:
## The problem

Have a look to https://github.com/YunoHost-Apps/nextcloud_ynh/issues/138 for more information.

## Solution

Do not use separate ini file anymore.

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [x] Simple test 0/1 : Josué
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -222,11 +222,7 @@ ynh_add_fpm_config () {
 
 	if [ -e "../conf/php-fpm.ini" ]
 	then
-		finalphpini="$fpm_config_dir/conf.d/20-$app.ini"
-		ynh_backup_if_checksum_is_different "$finalphpini"
-		sudo cp ../conf/php-fpm.ini "$finalphpini"
-		sudo chown root: "$finalphpini"
-		ynh_store_file_checksum "$finalphpini"
+		echo "Please do not use a separate ini file, merge you directives in the pool file instead." &>2
 	fi
 	sudo systemctl reload $fpm_service
 }


data/helpers.d/backend
```diff
@@ -222,11 +222,7 @@ ynh_add_fpm_config () {
 
 	if [ -e "../conf/php-fpm.ini" ]
 	then
-		finalphpini="$fpm_config_dir/conf.d/20-$app.ini"
-		ynh_backup_if_checksum_is_different "$finalphpini"
-		sudo cp ../conf/php-fpm.ini "$finalphpini"
-		sudo chown root: "$finalphpini"
-		ynh_store_file_checksum "$finalphpini"
+		echo "Please do not use a separate ini file, merge you directives in the pool file instead." &>2
```
zamentur: Instead of just display an Error, i suggest something like this: 

ynh_print_debug "Please do not use a separate ini file, merge you directives in the pool file instead."
sed s/=/]=/g ../conf/php-fpm.ini | sed s/^/php_value[/g >> "$finalphpconf"

And to move ynh_store_file_checksum "$finalphpconf" at the end of the function.

maniackcrudelis: It's not as easy, it could be php_admin_value or php_admin_flag, I don't think playing with the pool file like that is really a good idea.

However, ignore the ini file is probably too drastic. We should maybe just add a warning to say that it should be modified before definitively removing it.

alexAubin:
Bumpitybump, any update on this ? Should we go for Maniack's solution to only have a warning instead ?

data/helpers.d/backend
Josue-T: I think too that to not try to change that dynamically is better. It will force all packager to adapt the app.

zamentur: ok

maniackcrudelis:
> Should we go for Maniack's solution to only have a warning instead ?

Yep, we should have done that instead...
That's actually a mistake, because any application with such a ini file will fail to be backup because this file is missing...
Also, there was an error in my redirection, `&>2` instead of `>&2`. The warning isn't print on stderr but stdout... So I've missed it...

PR #549, opened 2018-10-02
[enh] allows to have several domains/path per app

Psycojoker:
## The problem

Some apps wants to be able to have several domains/path because they ship several components.

## Solution

Allows up to 9 domains per applications. using "domain2/path2" "domain3/path3" ... notation for keys.

I could have made it generic and allows infinite additional domains but that would have made the code way more complex than needed and the case where an app wants 2 domains is already super rare, I can't think of someone wanting more than 3-4 so that should be a good enough marging.

Worst case those people will contact us.

## PR Status

Test and working.

BUT, there might be other places in the code where this modification need to be done but I don't know where.

## How to test

Edit the config of an app, add "domain2/path2" kind of keys, then do "yunohost app map" and "yunohost app ssowatconf" and see the result.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -731,6 +731,10 @@ app:
                     help: App ID
                 new_label:
                     help: New app label
+                -n:
+                    full: --number
+                    help: Label number for multi-urls apps
+                    type: int
 
         ### app_addaccess() TODO: Write help
         addaccess:
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -427,17 +427,17 @@ def app_map(app=None, raw=False, user=None):
 
             domain = app_settings['domain' + number]
             path = app_settings.get('path' + number, '/')
+            label = app_settings.get('label' + number, app_settings['label'])
 
             if raw:
                 if domain not in result:
                     result[domain] = {}
                 result[domain][path] = {
-                    'label': app_settings['label'],
+                    'label': label,
                     'id': app_settings['id']
                 }
             else:
-                result[domain + path] = app_settings['label']
-
+                result[domain + path] = label
     return result
 
 
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -416,18 +416,27 @@ def app_map(app=None, raw=False, user=None):
                     and user not in app_settings['allowed_users'].split(','):
                 continue
 
-        domain = app_settings['domain']
-        path = app_settings.get('path', '/')
+        # let supports up to 9 additional domains.
+        #
+        # we could easily parse the key to support an infinite number of
+        # domains but that would make the code way more complex for very
+        # very marginal benefit (who needs more than 2-3 domains?)
+        for number in [""] + [str(x) for x in range(2, 10)]:
+            if 'domain' + number not in app_settings:
+                continue
 
-        if raw:
-            if domain not in result:
-                result[domain] = {}
-            result[domain][path] = {
-                'label': app_settings['label'],
-                'id': app_settings['id']
-            }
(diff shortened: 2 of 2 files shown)


Psycojoker:
So, @julienmalik tested it, said it was nearly perfect, it just also needs to handle labels, this raise unexpected new work, todo:

- [ ] handle label[2-9], easy
- [ ] modify change-label behavior, still easy, add an argument like `--number`
- [ ] modify the admin interface to handle modifications of more than one label: way less easy
- [ ] we probably need to modify the installation part that always ask for a label

@julienmalik args that since this doesn't break anything we can still have the strategy to merge it then add needed things one by one which would make things easier, I still think we need TODO element 1 (label[2-9]) and explore 4.

julienmalik:
Also app_change_url would need modification. Same kind of changes as for change-label

Psycojoker:
Working on the code I realize that:

- we also need to handle domain/url conflict handling here https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/app.py#L2275
- register_url also needs to be modified
- check_url too if we don't remove it in the meantime...

There are probably other things hiding around.

zamentur:
ssowatconf should be modified too

Psycojoker:
> ssowatconf should be modified too

It is already modified, its subcalls are modified.

alexAubin:
Since the permission system is very related to this and I'm wondering about this "multiple domain/path" use case ...

Do you have examples of such requirements for apps ? And shall we really handle the case where there are multiple domains (and not just multiple paths) ?

Josue-T:
> Since the permission system is very related to this and I'm wondering about this "multiple domain/path" use case ...

I just give an example of one app that I'm managing. Seafile need multiple path. The first path is the main path used for the web interface. Generally it's `/seafile` but it could be anything. The second path is fixed (I saw somewhere that it's hardcoded) it's `/seafhttp` it's used for all file exchange.

Psycojoker:
It was asked by @opi. Generally those apps are the ones that comme with a backend (rest api) and a frontend.

It is packaged as 2 apps right now I think but riot/synapse (matrix) is an example.

Also having a separate domains for static files optimization is a very common practice (like a mastodon instance for the last one I've saw).

(related to https://github.com/YunoHost/yunohost/pull/798)

alexAubin:
Hmokay, I believe we should clarify what exactly it is we want to address (and how those use case related to this) because there's really 3 differents things - though all related - in these stories : 
1. handling the **conflicts** properly with apps that wanna book multiple domain/path
2. having (or not) a **tile** displayed in the SSO panel
3. managing the effective access **permission** for users to these urls

As explained on https://github.com/YunoHost/yunohost/pull/797 currently all this is indirectly managed by `app_map` and in the current implementation of yunohost and ssowat, you cannot having conflict handling without effectively allowing access to the url which in turns creates a tile in ssowat.

Anway there's a lot of movement in `app_map` inside https://github.com/YunoHost/yunohost/pull/797 ... at least to address the permission thing, and the tile thing (at least partially)  but not the conflict thing.

It's not clear to me how to design the whole thing. We could imagine changing the way we manage app conflicts by looping over all permissions, though that shows that the permission system becomes more general than just permissions but more of a "route management system" (which could be applied not just to web routes/path but also to book ports) ...

zamentur:
The new permission system allows to define several urls in ssocat.
For instance it doesn't allow to define the label of a specific permission (only main supported).
But i think this PR is quite old and outdated, because even for the change-label feature, the feature need to be rewritten.

SO i close this PR

PR #551, opened 2018-10-05
Fix potential key error when retrieving install_time

flahemade:
## The problem

Sometimes, on edge cases, the "install time" key is not in settings, it then triggers a KeyError.

## Solution

Prevent from triggering such an error adding a fallback on 0, which will cause the app to be considered "upgradable".

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


zamentur:
Will be merge in 3 days

Psycojoker:
Thx a lot for your contribution :heart:

PR #553, opened 2018-10-07
[fix] Improve the message w.a. when admins tries to create users with a reserved Email Address (Issue 1216)

frju365:
## The problem

- Error message is not useful to inform user that email adress will be created by default ( fixes https://github.com/YunoHost/issues/issues/1216)

## Solution

- Add a custom Message.

## PR Status

Finished. Not tested.

## How to test

- Install Yunohost (ynh-dev)
- Modify app.py
- Try to create a user and add an already-used alias for this email

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -255,7 +255,7 @@
     "mail_domain_unknown": "Unknown mail address domain '{domain:s}'",
     "mail_forward_remove_failed": "Unable to remove mail forward '{mail:s}'",
     "mailbox_used_space_dovecot_down": "Dovecot mailbox service need to be up, if you want to get mailbox used space",
-    "mail_unavailable": "This email address is reserved. It will be assigned to the first user",
+    "mail_unavailable": "This email address is reserved and shall be automatically allocated to the very first user",
     "maindomain_change_failed": "Unable to change the main domain",
     "maindomain_changed": "The main domain has been changed",
     "migrate_tsig_end": "Migration to hmac-sha512 finished",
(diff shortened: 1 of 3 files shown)


locales/fr.json
```diff
@@ -122,6 +122,7 @@
     "mail_alias_remove_failed": "Impossible de supprimer l'alias courriel « {mail:s} »",
     "mail_domain_unknown": "Le domaine « {domain:s} » du courriel est inconnu",
     "mail_forward_remove_failed": "Impossible de supprimer le courriel de transfert « {mail:s} »",
+    "mail_unavailable": "Adresse mail indisponible",
```
alexAubin: (We are not supposed to edit fr.json, this usually breaks weblate :s)

locales/en.json
```diff
@@ -255,6 +255,7 @@
     "mail_domain_unknown": "Unknown mail address domain '{domain:s}'",
     "mail_forward_remove_failed": "Unable to remove mail forward '{mail:s}'",
     "mailbox_used_space_dovecot_down": "Dovecot mailbox service need to be up, if you want to get mailbox used space",
+    "mail_unavailable": "Mail adress unavailable",
```
alexAubin: I think this message could be improved, e.g. by clarifying that those addresses are reserved and automatically assigned to the first user created

locales/fr.json
frju365: well... it's because weblate is already broken that I did that. But ok. Will be removed. Only english will stay.

locales/en.json
frju365: done.

locales/fr.json
frju365: Done

alexAubin:
@zamentur : not sure what you mean : those emails are to be allocated to the first user anyway, so ?

alexAubin:
Proposing to merge in a few days

PR #555, opened 2018-10-07
[fix] Add --verbose again for backward compatibility (issue 1215)

frju365:
## The problem

fixes https://github.com/YunoHost/issues/issues/1215

## Solution

- Add --verbose option again but with a Warning. + Same result as the --debug option.

## PR Status

Work finished.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/bin/yunohost
+++ b/bin/yunohost
@@ -197,7 +197,7 @@ if __name__ == '__main__':
         sys.exit(1)
 
     parser, opts, args = _parse_cli_args()
-    _init_moulinette(opts.debug, opts.quiet)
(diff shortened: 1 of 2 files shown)


frju365:
I agree with you. It was just a request from the community. I don't have any idea if it's good or not. ;)

zamentur:
I agree with aleks.

frju365:
well, it was an issue. I tried to do something... :/

PR #556, opened 2018-10-07
[fix] Allow user to close ssh port (issue 1210)

frju365:
## The problem

...

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/firewall.py
+++ b/src/yunohost/firewall.py
@@ -203,7 +203,18 @@ def firewall_reload(skip_upnp=False):
     # Check if SSH port is allowed
     ssh_port = _get_ssh_port()
     if ssh_port not in firewall_list()['opened_ports']:
-        firewall_allow('TCP', ssh_port, no_reload=True)
+        _run_service_command('stop', 'ssh')
+        _run_service_command('disable', 'ssh')
+        logger.warning(m18n.n('ssh_port_not_opened'))
+        logger.warning(m18n.n('service_disabled', service="ssh"))
+    else
+        try:
+            _run_service_command('start', 'ssh')
+            _run_service_command('enable', 'ssh')
+        except process.CalledProcessError as e:
+            logger.warning(m18n.n('service_already_started', service="ssh"))
+        
+        
 
     # Retrieve firewall rules and UPnP status
     firewall = firewall_list(raw=True)


frju365:
For me, I'm for. Because everybody has to be free to open or close ssh port as it's a _sensitive_ port.

src/yunohost/firewall.py
```diff
@@ -203,7 +203,13 @@ def firewall_reload(skip_upnp=False):
     # Check if SSH port is allowed
     ssh_port = _get_ssh_port()
     if ssh_port not in firewall_list()['opened_ports']:
-        firewall_allow('TCP', ssh_port, no_reload=True)
+        _run_service_command('stop', 'ssh')
+        _run_service_command('disable', 'ssh')
+        logger.warning(m18n.n('ssh_port_not_opened'))
+    else:
+        _run_service_command('enable', 'ssh') 
+        _run_service_command('start', 'ssh')
```
alexAubin: Uhhhh, I'm not particularly against allowing the user to close the SSH port, but why stop and disable the SSH service ? :|

alexAubin: Yes and no, i dunno, this is a touchy piece of code ... To me asking to close port 22 should result in port 22 being closed, not resulting in the system deciding automatically to stop and disable the SSH service :/ 

Later, a user could be willing to somehow regain SSH access to its server, might end up changing the port in the SSH config, restart the SSH server and then restart the SSH daemon ... until the day its server reboots but the service is "disabled" and the SSH service won't start at boot and the user ends up being locked / requiring physical access to the machine.

That's a specific use case but I'm just not really comfortable with doing this automagically :/ The user can still manually stop and disable the SSH daemon manually (using the yunohost cli or webadmin) if he/she sure about that...

frju365: yes, but if he wanted to "regain SSH access to its server", he will have to open ssh port, no ? I prefer first : 
- not generating error log for a service which won't be used.
- securing the ssh service

But ok, if you want really me to remove those two lines, it's ok. I just find it unsecure, and really even more touchy than enable it when port is closed. It's my point of view.

Josue-T: > because ssh can't work if ssh port is not opened... did I forget something ?

ssh is independent of the firewall. In any case you can access to ssh localy (yes it's it really specific use case but it's possible.

I think too that it's better to keep ssh enabled when the firewall is closed.

frju365: ok. Let's do a little correction.

alexAubin: Well I dunno, thinking a bit more about this, I think what really troubles me is that it seems easy to do the "mistake" of disabling the SSH port for some reason, and then getting locked out of your server ...

Imho the ideal solution would be something like this : 
- allowing the closing of the SSH port only via CLI
- asking for explicit confirmation for this specific port ?

frju365: mh... yes. Need work. For the first PR, I removed "disable/stop" commands.

frju365: But I let the warning.

zamentur:
For me, the danger is those users could close the port by lack of knowledge... 

A user could consider not needed to keep ssh open because he or she don't use it! The problem is when the web admin will be done he or she will reconsider it but it will be too late. Not sure those person will be available to activate a rescue boot or other things like that.

I understand the idea of ths PR but i think we should be careful.

Put an option to change the SSH port: ok, deactivate SSH: no, we shouldn't.

zamentur:
Other idea: we could give the possibility to specify which ips can connect with ssh. In some context you could want to limit this to your local network.

frju365:
well I keep thinking a user must be able to close or open what he wants. Perhaps add a warning and be able to know if he is connected through ssh or admin interface. Anyway, he will be able to reopen it through admin interface, so I don't understand the problem ?

alexAubin:
> he will be able to reopen it through admin interface, so I don't understand the problem ?

That's assuming the webadmin is up and working. We regularly encounter issues where users have a broken nginx, or can't connect because of some HTTPS stuff, or the yunohost-api is down/broken somehow (c.f. https://github.com/YunoHost/moulinette/pull/182 )

frju365:
Mh : we could just allows the Local ip 192.168.1.* to connect to the server through ssh ?

zamentur:
I add that on meeting topics (4th december) https://pad.lqdn.fr/p/yunohost-04-12-2018

alexAubin:
Conclusion of today's meeting : we definitely don't want to allow people to just close their ssh port just like that ...

Proposed alternatives are : 
- implement a setting to close (or allow to close) SSH port
- or ask explicit confirmation via the CLI like it's done for a few other things

PR #558, opened 2018-10-10
Do not fail on missing fail2ban config during the backup

maniackcrudelis:
## The problem

Especially with the migration to stretch, it happens quite often that a file is missing in an app for php or fail2ban.
But, if a fail2ban config is missing, the app will still be working.
The current behavior can break an upgrade without any restoration of the backup, because a file is missing during the backup.

## Solution

Print a warning but do not fail if a file is missing, except if this file is really important for the app.

## PR Status

Tested on a VM.
Can be reviewed.

## How to test

Replace the helper, and remove a file used during the backup.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -46,7 +46,6 @@ ynh_backup() {
     local SRC_PATH="$1"
     local DEST_PATH="${2:-}"
     local IS_BIG="${3:-0}"
-    local NEEDED="${4:-0}"
     BACKUP_CORE_ONLY=${BACKUP_CORE_ONLY:-0}
 
     # If backing up core only (used by ynh_backup_before_upgrade),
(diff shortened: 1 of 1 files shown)


data/helpers.d/filesystem
```diff
@@ -60,8 +61,14 @@ ynh_backup() {
...
     [[ -e "${SRC_PATH}" ]] || {
-        echo "Source path '${SRC_PATH}' does not exist" >&2
-        return 1
+        echo "!!! Source path '${SRC_PATH}' does not exist !!!" >&2
+        if [ $NEEDED -eq 1 ]
+        then
+                return 1
+        else
+                echo "The missing file will be replaced by a dummy one for the backup !!!" >&2
+                echo "This file is dummy... Please contact the maintainer of this app..." | tee "${SRC_PATH}" > /dev/null
```
zamentur: I suggest to remove this line. It could break the system by creating a conf file with a bad content not interpretable by the service that manage this conf file

maniackcrudelis: Indeed, forgot that...
Still, an admin should be informed that there were an error with the backup.

Josue-T: Mybe we just can inform the admin that the file is missing and explain how to create a dummy file.

maniackcrudelis: Create a dummy file is what is going to save the process. So we can't avoid that otherwise this PR has no meaning.
But the question was more, how to inform an admin ?

maniackcrudelis:
Well, I change this PR to not fail only on fail2ban config files, and replace it by blank files, which is not going to cause any issues with fail2ban.
I tried that fix with wordpress. It works as expected.

alexAubin:
So shall we merge this for 3.3 stable ?

maniackcrudelis:
This should be merged as soon as possible. To prevent any further failures for our users.

PR #559, opened 2018-10-13
Update Metronome configuration file to v3.11 standard

maranda:
## The problem
Some now wanted features aren't loaded because Metronome uses a very old file

## Solution
Conform configuration file to latest v3.11 release file, Metronome v3.11.x is best package or Metronome might throw a few (non-fatal) errors on startup

- [ ] Principle agreement 1/2: @M5oul
- [ ] Quick review 0/1:
- [x] Simple test 1/1: @M5oul 
- [x] Deep review 1/1: @M5oul

### DIFF
--- a/data/templates/metronome/metronome.cfg.lua
+++ b/data/templates/metronome/metronome.cfg.lua
@@ -65,7 +65,7 @@ modules_enabled = {
 		--"service_directory"; -- Enables Service Directories extension support
 		--"public_service"; -- Enables Server vCard support for public services in directories and advertises in features
 		--"register_api"; -- Provides secure API for both Out-Of-Band and In-Band registration for E-Mail verification
-		"websocket"; -- Enable support for WebSocket clients, aka "XMPP over WebSockets"
+		--"websocket"; -- Enable support for WebSocket clients, aka "XMPP over WebSockets"
 };
 
 -- Server PID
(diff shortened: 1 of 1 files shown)


M5oul:
To test it on a production instance:
```bash
cd /usr/share/yunohost/templates/metronome
rm metronome.cfg.lua
wget https://raw.githubusercontent.com/YunoHost/yunohost/19120cfa60c02303eb7a0632ffb10f0ad5c48b37/data/templates/metronome/metronome.cfg.lua
sudo yunohost service regen-conf metronome --force
```

M5oul:
I don’t think:
```bash
sudo yunohost service regen-conf metronome --force
```
will happen.
Is there a way to run it automatically. Using a migration script?
It may erase some custom configurations.

alexAubin:
>I don’t think:
>
>`sudo yunohost service regen-conf metronome --force`
>
>will happen.
>Is there a way to run it automatically. Using a migration script?
>It may erase some custom configurations.

Is the `--force` needed here ? Won't it be regenerated by a simple regen-conf (which is done at every regular upgrade of the yunohost package) as long as the conf was not manually edited ?

M5oul:
> Is the --force needed here ? Won't it be regenerated by a simple regen-conf (which is done at every regular upgrade of the yunohost package) as long as the conf was not manually edited ?

Good question. Without the `--force` it wasn’t working. May be because I did change the configuration file. It could be a security avoiding custom configuration to be erased.

In case someone did modify this file, they won’t profit from all Metronome’s features.

M5oul:
Ok, looks good to merge soon.
Concerning the regen-conf, I propose to notify people in the release annoncement that they can regenerate Metronome configuration with upper command to enable all the features.

alexAubin:
Tested today and a regular regenconf was enough.

Yolomergin' /o/

PR #560, opened 2018-10-15
Pass Host header to YunoHost API

randomstuff:
This is useful to validate Origin/Referer headers in order to prevent CSRF.

## The problem

We might want to validate `Referer`/`Origin` headers to prevent CSRF but the `Host` header passed to the application is `localhost:$port` instead of the original one.

## Solution

Let nginx pass the original `Host` header.

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Yolomergin'

alexAubin:
(Uh dunno why the target branch was master, but anyway I cherry-picked this to stretch-unstable)

PR #563, opened 2018-10-19
[enh] Set Path as full-path

frju365:
## The problem

- Problem during test because of this. I got /etc/nginx/conf.d/conf.d/...

## Solution

- Set a real path

## PR Status

Tested. Micro-decision.

## Future

- Perhaps add gixy tool in core

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -73,6 +73,6 @@ server {
         access_by_lua_file /usr/share/ssowat/access.lua;
     }
 
-    include conf.d/yunohost_admin.conf.inc;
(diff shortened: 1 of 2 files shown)


alexAubin:
Uh but how did you end up with this `/etc/nginx/conf.d/conf.d/` thing ? o.O"

Josue-T:
I thought about using gixy in the CI. I think it could be really a good improvement about security.

frju365:
Yes. Indeed.

frju365:
done.

alexAubin:
Yolomergin' ¯\\\_(ツ)\_/¯

PR #564, opened 2018-10-19
Better Configuration of nginx

frju365:
## The problem

- Path-traversal not fixed (not really vulnerable in these cases)
- "Nested "add_header" drops parent headers." (Gixy test)

## Solution

- Use more_set_headers which allows us to set and clear output and input headers (https://github.com/openresty/headers-more-nginx-module#synopsis)
- Fix Path_traversal. 

## PR Status

Tested. Need review

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -11,7 +11,7 @@ server {
         return 301 https://$http_host$request_uri;
     }
 
-    location /.well-known/autoconfig/mail {
+    location /.well-known/autoconfig/mail/ {
         alias /var/www/.well-known/{{ domain }}/autoconfig/mail;
     }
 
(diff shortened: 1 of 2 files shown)


frju365:
I forgot to say : but to test it you can use this website : https://securityheaders.com/

silkevicious:
Hi everyone.

I checked my YNH installation (both the "stable" release and the "testing" release) with the security headers.

It's not that bad actually, i got an A, we're missing the Referrer-Policy and the Feature-Policy ( https://securityheaders.com/?q=luras.space&hide=on&followRedirects=on ), but the rest looks good, i think?!

For reference, in order to improve even more, i've looked at that website headers: https://securityheaders.com/?q=https%3A%2F%2Fsecurityheaders.com%2F&hide=on&followRedirects=on

Maybe contains some hint! Hope it helps!

frju365:
Yeah, I know. For referrer : we disabled it. Perhaps it was too strict.
For feature-policy, it's quite new. I will perhaps see the topic, but after this PR.

PR #566, opened 2018-10-25
[fix] php conf files not properly removed when an app is uninstalled

alexAubin:
## The problem

As reported here https://github.com/YunoHost/issues/issues/1213, here https://blog.genma.fr/?Yunohost-Soucis-a-la-suppression-d-une-webapp and in several posts on the forum, we currently have an important bug with PHP apps where a file is not removed and makes php7.0-fpm crashes.

## Solution

Turns out `ynh_remove_fpm_config` still assumed we were in jessie with Php5 :|

## PR Status

Tested and ready

## How to test

Follow Genma's instructions with and without the PR : https://blog.genma.fr/?Yunohost-Soucis-a-la-suppression-d-une-webapp 

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/backend
+++ b/data/helpers.d/backend
@@ -237,10 +237,10 @@ ynh_add_fpm_config () {
 ynh_remove_fpm_config () {
 	local fpm_config_dir=$(ynh_app_setting_get $app fpm_config_dir)
 	local fpm_service=$(ynh_app_setting_get $app fpm_service)
-	# Assume php version 5 if not set
+	# Assume php version 7 if not set
 	if [ -z "$fpm_config_dir" ]; then
-		fpm_config_dir="/etc/php5/fpm"
-		fpm_service="php5-fpm"
+		fpm_config_dir="/etc/php/7.0/fpm"
+		fpm_service="php7.0-fpm"
 	fi
 	ynh_secure_remove "$fpm_config_dir/pool.d/$app.conf"
 	ynh_secure_remove "$fpm_config_dir/conf.d/20-$app.ini" 2>&1


alexAubin:
(As it's an important bug, I propose to merge this soon, and maybe push a hotfix on 3.2.x)

alexAubin:
Uh but there are no php5 file anyway, are there ...?

alexAubin:
Weeell what I can say is that during the migration to stretch, all php5 pool conf file are automatically moved to php7 pool conf files, and stuff like nginx conf files are also edited to use php7 socket.

So no PHP5 stuff should be "active" anymore, though conf file might still be there. But the php5-fpm service should not even be started (if it's still installed), let alone being used by anything..

maniackcrudelis:
OK, looks good to me then.
And what if the path in settings.yml is on php5 ? Is it even possible ?

maniackcrudelis:
My question was more about that, https://github.com/YunoHost/yunohost/commit/8590d6b5c6112581c021f81a21c45bfe7e140610
So the answer is no.

EDIT: Actually, looks like it was merge in jessie as well, https://github.com/YunoHost/yunohost/blob/stable/data/helpers.d/backend
So, yes an app could have store the php5 path of its config in settings.yml

alexAubin:
Hm I didnt notice those lines from the helper that add the config : 

```
ynh_app_setting_set $app fpm_config_dir "$fpm_config_dir"
ynh_app_setting_set $app fpm_service "$fpm_service"
```

Now I don't understand anymore why it wasnt working (in the context of installing then removing the app on Stretch as described in Genma's post) ... but if I understand correctly, this could also happen that this var is not empty because defined while being on Jessie ? 

If so, I don't know what's the proper fix except something like : 

```
if (fpm_config_dir is empty) or (fpm_config_dir == "/etc/php5/fpm")
then
    fpm_config_dir = "/etc/php/7.0/fpm"
endif
```

maniackcrudelis:
Hum, since it's a remove helper and not removing a config file could create an error.
I think removing all possible config file would be the safest way to do it.

alexAubin:
Meh I don't understand the other part of the bug related to old apps from a jessie install (c.f. the fpm_config_dir in https://github.com/YunoHost/yunohost/commit/8590d6b5c6112581c021f81a21c45bfe7e140610 ) and too tired to dig this ... but at least the current PR should fix the issue of new app install in stretch as described by Genma, so merging this for now...

PR #567, opened 2018-10-25
Cracklib enh

alexAubin:
c.f. comments https://github.com/YunoHost/yunohost/pull/196 , since there are many fixes to be done imho, this was a bit too much to merge directly into the PR. So opening this instead. Everything is up to discussion.

Basically I tried to simplify the code, make it more semantic, fixed/cleaned a few weird/unused things and a few other arbitrary changes :s


alexAubin:
Merging with ljf's agreement

PR #571, opened 2018-11-06
[enh] Wait for dpkg lock to be free

maniackcrudelis:
## The problem

We have a lot of failed upgrade because of apt couldn't get its lock file because of another apt running in background.

## Solution

Add an internal helper to check before any apt helper if the lock is free. Or wait to run apt.

## PR Status

Tested on a VM
Ready to be reviewed.

## How to test

Run an apt command that will hang, waiting for an answer from the user. Like an install or an upgrade.
Then try to use an apt helper.

## Validation

- [x] Principle agreement 2/2 : frju365, ljf
- [x] Quick review 1/1 : Josué
- [ ] Simple test 0/1 : 
- [x] Deep review 1/1 : ljf

### DIFF
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -2,8 +2,8 @@
 #
 # [internal]
 #
-# usage: ynh_is_dpkg_free
-ynh_is_dpkg_free() {
+# usage: ynh_wait_dpkg_free
+ynh_wait_dpkg_free() {
     local try
     # With seq 1 17, timeout will be almost 30 minutes
     for try in `seq 1 17`
@@ -28,7 +28,7 @@ ynh_is_dpkg_free() {
 # usage: ynh_package_is_installed name
 # | arg: name - the package name to check
 ynh_package_is_installed() {
-    ynh_is_dpkg_free
+    ynh_wait_dpkg_free
     dpkg-query -W -f '${Status}' "$1" 2>/dev/null \
         | grep -c "ok installed" &>/dev/null
 }
@@ -54,7 +54,7 @@ ynh_package_version() {
 #
 # usage: ynh_apt update
 ynh_apt() {
-    ynh_is_dpkg_free
+    ynh_wait_dpkg_free
     DEBIAN_FRONTEND=noninteractive sudo apt-get -y $@
 }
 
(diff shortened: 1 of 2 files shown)


data/helpers.d/package
```diff
@@ -1,10 +1,34 @@
+# Check if apt is free to use, or wait, until timeout.
+#
+# [internal]
+#
+# usage: ynh_is_dpkg_free
+ynh_is_dpkg_free() {
```
Josue-T: Is it better to use a name like `ynh_wait_dpkg_free` to explain that it's not just a check.

data/helpers.d/package
```diff
@@ -1,10 +1,34 @@
...
+    local try
+    # With seq 1 17, timeout will be almost 30 minutes
+    for try in `seq 1 17`
+    do
+        # Check if /var/lib/dpkg/lock is used by another process
+        if sudo lsof /var/lib/dpkg/lock > /dev/null
```
Josue-T: Does we really need `sudo` as normally all helper should be called as root ?

data/helpers.d/package
maniackcrudelis: Honestly, I don't care about the name of the helper.

data/helpers.d/package
maniackcrudelis: lsof won't work if it's not root, I added it for my own tests and I forgot to remove it...
Still, It's maybe not useless if an user want to use it on his own.
The problem is, without root, lsof will just don't see any process using the lock. So I'd rather prefer to keep it, even if it's going to be useless in most cases.

frju365:
LGTM for me (principle argreement) but as I can't edit post, I can't tick the box. ;)

Josue-T:
> LGTM for me (principle argreement) but as I can't edit post, I can't tick the box. ;)

I wrote your name in the "principle argreement".

frju365:
Thanks ;)

data/helpers.d/package
zamentur: Change done

data/helpers.d/package
zamentur: Note lsof command is here because mariadb is installed. Might we add this requirement in our packages in case of mariadb will be remove in future ?

maniackcrudelis: I have this package on my own debian (buster though), but I guess it could be better to have it as a dependency.

PR #573, opened 2018-11-13
[fix] app_removeaccess call set.add

airwoodix:
## The problem

In `yunohost/app.py`, `app_removeaccess` calls `allowed_users.append` which doesn't exist for the `set` `allowed_users`.

## Solution

Call `allowed_users.add`.

## PR Status

Trivial change. Tested in production. Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Many thanks !

PR #576, opened 2018-11-20
[enh] Allow to not fail on backup and restore for some files

maniackcrudelis:
## The problem

Following https://github.com/YunoHost/yunohost/pull/558, this PR allow to specify that, for a file, the backup process can continue if the file is missing.
The same argument has to be added to the restore command to ignore it as well.

## Solution

Simply ignore the file during the backup and the restore process, and do not fail.

## PR Status

Tested on a VM.
Can be reviewed.

## How to test

Add the argument in the backup and restore script for a file.
Replace the helper, and remove a file used during the backup.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -10,12 +10,13 @@ CAN_BIND=${CAN_BIND:-1}
 # 
 # If DEST is ended by a slash it complete this path with the basename of SRC.
 #
-# usage: ynh_backup src [dest [is_big [arg]]]
+# usage: ynh_backup src [dest [is_big [not_mandatory [arg]]]]
 # | arg: src - file or directory to bind or symlink or copy. it shouldn't be in
 # the backup dir. 
 # | arg: dest - destination file or directory inside the
 # backup dir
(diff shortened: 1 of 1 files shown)


data/helpers.d/filesystem
```diff
@@ -46,6 +47,7 @@ ynh_backup() {
     local SRC_PATH="$1"
     local DEST_PATH="${2:-}"
     local IS_BIG="${3:-0}"
+    local NOT_MANDATORY="${4:-0}"
```
zamentur: If there isn't a 4th argument, could i suggest to put 1 by default if $SRC_PATH contains "fail2ban" ? (same in restore)

maniackcrudelis: There's already this PR  (https://github.com/YunoHost/yunohost/pull/558) for fail2ban.

zamentur: ok

alexAubin:
So uh, not sure what to do with the conflict with the fail2ban thing

PR #577, opened 2018-11-21
[fix] Unknown variable password in user_update

zamentur:
## The problem
Error on missing var

## Solution

Rename the var with the correct name

## PR Status

Ready


zamentur:
I merge that because it's a micro decision

PR #579, opened 2018-11-22
Update ECDH curves recommended by Mozilla, now that we are on stretch

liberodark:
Better security is option is not recommended.

## The problem

...

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -37,7 +37,7 @@ server {
     # ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
 
     # As suggested by https://cipherli.st/
-    ssl_ecdh_curve secp384r1;
+    #ssl_ecdh_curve secp384r1;
 
     ssl_prefer_server_ciphers on;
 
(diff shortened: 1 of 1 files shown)


frju365:
Can you add a description of source and perhaps of example of online conf which had removed it ?

For the conf I based my conf on this [conf](https://github.com/stylersnico/nginx-secure-config/blob/master/nginx.conf) but it's doesn't remove this part.

liberodark:
Yes based on this work https://doc.probetech.ovh/services_sec/services_nginx/#our_ssl_certificate_a_approved_with_full_compliance

data/templates/nginx/server.tpl.conf
```diff
@@ -37,7 +37,7 @@ server {
     # ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
```
alexAubin: Note the comment here, which apparently was meant to be used once on Stretch...

frju365:
But, we can perhaps enable stretch compatibility (the curves which were in comment)

alexAubin:
>But, we can perhaps enable stretch compatibility (the curves which were in comment)

Yes :+1:

alexAubin:
So as discussed, we shall instead use the curves recommended by Mozilla, now that they are available on Stretch.

I'm yolomerging this as it was already agreed in previous PR (we just fell back to curves available on jessie while we were waiting for stretch)

PR #581, opened 2018-11-22
New Cyphers

liberodark:
Better Security and active for moderne navigator and support new web standard

TLS 1.0 / 1.1 would be remove in some navigators FF / Chrome / Safari / Edge

## The problem

...

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


zamentur:
The last discussion about cyphers conclude we shouldn't remove some protocols cause some phones are not updatable and couldn't connect if we strengthen the security.

The exact conclusion was: we need to let the admin make a choice between strengthenning security and limiting potential obsolescence.

So we should create a new security parameter and use it during the nginx configuration regeneration. See the command yunohost setting for more details.

So this PR should be completed to manage almost 2 levels of security in order to avoid some user dropping their smartphone.

zamentur:
Note about the cypher used, we prefer use some well known suite as those one : https://mozilla.github.io/server-side-tls/ssl-config-generator

PR #585, opened 2018-11-26
Group permission

Josue-T:
## The problem

- We can't have a group of user.
- In the application we can't know which user is allowed to access to.
- The metronome, sftp and mail server allow all user to access. There are no possibility to limite the user to access.

## Solution

- Implement a group management.
- Implement a concept of permission which will give many new feature and flexibility.

## PR Status

Note that this PR depends of : 
- https://github.com/YunoHost/moulinette/pull/181
- https://github.com/YunoHost/moulinette/pull/183
- https://github.com/YunoHost/moulinette/pull/186
- https://github.com/YunoHost/moulinette/pull/201
- https://github.com/YunoHost/yunohost/pull/599
- https://github.com/YunoHost/yunohost/pull/608 (for the tests)
- https://github.com/YunoHost/test_apps/pull/3
- https://github.com/YunoHost/moulinette/pull/189 (Just for documentation)

It can be tested and reviewed. Theses following things will come nearly : 

- [ ] Check decoration for operation logger
- [ ] More Test
- [x] Update documentation
- [x] Implement a migration for the actual instances
- [x] Unit Test

Optional (depends of my time) :

- [ ] ~~Implement 2 password for admin (one for the user and one for the LDAP authentication by the moulinette)~~
- [x] Use root id for LDAP authentication (see PR ...)
- [x] Create some helper to manage the permission (create, remove, manage permission URL)
- [ ] ~~Update SSOWAT to manage the new ssowat config~~ (will be in a next project)
- [ ] ~~Add theses new features in the webadmin~~ (will be in a next project)

## How to test

Actually it's not possible to use theses new feature with the actual LDAP schema, so you need to do the postinstall to have clean LDAP schema with the new features.

With ynhdev :

- for the yunohost repos checkout the branch "group_permission"
- For the moulinette repos checkout the branch "LDAP_validate_uniqueness" and "sasl_authentication"
- Do the postinstall
- Play with the new features

#### The group management

You should be able to create a group of a user by this command :

```
yunohost user group add YOUR_GROUP
```
After you should be able add a user in your group :

```
yunohost user group update YOUR_GROUP -a YOUR_USER
```

You can also list all of group in Yunohost by this command :
```
# yunohost user group list
groups:
  ALL:
    groupname: ALL
    members: 
      - YOUR_USER1
      - YOUR_USER2
  YOUR_USER1: 
    groupname: YOUR_USER1
    members: YOUR_USER1
  YOUR_USER2: 
    groupname: YOUR_USER2
    members: YOUR_USER2
  YOUR_GROUP: 
    groupname: YOUR_GROUP
    members: YOUR_USER
```

#### The permission management

By default you have 3 permission :
- mail (main)
- sftp (main)
- metronome (main)

By default all user is permitted. If you want to specify the permission you can use the app access like concept :

```
# Allow YOUR_USER1 to access to the mail
yunohost user permission add mail -u YOUR_USER1
# Disallow all user of group YOUR_GROUP to access to the mail
yunohost user permission remove mail -g YOUR_GROUP
# Clear the permission -> Allow all user to access to the mail
yunohost user permission clear mail
```

Note that it's just some example.

### Integration in application

To give the possiblitly to the application to know which user is allowed access to it you need to configure the LDAP connector with theses parameter : 
```
Host:               ldap://localhost
Port:               389
Base DN:            dc=yunohost,dc=org
User DN:            ou=users,dc=yunohost,dc=org
fiter :             (&(objectClass=posixAccount)
                        (permission=cn=main.APP1,ou=permission,dc=yunohost,dc=org))
LDAP Username:      uid
LDAP Email Address: mail
```

Note that the main diffrence is a new filter.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


dromer:
Very cool that you're working on this @Josue-T 

To know how to get SSOwat to read and display group-based apps of the user have a look at my Proof of Concept PR here: https://github.com/YunoHost/SSOwat/pull/62/files

zamentur:
Sounds great, I hope find time to test it during the week end :)

arthurlogilab:
:tada: Great news to see this patch arrive. Will try to find the time to review it and test it since it's something we needs.

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,108 @@
...
+        migrate_LDAP_db(auth)
+
+        # Migrate permission
+        migrate_app_permission(auth)
+
+        permission_sync_to_user(auth)
+        logger.info(m18n.n("migration_0009_done"))
+
+    @property
+    def disclaimer(self):
```
zamentur: This migration should be done automatically no ?

Josue-T:
I think It should be ready for a first review.

I also updated the doc here : https://github.com/YunoHost/moulinette/pull/189

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+                            entry[group_attr[attr]][permission] = [pType]
+                else:
+                    entry[group_attr[attr]] = values[0]
+
+        groupname = entry[group_attr['cn']]
+        groups[groupname] = entry
+    return {'groups' : groups}
+
+
+@is_unit_operation([('groupname', 'user')])
```
alexAubin: Uh it's not clear to me how this `is_unit_operation` works but eh, what `'user'` are we talking about ? Maybe something missing here, idk... #drunk

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+
+    if (groupname == 'all_users' or groupname == 'admins') and not force:
+        raise YunohostError('edit_group_not_allowed', group=groupname)
+
+    # Populate group informations
+    result = auth.search(base='ou=groups,dc=yunohost,dc=org',
+                         filter='cn=' + groupname, attrs=attrs_to_fetch)
+    if not result:
+        raise YunohostError('group_unknown', group=groupname)
+    group = result[0]
```
alexAubin: Uh what happens if there's more than 1 entry ? Or should it never happen ?

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+
+    new_group_list = {'member': set(), 'memberUid': set()}
+    if 'member' in group:
+        new_group_list['member'] = set(group['member'])
+    else:
+        group['member'] = []
+
+    user_l = user_list(auth, ['uid'])['users']
```
alexAubin: Eh, when possible please let's explicit the arguments to improve readability :S 

So in that case : `user_list(auth, fields=['uid'])` instead of just `user_list(auth, ['uid'])`

alexAubin: Let's call this list `existing_users` for instance ?

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+
+    if add_user:
+        if not isinstance(add_user, list):
+            add_user = [add_user]
+        for user in add_user:
+            if not user in user_l:
+                raise YunohostError('user_unknown', user=user)
```
alexAubin: Hm I think I would split this check in a separate step, in order to avoid starting doing stuff for the first user, and the in fact in the middle of the list, realizing that a user does not exists and raising an error.

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+                logger.warning(m18n.n('user_already_in_group', user=user, group=groupname))
+            new_group_list['member'].add(userDN)
+
+    if remove_user:
+        if not isinstance(remove_user, list):
+            remove_user = [remove_user]
+        for user in remove_user:
+            userDN = "uid=" + user + ",ou=users,dc=yunohost,dc=org"
+            if user == groupname:
+                raise YunohostError('remove_user_of_group_not_allowed', user=user, group=groupname)
```
alexAubin: same as for the previous block : let's have do this check independently of starting to actually do stuff ?

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+        member_Uid = member.split("=")[1].split(",")[0]
+        # Don't add main user in the group.
+        # Note that in the Unix system the main user of the group is linked by the gid in the user attribute.
+        # So the main user need to be not in the memberUid list of his own group.
+        if member_Uid != groupname:
+            new_group_list['memberUid'].add(member_Uid)
+
+    operation_logger.start()
+
+    if new_group_list['member'] != set(group['member']):
```
alexAubin: Uh can we add a comment here to explain what's going on ? I guess we want to check the two sets are different but why ... (might be obvious but im kinda drunk)

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+        if not auth.update('cn=%s,ou=groups' % groupname, new_group_list):
+            raise YunohostError('group_update_failed')
```
alexAubin: Ideally it would be nice to include details on why it failed otherwise we won't be able to debug :/ (Maybe looking at similar action, I dunno if we can easily get details on why it failed)

src/yunohost/user.py
```diff
@@ -462,6 +469,276 @@ def user_info(auth, username):
...
+
+    """
+    group_attrs = [
+        'cn', 'member', 'permission'
+    ]
+    result = auth.search('ou=groups,dc=yunohost,dc=org', "cn=" + groupname, group_attrs)
+
+    if not result:
+        raise YunohostError('group_unknown', group=groupname)
+    else:
```
alexAubin: Since there's a raise in the if, you don't really need the `else` here and therefore can avoid an unecessary indent level :wink:

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+from yunohost.permission import permission_add, permission_sync_to_user
+from yunohost.user import user_permission_add
+
+logger = getActionLogger('yunohost.migration')
+
+###################################################
+# Tools used also for restoration
+###################################################
+
+def migrate_LDAP_db(auth):
```
alexAubin: You might want to move this inside the migration ? I guess you put it outside because you wanted to access it by just loading the module but maybe with `_get_migration_by_name` that should be okay :stuck_out_tongue_winking_eye:

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+    user_list = auth.search('ou=users,dc=yunohost,dc=org',
+                            '(&(objectclass=person)(!(uid=root))(!(uid=nobody)))',
+                            ['uid', 'uidNumber'])
+    for user_info in user_list:
+        username = user_info['uid'][0]
+        auth.update('uid=%s,ou=users' % username,
+                    {'objectClass': ['mailAccount', 'inetOrgPerson', 'posixAccount', 'userPermissionYnh']})
+        user_group_add(auth, username, gid=user_info['uidNumber'][0], sync_perm=False)
+        user_group_update(auth, groupname=username, add_user=username, force=True, sync_perm=False)
+        user_group_update(auth, 'all_users', add_user=username, force=True, sync_perm=False)
```
alexAubin: `groupname="all_users"` please :sweat_smile:

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def migrate_app_permission(auth, app=None):
+    logger.info(m18n.n("migration_0009_migrate_permission"))
+
+    if app:
+        apps = app_list(installed=True, filter=app)['apps']
```
alexAubin: Don't we want to assert that the result is not empty here ? Otherwise that mean the app does not exists ... Or maybe we don't care so much, idk

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+        path = app_setting(app, 'path')
+        domain = app_setting(app, 'domain')
+
+        url = None
+        if domain and path:
+            url = domain + path
+        permission_add(auth, app, 'main', urls=[url], default_allow=True, sync_perm=False)
+        if permission:
+            allowed_group = permission.split(',')
+            user_permission_add(auth, [app], 'main', group=allowed_group, sync_perm=False)
```
alexAubin: `permission="main"` ? :S

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+    for app_info in apps:
+        app = app_info['id']
+        permission = app_setting(app, 'allowed_users')
+        path = app_setting(app, 'path')
+        domain = app_setting(app, 'domain')
+
+        url = None
+        if domain and path:
+            url = domain + path
+        permission_add(auth, app, 'main', urls=[url], default_allow=True, sync_perm=False)
```
alexAubin: `permission="main"` ? :S

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            backup_folder = "/home/yunohost.backup/premigration/" + time.strftime('%Y%m%d-%H%M%S', time.gmtime())
+            os.makedirs(backup_folder, 0o750)
+            os.system("systemctl stop slapd")
+            os.system("cp -r --preserve /etc/ldap %s/ldap_config" % backup_folder)
+            os.system("cp -r --preserve /var/lib/ldap %s/ldap_db" % backup_folder)
+            os.system("cp -r --preserve /etc/yunohost/apps %s/apps_settings" % backup_folder)
+            os.system("systemctl start slapd")
+        except Exception as e:
+            raise YunohostError("migration_0009_can_not_backup_before_migration", error=e)
```
alexAubin: I dunno if that's really needed, but maybe to be secure we might want to add a : 
```
   os.system("systemctl restart slapd")
```
in the exception block ?

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Update LDAP schema restart slapd
+        logger.info(m18n.n("migration_0009_update_LDAP_schema"))
+        service_regen_conf(names=['slapd'], force=True)
```
alexAubin: Hmmm I dunno what to do about this ... I'm trying to think about case where the admin tweaked the slapd conf (which is likely a very bad idea). Maybe we could think about a mechanism to check that the conf was not manually edited, and if that's the case abort the migration at the very beginning, instead of forcing a regenconf here ?

src/yunohost/data_migrations/0009_setup_group_permission.py
```diff
@@ -0,0 +1,142 @@
...
+    def migrate(self):
+        # Backup LDAP and the apps settings before to do the migration
+        logger.info(m18n.n("migration_0009_backup_before_migration"))
+        try:
+            backup_folder = "/home/yunohost.backup/premigration/" + time.strftime('%Y%m%d-%H%M%S', time.gmtime())
+            os.makedirs(backup_folder, 0o750)
+            os.system("systemctl stop slapd")
+            os.system("cp -r --preserve /etc/ldap %s/ldap_config" % backup_folder)
+            os.system("cp -r --preserve /var/lib/ldap %s/ldap_db" % backup_folder)
+            os.system("cp -r --preserve /etc/yunohost/apps %s/apps_settings" % backup_folder)
```
alexAubin: Would a `slapcat` dump be also meaningful here to backup the existing slapd database ?

src/yunohost/app.py
```diff
@@ -896,7 +909,14 @@ def confirm_install(confirm):
     os.system('chown -R root: %s' % app_setting_path)
     os.system('chown -R admin: %s/scripts' % app_setting_path)
 
-    app_ssowatconf(auth)
+    # Add path in permission if it's defined in the app install script
+    app_settings = _get_app_settings(app_instance_name)
+    domain = app_settings['domain']
```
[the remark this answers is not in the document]
Josue-T: Yes like that ?
```suggestion
    if app_settings['domain']:
        domain = app_settings['domain']
```

src/yunohost/app.py
```diff
@@ -1032,93 +1014,31 @@ def app_removeaccess(auth, apps, users=[]):
...
-
-        if 'allowed_users' in app_settings:
-            app_setting(app, 'allowed_users', delete=True)
-
-        hook_callback('post_app_clearaccess', args=[app])
-
-        operation_logger.success()
-
-    app_ssowatconf(auth)
+    result = {p : v['main']['allowed_users'] for p, v in permission['permissions'].items()}
```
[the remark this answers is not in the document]
Josue-T: Should be fixed

src/yunohost/backup.py
```diff
@@ -887,11 +893,29 @@ def _postinstall_if_needed(self):
...
+        Do some migration if needed
+        """
+
+        # Check if we need to do the migration 0009 : setup group and permission
+        result = auth.search('ou=groups,dc=yunohost,dc=org',
+                             '(&(objectclass=groupOfNamesYnh)(cn=all_users))',
+                             ['cn'])
+        if not result:
+            from importlib import import_module
+            migration_0009 = import_module('yunohost.data_migrations.0009_setup_group_permission')
```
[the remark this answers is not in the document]
Josue-T: Yes probably.

Didn't know how to do that...

src/yunohost/app.py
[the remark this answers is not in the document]
alexAubin: Was more thinking about `domain = app_settings.get("domain", None)` (otherwise `domain` might not be defined later and you are in trouble). Again, pretty small edge case anyway

src/yunohost/backup.py
```diff
@@ -900,6 +924,8 @@ def clean(self):
             # ldap restore hooks
             os.system('sudo yunohost app ssowatconf')
 
+        permission_sync_to_user(auth, force=True)
```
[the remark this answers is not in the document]
Josue-T: Done

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: Yes because the main permission for each apps is "main".

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: Yes, a group which contains all users.

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: It should be already done by this `os.system("cp -r --preserve /var/lib/ldap %s/ldap_db" % backup_folder)`

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: Yes, but what happens if `systemctl restart slapd` generate an exception too ?

src/yunohost/tools.py
```diff
@@ -68,11 +68,11 @@ def tools_ldapinit():
...
-    auth = init_authenticator(('ldap', 'default'),
-                              {'uri': "ldap://localhost:389",
-                               'base_dn': "dc=yunohost,dc=org",
-                               'user_rdn': "cn=admin"})
-    auth.authenticate('yunohost')
+    AUTH_IDENTIFIER = ('ldap', 'as-root')
+    AUTH_PARAMETERS = {'uri': 'ldapi://%2Fvar%2Frun%2Fslapd%2Fldapi',
+                       'base_dn': 'dc=yunohost,dc=org',
+                       'user_rdn': 'gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth'}
+    auth = init_authenticator(AUTH_IDENTIFIER, AUTH_PARAMETERS)
```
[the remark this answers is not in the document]
Josue-T: I thought about doing something like that but as I remember it was not easily possible. But not really sure...

src/yunohost/data_migrations/0009_setup_group_permission.py
alexAubin: I mean my point was to add the `groupname=` to be more explicit (and symetric with the previous commands)

src/yunohost/data_migrations/0009_setup_group_permission.py
alexAubin: I mean my point was to add the `permission=` to be more explicit :s

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: Ok

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: Ok

src/yunohost/app.py
```diff
@@ -828,6 +831,9 @@ def confirm_install(confirm):
         if os.path.exists(os.path.join(extracted_app_folder, file_to_copy)):
             os.system('cp -R %s/%s %s' % (extracted_app_folder, file_to_copy, app_setting_path))
 
+    # Create permission before the install (useful if the install script redefine the permission)
+    permission_add(auth, app=app_instance_name, permission="main")
```
alexAubin: Any chance that we can add `sync_perm=False` here ? Like, to avoid triggering a whole bunch of code and messages ? I dunno what are the implications tho :s

src/yunohost/permission.py
```diff
@@ -0,0 +1,510 @@
...
+        if sync_perm:
+            permission_sync_to_user(auth)
+        logger.success(m18n.n('permission_created', permission=permission, app=app))
+        return user_permission_list(auth, app, permission)
+
+    raise YunohostError('permission_creation_failed')
+
+
+@is_unit_operation(['permission','app'])
+def permission_update(operation_logger, auth, app, permission, add_url=None, remove_url=None, sync_perm=True):
```
alexAubin: It seems to me that this `update` action ain't available from the actionmap and should be ? Or am I missing something ?

src/yunohost/permission.py
```diff
@@ -0,0 +1,510 @@
...
+            permissions[app_name][permission_name]['allowed_users'].append(u)
+        if 'URL' in res:
+            permissions[app_name][permission_name]['URL'] = []
+            for u in res['URL']:
+                permissions[app_name][permission_name]['URL'].append(u)
+
+    return {'permissions': permissions}
+
+
+def user_permission_update(operation_logger, auth, app=[], permission=None, add_username=None, add_group=None, del_username=None, del_group=None, sync_perm=True):
```
alexAubin: Same question : shouldn't this be available from the actionmap ? Or maybe it is available through the `yunohost app [addaccess|removeaccess|...]` but it's a bit confusing ...

src/yunohost/tools.py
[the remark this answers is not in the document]
alexAubin: Okay let's see this later

src/yunohost/permission.py
```diff
@@ -0,0 +1,513 @@
...
+
+    if permission == "main" and not force:
+        raise YunohostError('remove_main_permission_not_allowed')
+
+    operation_logger.start()
+    if not auth.remove('cn=%s,ou=permission' % str(permission + '.' + app)):
+        raise YunohostError('permission_deletion_failed', permission=permission, app=app)
+    if sync_perm:
+        permission_sync_to_user(auth)
+    logger.success(m18n.n('permission_deleted', permission=permission, app=app))
```
alexAubin: I apparently triggered a bug when installing then removing an app (well actually two, but should work with one)

After removal of an app, there still are some references in the 'all_users' group (and probably in other group if those existed I guess). Spotted this thanks to the `check_LDAP_db_integrity` test :wink: 

c.f. the references to h5ai and nextcloud in this search : 

```
(Pdb) pp auth.search('ou=groups,dc=yunohost,dc=org', filter='(objectclass=groupOfNamesYnh)', attrs=['cn', 'member', 'memberUid', 'permission'])
[{'cn': ['all_users'],
  'member': ['uid=alice,ou=users,dc=yunohost,dc=org',
             'uid=jack,ou=users,dc=yunohost,dc=org',
             'uid=bob,ou=users,dc=yunohost,dc=org'],
  'memberUid': ['bob', 'alice', 'jack'],
  'permission': ['cn=main.h5ai,ou=permission,dc=yunohost,dc=org',
                 'cn=main.nextcloud,ou=permission,dc=yunohost,dc=org',
                 'cn=main.mail,ou=permission,dc=yunohost,dc=org',
                 'cn=main.metronome,ou=permission,dc=yunohost,dc=org']},
 {'cn': ['alice'], 'member': ['uid=alice,ou=users,dc=yunohost,dc=org']},
 {'cn': ['bob'], 'member': ['uid=bob,ou=users,dc=yunohost,dc=org']},
 {'cn': ['jack'], 'member': ['uid=jack,ou=users,dc=yunohost,dc=org']},
 {'cn': ['dev'],
  'member': ['uid=alice,ou=users,dc=yunohost,dc=org'],
  'memberUid': ['alice']},
 {'cn': ['apps'],
  'member': ['uid=bob,ou=users,dc=yunohost,dc=org'],
  'memberUid': ['bob']}]
```

Not sure how that happens, maybe some stuff not cleared correctly in `permission_sync_to_user`

src/yunohost/permission.py
```diff
@@ -0,0 +1,513 @@
...
+
+    permissions = {}
+
+    result = auth.search('ou=permission,dc=yunohost,dc=org',
+                         '(objectclass=permissionYnh)', permission_attrs)
+
+    for res in result:
+        permission_name = res['cn'][0].split('.')[0]
+        try:
+            app_name = res['cn'][0].split('.')[1]
```
alexAubin: Can't we do something more elegant like 

```python
permission_name, app_name = res['cn'][0].split('.')
```

?

src/yunohost/permission.py
```diff
@@ -0,0 +1,513 @@
...
+        except:
+            logger.warning(m18n.n('permission_name_not_valid', permission=per))
```
alexAubin: `per` is not defined :confused:

src/yunohost/app.py
Josue-T: Should be possible I think. The only issue that I think we can if with this idea is that the application try to update the permission and not everything is synced. But I not sure about if it's possible and if yes it should be probably a really edge case...

So I would say yes we can add `sync_perm=False` and if we have some issue we can remove it.

src/yunohost/permission.py
Josue-T: Yes because it's internal. So no user should use this function.

src/yunohost/permission.py
Josue-T: Well, this function is a merge of `yunohost user permission add, remove and clear`. So it should be not needed.

By example it's used here : https://github.com/YunoHost/yunohost/pull/585/files#diff-f4b72011f32f548ac5655aec63be4d7fR734

src/yunohost/permission.py
Josue-T: Tried with my_webapp and didn't see this issue.

Are you able to reproduce this bug ?

src/yunohost/backup.py
```diff
@@ -887,11 +893,29 @@ def _postinstall_if_needed(self):
             logger.debug("executing the post-install...")
             tools_postinstall(domain, 'yunohost', True)
 
-    def clean(self):
+    def _migrate_system_if_needed(self, auth):
```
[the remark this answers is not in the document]
Josue-T: Should be done by 975341a0

src/yunohost/app.py
Josue-T: Done with some comment to explain the question (d0f422ee)

src/yunohost/permission.py
Josue-T: Done by fd76f554

src/yunohost/permission.py
Josue-T: Done by fd76f554

src/yunohost/data_migrations/0009_setup_group_permission.py
Josue-T: I've add come code to check that and to add a message (0c0db4c8).

src/yunohost/permission.py
alexAubin: I think I was able to retrigger it using the following steps : 
- Setup a fresh install then pull this PR (and run the migration etc)
- Install Nextcloud
- Run the `backup_and_restore_all_sys` test : `./ynh-dev test yunohost/backuprestore:backup_only_ldap`

locales/en.json
```diff
@@ -460,7 +461,6 @@
     "regenconf_dry_pending_applying": "Checking pending configuration which would have been applied for category '{category}'…",
     "regenconf_failed": "Unable to regenerate the configuration for category(s): {categories}",
     "regenconf_pending_applying": "Applying pending configuration for category '{category}'…",
->>>>>>> upstream/stretch-unstable
```
kay0u: /!\

src/yunohost/permission.py
Josue-T: I've probably found the bug :smile: 

The problem is that actually when we restore only the system (without the app) we only restore the database **without** the permission (because the backup the database without the app permission). The problem is that if we have an app which is installed (before the restoration) and which is not restored, the permission in the "old" database will be replaced by the new database which don't contains any permission !! It's an edge case but when I'm seeing this bug I'm really thinking that the restoration need to be managed more cleanly.

The other problem is that we use in some place the `slapadd` command which is really bad if we wan't to keep the database integrity. In the man of `slapadd` it's written this:
> Your slapd(8) should not be running when you do this to ensure consistency of the database.

So I purpose this solution of that:

- In the backup : save all content of the database
- In the slapd restoration we do this:
  - Save the permission
  - Restore all content of the database (with `slapadd`, but we can because It's all the database so we shouln't have any problem with integrity).
  - Remove all permission in the database
  - Restore the permission which was saved (with `ldapadd` (wich keep the integrity) or with python) for only the apps which is already installed.

With this we should keep a better integrity of the database.

Josue-T: So, now it should be fixed by f839ec51

alexAubin: Looks like it's fixed indeed, thanks !

alexAubin:
Some update on this : I kinda plan to merge this as soon as we released the 3.6 (maybe do a small iteration after the initial stable release to merge the password redact stuff ... though anyway we can cherrypick it later if needed) so that we move forward

As discussed, we still need to rework the migration framework to have a more flexible way of handling independent migrations - gotta find and allocate time to work on this

Josue-T:
> Some update on this : I kinda plan to merge this as soon as we released the 3.6 (maybe do a small iteration after the initial stable release to merge the password redact stuff ... though anyway we can cherrypick it later if needed) so that we move forward
> 
> As discussed, we still need to rework the migration framework to have a more flexible way of handling independent migrations - gotta find and allocate time to work on this

Happy to know that. For the migration it still the problem... but we firstly need to decide how would work the next migration process, before to implement it :smile:

src/yunohost/data_migrations/0009_setup_group_permission.py
alexAubin: Marking this as resolved, to be addressed in anothah PR

alexAubin:
Thanks for all the work and patience @Josue-T ! Sorry about the huge delay /o\

Josue-T:
Happy to know that is merged !!

src/yunohost/backup.py
```diff
@@ -1214,12 +1232,44 @@ def _restore_system(self):
...
+            logger.info(m18n.n("migration_0011_update_LDAP_schema"))
+            regen_conf(names=['slapd'], force=True)
+            setup_group_permission.migrate_LDAP_db()
+
+        # Remove all permission for all app which sill in the LDAP
+        for per in ldap.search('ou=permission,dc=yunohost,dc=org',
+                               '(&(objectClass=permissionYnh)(!(cn=main.mail))(!(cn=main.metronome))(!(cn=main.sftp)))',
+                               ['cn']):
+            if not ldap.remove('cn=%s,ou=permission' % per['cn'][0]):
+                raise YunohostError('permission_deletion_failed', permission=permission, app=app)
```
alexAubin: There are several issues in this file ... e.g. here `permission` and `app` are not defined in this context :/ 

And a few line below, `permission_name` and `app_name` aren't defined either :/

Josue-T: Normally this should fix the issue:

```suggestion
                raise YunohostError('permission_deletion_failed', permission=per['cn'][0].split('.')[0], app=per['cn'][0].split('.')[1])
```

Josue-T: Should be fixed by https://github.com/YunoHost/yunohost/pull/763

Psycojoker: I have to admit that this is not really the more readable code :x (but it might be in context)

PR #588, opened 2018-11-28
[fix] Regen nginx conf to be sure it integrates OCSP Stapling

frju365:
## The problem

- When a LE certificate is created, user must run `yunohost service regen-conf` to add OCSP Block in nginx template.  

## Solution

- Regen_conf automatically

## PR Status

Work finished. Not tested.

## How to test

Install yunohost. Create a LE cert

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -304,6 +304,8 @@ def _certificate_install_letsencrypt(auth, domain_list, force=False, no_checks=F
             logger.error(msg)
             operation_logger.error(msg)
 
+      service_regen_conf(names=['nginx')
+
 def certificate_renew(auth, domain_list, force=False, no_checks=False, email=False, staging=False):
     """
     Renew Let's Encrypt certificate for given domains (all by default)


src/yunohost/certificate.py
```diff
@@ -304,6 +304,8 @@ def _certificate_install_letsencrypt(auth, domain_list, force=False, no_checks=F
             logger.error(msg)
             operation_logger.error(msg)
 
+      service_regen_conf(names=['nginx')
```
alexAubin: Uuuh, there's a typo in there :/

frju365: oups :/

src/yunohost/certificate.py
```diff
@@ -803,6 +803,10 @@ def _enable_certificate(domain, new_cert_folder):
     for service in ("postfix", "dovecot", "metronome"):
         _run_service_command("restart", service)
 
+    if os.path.isfile('/etc/yunohost/installed'):
+        # (We don't do this yet if postinstall is not finished yet)
```
Psycojoker:
```suggestion
        # regen nginx conf to be sure it integrates OCSP Stapling
        # (We don't do this yet if postinstall is not finished yet)
```

Integrate reason on why we do that so we don't forget in 2 years.

frju365: done

PR #589, opened 2018-11-28
[fix] Password constrains and checks not behaving properly

alexAubin:
## The problem

`--force-password` not working during postinstall, because there's a call to `tools_adminpw` which also performs a check of the strength. 

## Solution

Pass a new `check_strength` argument to `tools_adminpw` so that we can disable the check if `--force-password` is used.

## PR Status

Yolocomitted, not tested :s 

## How to test

Run a postinstall with a dummy password and `--force-install`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -118,7 +118,7 @@ def tools_ldapinit():
     return auth
 
 
-def tools_adminpw(auth, new_password):
+def tools_adminpw(auth, new_password, check_strength=True):
     """
     Change admin password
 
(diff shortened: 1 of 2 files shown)


alexAubin:
Tested both fixes and that's working

maniackcrudelis:
Is it "--force-password" or "--force-install" ???
During a scripted post install, I had
```
yunohost: error: unrecognized arguments: --force-password
```
Which also means that our unstable CI will wait again...

alexAubin:
Should be `yunohost tools postinstall --force-password` ? Are you sure the version is >= 3.3 ?

maniackcrudelis:
The script get the install script directly.
Anyway, I'm going to delete the complete directory and restart from scratch.

PR #590, opened 2018-11-28
[enh] Clean + harden sshd config using Mozilla recommendation

alexAubin:
## The problem

There are various settings which can be improved in the sshd configuration to increase security.

## Solution

I used the Mozilla recommendation which I found here : https://infosec.mozilla.org/guidelines/openssh

I also cleaned a few things : many weird commented settings which are unlikely to be used, or settings which corresponds to the default setting and therefore unnecessary to state them.

Of course, the best security gain in here would be to disable password authentication... To quote a guy from Stack Overflow : « The authentication and negotiation ciphers are far more important than the symmetric algorithm for the overall security »

## PR Status

Tested in a dev environnement ... but we should be careful and test this in various setups. (e.g. does this break putty ? etc...)

## How to test

This is built on top of https://github.com/YunoHost/yunohost/pull/518 - you'll need to apply the migration, make sure you are using the new conf and see if everything works as expected

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/bin/yunoprompt
+++ b/bin/yunoprompt
@@ -5,7 +5,7 @@ ip=$(hostname --all-ip-address)
 
 # Fetch SSH fingerprints
 i=0
-for key in $(ls /etc/ssh/ssh_host_{rsa,ecdsa,ed25519}_key.pub 2> /dev/null) ; do 
+for key in $(ls /etc/ssh/ssh_host_{ed25519,rsa,ecdsa}_key.pub 2> /dev/null) ; do 
     output=$(ssh-keygen -l -f $key)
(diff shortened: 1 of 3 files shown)


alexAubin:
Note that https://cipherli.st/ and https://github.com/arthepsy/ssh-audit (linked by @kitoy30 and @frju365 ) are unhappy about allowing ECDSA keys (and related key exchange algorithms because those are using weak elliptic curves)

Dunno how that may impact clients compatibilities

PR #591, opened 2018-11-28
[fix] Log dyndns update only if we really update something

alexAubin:
## The problem

`dyndns update` update rate is too damn high, and generates two log file every two minutes even if it doesn't update anything.

```
ls -l /var/log/yunohost/categories/operation/ | wc -l
43915
ls -l /var/log/yunohost/categories/operation/ | grep dyndns | wc -l
43678
```

## Solution

Stop this madness and only start logging if we really do update something.

## PR Status

Yolocomitted but should be fine :< 

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/dyndns.py
+++ b/src/yunohost/dyndns.py
@@ -206,9 +206,6 @@ def dyndns_update(operation_logger, dyn_host="dyndns.yunohost.org", domain=None,
 
             key = keys[0]
 
-    operation_logger.related_to.append(('domain', domain))
-    operation_logger.start()
-
     # This mean that hmac-md5 is used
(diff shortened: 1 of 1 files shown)


Psycojoker:
Shouldn't we also write a migration to trash all old logs? That can be in another PR.

alexAubin:
>Shouldn't we also write a migration to trash all old logs? That can be in another PR.

Yes, but that doesn't prevent stopping the flooding already :s 

If anybody is motivated then please go ahead ... (I won't, I already have too many things to do :/ )

PR #592, opened 2018-11-29
Fix multiple value for getopts

maniackcrudelis:
## The problem

If an option has multiple values, the variable use to store the values will be erased at each loop.

## Solution

Move the creation of the variable before the loop.

## PR Status

Ready to be reviewed.
Tested.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 
I think it can be merged quickly, it's just a simple fix.

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -129,6 +129,8 @@ ynh_handle_getopts_args () {
 							shift_value=$(( shift_value - 1 ))
 						fi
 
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Yep, it was a request from @Josue-T. It's indeed a bug in the code which is supposed to handle that behavior.
I tested the code, it doesn't induce any regression.

PR #595, opened 2018-12-07
[fix] Repair tests

Josue-T:
## The problem

Many tests are broken

## Solution

Fix it

Note that a fix is solved in moulinette in [#185](https://github.com/YunoHost/moulinette/pull/185)

## PR Status

All test should be fixed

## How to test

In the moulinette repos, use the branch `fix_ldap_authenticator`.

run `./ynh-dev test yunohost`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tests/test_changeurl.py
+++ b/src/yunohost/tests/test_changeurl.py
@@ -19,8 +19,10 @@ maindomain = _get_maindomain()
 
 
 def setup_function(function):
-    pass
-
+    # For some reason the nginx reload can take some time to propagate
+    time.sleep(1)
+    global auth
+    auth = init_authenticator(AUTH_IDENTIFIER, AUTH_PARAMETERS)
 
 def teardown_function(function):
     app_remove(auth, "change_url_app")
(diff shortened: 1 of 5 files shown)


src/yunohost/tests/test_changeurl.py
```diff
@@ -19,8 +19,10 @@
 
 
 def setup_function(function):
-    pass
-
+    # For some reason the nginx reload can take some time to propagate
+    time.sleep(1)
+    global auth
+    auth = init_authenticator(AUTH_IDENTIFIER, AUTH_PARAMETERS)
```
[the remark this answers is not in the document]
Josue-T: Should be ok now by https://github.com/YunoHost/moulinette/pull/185

src/yunohost/tests/test_changeurl.py
```diff
@@ -19,10 +19,8 @@
 
 
 def setup_function(function):
-    # For some reason the nginx reload can take some time to propagate
-    time.sleep(1)
```
alexAubin: I still think we might want to keep that `time.sleep(1)`, apparently that was related to nginx starting. What do you think ? :wink:

Josue-T: Well it was a copy past. And it was to try to fix the authenticator. But know I've found an other way to fix that. Before this PR this line wasn't here...

PR #599, opened 2018-12-12
Check migration before yunohost operation

Josue-T:
## The problem

Doing the migration is not mandatory to be able to use Yunohost.

By example we can have a use case of a user which upgrade the debian package but don't do the migration.

If the migration change the structure of data in yunohost, using Yunohost without doing the migration could make some data corruption.

## Solution

Add a mechanism to check if a migration is mandatory before to do the actions.

## PR Status

Should be ready. Need more tests.

## How to test

Create a migration mandatory. If your migration contains this code is mandatory : 
```python
    @property
    def required(self):
        return True
```
Try to do some yunohost action before before the migration is done. Normally you should have this message : 
```
Erreur : To be able do the yunohost operation you need to do the migration
```

After do the migration and check that you can do everything you want.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -50,6 +50,10 @@ _global:
                     uri: ldap://localhost:389
                     base_dn: dc=yunohost,dc=org
         argument_auth: true
+
+    preaction:
+        yunohost.preactions.checkmigrations:
+            - main
     arguments:
         -v:
             full: --version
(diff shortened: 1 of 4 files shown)


Psycojoker:
> Create a migration mandatory. If your migration contains this code is mandatory :
> 
> ```
>     @property
>     def required(self):
>         return True
> ```

FYI you can simply do:

```
class Migration(...):
    required = True
```

And it will do the same thing.

Josue-T:
Should be changed

alexAubin:
Sooo, following discussions about the way to handle this, the migration mechanism should be changed as proposed in another PR instead of locking half of yunohost :s

PR #603, opened 2018-12-19
[fix] DEBUG-level messages not appearing in actions performed via the API

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1251 : this is annoying as fuck that people show logs from failed app install without the DEBUG-level messages, which defeats the purpose of log feature from 3.2 x_x ...

## Solution

Fix the loggers/handlers log level such that it's consistent with what was expected ...

## PR Status

Tested and working

## How to test

Launch an app install that will fail (or any action that shall add DEBUG stuff in the logs) and check the produced logs in /var/log/yunohost/categories/operations with and without this.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/bin/yunohost-api
+++ b/bin/yunohost-api
@@ -13,7 +13,8 @@ DEFAULT_HOST = 'localhost'
 DEFAULT_PORT = 6787
 
 # Level for which loggers will log
-LOGGERS_LEVEL = 'INFO'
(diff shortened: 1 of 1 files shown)


alexAubin:
Yolomerging to have this in 3.4.0 testing

PR #606, opened 2018-12-28
Add SFTP / SSH permissions

Josue-T:
## The problem

Actually the can't restrict the user allowed to access by sftp.

All user are allowed to read the content of the root directory.

The directive `Match User sftpusers` in the sshd config never worked.

## Solution

Update the ssh config.

Add the possibility to define which user is allowed to access by sftp with the permission concept see #585.

## PR Status

This PR depends of #585 .

Can be tested and reviewed.

## How to test

By default this feature is disabled. To enable it type : 
```
yunohost user ssh enable-permission
yunohost service regen-conf -f ssh
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/actionsmap/yunohost.yml
```diff
@@ -288,6 +288,21 @@ user:
                         key:
                             help: The key to be removed
 
+                ### user_ssh_enable_permission()
+                enable-permission:
```
alexAubin: Eh, I'm confused about : 
- the naming : if the purpose is to enable sftp, why call this `enable-permission` and not `enable-sftp` ?
- from what I understand from the PR description, the purpose it to enable sftp per user .. but this does something on a global basis - so is the purpose to enable it per user, or globally ...?

Josue-T: No, by default we keep sftp enabled as it's the case now. We just add the possibility define which user/group is allowed to access by sftp.

alexAubin: But this doesn't take any 'user' argument so in what sense is it per user/group ? To me this look just like a toggle on/off to enable or not SFTP ? (And therefore makes me wonder if it shouldn't just be a setting instead ?)

Josue-T: Well, 

Maybe I forgot to explain this.

When you enable the permission, you have a new permission named `main.sftp`.and after you will manage the permission by the permission command see #585 .

alexAubin:
So idk when's the right time to merge this ... I'm still heavily confused by the naming ... Postponing to 3.8 because there's already too many stuff to finish for 3.7 :/

Psycojoker:
I remember a meeting where we've talked about this kind of modifications, I'm not sure if it was this one but the conclusion was that it would prevent existing non-yunohost user to log by ssh by default and they needed to be added to the correct group (and we concluded that we couldn't do that), is it this PR?

Josue-T:
> I remember a meeting where we've talked about this kind of modifications, I'm not sure if it was this one but the conclusion was that it would prevent existing non-yunohost user to log by ssh by default and they needed to be added to the correct group (and we concluded that we couldn't do that), is it this PR?

Yes, but it's disabled by default.

Psycojoker:
> > I remember a meeting where we've talked about this kind of modifications, I'm not sure if it was this one but the conclusion was that it would prevent existing non-yunohost user to log by ssh by default and they needed to be added to the correct group (and we concluded that we couldn't do that), is it this PR?
> 
> Yes, but it's disabled by default.

Ok, I remember we decided against that because that could break a lot of users existing configuration without any warning.

zamentur:
I did what we discussed during the last meeting, the admin manage SSH through permission view.
I create 2 permission:
- SFTP : chrooted on /home and give only SFTP"(no tunnel)
- SSH : a full ssh access

I don't chroot on /home/USER because the dir need to be writable only by root user, if we did that the user will not be able to put file in his dir from nextcloud or ssh...

It use the current permission but could be quickly adapted with the new system, i think.

data/templates/ssh/sshd_config
```diff
@@ -64,21 +64,33 @@ PrintLastLog yes
 ClientAliveInterval 60
 AcceptEnv LANG LC_*
 
+# Disallow user without ssh or sftp permissions
+AllowGroups ssh.main sftp.main admins root
```
Josue-T, line 68: Now, remember why I implemented a mechanism to enable this feature (and I disabled it by default). The issue with this is that a user manually created by the admin (with `useradd`) will be blocked. And I think by example I in ynh-dev with virtualbox there are some year ago, a user was created by ynh-dev and by this PR this user will be blocked. In the ynh-dev case it's not a really big issue, but we can see some other user in production which have a same setup like this.

Josue-T (commented):
Thanks for your work :+1:

data/templates/ssh/sshd_config
alexAubin: Hmmm imho the fact that a user created with useradd doesn't automagically get ssh access is a good thing ? :thinking:

Josue-T: For me it's OK, and it improve really the security, as for me actually it like a security holes. But we just need to know that there might be some impacted user by this change.

alexAubin: Yup indeed :+1:

zamentur:
I rebased the code with the recent 4.1 (dev branch).
I think this PR is ready.

zamentur (approved):

Josue-T:
OK for me, thanks :+1:

alexAubin:
Remaining points to address:
- are we really removing `yunohost user ssh allow`, or are we deprecating it (or just put a warning pointing to the new permission)
- for chroot/sftp, do we want `/home`, `/home/%u`, or something else? 
- we should have a warning/confirmation when adding people to the ssh perm ? :/

src/yunohost/data_migrations/0020_ssh_sftp_permissions.py
```diff
@@ -0,0 +1,50 @@
...
+        # Add a bash terminal to each users
+        users = ldap.search('ou=users,dc=yunohost,dc=org', filter="(loginShell=*)", attrs=["dn", "uid", "loginShell"])
+        for user in users:
+            if user['loginShell'][0] == '/bin/false':
+                dn=user['dn'][0].replace(',dc=yunohost,dc=org', '')
+                ldap.update(dn, {'loginShell': ['/bin/bash']})
+            else:
+                user_permission_update("ssh.main", add=user["uid"][0], sync_perm=False)
+
+        permission_sync_to_user()
```
alexAubin, line 45: So I'm once more reviewing this and this ends up being catastrophic in the not-so-uncommon case where people manually edited their ssh conf.

In such a case, the SSH restriction on root+admin+members of ssh.main won't be applied, yet we're setting /bin/bash as the shell for all users. Which ends up with every users obtaining ssh access, which is not acceptable.

On the other hand, we can't just force-update the ssh configuration, because that may simply break the SSH workflow etc...

And ideally we don't want to just not change the loginShell, because /bin/false is legacy stuff and we won't have any way to reset the loginShell ...

alexAubin: Follow-up of this discussion,

considering that we addressed a large bunch of the issue of "users can browse /var/www/$app" with the recent security fix, and after discussing with ljf, we propose to handle this situation like this: 
- keep applying what the current migration does
- the sshd_config won't be updated because the user manually edited it
- hence every user end up with an actual SSH access, which is "not cool" but not that bad because they cant browse /var/www/ or other critical locations
- we shall add a check + error message in the diagnosis if the sshd_config contains no `AllowGroups` or `AllowUsers` instructions to incentivize the admin to fix its config
- we can also perform this check and display this error during the migration

PR #608, opened 2018-12-30
[fix] Fix the way change_url updates the domain/path

Josue-T:
## The problem

- In the `change-url` function we use a old function `check_url`.
- The change-url function call the change-url script with variable with empty value (`$YNH_APP_OLD_PATH`, `$YNH_APP_NEW_PATH
`) when the path is the root path. By this we need write a code like that : https://github.com/YunoHost-Apps/wordpress_ynh/blob/master/scripts/change_url#L41-L44 . In the same time we have in the core a function which do that : `_normalize_domain_path`.

## Solution

- Replace the `check_url` by `_get_conflicting_apps`.
- Give the argument about the url as same as in the install.

## PR Status

Tested. Work on my side.

Note that this PR should also fix https://github.com/YunoHost/yunohost/pull/376

## How to test

Test with an app which has a change-url script.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1284,7 +1284,7 @@ def app_checkurl(auth, url, app=None):
 
     logger.error("Packagers /!\\ : 'app checkurl' is deprecated ! Please use the helper 'ynh_webpath_register' instead !")
 
-    from yunohost.domain import domain_list
+    from yunohost.domain import domain_list, _normalize_domain_path
 
     if "https://" == url[:8]:
         url = url[8:]
@@ -1298,8 +1298,7 @@ def app_checkurl(auth, url, app=None):
     path = url[url.index('/'):]
     installed = False
 
-    if path[-1:] != '/':
-        path = path + '/'
+    domain, path = _normalize_domain_path(domain, path)
 
     apps_map = app_map(raw=True)
 
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -445,6 +445,7 @@ def app_change_url(operation_logger, auth, app, domain, path):
 
     """
     from yunohost.hook import hook_exec, hook_callback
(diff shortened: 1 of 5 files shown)


src/yunohost/app.py
```diff
@@ -457,18 +458,13 @@ def app_change_url(operation_logger, auth, app, domain, path):
     old_path = app_setting(app, "path")
 
     # Normalize path and domain format
-    domain = domain.strip().lower()
-
-    old_path = normalize_url_path(old_path)
-    path = normalize_url_path(path)
```
Psycojoker: You should really avoid merging several different modifications into one commit and prefer atomic commits instead, this make review harder.

src/yunohost/app.py
```diff
@@ -1303,8 +1309,7 @@ def app_checkurl(auth, url, app=None):
     path = url[url.index('/'):]
     installed = False
 
-    if path[-1:] != '/':
-        path = path + '/'
+    domain, path = _normalize_domain_path(domain, path)
```
Psycojoker: `_normalize_domain_path` doesn't had a trailing "/", this code replacement is not equivalent.

src/yunohost/app.py
Josue-T: Sorry,

Will try to do better the next time

src/yunohost/app.py
Josue-T: This is to keep the same result than the `register_url`. Actually the `register_url` function use this function. I think we need to be coherent.

src/yunohost/app.py
Psycojoker: Thx <3

PR #609, opened 2019-01-07
Fix several issues with bootprompt

alexAubin:
## The problem

Currently there are several issues with the bootprompt, namely : 
- the IP field being empty (https://github.com/YunoHost/issues/issues/1248)
- some `\n` being displayed as literal `\n` (also mentionned in https://github.com/YunoHost/issues/issues/1248)
- after postinstall being successful, the bootprompt doesn't give back a working terminal hence users end up being stucked and have to reboot to get a working terminal (can't find the corresponding issue for this issue)

## Solution

I reworked and tweaked a few things : 
- fetch the ip after a sleep 5, which gives better changes that the network stack will be up
- add a message similar introducing the postinstall and other things in a similar way to what's done after the install script
- use `chvt 1` to switch to a fresh tty after postinstall is done.

## PR Status

Tested, ready for rewiews

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/bin/yunoprompt
+++ b/bin/yunoprompt
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-# Fetch ips
-ip=$(hostname --all-ip-address)
-
 # Fetch SSH fingerprints
 i=0
 for key in $(ls /etc/ssh/ssh_host_{ed25519,rsa,ecdsa}_key.pub 2> /dev/null) ; do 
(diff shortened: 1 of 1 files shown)


bin/yunoprompt
```diff
@@ -51,17 +54,35 @@ if [[ -f /etc/yunohost/installed ]]
...
+You can perform this step :
+  - from your web browser, by accessing : ${local_ip}
+  - or in this terminal by answering 'yes' to the following question
+
+If this is your first time with YunoHost, it is strongly recommended to take
+time to read the administator documentation and in particular the sections
+'Finalizing your setup' and 'Getting to know YunoHost'. It is available at
+the following URL : https://yunohost.org/admindoc
+===============================================================================
+EOF
```
zamentur: Isn't it too long to be displayed on screen ?

zamentur: We could make a smaller logo on 3 lines to get more lines to explain.

alexAubin: Idk, tested it in a virtualbox with the default "small" screen and it looked fine

zamentur: ok cool

alexAubin:
Well in term of height, yes : on the small default virtualbox screen, the yunohost logo is truncated. But imho that's not so much a big deal...

PR #610, opened 2019-01-07
[fix] propagate --no-checks cert-install option to renew crontab

Psycojoker:
## The problem

Sometime, user needs to uses the option `--no-checks` while installing new LE certificates because of their configuration, but this option is not put inside the crontab file thus the renew will never work and they'll either have to modify the cronfile by hand and do a renew by hand with `--no-checks`.

Should we propagate that to the "renew" command to?

## Solution

Add `--no-checks` inside the cronfile if it's used in the command line.

## PR Status

Done but untested (I don't have the setup for that :().

## How to test

Have a instance where you can test this PR and run "cert-install" with "--no-checks" and check the cronfile.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -286,7 +286,7 @@ def _certificate_install_letsencrypt(auth, domain_list, force=False, no_checks=F
 
             _configure_for_acme_challenge(auth, domain)
             _fetch_and_enable_new_certificate(domain, staging, no_checks=no_checks)
-            _install_cron()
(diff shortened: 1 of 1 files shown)


alexAubin:
Yolomerging so that Bram can test using unstable packages

Psycojoker:
Seems to be working as expected :+1:

PR #611, opened 2019-01-08
[enh] add reload and restart service

bourreP:
## The problem

Unable to use `yunohost service [...]` to restart, reload or reloard-or-restart like [systemctl](https://www.freedesktop.org/software/systemd/man/systemctl.html#reload%20PATTERN%E2%80%A6)

## Solution

Adding this possibility

## PR Status

...

## How to test

`yunohost service restart`
`yunohost service reload`
`yunohost service reload_or_restart`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


locales/en.json
```diff
@@ -220,7 +220,6 @@
     "log_help_to_get_log": "To view the log of the operation '{desc}', use the command 'yunohost log display {name}'",
     "log_link_to_failed_log": "The operation '{desc}' has failed ! To get help, please <a href=\"#/tools/logs/{name}\">provide the full log of this operation by clicking here</a>",
     "log_help_to_get_failed_log": "The operation '{desc}' has failed ! To get help, please share the full log of this operation using the command 'yunohost log display {name} --share'",
-    "log_category_404": "The log category '{category}' does not exist",
```
Psycojoker: Why this change? I don't see how it is related to this PR :o

bourreP: That it's true. I've taken the liberty to remove this line cause it's a duplicate of the line 218. But I can revert that.

Psycojoker: It's more that we shouldn't mix different things in the same PR and do atomic commits, that makes things way easier to handle :)

bourreP: I understand that, you what I revert this line and submit another PR with this line?

bourreP:
Thx for the review!

data/actionsmap/yunohost.yml
```diff
@@ -1178,6 +1178,33 @@ service:
...
+        restart:
+            action_help: Restart one or more services. If the services are not running yet, they will be started.
+            arguments:
+                names:
+                    help: Service name to restart
+                    nargs: "+"
+                    metavar: NAME
+
+        ### service_reload_or_restart()
+        reload_or_restart:
```
alexAubin: Can you elaborate on the intention behind this `reload_or_restart` ? That's not such a big deal I guess, but I'm just scared by this considering that for example for nginx, doing a `restart` instead of `reload` has serious implications as in "this will disconnect every ongoing session".

I don't really understand the use case behind it. Is it just to copy-paste some systemd api ?

bourreP: I'm not an expert in systemd and I copied pasted the api.

But the use case of this command is if you are unsure whether the service has the functionality to reload its configuration, you can use the `reload-or-restart` command. This will reload the configuration in-place if available. Otherwise, it will restart the service.

If you use this on nginx, as nginx can be reloaded, it should not be restarted.

locales/en.json
Psycojoker: No, that's for next time :)

PR #613, opened 2019-01-14
[fix] Set owner of archives folder to 'admin'

alexAubin:
## The problem

c.f. this issue https://github.com/YunoHost/issues/issues/1266

Currently, `admin` cannot access files in the archive folder, `/home/yunohost.backup/archives`. This is not such a big deal for now because I guess most people still log as root. But with 3.4 coming where we really discourage root login and encourage admin login, this prevent people from scp-ing backup archives out of their server.

## Solution

When creating the folder, set `admin` as the owner of the folder. + Add a thing in migration n°8 to change the permission for already existing instances 

## PR Status

Not tested x_x

## How to test

Test on a fresh instance, and on a migrated instance that admin can indeed access those files

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -40,7 +40,7 @@ from moulinette import msignals, m18n
 from yunohost.utils.error import YunohostError
 from moulinette.utils import filesystem
 from moulinette.utils.log import getActionLogger
-from moulinette.utils.filesystem import read_file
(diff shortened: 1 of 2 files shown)


alexAubin:
Tested and works as expected

PR #614, opened 2019-01-15
Enforce --force-confold during apt upgrade

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1232

## Solution

Bram digged into the madness of `python-apt` and its (lack of) documentation to find that we could use some magic stuff to change the configuration supposedly being used during the actual upgrade...

This is not tested yet tho.

## PR Status

Ready ? But not tested ...

## How to test

Launch a `yunohost tools upgrade` and somehow try to detect if the option is correctly being used :/ 

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -563,6 +563,7 @@ def tools_upgrade(operation_logger, auth, ignore_apps=False, ignore_packages=Fal
 
             operation_logger.start()
             try:
+                os.environ["DEBIAN_FRONTEND"] = "noninteractive"
                 # Apply APT changes
                 # TODO: Logs output for the API
                 cache.commit(apt.progress.text.AcquireProgress(),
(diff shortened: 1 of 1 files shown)


src/yunohost/tools.py
```diff
@@ -530,6 +530,10 @@ def tools_upgrade(operation_logger, auth, ignore_apps=False, ignore_packages=Fal
     is_api = True if msettings.get('interface') == 'api' else False
 
     if not ignore_packages:
+
+        apt.apt_pkg.config.init()
```
Psycojoker:
```suggestion
        apt.apt_pkg.init()
```

You need to init apt_pkg, not the config (while this will init the "main config" and the "system" ...)

alexAubin: Indeed :+1:

Psycojoker:
Shouldn't we export this env variable too? https://github.com/YunoHost/install_script/pull/55/files#diff-4e345db22ccb012d388fdee689c9a565R258

alexAubin:
@Psycojoker : you mean `DEBIAN_FRONTEND` ?

alexAubin:
(Proposing to have this as a 3.3 hotfix - though I still need to find a way to test and validate this)

alexAubin:
OK, I made a quick by-hand test calling manually pieces of code from a yunohost shell, while monitoring `ps -ef --forest`. I indeed saw a `dpkg --force-confold` spawning so I'd say it's good

PR #618, opened 2019-01-18
Admin conf improvement

frju365:
## The problem

- We did the configuration with normal conf, but not admin

## Solution

~~

## PR Status

Need test. Not tested yet. Need a fix before merging which need discussion.

## How to test

Install Yunohost, apply conf

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -13,10 +13,8 @@ server {
 
 server {
     # Disabling http2 for now as it's causing weird issues with curl
-    #listen 443 ssl http2 default_server;
-    #listen [::]:443 ssl http2 default_server;
-    listen 443 ssl default_server;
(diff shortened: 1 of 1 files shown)


frju365:
I've some strange thing when I test my server... every server has 2(!) SSL certificate, one is set as Yunohost.org (so, invalid) and the other is valid (from the nginx conf). Perhaps the ssl cert of admin_conf makes a conflict with the valid certificate. Did you have already this issue ?

For instance, don't merge, I wanted seriously to solve this problem.

frju365:
Well ok for merging. Let's talk about this issue in another issue.

alexAubin:
Well, if this PR causes a bug, we shouldn't merge the PR until we understand the bug ...

So is there a bug, or not ?

PR #619, opened 2019-01-19
New way to prevent Robot from watching files on the server + ECDH Curve upgrade (same goal as #618)

frju365:
## The problem

- No precise problem, just a clean way to replace one day User-Agent
- ECDH upgrade (same goal as in the #618)

## Solution

- Add the dedicated nginx header
- Change ECDH (for more secure one)

## PR Status

Work. Tested (but difficult to know if it works, that's the reason, I set to one day the definitive replacement)

## How to test

Install Yunohost

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -68,7 +68,7 @@ server {
         if ($http_user_agent ~ (crawl|Googlebot|Slurp|spider|bingbot|tracker|click|parser|spider|facebookexternalhit) ) {
             return 403;
         }
-
+        add_header  X-Robots-Tag "nofollow, noindex, noarchive, nosnippet";
         # Redirect most of 404 to maindomain.tld/yunohost/sso
         access_by_lua_file /usr/share/ssowat/access.lua;
     }
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -13,10 +13,8 @@ server {
 
 server {
     # Disabling http2 for now as it's causing weird issues with curl
-    #listen 443 ssl http2 default_server;
-    #listen [::]:443 ssl http2 default_server;
-    listen 443 ssl default_server;
-    listen [::]:443 ssl default_server;
+    listen 443 ssl http2 default_server;
+    listen [::]:443 ssl http2 default_server;
 
     ssl_certificate /etc/yunohost/certs/yunohost.org/crt.pem;
     ssl_certificate_key /etc/yunohost/certs/yunohost.org/key.pem;
@@ -24,12 +22,7 @@ server {
     ssl_session_cache shared:SSL:50m;
 
     # As suggested by Mozilla : https://wiki.mozilla.org/Security/Server_Side_TLS and https://en.wikipedia.org/wiki/Curve25519
-    # (this doesn't work on jessie though ...?)
-    # ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
-
-    # As suggested by https://cipherli.st/
-    ssl_ecdh_curve secp384r1;
-
+    ssl_ecdh_curve X25519:sect571r1:secp521r1:secp384r1;
     ssl_prefer_server_ciphers on;
 
     # Ciphers with intermediate compatibility
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -22,7 +22,7 @@ server {
     ssl_session_cache shared:SSL:50m;
 
     # As suggested by Mozilla : https://wiki.mozilla.org/Security/Server_Side_TLS and https://en.wikipedia.org/wiki/Curve25519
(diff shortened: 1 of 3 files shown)


data/templates/nginx/plain/yunohost_admin.conf
```diff
@@ -13,23 +13,16 @@ server {
...
     ssl_session_cache shared:SSL:50m;
 
     # As suggested by Mozilla : https://wiki.mozilla.org/Security/Server_Side_TLS and https://en.wikipedia.org/wiki/Curve25519
-    # (this doesn't work on jessie though ...?)
-    # ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
-
-    # As suggested by https://cipherli.st/
-    ssl_ecdh_curve secp384r1;
-
+    ssl_ecdh_curve X25519:sect571r1:secp521r1:secp384r1:prime256v1;
```
alexAubin: This is not the recommendation from Mozilla ... Where is the source explaining why we should have X25519 and sect571r1

frju365: Anyway : if you see on some source : these 2 curves are not supported by all browser, and the changes may need to be applied in openssl conf). I kept the 2 other ecdh for backward compatibility. For X25519, it's in the wiki page (link is already in comment of the PR) : 

For source : 
- https://github.com/h5bp/server-configs-nginx 
- https://github.com/stylersnico/nginx-secure-config
These are conf which have the goal to be updated following the latest security features.

frju365: ok. So, I close the PR, and never will touch the nginx conf anymore (we have already good conf). I don't even understand why I'm spending hours in doing research for nothing.

Just one point : here I don't understand what is the problem as there is backward compatibility already, and if you see X25519 wikipedia, you would have plenty of source of researcher from several university which will tell you that it has quite good feature (robust, quick...). :confused: 

I trust several other conf too... Being blocked by Mozilla configuration, is a bit strange, as the mozilla conf is not the best at all today.

frju365:
So, I reopen this PR. To sum up : This PR just X-Robot tag, and an ECDH curve (http2 is just to follow the #618 PR)

frju365:
@alexAubin if you want me to do something for this PR(remove ecdh curve), tell me :-)

Psycojoker:
I'm with Aleks on basically all points, I'm really surprised that after all this time we still need to ask (multiple time) for a clear argumentation on those modification and on having "one issue per PR".

The PR template is here to fulfill the role of communication with the rest of the team, when you create a PR you need to make it clear for all of us on why you make those modifications, especially for non trivial things that are internal to the way YunoHost works but on external tools that we aren't expert with. I don't even see a link to the documentation or an explanation of the behavior of this `X-Robots-Tag` on what it is needed for and why it is the good solution.

We are pretty relax on the requirements for opening and submitting a PR compare to a shitload of other projects (you don't imagine how bureaucratic it can be) but at least make some effort on communication, we are a team and a collective, we need to understand your modification and their justification to accept them and if we have to spent a lot of time to do that job on our side because you haven't provided us with this information, don't expect this kind of PR to be merged soon or at all.

frju365:
ok. Let's split this PR, as you want (in my mind there is just one iddue on this PR...)

You ask me to split, but so, in the other PR I made, I will have to split it, explaining why I chose more_set_header and why we have to have http2 in admin conf too... it's a bit useless. Here Robot tag is not a really important thing, which is in my mind a thing we could merge quickly :/ But ok.

Psycojoker:
It's a bit annoying that every time we ask you to do something it's meet with this kind of reaction "it's a bit useless" "ok. So, I close the PR, and never will touch the nginx conf anymore", we aren't against you, we just ask you to work in a way that is easier for the whole team like we all do.

> Here Robot tag is not a really important thing, which is in my mind a thing we could merge quickly :/ But ok.

Yes if it's in its own PR with an explanation and justification, if we don't understand it because you haven't explained it we can't do that.

frju365:
> It's a bit annoying that every time we ask you to do something it's meet with this kind of reaction "it's a bit useless" 

I don't find this nice and completely true. I agree for nginx conf, I have kind of habits (because it's always the same thing -- I open and close PR) and I go really fast unmotivated when somebody ask me to explain something I thought I already explained (which is false apparently in that case).

> we aren't against you, we just ask you to work in a way that is easier for the whole team like we all do.

Well, for example, @alexAubin asks me to explain the ECDH curve, but there is the comment in the file, I can't do much more as every bottom notes on wikipedia could have its place here :-/ and I think wikipedia page is a way to understand better crypto works of researchers (for a first approach).

> Yes if it's in its own PR with an explanation and justification, if we don't understand it because you haven't explained it we can't do that.

Well, it's another syntax of what is still here, and I think it's better (can be complementary to the first and perhaps avoid the condition in nginx file -- list of known bots can change, so the general thing will stay).

And well, I try to help and to be honnest (I'm surely wrong), I find some comment really close-minded about crypto (even if I agree totally that is totally my fault because didn't explain clearly what I wanted to do) and in some way that has made me quite unhappy for the other comments. :-/ Really sorry.

PR #464, opened 2018-05-10
Update spectre meltdown checker

alexAubin:
## The problem

Seems like there was a lot of activity on the spectre/meltdown checker repo so this needs update

## Solution

Ran : `git subtree pull --prefix src/yunohost/vendor/spectre-meltdown-checker https://github.com/speed47/spectre-meltdown-checker/ v0.37 --squash`

## PR Status

Not tested at all, we should at least check that the API hasn't changed :sweat_smile: 

## How to test

Run the diagnostic in CLI, and we should also verify that the check when arriving on the webadmin home page still works

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #620, opened 2019-01-19
Update spectre meltdown checker

alexAubin:
## The problem

Spectre Meltdown checker was last updated around May : https://github.com/YunoHost/yunohost/pull/464

## Solution

Update to most recent version with `git subtree pull --prefix src/yunohost/vendor/spectre-meltdown-checker https://github.com/speed47/spectre-meltdown-checker/ master --squash`

## PR Status

Tested, works as expected

## How to test

Pull the branch and run `yunohost tools diagnosis`, see if there's an output in `security` section

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/vendor/spectre-meltdown-checker/Dockerfile
+++ b/src/yunohost/vendor/spectre-meltdown-checker/Dockerfile
@@ -1,0 +1,7 @@
+FROM alpine:3.7
+
+RUN apk --update --no-cache add kmod binutils grep perl
+
+COPY . /check
+
+ENTRYPOINT ["/check/spectre-meltdown-checker.sh"]
--- a/src/yunohost/vendor/spectre-meltdown-checker/README.md
+++ b/src/yunohost/vendor/spectre-meltdown-checker/README.md
@@ -1,7 +1,15 @@
 Spectre & Meltdown Checker
 ==========================
 
-A shell script to tell if your system is vulnerable against the 3 "speculative execution" CVEs that were made public early 2018.
+A shell script to tell if your system is vulnerable against the several "speculative execution" CVEs that were made public in 2018.
+- CVE-2017-5753 [bounds check bypass] aka 'Spectre Variant 1'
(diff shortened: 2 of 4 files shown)


Psycojoker:
> What about other variants ? In particular, on one of my test machines (a VPS), I noticed that it is vulnerable to newer variants, though it's not clear how I'm supposed to fix them (maybe they can't be fixed for now)

It's a hard question: those other variants aren't fixed everywhere according to debian CVE pages. So, what should we do? Options that I see:

1. don't handle this situation, it's been so long since that
1. show a warning if possible
1. show a warning only if it's possible to fix it? That would require quite moar code that needs to evolve with debian data (I know this website allow to download a json with all its data)

For me it's either 1 or 3 and probably 1 in the meantime, 2 wouldn't be god as a UX stand point because informing users that they are vulnerable with not way to fix that would sucks I think (but ... they are still vulnerable and I don't like that x( but that's on debian/hardware builder side I think :/) and only generate problems that can't be solved and support for us.

That's not really a trivial situation :/

alexAubin:
I dunno much what to do either :/ 

An interesting anecdote is a user who recently checked this and the script reported that the instance was vulnerable to variant 3. After contacting the VPS provider (Gandi), they told that "everything has been done to mitigate the issue and this is probably the tool misdiagnosis this" (paraphrasing what they wrote but this is pretty close).

If they are right, then that is disappointing that they are not contacting the script author in order to improve the accuracy of the diagnosis ... But that basically adds more questions to what should be done about those diagnosis results.

Naively I would just test all variants and inform the user about vulnerabilities found, but just make it "less scary" somehow, and provide link to human-readable infos about this, idk @.@

alexAubin:
Merging this meanwhile ... I guess we can postpone the discussion :/

Psycojoker:
Yes, we should copy/paste/move that discussion to somewhere else like an issue.

PR #621, opened 2019-01-19
[enh] Generate custom Diffie-Hellman primes

frju365:
## The problem

Weak dh_params == weak security

## Solution

Add dh_params in background

## How to test

Install Yunohost / ynh-dev

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/tools.py
```diff
@@ -322,6 +322,17 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
     else:
         dyndns = False
 
+    cron_job_file = "/etc/cron.hourly/yunohost-generate-dh-params"
```
Psycojoker: The max time taken by this needs to be tested and maybe moved to daily, on super small slow board that might be too long.

src/yunohost/tools.py
```diff
@@ -322,6 +322,17 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
...
+
+    command = "openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
```
Psycojoker: You very likely wants to "nice" this command since it can consume a lot of resources and doesn't have the most priority.

src/yunohost/tools.py
frju365: Yes. At the begining I set it to dayly... but didn't know if it was too long or not. :-/

alexAubin: As discussed in https://github.com/YunoHost/issues/issues/93 , it's ~1min on a RPi2 so I think hourly is fine. But re-testing it would be good.

src/yunohost/tools.py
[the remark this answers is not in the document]
frju365: ok I added : 2> /dev/null to avoid errors... but perhaps redirecting in a file, can be good ? or not ?

frju365:
I add regen-conf at the end of postinstall but I can't be sure it will be generated at the end of it (surely not).

src/yunohost/tools.py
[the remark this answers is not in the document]
Psycojoker: That would be indeed way better, something like `/var/log/yunohost/dhparam_generation.log` I guess (don't have strong opinion on that)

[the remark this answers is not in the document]
frju365: done

data/hooks/conf_regen/15-nginx
```diff
@@ -49,7 +49,14 @@ do_pre_regen() {
...
                             | tr -d '"')
 
+    if [[ -e "/etc/ssl/private/dh2048.pem" ]]; then
+      export dh_params=true
+    else
+      export dh_params=false
+    fi                  
+
     ynh_render_template "server.tpl.conf" "${nginx_conf_dir}/${domain}.conf"
+    ynh_render_template "server.tpl.conf" "${nginx_conf_dir}/yunohost-admin.conf"
```
alexAubin: This probably does not do what you think it does :/

src/yunohost/tools.py
```diff
@@ -322,6 +322,17 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
     else:
         dyndns = False
 
+    cron_job_file = "/etc/cron.hourly/yunohost-generate-dh-params"
+
+    command = "nice -n 19 openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam 2> /var/log/yunohost/dhparam_generation.log && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
```
alexAubin: Note that a migration should be implemented so that it also gets applied to existing instances

data/hooks/conf_regen/15-nginx
frju365: indeed, fixed.

src/yunohost/tools.py
Josue-T: You also probably need to change the owner for this file like : `root:ssl-cert`

frju365: mh ? On my instance it's root:root, and it works. If you think it's better root:ssl-cert... but it works like that (perhaps I miss something).

alexAubin: That essentially depends of wether or not we will need some other service to use this DH param. For instance, SSL certificates have root:ssl-cert as permission because metronome needs access to it while not having root privileges... I don't know if dovecot or postfix or metronome or whatever service needs DH params but having ssl-cert as group for this file provides more flexibility I guess

Josue-T: I say this because I'll need this for synapse :smiley:

frju365: oh ok. Well, I was only thinking about nginx :-) For other use, ok ;-) 
Well I don't think it will be use for Dovecot or postifx...but I agree with you

frju365: normalement corrigé.

alexAubin: Note that the original comment is still valid : a migration should be implemented so that it also gets applied to existing instances

frju365: yes, I know. I just didn't have time to do it :/ Let's go to try.

alexAubin: That's okay :wink: Just be aware of it :stuck_out_tongue:

frju365: Normally, quite done :/ Not tested as always (I'm at work for now).

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,36 @@
...
+dhparams_file = "/etc/ssl/private/dh2048.pem"
+
+class MyMigration(Migration):
+    "This migration will add dh_params line and generate it in installed instance"
+
+    def migrate(self):
+
+        try:
+            file_open = open(dhparams_file)
+            service_regen_conf(['nginx'], force=True)
```
alexAubin: I really don't think you want to do this :wink: It's only relevant to regen nginx *after* the DH generation is done, so it should happen in the cron job itself.

Also you should absolutely not use `--force` as it's gonna override manual changes, which may break some critical stuff for people who customized stuff.

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,36 @@
...
+            
+        except:
+            with open(cron_job_file, "w") as f:
+                        f.write("#!/bin/bash\n")
+                        f.write(command)
+
+            _set_permissions(cron_job_file, "root", "root", 0o755)
+            service_regen_conf(['nginx'], force=True)
```
alexAubin: (same here)

src/yunohost/tools.py
```diff
@@ -452,6 +463,7 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
         # (by default, i.e. first argument = None, it won't because it's too touchy)
         service_regen_conf(names=["ssh"], force=True)
 
+    service_regen_conf(['nginx'], force=True)
```
alexAubin: You don't need this since it's already done a few lines up (`service_regen_conf(force=True)`). Note that it's fine to use `--force` inside the postinstall because that's the initial configuration, but you should have a really good reason to do it elsewhere :stuck_out_tongue_winking_eye:

data/hooks/conf_regen/15-nginx
```diff
@@ -49,7 +49,14 @@ do_pre_regen() {
...
                             | tr -d '"')
 
+    if [[ -e "/etc/ssl/private/dh2048.pem" ]]; then
+      export dh_params=true
+    else
+      export dh_params=false
+    fi                  
+
     ynh_render_template "server.tpl.conf" "${nginx_conf_dir}/${domain}.conf"
+    ynh_render_template "plain/yunohost_admin.conf" "${nginx_conf_dir}/yunohost-admin.conf"
```
alexAubin: If the template is to be customized, I think it shouldn't be in "plain" anymore

frju365: mh... ok. So in the same directory as server.tpl.conf ?

alexAubin: Yup

src/yunohost/data_migrations/0009_dh_params.py
frju365: Normally, here it's the case where dh_params is already  generated. So I would say that it's useful here, am I wrong ?

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,36 @@
...
+command = "nice -n 19 openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam 2> /var/log/yunohost/dhparam_generation.log && chown root:ssl-cert /etc/ssl/private/dh2048.pem && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
+dhparams_file = "/etc/ssl/private/dh2048.pem"
+
+class MyMigration(Migration):
+    "This migration will add dh_params line and generate it in installed instance"
+
+    def migrate(self):
+
+        try:
+            file_open = open(dhparams_file)
```
alexAubin: Ah ... Now I realize what you are doing here with this try/except ...

Personally I would rather just have a 

```
if os.path.exists(dhparams_file):
    return
```

(and just assume that if the file exist, the admin edited its nginx conf file appropriately already) instead of those "pythonic" try/except where the intention is implicit :/

frju365: You are true. I just didn't know the function :/ (shame on me). Well I tried to find a solution. :D

data/hooks/conf_regen/15-nginx
frju365: done

src/yunohost/data_migrations/0009_dh_params.py
frju365: done

src/yunohost/data_migrations/0009_dh_params.py
frju365: I've added it in command. Here there is not cron... do it's normal. Or, perhaps I didn't understand what you wanted  ?

src/yunohost/data_migrations/0009_dh_params.py
frju365: done.

frju365:
Why is there conflict ?

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,30 @@
...
+command = "nice -n 19 openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam 2> /var/log/yunohost/dhparam_generation.log && chown root:ssl-cert /etc/ssl/private/dh2048.pem && yunohost service regen-conf >> /var/log/yunohost/dhparam_generation.log && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
+dhparams_file = "/etc/ssl/private/dh2048.pem"
+
+class MyMigration(Migration):
+    "This migration will add dh_params line and generate it in installed instance"
+
+    def migrate(self):
+
+        if os.path.exists(dhparams_file):
+            service_regen_conf(['nginx'])
```
alexAubin: I still don't know what you expect to achieve by running this regen-conf :/ If the admin already generated a dhparam himself, either he modified the nginx conf already and that's not useful to run a regenconf here, or he didn't and the regen-conf which happen during the debian package upgrade will take care of it ...

Maybe we just wanna display a warning or info stating that the dhparam file already exists and there's nothing to do and that's it. Or just a return is fine as well I guess.

frju365: mh... yes. Forgot there were a regen-conf during upgrade (I was alone with my migration in my head). ok.

frju365: ok. Done.

Psycojoker:
I'm sorry to always seems to be complaining but could you write commit messages that make more sens and actually tell what they are about plz :x? Right now it's not really that :/ https://github.com/YunoHost/yunohost/pull/621/commits (and it looks like most of them are done in the github ui admin without testing :x)

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,28 @@
...
+
+command = "nice -n 19 openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam 2> /var/log/yunohost/dhparam_generation.log && chown root:ssl-cert /etc/ssl/private/dh2048.pem && yunohost service regen-conf >> /var/log/yunohost/dhparam_generation.log && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
+dhparams_file = "/etc/ssl/private/dh2048.pem"
+
+class MyMigration(Migration):
+    "This migration will add dh_params line and generate it in installed instance"
+
+    def migrate(self):
+
+        if os.path.exists(dhparams_file) is False:
```
Psycojoker:
```suggestion
        if not os.path.exists(dhparams_file):
```

Proper way of writing this kind of condition in python.

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,28 @@
...
+            with open(cron_job_file, "w") as f:
```
Psycojoker: "cron_job_file" isn't declared in this file, have you run this migration :neutral_face: ?

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,28 @@
+import re
+import os
+
+from moulinette.utils.log import getActionLogger
+from moulinette.utils.filesystem import chown
+
+from yunohost.tools import Migration
+from yunohost.service import service_regen_conf
+
+command = "nice -n 19 openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam 2> /var/log/yunohost/dhparam_generation.log && chown root:ssl-cert /etc/ssl/private/dh2048.pem && yunohost service regen-conf >> /var/log/yunohost/dhparam_generation.log && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
```
Psycojoker: Do we really want to run that in hourly and not in daily instead?

Psycojoker: You also probably want to "%s" in the path the file to remove once it has been declared in this file for stronger code.

src/yunohost/data_migrations/0009_dh_params.py
```diff
@@ -0,0 +1,28 @@
...
+dhparams_file = "/etc/ssl/private/dh2048.pem"
+
+class MyMigration(Migration):
+    "This migration will add dh_params line and generate it in installed instance"
+
+    def migrate(self):
+
+        if os.path.exists(dhparams_file) is False:
+            with open(cron_job_file, "w") as f:
+                        f.write("#!/bin/bash\n")
```
Psycojoker: Might also want to `set -eux` in this file for easier debugging.

src/yunohost/data_migrations/0009_dh_params.py
frju365: don't understand you lst comment. 

For hourly, for me it's quite correct. If we set it to daily, the user will have to wait 1 day whereas it's quite long. Moreover we said that hourly was ok.

frju365:
Yeah. Sorry for my commit message. I will improve them.

src/yunohost/data_migrations/0009_dh_params.py
frju365: done.

src/yunohost/data_migrations/0009_dh_params.py
frju365: I didn't understand : "You also probably want to "%s" in the path the file to remove once it has been declared in this file for stronger code.". Could you explain me what is the goal of adding %s ?

And, I wanted to know if I have to set it to daily or hourly as we had a conversation about that and chose hourly. So, I really don't know what to do.

frju365:
Normalement, j'ai corrigé ma PR. Elle semble fonctionné d'après mes tests. 

Le pb réside dans le fait que le fichier est ajouté (comme prévu) au moment du cron, sauf, que si on recherche juste après la migration (conf_regen), le regen_conf échoue. On pourrait mettre un test dans le regen_conf peut-être pour vérifier si le fichier est là, et en cas de regen_conf le générer. Qu'en pensez-vous ?

src/yunohost/tools.py
```diff
@@ -321,6 +322,17 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
...
+    cron_job_file = "/etc/cron.hourly/yunohost-generate-dh-params"
+
+    command = "nice -n 19 openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048 -dsaparam 2> /var/log/yunohost/dhparam_generation.log && chown root:ssl-cert /etc/ssl/private/dh2048.pem && yunohost service regen-conf >> /var/log/yunohost/dhparam_generation.log && rm /etc/cron.hourly/yunohost-generate-dh-params\n"
+
+
+    with open(cron_job_file, "w") as f:
+        f.write("#!/bin/bash\n")
+        f.write(command)
+
+    _set_permissions(cron_job_file, "root", "root", 0o755)
```
alexAubin: Est-ce qu'on pourrait mettre toute cette partie dans une fonction qui serait appeler par la migration plutôt que de dupliquer le code dans la migration ?

alexAubin:
>Le pb réside dans le fait que le fichier est ajouté (comme prévu) au moment du cron, sauf, que si on recherche juste après la migration (conf_regen), le regen_conf échoue. On pourrait mettre un test dans le regen_conf peut-être pour vérifier si le fichier est là, et en cas de regen_conf le générer. Qu'en pensez-vous ?

Uh beh je comprends pas ce que tu veux dire, pour moi c'est déjà ce qui est fait : il y a des "si le fichier existe, l'utiliser" dans la regenconf.

Tu dis que la regen conf échoue, mais pourquoi est-ce qu'elle échoue ?

frju365:
En fait, elle échoue car il considère que dh_params = true. Mais je referai un test, car, oui, le test -e dans le bash devrait faire le travail. :/ Si qqn peut tester aussi de son côté, ça serait super. Normalement la migration fonctionne.

src/yunohost/tools.py
frju365: Fait.

frju365:
I need help. I try but every test failed. I separate the two if... but it doesn't seem to do something. Has anybody an idea ?

alexAubin:
Closing because that's fixed by https://github.com/YunoHost/yunohost/pull/1007 (Mozilla in fact recommends using a predefined DH param instead of a custom one)

PR #622, opened 2019-01-21
Improve protection against indexation from the robots.

frju365:
## The problem

- $http_user_agent ne permet pas bloquer totalement tout, et sûrement ne bloque pas tous les robot.

## Solution

-  Ajouter une directive plus générale permettant de bloquer l'ensemble des contenu
- Code plus propre qu'une fonction nginx
- Évite de lister l'ensemble des User-Agent des robots d'indexation (ce qui est possible si c'est nécessaire)

## PR Status

Tested, but need test (I don't know how to test the effect of the header)

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

Sources (because it's needed) : 
- https://developers.google.com/search/reference/robots_meta_tag
- http://robots-txt.com/x-robots-tag/

### DIFF
--- a/data/templates/nginx/plain/yunohost_admin.conf
+++ b/data/templates/nginx/plain/yunohost_admin.conf
@@ -65,6 +65,10 @@ server {
 
     location /yunohost {
         # Block crawlers bot
+        if ($http_user_agent ~ (crawl|Googlebot|Slurp|spider|bingbot|tracker|click|parser|spider|facebookexternalhit) ) {
+            return 403;
+        }
+        # X-Robots-Tag to precise the rules applied.
         add_header  X-Robots-Tag "nofollow, noindex, noarchive, nosnippet";
         # Redirect most of 404 to maindomain.tld/yunohost/sso
         access_by_lua_file /usr/share/ssowat/access.lua;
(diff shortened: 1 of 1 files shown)


frju365:
I don't know why it's the second test which had problem.  

@alexAubin @Psycojoker : Do I need to add more informations about the feature ?

data/templates/nginx/plain/yunohost_admin.conf
```diff
@@ -65,10 +65,7 @@ server {
 
     location /yunohost {
         # Block crawlers bot
-        if ($http_user_agent ~ (crawl|Googlebot|Slurp|spider|bingbot|tracker|click|parser|spider|facebookexternalhit) ) {
-            return 403;
```
alexAubin: Reading the doc, I guess I'm okay with the X-Robots-Tag. Nevertheless I don't understand why we should remove this line ? Sure it probably does not cover all user agents, but yet : 
- we could improve this list
- I am more confident about this approach that simply tells "fuck off" to the bots as it better covers the existence of the /yunohost subpath in the first place than a simple header.

So imho those are complementary approach.

frju365: mh... ok. True. I can readd this line (I don't remember why I removed it).

frju365: done

PR #624, opened 2019-01-23
[enh] add systemd logs

kay0u:
## The problem

Some apps don't provide log files.

## Solution

I have added the possibility to use systemd log (with journalctl)

## PR Status

I either have to create a migration file to update the file `/etc/yunohost/services.yml` or check if `log_type` is defined, and if not, I'm considering that the log is a file (as old method)

#### Old method:
 - `yunohost service log service_name --log /var/log/service_name/service_name.log`
 - `yunohost service log service_name --log /var/log/service_name/service_name.log /var/log/service_name/service_name2.log`

#### New method:

1. The old one is still compatible:
 - `yunohost service log service_name --log /var/log/service_name/service_name.log`
 - `yunohost service log service_name --log /var/log/service_name/service_name.log /var/log/service_name/service_name2.log`

2. The new:
 - for a file: `yunohost service log service_name --log /var/log/service_name/service_name.log --log_type file`
 - for several files: `yunohost service log service_name --log /var/log/service_name/service_name.log /var/log/service_name/service_name2.log --log_type file`
 - for a systemd log: `yunohost service log service_name --log service_name --log_type systemd`
 - for both: `yunohost service log service_name --log /var/log/service_name/service_name.log service_name --log_type file systemd`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1150,7 +1150,7 @@ service:
                     full: --description
                     help: Description of the service
                 -t:
-                    full: --type_log
+                    full: --log_type
                     help: Type of the log (file or systemd)
                     nargs: "+"
                     choices:
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1149,6 +1149,14 @@ service:
                 -d:
                     full: --description
                     help: Description of the service
+                -t:
+                    full: --type_log
+                    help: Type of the log (file or systemd)
+                    nargs: "+"
+                    choices:
+                        - file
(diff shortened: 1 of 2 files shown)


Psycojoker:
I think you need to add `nargs` to the name argument to allow a list of logs :thinking: (I can't comment directly on this line on the diff tab)

https://github.com/YunoHost/yunohost/blob/bbb004e056fd8de1ebe8de5ac8b1ec0b58d55025/data/actionsmap/yunohost.yml#L365

PR #629, opened 2019-01-27
[fix] Fix home_dir unset in ynh_system_user_create

maniackcrudelis:
## The problem

ynh_system_user_create is failing with the error `home_dir: unbound variable`

## Solution

Set `home_dir` if not used.

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


maniackcrudelis:
Yep, can be, just a small fix

PR #631, opened 2019-01-28
First simple implementation to allow to set custom theming of yunohost.

eauchat:
## The problem

YNH's SSO look is not so straightforwardly customizable.

## Solution

Following on issue YunoHost/issues#1267, I was trying to bring the possibility to add custom themes.
For now it's just a simple CLI interface that allow to choose the currently active theme.
Installing the themes is still to be done copying files manually for now.

## How to test

- Get the changes from pull request YunoHost/SSOwat#116.
- ( Run `yunohost themes list` you should see "clouds", and "black". )
- Run `yunohost themes set clouds`, this will set the current theme to clouds, go check it out visiting the SSO in the browser.

You can create any custom theme by manually creating a directory in `/usr/share/ssowat/portal/assets/themes/` and adding the necessary css files to it (`css/ynhpanel.css` and `css/ynh-style.css`).

## Next steps

- a command to create a blank theme (that create the directory and necessary css files)
- a command to unset custom theme (switch back to default mode)
- create a package manager for pre-made themes that could be easily fetched and installed from any chosen repo. Something similar to what exist for apps.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1274,6 +1274,8 @@ themes:
       set:
         action_help: Set current active theme
         api: POST /themes/current/<name>
+        configuration:
+          authenticate: all
         arguments:
           name:
             help: Theme name
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1373,6 +1373,7 @@ def app_ssowatconf(auth):
     """
     from yunohost.domain import domain_list, _get_maindomain
     from yunohost.user import user_list
+    import settings
 
     main_domain = _get_maindomain()
     domains = domain_list(auth)['domains']
@@ -1442,6 +1443,7 @@ def app_ssowatconf(auth):
             'Email': 'mail'
         },
         'domains': domains,
+        'theme': settings.settings_get("ssowat.theme"),
         'skipped_urls': skipped_urls,
         'unprotected_urls': unprotected_urls,
         'protected_urls': protected_urls,
--- a/src/yunohost/settings.py
+++ b/src/yunohost/settings.py
@@ -39,6 +39,8 @@ DEFAULTS = OrderedDict([
     ("security.password.admin.strength", {"type": "int", "default": 1}),
     ("security.password.user.strength", {"type": "int", "default": 1}),
     ("service.ssh.allow_deprecated_dsa_hostkey", {"type": "bool", "default": False}),
+    # Theming
+    ("ssowat.theme", {"type": "string", "default": "default"}),
 ])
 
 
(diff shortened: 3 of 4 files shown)


eauchat:
I just realized that `/etc/ssowat/conf.json` is automatically refreshed by the function "app_ssowatconf" on many admin actions. It seems that the way I implemented the saving of the theme name is not good then, I guess it should use this function. However I don't really understand where would be good to save it so that app_ssowatconf could fetch it.

eauchat:
Yes, actually I realize that using settings could be much simpler than implementing the new themes interface. The thing is that if the themes purpose is to let install custom themes like a package manager, the settings interface won't be adapted for that.

I think I can understand the concept of decorator, but from very far ;/ I'm really discovering python, so I don't feel competent to implement such behaviour.

I don't know if that's over-complicated, but I can imagine that the themes (get/set/list) interface could trigger `yunohost settings [set|get] ssowat.theme`. And on set, also trigger app_ssowatconf for refreshing "/etc/ssowat/conf.json".
That could be a temporary fix before the decorator thing is implemented (it's just a hack, because if someone decided to use the settings interface to set the theme, they wouldn't understand why it doesn't take effect and that's no good).

eauchat:
I don't know, I think the specific theme CLI interface is cool with the "package manager" idea, but I don't know if anyone also likes that. It's one more option at the top level of yunohost CLI, and if no one is so interested in it, it might not make sense to create that.
It might just be my special attachment to wanting users* to easily customize the world they are living in.

*Admins in this case, but they are users of the yunohost distribution.

eauchat:
Ok, with the last commit, the theme's value persists, it's properly gotten from settings when the ssowat config is refreshed.

Using settings made it simple to save the value of the current theme.
However, at the moment, using `yunohost settings set "ssowat.theme" -v someThemeName` shouldn't be allowed, because it doesn't properly check if the asked theme is available, and it also doesn't refresh ssowat config.

Maybe it would be useful to implement a way to do those checks from `yunohost settings set` but it seems a bit complex no?
Or implement a new parameter like: `("ssowat.theme", {"type": "string", "default": "default", "customSet": "this setting must be set with 'yunohost theme set' interface" })`.
`customSet` if defined would disable the possibility to set a setting from `yunohost settings` interface and log the given error message if one tries to.
But maybe that's also too complicated.

alexAubin:
Rediscussed quickly today : would be nice to rework this PR from scratch using the new conf panel for global settings

eauchat:
Yep I'd love to do it, but it's been a while I don't have a ynh development environment working anymore. And didn't have time to get it back up.
If someone wants to go for it, feel free :)

alexAubin:
Alrighty, superseded by https://github.com/YunoHost/yunohost/pull/1545

PR #633, opened 2019-01-28
[enh] Add ynh_systemd_action helper

maniackcrudelis:
## The problem

I use this helper in all my own apps without any problem.
That's a really useful helper, especially for all our helpers with a systemctl restart or reload. By using this helper, we'll have directly the log when a service fail to start.

## Solution

Add ynh_systemd_action as an official helper.

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


maniackcrudelis:
@YunoHost/apps

data/helpers.d/system
```diff
@@ -53,3 +53,92 @@ ynh_abort_if_errors () {
...
+    if [[ -n "${line_match:-}" ]]
+    then
+        local templog="$(mktemp)"
+	# Following the starting of the app in its log
+	if [ "$log_path" == "systemd" ] ; then
+            # Read the systemd journal
+            journalctl -u $service_name -f --since=-45 > "$templog" &
+	else
+            # Read the specified log file
+            tail -F -n0 "$log_path" > "$templog" &
```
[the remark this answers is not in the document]
maniackcrudelis: Because that's exactly the point, we don't want to find a previous line.
We're going to restart the service [here](https://github.com/YunoHost/yunohost/pull/633/files#diff-43d93913a4448bf0d9e96b5bc480bf5aR103), so first we start to tail the log, then we restart the service.

data/helpers.d/system
```diff
@@ -53,3 +53,92 @@ ynh_abort_if_errors () {
...
+    local timeout=${timeout:-300}
+
+    # Start to read the log
+    if [[ -n "${line_match:-}" ]]
+    then
+        local templog="$(mktemp)"
+	# Following the starting of the app in its log
+	if [ "$log_path" == "systemd" ] ; then
+            # Read the systemd journal
+            journalctl -u $service_name -f --since=-45 > "$templog" &
```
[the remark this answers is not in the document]
maniackcrudelis: Considering my answer for tail, I'm wondering why there's a --since minus 45s here !?
@JimboJoe could you explain why you added that ? https://github.com/YunoHost-Apps/Experimental_helpers/pull/18/commits/97d6d5261c5745f361061beb64dd3a885215e287
Is `--lines=` not working the same way than for tail ?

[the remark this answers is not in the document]
kay0u: For me a -n0 -q (--lines=0 and --quiet) do exactly what we want.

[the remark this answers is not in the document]
maniackcrudelis: Is it `--quiet` which prevent the last previous line to be printed ?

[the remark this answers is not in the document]
maniackcrudelis: Retried right now, `-n0 -q` gives me immediately the last line of the log, even if nothing's happening in this log.
So I rather prefer `--since=0` which prints only following lines.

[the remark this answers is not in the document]
kay0u: Weird for `-n0` but ok, anyway, --since=0 is working

data/helpers.d/backend
```diff
@@ -158,10 +158,10 @@ ynh_remove_systemd_config () {
 
 	local finalsystemdconf="/etc/systemd/system/$service.service"
 	if [ -e "$finalsystemdconf" ]; then
-		sudo systemctl stop $service
-		sudo systemctl disable $service
+		ynh_systemd_action --service_name=$service_name --action=stop
+		systemctl disable $service_name
```
alexAubin: Eh there is no `$service_name` ? Shouldn't it be `$service` ?

maniackcrudelis: Indeed

maniackcrudelis:
Bump

@YunoHost/apps

Josue-T:
Well, I discovered some issues with this helper.

So I'll push some fix when I's tested

Josue-T:
I've pushed some small changes.

For me it's OK now.

maniackcrudelis:
What were the issues @Josue-T ?
Instead of 'fixed_string', did you try to simply escape characters !? Otherwise, we'll need that in almost every helpers !
And why adding 'systemd_log' ? That was already covered by $log_path at 'systemd'.

Josue-T:
For the fixed_string, I think It make really more easy to just add this option like it is. Probably escaping the caractere should also work.

And for the option "systemd" it didn't work because if the service fail to start, we get an error like `can't find the file "systemd"`.

maniackcrudelis:
I'd rather prefer to keep the usage of escaping characters...

I had seen also that error about systemd, but quickly and had never investigated it.
So your idea would be either to use --log_path or --systemd_log ? If so, it's clearly not going to work since log_path has a default value...

maniackcrudelis:
Huh, so what you revert everything ?
Do you think it would be possible to discuss before commiting when you find something ? Instead of just pushing commits without explaining before.

data/helpers.d/system
```diff
@@ -55,3 +55,92 @@ ynh_abort_if_errors () {
...
+            # Read the specified log file
+            tail -F -n0 "$log_path" > "$templog" &
+        fi
+        # Get the PID of the tail command
+        local pid_tail=$!
+    fi
+
+    echo "${action^} the service $service_name" >&2
+    systemctl $action $service_name \
+        || ( journalctl --lines=$length -u $service_name >&2 \
```
kay0u: If the service fail, `journalctl --lines=$length -u $service_name` is called, and this is a blocking command. I suggest to add a `--no-pager`

data/helpers.d/system
```diff
@@ -55,3 +55,92 @@ ynh_abort_if_errors () {
...
+                break
+            fi
+            echo -n "." >&2
+            sleep 1
+        done
+        if [ $i -eq $timeout ]
+        then
+            echo "The service $service_name didn't fully started before the timeout." >&2
+            echo "Please find here an extract of the end of the log of the service $service_name:"
+            journalctl --lines=$length -u $service_name >&2
```
kay0u: Same

data/helpers.d/system
maniackcrudelis: You're probably right, I read "The output is paged through less by default" in the man.
That part was added for syslog, but never used it. So never noticed that.

alexAubin:
Bump ? any love for this cool PR ;P ?

data/helpers.d/system
```diff
@@ -56,6 +56,99 @@ ynh_get_debian_release () {
...
+
+# Clean temporary process and file used by ynh_check_starting
+# (usually used in ynh_clean_setup scripts)
+#
+# usage: ynh_clean_check_starting
+ynh_clean_check_starting () {
+	# Stop the execution of tail.
+	kill -s 15 $pid_tail 2>&1
+	ynh_secure_remove "$templog" 2>&1
+
```
alexAubin: Eh it looks like there's a missing } in here

maniackcrudelis: Indeed

data/helpers.d/system
```diff
@@ -56,6 +56,100 @@ ynh_get_debian_release () {
...
+        else
+            # Read the specified log file
+            tail -F -n0 "$log_path" > "$templog" &
+            # Get the PID of the tail command
+            local pid_tail=$!
+        fi
+    fi
+
+    echo "${action^} the service $service_name" >&2
+    systemctl $action $service_name \
```
kay0u: I'm wondering, why not using `yunohost service $action $service_name instead`? To deal with the lock?

data/helpers.d/system
```diff
@@ -56,6 +56,100 @@ ynh_get_debian_release () {
...
+            local pid_tail=$!
+        else
+            # Read the specified log file
+            tail -F -n0 "$log_path" > "$templog" &
+            # Get the PID of the tail command
+            local pid_tail=$!
+        fi
+    fi
+
+    echo "${action^} the service $service_name" >&2
```
kay0u: Maybe `ynh_print_info`?

data/helpers.d/system
```diff
@@ -56,6 +56,100 @@ ynh_get_debian_release () {
...
+        local i=0
+        for i in $(seq 1 $timeout)
+        do
+            # Read the log until the sentence is found, that means the app finished to start. Or run until the timeout
+            if grep --quiet "$line_match" "$templog"
+            then
+                echo "The service $service_name has correctly started." >&2
+                break
+            fi
+            echo -n "." >&2
```
kay0u: I guess `ynh_print_warn` cant work with this argument?

data/helpers.d/system
```diff
@@ -56,6 +56,100 @@ ynh_get_debian_release () {
...
+    # Start the timeout and try to find line_match
+    if [[ -n "${line_match:-}" ]]
+    then
+        local i=0
+        for i in $(seq 1 $timeout)
+        do
+            # Read the log until the sentence is found, that means the app finished to start. Or run until the timeout
+            if grep --quiet "$line_match" "$templog"
+            then
+                echo "The service $service_name has correctly started." >&2
```
kay0u: Mayber `ynh_print_warn`

data/helpers.d/system
maniackcrudelis: Which lock are you talking about ?

data/helpers.d/system
maniackcrudelis: Sure you're right !

data/helpers.d/system
maniackcrudelis: That's not an error, I would rather prefer a `ynh_print_info`

data/helpers.d/system
maniackcrudelis: Indeed it can't.
I wonder though if we should keep that echo.

data/helpers.d/system
kay0u: I'm really sorry, I have to.

/!\ WARNING, PYTHON BELOW /!\

I'm talking about this: https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/service.py#L685

data/helpers.d/system
kay0u: As you wish, I found this small log useful to know if the script has crash or not, but I can understand if you want to remove it

data/helpers.d/system
maniackcrudelis: My eyes are bleeding...

I think this lock is only for YunoHost services.

data/helpers.d/system
kay0u: I'm saying that because you use `>&2`

data/helpers.d/system
maniackcrudelis: Well if you had at least once found it useful. Let's keep it.

data/helpers.d/system
maniackcrudelis: That's because otherwise you're not going to see it, except with --debug

data/helpers.d/system
kay0u: I know that hotspot_ynh from labriqueinternet use a lock.

data/helpers.d/system
kay0u: True :-)

data/helpers.d/system
```diff
@@ -56,6 +56,100 @@ ynh_get_debian_release () {
...
+                break
+            fi
+            echo -n "." >&2
+            sleep 1
+        done
+        if [ $i -eq $timeout ]
+        then
+            echo "The service $service_name didn't fully started before the timeout." >&2
```
kay0u: I think, this is a warning so: `ynh_print_warn`

maniackcrudelis: Here indeed at least a `ynh_print_warn`

data/helpers.d/system
kay0u: Gitlab really takes a long time to start -_-'

data/helpers.d/system
kay0u: I'm not really convinced when I'm asking that, so you can keep it like that!

alexAubin: Hmm, let's forget about this --need_lock for now, it's indeed basically only for vpnclient / hotspot and is related to the use of `systemctl` with service themselves triggering `yunohost` commands ...

Will see in the future if we really need to adapt this helper. But there are other strategies to investigate like getting rid of using `yunohost` in the service.

kay0u:
LGTM

maniackcrudelis:
While using this helper for searx. I noticed some annoying warnings.

- `tail -F` print a warning because the file doesn't exist yet and then a new one when the file is created.
- To see the progression is useless when the service is quick to start. It just shows one or two dots.

alexAubin:
Soooo are we good with this ? (@maniackcrudelis : I assume the thing you mentionned were fixed by the last 2 commits)

maniackcrudelis:
Yep everything's good for me

PR #634, opened 2019-01-28
Add ynh_script_progression

maniackcrudelis:
## The problem

I know that this helper is brand new, 2 days only, but there's no real danger with it and I really would like to see it widely used by all apps. Especially official ones.

## Solution

Add the wonderful new helper ynh_script_progression :D

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/print
+++ b/data/helpers.d/print
@@ -124,3 +124,81 @@ ynh_print_ON () {
 	# Print an echo only for the log, to be able to know that ynh_print_ON has been called.
 	echo ynh_print_ON > /dev/null
 }
+
+# Print a message as INFO and show progression during an app script
+#
+# usage: ynh_script_progression --message=message [--weight=weight] [--time]
+# | arg: -m, --message= - The text to print
+# | arg: -w, --weight=  - The weight for this progression. This value is 1 by default. Use a bigger value for a longer part of the script.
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
@YunoHost/apps

alexAubin:
About the usage of this script : please note that it might not be relevant for backup and restore scripts !

For instance, the way backup happens is to run the backup script only to get the list of files to backup (modulo the mysql dump) but not to *copy* them yet. So the bulk of the execution time is expected to be *after* running all those script. In that case, what the use might see is : 
- launch the backup thing
- all apps write progress bar very quickly
- everything seems to be stucked (because no feedback)
- user might wonder if backup is still going on or system is blocked somehow...

And a similar thing (in reverse) happening when restoring

So I would not recommend using this helper during backup (though I havent tried to confirm the behavior). If there's a progress bar to be implemented for this, it should be implemented in the core during the actual copy I think...

maniackcrudelis:
There's actually other steps in backup and restore, especially restore script with reinstallation of dependencies, restart of services and other things like that which take time.
However, we should indeed finish the backup script with a message to inform that it's not completely done.

PR #635, opened 2019-01-29
Warn the user that ynh_secure_remove should be used with only one arg…

alexAubin:
## The problem

This is a silly detail, but the mistake of calling `ynh_secure_remove` with more than one arg comes quickly ... Either because you transform a `rm -rf /foo/*` to `ynh_secure_remove /foo/*`, or `rm foo bar` into `ynh_secure_remove foo bar`

## Solution

Add a small warning for the packager to tell that other arguments given will be ignored

## PR Status

Not tested because laziness @.@

## How to test

Try to run `ynh_secure_remove` with multiple args

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -348,6 +348,11 @@ ynh_secure_remove () {
 	/var/www \
 	/home/yunohost.app"
 
+    if [[ -n "$2" ]]
+    then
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
> ```Warning: yunohost.hook <lambda> - [1003.1] /usr/share/yunohost/helpers.d/filesystem: line 351: $2: unbound variable```

PR #638, opened 2019-01-31
[enh] Check if dpkg is not broken when calling ynh_wait_dpkg_free

maniackcrudelis:
## The problem

Fix https://github.com/YunoHost/issues/issues/1284

## Solution

Use the way apt check that dpkg is available to know if `dpkg --configure -a` is needed.

## PR Status

Ready to be reviewed.

## How to test

As Aleks told me, update metronome or any other package that will ask you to replace a config file. And kill the process without answering the question.
Then, remove the lock file and you'll have a broken dpkg

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -15,6 +15,21 @@ ynh_wait_dpkg_free() {
             # Sleep an exponential time at each round
             sleep $(( try * try ))
         else
+            # Check if dpkg hasn't been interrupted and is fully available.
+            # See this for more information: https://sources.debian.org/src/apt/1.4.9/apt-pkg/deb/debsystem.cc/#L141-L174
+            local dpkg_dir="/var/lib/dpkg/updates/"
+
+            # For each file in $dpkg_dir
+            while read dpkg_file <&9
+            do
+                # Check if the name of this file contains only numbers.
+                if echo "$dpkg_file" | grep -Pq "^[[:digit:]]*$"
+                then
+                    # If so, that a remaining of dpkg.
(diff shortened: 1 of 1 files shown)


Psycojoker:
Shouldn't we run `sudo dpkg --configure -a` instead of breaking the install? I don't know if it's a good idea.

I have a concern because people that only uses the admin interface won't be able to do that by themselves but that's really out of the scope of this PR.

alexAubin:
It won't solve the issue. For instance, in the case of the metronome upgrade, it excepts to interact with the user so that it can choose what version of the config to choose ... That's one example but there might be other situation (e.g. not being able to install a package because it cant override a file, c.f. https://github.com/labriqueinternet/hotspot_ynh/issues/30 )

So there isn't really any obvious general automatic fix :/ But at least diagnosing what the actual issue is shall make support faster.

(For the anecdote : I reported this after a friend was not able to install an app. Then he ended up trying to install all conf files by hand because he thought the app was broken. Turns out it was just the metronome thing which broke apt :/)

Psycojoker:
Meh, ok x(

I don't know how we are going to make this solvable on the admin one day :/

alexAubin:
Yea :s I guess the best lead so far is to avoid breaking apt in the first place :sweat_smile:

Psycojoker:
If we want to explore real solutions we have pexpect for that and probably some dpkg python integration or ... maybe something in that horrible python-apt :sweat:

maniackcrudelis:
Bump

@YunoHost/apps

maniackcrudelis:
Let's merge in 3 days if it's ok

PR #639, opened 2019-02-03
[enh] Add IPv6 nameserver to resolv.dnsmasq.conf from diyisp.org

Taekiro:
## The problem

No DNS query to IPv6 servers available.

## Solution

Add IPv6 nameserver from http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver

## PR Status

Manually changed list tested on a server.

## How to test

Update the list, regen dnsmasq conf, check if domain name resolution

## Validation

- [x] Principle agreement 0/2 : 
- [x] Quick review 0/1 : 
- [x] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/dnsmasq/plain/dnsmasq.conf
+++ b/data/templates/dnsmasq/plain/dnsmasq.conf
@@ -2,6 +2,5 @@ domain-needed
 expand-hosts
 
 listen-address=127.0.0.1
-listen-address=::1
 resolv-file=/etc/resolv.dnsmasq.conf
 cache-size=256
--- a/data/templates/dnsmasq/plain/dnsmasq.conf
+++ b/data/templates/dnsmasq/plain/dnsmasq.conf
@@ -2,5 +2,6 @@ domain-needed
 expand-hosts
 
 listen-address=127.0.0.1
+listen-address=::1
 resolv-file=/etc/resolv.dnsmasq.conf
 cache-size=256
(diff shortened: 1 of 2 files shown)


alexAubin:
Soooo, that sounds nice. But we should check how things are gonna behave on IPv4-only instances ? :thinking:

Taekiro:
I just tested this on my server by disabling IPv6 and checking the destinations using dnstop, it seems that dnsmasq automatically switch to use what is available.
(I didn't try to disable IPv4 and have only IPv6 since there seems to be no simple command equivalent for `sysctl -w net.ipv6.conf.all.disable_ipv6=1`)

Taekiro:
I'm not sure that there need to be a change on dnsmasq.conf to allow query on ::1, or how to it will work if IPv6 is disable (that's why I reverted the change).

alexAubin:
Tested on my machine which is in ipv4 only : so far it seems to work transparently :+1:

alexAubin:
Well now I'm asking myself existential question such as : 
- if /etc/resolv.conf points to 127.0.0.1 to proxy the request to dnsmasq, will dnsmasq maybe do the actual request in ipv6 if it can ?
- otherwise maybe we need to make /etc/resolv.conf point to ::1 ... in which case we also do need to modify dnsmasq conf to listen to ::1 

Hopefully answer to item 1 is yes ... One could check this by looking at the outgoing traffic with tcpdump or something like this ...

alexAubin:
Yolomergin'

PR #640, opened 2019-02-04
[enh] use setting security_ciphers_compatibility to define security configu…

rds13:
This PR is proposed following discussion in #581 

## The problem

Try to support modern configuration for admin willing to.

## Solution

Define settings 'security.ciphers.compatibility'

## PR Status

Tested

### Modern settings for web

- Change default security policy
``
yunohost settings set security.ciphers.compatibility -v modern
``

- Add a domain to have yunohost build nginx conf file
``
yunohost domain add yunohostdev1.host 
``

- Ensure you already have an application or install one
``
yunohost app install opensondage
``

- Check configuration is applied on nginx
``
cat /etc/nginx/conf.d/yunohostdev1.host.conf
``
Ensure only TLS v1.2 is allowed.

### Intermediate settings for web
- Reverse security policy to default
``
yunohost settings set security.ciphers.compatibility -v intermediate
``

- Check yunohost have rebuild nginx conf file

- Check configuration is applied on both nginx configuration
``
cat /etc/nginx/conf.d/yunohostdev1.host.conf
``
Ensure TLSv1 TLSv1.1 TLSv1.2 are supported.

### Intermediate settings for ssh
- Set intermediate policy for ssh security
``
yunohost settings set service.ssh.ciphers.compatibility -v intermediate
``

- Check SSH configuration have been regenerated

- Check new SSH configuration
``
cat /etc/sshd/sshd_config
``

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


locales/fr.json
```diff
@@ -308,6 +308,7 @@
     "global_settings_setting_example_int": "Exemple d’option de type entier",
     "global_settings_setting_example_string": "Exemple d’option de type chaîne",
     "global_settings_setting_example_enum": "Exemple d’option de type énumération",
+    "global_settings_setting_security_ciphers_compatibility": "Stratégie de compatibilité des ciphers",
```
alexAubin: Sorry, that's not obvious but you should only edit en.json and not fr.json (or any other language) because supposedly weblate (our translation tool) might complain and stuff :s

src/yunohost/settings.py
```diff
@@ -39,6 +39,7 @@
     ("security.password.admin.strength", {"type": "int", "default": 1}),
     ("security.password.user.strength", {"type": "int", "default": 1}),
     ("service.ssh.allow_deprecated_dsa_hostkey", {"type": "bool", "default": False}),
+    ("security.ciphers.compatibility", {"type": "string", "default": "intermediate"}),
```
alexAubin: I think we should be using a choice here ? e.g. `"choices": ["intermediate", "modern"]` (c.f. "example.enum" a few line up)

data/templates/ssh/sshd_config
```diff
@@ -15,10 +15,17 @@ HostKey {{ key }}{% endfor %}
 # https://infosec.mozilla.org/guidelines/openssh
 # ##############################################
 
-# Keys, ciphers and MACS
-KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp521,ecdh-sha2-nistp384,ecdh-sha2-nistp256,diffie-hellman-group-exchange-sha256
-Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes192-ctr,aes128-ctr
-MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
+{%- if security_ciphers_compatibility == "intermediate" -%}
```
alexAubin: Hm thinking about this I really think we should split this setting into "nginx" and "ssh" config ... I know we technical people like consistency, but consider the following : 
- a yunohost admin might have users with not-so-recent devices and wants to keep intermediate compatibility
- but wants to harden the SSH conf (only used for administration in his/her use-case)

data/templates/nginx/server.tpl.conf
```diff
@@ -29,6 +29,14 @@ server {
     ssl_session_timeout 5m;
     ssl_session_cache shared:SSL:50m;
 
+    {% if security_ciphers_compatibility == "modern" %}
+    # Ciphers with modern compatibility
+    # https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.6.2&openssl=1.0.1t&hsts=yes&profile=modern
+    # Uncomment the following to use modern ciphers, but remove compatibility with some old clients (android < 5.0, Internet Explorer < 10, ...)
```
alexAubin: Small detail but I guess we can remove this comment ? Or maybe we should put a comment explaining the use of the yunohost setting ?

data/hooks/conf_regen/15-nginx
```diff
@@ -44,6 +49,8 @@ do_pre_regen() {
     mkdir -p "$mail_autoconfig_dir"
 
     # NGINX server configuration
+    echo "RDS Using security_ciphers_compatibility=$security_ciphers_compatibility"
```
alexAubin: Debug message to be removed ?

rds13: Of course !

locales/fr.json
rds13: No problem. It's fixed.

data/templates/ssh/sshd_config
rds13: Ok. Shall I make 2 differents PR ?

data/templates/nginx/server.tpl.conf
rds13: This is still informative and anybody can check against mozilla tool to see it matches. What do you think ?

data/templates/ssh/sshd_config
alexAubin: Nah I think that's the same "feature" :wink: It's just a matter of having two different settings instead of one. (But idk, that's a decision decision, maybe other people have different views on this..)

data/templates/nginx/server.tpl.conf
alexAubin: Absolutely :wink:  I was more referring about the "Uncomment the following [...]" which is no more relevant since it's not meant to be commented anymore but controlled via the setting

src/yunohost/settings.py
rds13: Definitely !

data/templates/ssh/sshd_config
rds13: Done.

data/templates/nginx/server.tpl.conf
rds13: I updated it due to new context.

src/yunohost/settings.py
```diff
@@ -39,6 +39,10 @@
     ("security.password.admin.strength", {"type": "int", "default": 1}),
     ("security.password.user.strength", {"type": "int", "default": 1}),
     ("service.ssh.allow_deprecated_dsa_hostkey", {"type": "bool", "default": False}),
+    ("service.ssh.ciphers.compatibility", {"type": "enum", "default": "modern",
+        "choices": ["intermediate", "modern"]}),
+    ("security.ciphers.compatibility", {"type": "enum", "default": "intermediate",
+        "choices": ["intermediate", "modern"]}),
```
alexAubin: Oh and then the question which goes beyond this PR is : what do we want to do when those settings are modified ... So e.g. when calling `yunohost setting set security.ciphers.compatibility modern` (whatever the exact syntax), ssh or nginx conf shall be regenerated ... we can't expect the user to guess that.

So, when this setting system was first designed, we though of using a decorator for this to create something like : 

```
@when_changing("some_setting_name")
def some_function_name():
   [...]
```

so that for this case that would be something like : 

```
@when_changing("security.ciphers.compatibility")
def some_function_name():
    service_regen_conf("nginx")
```

So this `@when_changing` decorator needs to be defined somewhere so that after using `setting_set`, the functions decorated in that way are called :wink: 

Dunno if you are at ease with decorators and understand this whole thing, otherwise I can work on it at some point.

rds13: I understand the mecanism but I'm not confortable enough with Yunohost stack to handle that at the moment.
If it's a new mecanism to design I bet you or someone with more experience on this subject is to be needed.

alexAubin: So I implemented the mechanism in https://github.com/YunoHost/yunohost/pull/654

If the design does not change, it should be as simple as adding this kind of lines : 

```python
@post_change_hook("service.ssh.ciphers.compatibility")
def foo(old_value, new_value):
    # (We don't really care about the old and new value for that use case)
    service_regen_conf(["ssh"])
```

rds13: As soon as I am able to test this mecanism in the context of this PR, I will report here. Thx @alexAubin for your help.

rds13: I have updated this PR to use post_change_hook. It works like a charm ! 👍

rds13:
Shall we need to document this feature in some documentation ?

alexAubin:
Hmyes I think if you want, you can propose to amend the page about security : https://yunohost.org/#/security

src/yunohost/settings.py
```diff
@@ -278,10 +283,12 @@ def trigger_post_change_hook(setting_name, old_value, new_value):
...
-#@post_change_hook("example.int")
-#def myfunc(setting_name, old_value, new_value):
-#    print("In hook")
-#    print(setting_name)
-#    print(old_value)
-#    print(new_value)
+@post_change_hook("service.ssh.ciphers.compatibility")
+def reconfigure_ssh(setting_name, old_value, new_value):
+    if old_value != new_value:
+        service_regen_conf(names=['ssh'], force=True)
```
alexAubin: Note that you should *not* use `force=True` if you don't have a good reason to do it... otherwise that will overwrite manually-edited conf which might have desastrous consequences. If the user manually edited in a conf file, then a message will be displayed accordingly, and it's his/her responsability to handle the file change. (We hope someday™ to have a smarter things such as merging manual changes and yunohost changes with a git-like merge, or user being able to create custom patch, but for now that's not the case)

src/yunohost/settings.py
```diff
@@ -39,6 +40,10 @@
     ("security.password.admin.strength", {"type": "int", "default": 1}),
     ("security.password.user.strength", {"type": "int", "default": 1}),
     ("service.ssh.allow_deprecated_dsa_hostkey", {"type": "bool", "default": False}),
+    ("service.ssh.ciphers.compatibility", {"type": "enum", "default": "modern",
+        "choices": ["intermediate", "modern"]}),
+    ("security.ciphers.compatibility", {"type": "enum", "default": "intermediate",
```
alexAubin: Eeh can we try to call those in a symetric fashion ? I don't know what's best ... from what I understand you wanted to keep `security` because there were already `security.password.*` settings... And then there was also a `service.ssh.*` setting ... (which now that I look at it, should have also been called security because it's related to security...)

Sooo meh, idk, I would call these just 

```
security.compatibility.ssh
security.compatibility.nginx
```

?

rds13: Did you say naming is hard ? 😄 
I think security is a major point so I would name all those settings to begin with security. Then there is a compatibility matrix which affects most of the yunohost services.
May be we need an advice from someone with a biggest picture ?

src/yunohost/settings.py
rds13: It was intended. I fixed it.

src/yunohost/settings.py
rds13: @alexAubin Do you think we should wait for more comments ?

data/templates/ssh/sshd_config
```diff
@@ -15,10 +15,17 @@ HostKey {{ key }}{% endfor %}
 # https://infosec.mozilla.org/guidelines/openssh
 # ##############################################
 
-# Keys, ciphers and MACS
-KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp521,ecdh-sha2-nistp384,ecdh-sha2-nistp256,diffie-hellman-group-exchange-sha256
-Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes192-ctr,aes128-ctr
-MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
+{% if ssh_ciphers_compatibility == "intermediate" %}
+  KexAlgorithms diffie-hellman-group-exchange-sha256
```
alexAubin: Re-reading this I'm wondering if this is really something people are going to use ... Mozilla's comment about it is : 

>This is mainly for use by RHEL6, CentOS6, etc. which run older versions of OpenSSH.

And I don't think that people in our userbase fall in that case...

However, it could be interesting to have something like `modern++` or `hardened` (or whatever name, idk) which increases stuff over the modern configuration, such as : 
- forbidding password logins (so only use asymetric keys)
- disabling root login on local network (assuming the admin is well aware of consequences about that)
- possibly removing ciphers with political speculation / superstition..., c.f. https://github.com/YunoHost/yunohost/pull/668

src/yunohost/settings.py
alexAubin: I implemented what I was thinking about in https://github.com/YunoHost/yunohost/pull/640/commits/fee79820e097ebefb342bd53a16098661359b196

alexAubin:
Will merge at the beggining of next week if there's no other request for change since then

rds13:
Nice @alexAubin.

PR #642, opened 2019-02-06
Fix #635 on ynh_secure_remove

maniackcrudelis:
## The problem

#635 has added a regression by using $2 directly, which fails if there's only one argument.
`/usr/share/yunohost/helpers.d/filesystem: line 351: $2: unbound variable`

## Solution

Use the number of args instead of testing $2
Anyway, by using getopts, we will already have a message like: `Attention : [WARN] Too many arguments ! /my/path will be ignored.`

## PR Status

Tested with my new "personal" ynh-dev ;)

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -392,10 +392,10 @@ ynh_secure_remove () {
 	/var/www \
 	/home/yunohost.app"
 
-    if [[ -n "$2" ]]
-    then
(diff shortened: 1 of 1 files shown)


alexAubin:
Merging as microdecision

PR #643, opened 2019-02-09
[fix] Get rid of annoying 'unable to initialize frontend' messages

alexAubin:
## The problem

In some situation, e.g. installing an app from the webadmin, the install dump some warnings : 

```text
dpkg-preconfigure: unable to re-open stdin:
debconf: falling back to frontend: Teletype
debconf: (This frontend requires a controlling tty.)
debconf: unable to initialize frontend: Readline
debconf: falling back to frontend: Readline
debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
debconf: unable to initialize frontend: Dialog
```

Those are pretty annoying and technical, would be nice to get rid of them ...

## Solution

My guess is that it's related to the use of dpkg inside `ynh_package_install_from_equivs` but I'm not that sure ... So tried to add `DEBIAN_FRONTEND=noninteractive` but not sure that's the issue / fix :wink: Mainly creating this PR to not forget about it later.

## PR Status

Not tested

## How to test

Pull the branch, try to install an app from the webadmin (e.g. Nextcloud). (Don't forget to restart yunohost-api)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/package
+++ b/data/helpers.d/package
@@ -71,7 +71,7 @@ ynh_package_version() {
 # usage: ynh_apt update
 ynh_apt() {
     ynh_wait_dpkg_free
-    DEBIAN_FRONTEND=noninteractive sudo apt-get -y $@
+    DEBIAN_FRONTEND=noninteractive apt-get -y $@
 }
 
 # Update package index files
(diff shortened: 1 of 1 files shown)


alexAubin:
So turns out my intention was wrong and it's much worse than that ... It's about the use of `sudo` in `ynh_apt`, which for some reason cause `DEBIAN_FRONTEND` to not be forwarded to the actual apt / dpkg command -_-

@maniackcrudelis : do you think it's okay to remove this sudo ?

maniackcrudelis:
sudo start another shell dedicated to its usage. In this command line, DEBIAN_FRONTEND is defined before sudo, so sudo doesn't have it.
I know that usually a sub shell heritate the variable from its parent. But not with sudo, which has a specific config regarding its variables.

Anyway, for apps, to not have sudo shouldn't be a problem.

alexAubin:
Flagging this as ready to merge because it should be a small benign change, but feel free to comment...

PR #645, opened 2019-02-09
 [add] ynh_url_join

kay0u:
## The problem

...

## Solution

Add a helper that can join several url together

## PR Status

Tested

## How to test

`ynh_url_join example admin index.php`
`ynh_url_join /example admin /index.php`
`ynh_url_join /`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -296,3 +296,36 @@ ynh_render_template() {
                     jinja2.Template(sys.stdin.read()
                  ).render(os.environ));' < $template_path > $output_path
 }
+
+# Join several urls together
+# Return a concatenate normalized url path
+#
+# example: url_path=$(ynh_url_join $url_path /admin /index.php)
+#          ynh_url_join example admin index.php -> /example/admin/index.php
+#          ynh_url_join /example admin /index.php -> /example/admin/index.php
+#          ynh_url_join / -> /
+#
+# usage: ynh_url_join url1 url2 ...
+ynh_url_join() {
+    if [ "$#" -eq 0 ]; then
+        ynh_die "Illegal number of parameters"
+    fi
+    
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
What's the purpose of this helper @kay0u ?

kay0u:
I needed it in an app, so I make a proposal...

maniackcrudelis:
Yep, so far I was guessing that.
My question was more what the use case of this helper?
Can't see any usage of it

kay0u:
For me, here:

https://github.com/YunoHost-Apps/dotclear2_ynh/blob/testing/scripts/install#L108

and:

https://github.com/YunoHost-Apps/dotclear2_ynh/blob/testing/scripts/install#L127

maniackcrudelis:
But! Why not simply concatenate the path together!?
If it's only because of a double path, remove it. It's would be easier

kay0u:
That's exactly what i've done before, but i was not happy with an `if path_url == "/"`

But, if you think it's not a helper really helpfull, you can close this PR without any problem :-)

maniackcrudelis:
For me it looks like an edge case. Never heard before of an app for which it was really a problem.
Meanwhile, you can add this helper to the [experimental repo](https://github.com/YunoHost-Apps/Experimental_helpers) and use it in dotclear.
While in experimental helpers repo, maybe it will meet its public.

Maybe other from @YunoHost/apps group would have a different opinion ?

JimboJoe:
Well I agree with @maniackcrudelis that you could submit it to experimental helpers first. Although it's ugly, double slashes have usually no impact... :man_shrugging:

PR #646, opened 2019-02-10
[fix] Escape double quote before eval in getopts

maniackcrudelis:
## The problem

Transmission fails on this [line](https://github.com/YunoHost-Apps/transmission_ynh/blob/master/scripts/install#L138) while using getopts.

## Solution

Escape double quotes before the eval to prevent an interpretation.

## PR Status

Ready to be reviewed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -150,6 +150,9 @@ ynh_handle_getopts_args () {
 									# If there's already another value for this option, add a ; before adding the new value
 									eval ${option_var}+="\;"
 								fi
+								# Escape double quote to prevent any interpretation during the eval
+								all_args[$i]="${all_args[$i]//\"/\\\"}"
+
 								eval ${option_var}+=\"${all_args[$i]}\"
 								shift_value=$(( shift_value + 1 ))
 							fi
@@ -187,6 +190,9 @@ ynh_handle_getopts_args () {
 				# Also, remove '=' at the end of the long option
 				# The variable name will be stored in 'option_var'
 				local option_var="${args_array[$option_flag]%=}"
+
+				# Escape double quote to prevent any interpretation during the eval
+				arguments[$i]="${arguments[$i]//\"/\\\"}"
 
 				# Store each value given as argument in the corresponding variable
 				# The values will be stored in the same order than $args_array


JimboJoe:
getops fails on [this line](https://github.com/YunoHost-Apps/discourse_ynh/blob/testing/scripts/upgrade#L78) in discourse app on YunoHost 3.5.0 (full log [here](https://paste.yunohost.org/raw/olozudoqof)). Is there a chance that it could be related to that PR...?

randomstuff:
Quoting double quotes is not enough when the argument is in double quotes (you can inject `$(...)`, you can `\"`, etc.). A more correct solution is to do something like `shell_quote` in python.

maniackcrudelis:
Hum, interesting case...
Clearly what's happening here is that your argument has eventually been interpreted as a variable because of the $ at the beginning. Did you try with a simple escape ?
Quote are interpreted anyway when sent to a function.

If you're comfortable with getopts, try to debug by removing the set +x in getopts, you'll see then what's happening.

JimboJoe:
I tried with
```
 ynh_replace_string "\$proxy_add_x_forwarded_for" "\$http_your_original_ip_header" "/etc/nginx/conf.d/$domain.d/$app.conf"
```
with no more luck...
I'm not comfortable at all with getops :sweat_smile: and agree to @randomstuff general approach because there may be other corner cases ($, ${, etc.)

maniackcrudelis:
Except that it's absolutely not related to how python is passing arguments or whatever to bash...
Here it's about how bash is passing arguments to each functions. And that's not that simple to not interpret special characters.

I think in your case JimboJoe, it's about the eval in getopts, I'll have a look to it.

randomstuff:
> Here it's about how bash is passing arguments to each functions.

Well in normal bash commands it's note super difficile. The problem here is using `eval` which its potential to shell metacharacters interpretation (and shell command injection). `eval` is really quite evil and should be avoided if possible.

If you're restricting to bash, you should be able to do something like:

    declare ${option_var}+="${arguments[$i]}"

maniackcrudelis:
`eval` is evil because it can interpret every bash command and execute it, I know. But there's no reason here to have any injection. And as explained before, I can't see any reason for an admin with a root permission (needed to run those scripts) to try to inject anything since he can simply do whatever he wants.

We're using `eval` here to get the content of the variable stored as a string in `arguments[$i]`, not to get the content of this cell of the array.
```
eval ${option_var}+=\"${arguments[$i]}\"
```
Gives the content of the variable in `${arguments[$i]}`
```
declare ${option_var}+="${arguments[$i]}"
```
Would gives the content of `${arguments[$i]}` as a string.

If ${arguments[$i]}` contains val1, with `eval` we'd have the content of `$val1`, without we'd have `val1` as a simple string.

@JimboJoe I do not reproduce, whether with the stable version of getopts or with the unstable version.
In both case, I got
```
sudo sed --in-place 's@$proxy_add_x_forwarded_for@$http_your_original_ip_header@g' /etc/nginx/conf.d/crudelis-test6.fr.d/discourse.conf
```

randomstuff:
@maniackcrudelis, are you sure?

I'm trying this:

```sh
#!/bin/sh

val1=whetever
arguments=(a b c val1 e f g)
option_var=res
i=3

echo Test 0
echo ${arguments[$i]}
echo

echo Test 1
res=""
eval ${option_var}+=\"${arguments[$i]}\"
echo $res
echo

echo Test 2
res=""
declare ${option_var}+="${arguments[$i]}"
echo $res
echo
```

I get:

<pre>
Test 0
val1

Test 1
val1

Test 2
val1

</pre>

maniackcrudelis:
I reread the code, indeed @randomstuff, that's the opposite, that's `${option_var}` which is interpreted to get the variable inside. (Just written above in the code actually)

Hum ok JimboJoe, my ynh_replace_string was still on stable. I now reproduce.
I see that it's happening indeed during the eval
```
getopts-L199: eval 'match_string+="$proxy_add_x_forwarded_for"'
```

I'll try other ways to get the variable contents, the thing is that I know that I put aside the 2 other ways I know. But what was the reason by this time...

maniackcrudelis:
So, from my tests (With the unstable/testing version of getopts)

- It works by using `'\$proxy_add_x_forwarded_for'` instead of `'$proxy_add_x_forwarded_for'`
So just by escaping the $.
- With `"\$proxy_add_x_forwarded_for"`, double quotes instead of simple, it doesn't, which is expected because double quotes are immediately interpreted by bash. So the backslash after.
- `${!option_var}` without eval doesn't work in the left part of a assignation.
- `declare ${option_var}+=\"${arguments[$i]}\"` doesn't work either because `declare` create a local variable.
- `declare -g ${option_var}+=\"${arguments[$i]}\"` doesn't work either.
`-g` is supposed to fix the behavior of `declare` by forcing the variable to a global scope.

There was indeed a good reason to decide at this time to use `eval` instead of those 2 other ways to get the content of the variable...

maniackcrudelis:
As your syntax is correct @JimboJoe, escaping $ should be done on getopts part.
By the way, I tried many other characters (exactly those: `\ ' & % + - ) ( } { ] [ = # @ ! / , ;`) in order to be sure, either inside double or simple quotes for the entire string.
Tried with `ynh_replace_special_string` to not trig any sed command.

Only $ seems to be a problem, as eval tries to interpret the text following $ as a variable.

Thanks JimboJoe for this bug report, I was expecting some edge cases from this testing release.

randomstuff:
@maniackcrudelis, what about:

~~~sh
declare -g ${option_var}+="${arguments[$i]}"
~~~

randomstuff:
You still have backticksbases shell commands and dollar based shell command subtitution working:

```sh
my_helper --arg1 'aa`ls`' -b "'" -c
my_helper --arg1 'aa$(ls)' -b "'" -c
```

Gives (both);

```
arg1=aabackend debug filesystem getopts ip mysql network nodejs package print psql setting string system testit.sh user utils
```

randomstuff:
> -g is supposed to fix the behavior of declare by forcing the variable to a global scope.

For reference, the reason `-g` does not work it that it makes the variable appear in the global scope (top-level), not the caller local scope.

randomstuff:
@maniackcrudelis, I think you did not understand what I intended to say (my formulation was somewhat ambiguous) in:

> A more correct solution is to do something like `shell_quote` in python.

What I meant to say that a similar quoting strategy should be implemented in bash script than the one which is implemented in Python in `shell_quote`. (Which is what #685 is attempting to do.)

randomstuff:
> I can't see any reason for an admin with a root permission (needed to run those scripts) to try to inject anything since he can simply do whatever he wants.

Of course the admin does not need to inject itself. However [he may need to use shell metacharacters in arguments](https://github.com/YunoHost-Apps/discourse_ynh/blob/testing/scripts/upgrade#L78).

At some point the admin may want to use these helpers in scripts (that's the reason CLI are nice after all). Suppose someone does something like that:

```sh
ynh_print_warn -m "Status of remote service is $(curl https://remoteserver/status)"
```
`ynh_print_warn` is currently relying on `ynh_handle_getopts_args` which does backtick shell command expansion *under the hood*:

```sh
ynh_print_warn -m '`ls`'  # Compare to: echo '`ls`'
```

maniackcrudelis:
> By the way, I tried many other characters (exactly those: \ ' & % + - ) ( } { ] [ = # @ ! / , ;) in order to be sure, either inside double or simple quotes for the entire string.
Tried with ynh_replace_special_string to not trig any sed command.

> Only $ seems to be a problem, as eval tries to interpret the text following $ as a variable.

It works perfectly as it is.
Using `$()` will start a sub shell, which could have side effects, adding simple quote could have also side effects, should be tested to be sure.
Anyway, getopts is here a helper used by other helpers, there's no meaning in testing solution into a simple bash script. Fixes have to be tested in real situation with an app script.

With the current fix it works. With all characters.

randomstuff:
> With the current fix it works. With all characters.

Not with backticks.

maniackcrudelis:
Backticks are deprecated in favor of `$()`.
As getopts is a helper of helpers, finally meant to be used by apps. To not cover that edge case isn't a problem for me.
Again, it works as it is. We'll see if other cases emerge from the testing or usages.
But I'm interested in usage according to the purpose of this helper. Not in diverted usages for other purposes.

PR #647, opened 2019-02-10
Add app debugger helper

maniackcrudelis:
Just wrote this helper to help me debugging [getopts](https://github.com/YunoHost/yunohost/pull/646). Figured it would be really useful for every packager while trying to debug an app.
This helper can simply print a message or a variable, marked as debug. But also trace a defined part of a script instead of simply using --debug while running an install.

As this helper isn't meant to be used permanently in app packages, I think we can't merge it quickly, so it will be easily available for debugging.

## PR Status

Ready to be reviewed.

## How to test

Try this helper in any app.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/debug
+++ b/data/helpers.d/debug
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# Debugger for app packagers
+#
+# usage: ynh_debug [--message=message] [--trace=1/0]
+# | arg: -m, --message= - The text to print
+# | arg: -t, --trace=   - Turn on or off the trace of the script. Usefull to trace nonly a small part of a script.
+ynh_debug () {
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Bump

This helper is really helpful.
If I could not copy it each time I need it, it would be wonderful ;)

maniackcrudelis:
I mainly use that helper with `--trace` option, to trace only the part of the code I want to debug.
So, without modifying all lines of the code, I can see what's going on a specific part.
In opposition with --debug that print the entire code. Also, this helper works in package check.

maniackcrudelis:
Thanks @alexAubin \o/

PR #648, opened 2019-02-12
[enh] Print diff of files when backup by ynh_backup_if_checksum_is_different

maniackcrudelis:
## The problem

When `ynh_backup_if_checksum_is_different` backup a file because it was modified, an admin can't know easily what are the differences.

## Solution

Print a diff of the differences between the previous file and the new one.

## PR Status

Ready to be reviewed.

## How to test

Change a config file, then upgrade the app.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/filesystem
+++ b/data/helpers.d/filesystem
@@ -352,7 +352,8 @@ ynh_backup_if_checksum_is_different () {
 			local backup_file="/home/yunohost.conf/backup/$file.backup.$(date '+%Y%m%d.%H%M%S')"
 			sudo mkdir -p "$(dirname "$backup_file")"
 			sudo cp -a "$file" "$backup_file"	# Backup the current file
-			echo "File $file has been manually modified since the installation or last upgrade. So it has been duplicated in $backup_file" >&2
+			ynh_print_info "File $file has been manually modified since the installation or last upgrade. So it has been duplicated in $backup_file"
+			ynh_print_info "$(git diff --no-index --patch-with-stat $backup_file $file)"
 			echo "$backup_file"	# Return the name of the backup file
 		fi
 	fi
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Bump

@YunoHost/apps

data/helpers.d/filesystem
```diff
@@ -345,15 +355,17 @@ ynh_backup_if_checksum_is_different () {
...
 		then	# If the checksum is now different
-			local backup_file="/home/yunohost.conf/backup/$file.backup.$(date '+%Y%m%d.%H%M%S')"
-			sudo mkdir -p "$(dirname "$backup_file")"
-			sudo cp -a "$file" "$backup_file"	# Backup the current file
-			echo "File $file has been manually modified since the installation or last upgrade. So it has been duplicated in $backup_file" >&2
-			echo "$backup_file"	# Return the name of the backup file
+			backup_file_checksum="/home/yunohost.conf/backup/$file.backup.$(date '+%Y%m%d.%H%M%S')"
+			sudo mkdir -p "$(dirname "$backup_file_checksum")"
+			sudo cp -a "$file" "$backup_file_checksum"	# Backup the current file
+			ynh_print_info "File $file has been manually modified since the installation or last upgrade. So it has been duplicated in $backup_file_checksum"
```
alexAubin: Hmm, don't we want this to be a warning ? Naively that sounds like a "not normal" situation and the user should be "encouraged" to read this message :P

maniackcrudelis: Not sure we should say "not normal", that mainly means that the admin has changed the config somehow.
Now about printing that at a warning or an info... Well don't know...
Considering what ynh_print_info is now used for, maybe indeed a warning would be better.

PR #650, opened 2019-02-16
[enh] Improve / add some messages to improve UX

alexAubin:
## The problem

Weird UX stuff, especially the "No app to upgrade" error when some app upgrade failed ... Also had this duplicate thing when just upgrading nextcloud : 

```
Info: Upgrading apps nextcloud
Info: Upgrading app nextcloud…
```

## Solution

Various tweaks to improve UX when dealing with app upgrades

## PR Status

To be tested :D 

## How to test

Trigger some app upgrades with single or several apps, and also check what happens when an upgrade fails

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Josue-T:
Not directly related to this PR, but I think that adding a possibility to fore the upgrade even if the app is up to date, could be really useful in some cases.

alexAubin:
@Josue-T : so like a `--force` option which upgrades anyway ?

Psycojoker:
Great! Doing exactly this was on my to-do list so 👍 (I just can't correctly review right now)

Josue-T:
> @Josue-T : so like a `--force` option which upgrades anyway ?

Yes

alexAubin:
About the --force option for app upgrade, there's an issue already opened here : https://github.com/YunoHost/issues/issues/33

alexAubin:
Flagging this as ready to merge

PR #651, opened 2019-02-16
[fix] Loading only one helper file leads to errors because missing getopts

alexAubin:
## The problem

Ran a postinstall on unstable and realized that some regenconf hooks were broken because they use a helper, and getopts wasn't sourced...

## Solution

Well that's a solution where I just load all helpers instead as in apps. Dunno that's what we want, maybe there's a significant overhead compared to sourcing getopts in each file ?

## PR Status

Not tested

## How to test

Run a postinstall with and without this branch

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/backup/05-conf_ldap
+++ b/data/hooks/backup/05-conf_ldap
@@ -4,7 +4,7 @@
 set -eu
 
 # Source YNH helpers
-source /usr/share/yunohost/helpers.d/filesystem
+source /usr/share/yunohost/helpers
 
 # Backup destination
 backup_dir="${1}/conf/ldap"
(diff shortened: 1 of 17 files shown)


maniackcrudelis:
That's probably a better idea to load all helpers, considering helpers are more and more related to each others.

alexAubin:
Some numbers trying on a RPi 2 : 

`time for I in `seq 1 10`; do source /usr/share/yunohost/helpers.d/ip; done`

-> 0.03 secs

`time for I in `seq 1 10`; do source /usr/share/yunohost/helpers; done`

->1.2 secs

So I guess the relative difference is huge, but in terms of absolute value I think it's not a big deal to add 1 sec to postinstall for this ...

alexAubin:
Yolomergin'

PR #652, opened 2019-02-18
[enh] Assert dpkg is not broken before app install

alexAubin:
## The problem

Follow-up of https://github.com/YunoHost/issues/issues/1284 and https://github.com/YunoHost/yunohost/pull/638.

## Solution

Make sure that dpkg/apt ain't broken before starting to install an app. This could also be done for other scripts which are quite likely to rely on apt : upgrade, remove and restore. I don't know if that's relevant though ... e.g. maybe we want people (or yunohost) to trigger removals even if dpkg/apt is broken ? I dunno ...

Note that the plan is anyway to add this check in ynh_apt though https://github.com/YunoHost/yunohost/pull/638 but the plan is to, where relevant, avoid starting to run a whole script if we can know it's gonna fail anyway.

## PR Status

Tested and working

## How to test

c.f. https://github.com/YunoHost/issues/issues/1284#issuecomment-459375469

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -146,6 +146,7 @@
     "diagnosis_monitor_network_error": "Can't monitor network: {error}",
     "diagnosis_monitor_system_error": "Can't monitor system: {error}",
     "diagnosis_no_apps": "No installed application",
+    "dpkg_is_broken": "You cannot do this right now because dpkg/apt (the system package managers) seems to be in a broken state... You can try to solve this issue by connecting through SSH and running `sudo dpkg --configure -a`.",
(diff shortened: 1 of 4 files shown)


rds13:
I find it relevant that Yunohost doesn't try to install some packages on a broken system.

alexAubin:
So today there was an interesting use case where : 
- someone installed the collaboraofficedocker app
- when uninstall it, it broke apt for some reason related to docker-ce and whatever
- later tried to upgrade nextcloud
- but apt was broken, so the upgrade failed
- and the restore of nextcloud's backup also failed, resulting in the app getting removed

TL;DR : one should not be able to try to upgrade an app if dpkg is broken or that will result in absolute disaster

PR #653, opened 2019-02-18
[enh] Decouple the regen-conf mechanism from services

alexAubin:
## The problem

This is a rewriting of https://github.com/YunoHost/yunohost/pull/480 because the original PR got stale after too many changes affecting the codebase as a whole (operation logs, error handling, ...)

**N.B. : I intend to merge this for the 3.5 release ... Otherwise it's gonna be the same story all over again :/**

Original problem : 

Regen-conf is currently heavily coupled to the notion of services, and this is problematic / annoying for many reasons : 
- we have "pseudo-services" like "ssl", "nsswitch" and "yunohost" which are not actual services but just here to handle some configuration files
- it's annoying to create a whole service, then add all the hack to make it "not a real service" when you in fact just want to manage a single file. This is why I ended up putting the management of etckeeper.conf inside the "yunohost" regen-conf hook
- we have other examples of files we would like to manage independently of the notion of service, for instance the dns resolvers (c.f. #352 ) and cron jobs (dyndns, cert-renewal)

## Solution

Decouple the regen-conf mechanism from the concept of services. To do this, the regen-conf related code was moved in a regenconf.py and file hashes are now stored in a regenconf.yml, independently from services.yml.

Of course that's a delicate operation since the regen-conf is one of the core feature of the core :o and a migration is required to properly handle this ... (which is why I ended up working on stretch-unstable to add a 6th migration)

The API also gets changed from `yunohost service regen-conf` (still supported but deprecated) to `yunohost tools regen-conf`.

Nevertheless, after this we'll be able to safely add simple conf-regen hooks for new files without all the weird "pseudo-service" thing.

## PR Status

Still needs some heavy testing to make sure this doesn't break anything, but should be relatively stable so can be reviewed

## How to test

Pull this branch ... the first time you run a regen-conf, it should automatically migrate. Though we also need to test postinstall and a "real-life upgrade" (which triggers the postinst script and so on)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/regenconf.py
```diff
@@ -43,14 +43,16 @@
 logger = log.getActionLogger('yunohost.regenconf')
 
 
+# FIXME : those ain't just services anymore ... what are we supposed to do with this ...
```
Psycojoker: Can't we migrate those?

src/yunohost/regenconf.py
```diff
@@ -290,39 +291,28 @@ def _pre_call(name, priority, path, args):
...
-    Get a dict of managed services with their parameters
-
+    Get a dict of regen conf informations
     """
     try:
-        with open('/etc/yunohost/services.yml', 'r') as f:
-            services = yaml.load(f)
+        with open(REGEN_CONF_FILE, 'r') as f:
+            return yaml.load(f)
     except:
```
Psycojoker: "except:" T_T

(I know it's not related to this PR)

src/yunohost/regenconf.py
alexAubin: You mean the existing logs ? Meh I guess it's not so much of a big deal to not migrate them, usually you only care about the recent logs

I was more thinking about if we can change this `'service'` key in the `is_unit_operation` and I guess we can indeed replace it to something like `configuration` ?

src/yunohost/service.py
```diff
@@ -414,12 +414,20 @@ def service_regen_conf(names=[], with_diff=False, force=False, dry_run=False,
...
 
     services = _get_services()
+
+    if isinstance(names, str):
+        names = [names]
+
     for name in names:
-        if name not in services:
-            raise YunohostError('service_unknown', service=service)
+        if name not in services.keys():
```
Psycojoker: Misc: the `.keys()` is not needed here normally (if it's a dict "a in some_dict" == "a in some_dict.keys()"

src/yunohost/regenconf.py
Psycojoker: or "regen-conf"?

Psycojoker:
I'm not 100% sure but shouldn't we ship a default regenconf.yaml file like we do for service for new installations? It won't be created since we don't run migrations on installation.

Or those this magic code is going to create it? https://github.com/YunoHost/yunohost/pull/653/commits/0ebbb83191273ca716ecaf1c128ece0705b84c82#diff-3cee47e1d870699f286cf66e3a4e04d9R67

alexAubin:
>
>I'm not 100% sure but shouldn't we ship a default  regenconf.yaml file like we do for service for new installations? It  won't be created since we don't run migrations on installation.

Ah yes indeed, we should probably do something explicitly during the postinstall. I forgot to look into this but I doubt that just 0ebbb83 will take care of it

alexAubin:
Alrighty, ran a few tests and fixed a few things. Both the migration and fresh postinstall seem to work fine and behave as expected so it seems ready for testing.

As discussed in meetings a few weeks ago, I'm merging this as we're at the beginning of 3.6 so we'll have time to fully validate this in testing until the stable release

PR #654, opened 2019-02-19
[enh] Be able to define hook to trigger when changing a setting

alexAubin:
## The problem

c.f. https://github.com/YunoHost/yunohost/pull/640 and other similar use case : we need to be able to trigger specific code when updating a setting value. For instance, to trigger a regenconf of nginx or ssh or whatever.

## Solution

Introduce a decorator `@post_change_hook("your.setting.name")` to define what code to trigger after the setting is changed. A dummy example is : 

```python
@post_change_hook("example.int")
def myfunc(old_value, new_value):
    print("In hook")
    print(old_value)
    print(new_value) 
```

## PR Status

Tested, ready for review

## How to test

Uncomment the example at the end and try to run for instance `yunohost settings set example.int -v 1664`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/settings.py
+++ b/src/yunohost/settings.py
@@ -264,7 +264,7 @@ def trigger_post_change_hook(setting_name, old_value, new_value):
         return
 
     f = post_change_hooks[setting_name]
-    f(old_value, new_value)
+    f(setting_name, old_value, new_value)
 
 
 # ===========================================
@@ -273,14 +273,15 @@ def trigger_post_change_hook(setting_name, old_value, new_value):
 # You can define such an action with :
 #
 # @post_change_hook("your.setting.name")
-# def some_function_name(old_value, new_value):
+# def some_function_name(setting_name, old_value, new_value):
 #     # Do some stuff
 #
 # ===========================================
 
 
 #@post_change_hook("example.int")
-#def myfunc(old_value, new_value):
+#def myfunc(setting_name, old_value, new_value):
 #    print("In hook")
+#    print(setting_name)
 #    print(old_value)
 #    print(new_value)
--- a/src/yunohost/settings.py
+++ b/src/yunohost/settings.py
@@ -115,9 +115,17 @@ def settings_set(key, value):
         raise YunohostError('global_settings_unknown_type', setting=key,
                             unknown_type=key_type)
 
+    old_value = settings[key].get("value")
     settings[key]["value"] = value
-
     _save_settings(settings)
+
+    # TODO : whatdo if the old value is the same as
+    # the new value...
+    try:
+        trigger_post_change_hook(key, old_value, value)
+    except Exception as e:
+        logger.error("Post-change hook for setting %s failed : %s" % (key, e))
+        raise
 
 
 def settings_reset(key):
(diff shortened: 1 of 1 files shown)


src/yunohost/settings.py
```diff
@@ -235,3 +243,44 @@ def _save_settings(settings, location=SETTINGS_PATH):
...
+    f(old_value, new_value)
+
+
+# ===========================================
+#
+# Actions to trigger when changing a setting
+# You can define such an action with :
+#
+# @post_change_hook("your.setting.name")
+# def some_function_name(old_value, new_value):
```
rds13: We might need to have the setting name in the parameter of this function.
You could share the same hook between two settings. You could avoid having the setting name hardcoded when you want to log a context for example.

src/yunohost/settings.py
```diff
@@ -235,3 +243,44 @@ def _save_settings(settings, location=SETTINGS_PATH):
...
+#     # Do some stuff
+#
+# ===========================================
+
+
+#@post_change_hook("example.int")
+#def myfunc(old_value, new_value):
+#    print("In hook")
```
rds13: Shall we protect this function from triggerring another setting change ?

src/yunohost/settings.py
```diff
@@ -235,3 +243,44 @@ def _save_settings(settings, location=SETTINGS_PATH):
...
+
+
+# Meant to be a dict of setting_name -> function to call
+post_change_hooks = {}
+
+
+def post_change_hook(setting_name):
+    def decorator(func):
+        assert setting_name in DEFAULTS.keys(), "The setting %s does not exists" % setting_name
+        assert setting_name not in post_change_hooks, "You can only register one post change hook per setting (in particular for %s)" % setting_name
```
rds13: I understand this cover use case for #640. But in other use case it could be useful to have a list of hook functions associated with a setting name. What do you think ?

src/yunohost/settings.py
alexAubin: Indeed ;) Done !

src/yunohost/settings.py
alexAubin: Hmyea I was a bit lazy to think about that so I just added this constrain :sweat_smile: 

My idea is that it should be easy to implement once we need it

src/yunohost/settings.py
alexAubin: uuuh, I guess, but how would you do that ?

I'm not sure that we really are going to implement setting triggers likely to trigger other setting changes ?

rds13: With decorator dependencies are not obvious. We could have a map that remembers what setting is currently triggering a hook and check that before entering hook and remove from map on hook exit.

src/yunohost/settings.py
rds13: I see you have already done it ! 😄  I cannot fix the documentation at the top of this PR but may be it's not very important.

src/yunohost/settings.py
alexAubin: Uuuuuh yes I guess it's technically doable, but I doubt that we're going to need this anytime soon ...

Really most of the hook are going to be : 

```
@post_change_hook("some.setting")
def myfunc(...):
    ... regen some configuration like nginx ...
```

If we ever do very complicated stuff in the future (again I doubt it é_è) we'll still be able to implement it later

rds13: Ok let's see how much time it will live before you ought me a 🍺 😆 !

rds13: I'm ready to fix merge conflict on #640 once this branch is merged.

alexAubin: I will sure you pay a beer if I can even if we don't end up implementing this :stuck_out_tongue:

src/yunohost/settings.py
alexAubin: Uuuuh, not sure we're talking about the same thing ? 

Ain't we talking about the ability or not to have multiple @post_change_hook("name") for the same "name" ?

e.g. 

```python
@post_change_hook("foo")
def foo1(...):
   ...

@post_change_hook("foo")
def foo3(...):
   ...
```

?

rds13: Sorry I messed up. I was indeed talking first about having multiple hook functions for the same setting but it’s not mandatory.
Then I switched to talking about having the setting name as a parameter.

rds13:
It looks like this can be merged now.

PR #656, opened 2019-02-19
Cache results from meltdown checker

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1304

## Solution

Cache the results from meltdown checker instead of re-running the script everytime a diagnosis is done...

The cache is stored in `/tmp/` with the idea that it shall be cleared upon a reboot (which might change the result from the script). Also since we might want to change what we do with this diagnosis, I check the modification time of the cached file against `/var/log/dpkg.log` such that the diagnosis is re-ran if there was an upgrade (of yunohost for example)

## PR Status

Tested and working

## How to test

Run `yunohost tools diagnosis --debug` a few time, you should see a message indicating the result is write to / read from the cached file

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -724,7 +724,6 @@ def _check_if_vulnerable_to_meltdown():
     # (e.g. from yunohost)
     cache_file = "/tmp/yunohost-meltdown-diagnosis"
     dpkg_log = "/var/log/dpkg.log"
-    print(os.path.exists(cache_file))
     if os.path.exists(cache_file):
         if not os.path.exists(dpkg_log) or os.path.getmtime(cache_file) > os.path.getmtime(dpkg_log):
             logger.debug("Using cached results for meltdown checker, from %s" % cache_file)
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -713,6 +713,23 @@ def tools_diagnosis(auth, private=False):
 def _check_if_vulnerable_to_meltdown():
     # meltdown CVE: https://security-tracker.debian.org/tracker/CVE-2017-5754
 
+    # We use a cache file to avoid re-running the script so many times,
+    # which can be expensive (up to around 5 seconds on ARM)
+    # and make the admin appear to be slow (c.f. the calls to diagnosis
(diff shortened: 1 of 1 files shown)


src/yunohost/tools.py
```diff
@@ -713,6 +713,23 @@ def tools_diagnosis(auth, private=False):
...
+    # and make the admin appear to be slow (c.f. the calls to diagnosis
+    # from the webadmin)
+    #
+    # The cache is in /tmp and shall disappear upon reboot
+    # *or* we compare it to dpkg.log modification time
+    # such that it's re-ran if there was package upgrades
+    # (e.g. from yunohost)
+    cache_file = "/tmp/yunohost-meltdown-diagnosis"
+    dpkg_log = "/var/log/dpkg.log"
+    print(os.path.exists(cache_file))
```
Psycojoker: Looks like  a debug print.

alexAubin: Indeed ;D

Psycojoker:
I think there are quite other things that could be optimized in diagnosis (like, I've added "yunohost service regenconf --dry-run -d" which is super slow :/)

I don't think that the super sequential architecture of this command is really good at scaling :/

But I guess this will be changed when we'll move to the global diagnosis system (one day ™)

alexAubin:
>I've added "yunohost service regenconf --dry-run -d" which is super slow :/

Uuh where though ? o.O

But yup the global diagnosis system should be more modular with cache and rainbows and stuff, someday™

Psycojoker:
> > I've added "yunohost service regenconf --dry-run -d" which is super slow :/
> 
> Uuh where though ? o.O

Here https://github.com/YunoHost/yunohost/blob/10f33010614ea507dbdd557cdf3313798dd9bc44/src/yunohost/tools.py#L696

alexAubin:
Ah yes, but that's for the "private" diagnosis so it happens less often ;P

Psycojoker:
Yes, fortunately :sweat_smile:

PR #657, opened 2019-02-20
[enh] Update psql helper

kay0u:
## The problem

...

## Solution

Update psql helper

## PR Status

Tested in local. ~~I spotted a silent error when restore a database from a file `ynh_psql_connect_as $db_name $db_pwd $db_name < ./db.sql` but already present with the old version:
"ERROR:  must be owner of extension plpgsql"~~
I think I've fixed that

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/psql
+++ b/data/helpers.d/psql
@@ -12,15 +12,15 @@ PSQL_ROOT_PWD_FILE=/etc/yunohost/psql
 # | arg: -p, --password - the user password
 # | arg: -d, --database - the database to connect to
 ynh_psql_connect_as() {
-    # Declare an array to define the options of this helper.
-    local legacy_args=upd
-    declare -Ar args_array=( [u]=user= [p]=password= [d]=database= )
-    local user
-    local password
-    local database
-    # Manage arguments with getopts
-    ynh_handle_getopts_args "$@"
-    database="${database:-}"
+	# Declare an array to define the options of this helper.
+	local legacy_args=upd
+	declare -Ar args_array=([u]=user= [p]=password= [d]=database=)
+	local user
+	local password
+	local database
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+	database="${database:-}"
 
 	sudo --login --user=postgres PGUSER="$user" PGPASSWORD="$password" psql "$database"
 }
@@ -30,18 +30,18 @@ ynh_psql_connect_as() {
 # usage: ynh_psql_execute_as_root --sql=sql [--database=database]
 # | arg: -s, --sql - the SQL command to execute
 # | arg: -d, --database - the database to connect to
-ynh_psql_execute_as_root () {
-    # Declare an array to define the options of this helper.
-    local legacy_args=sd
-    declare -Ar args_array=( [s]=sql= [d]=database= )
-    local sql
-    local database
-    # Manage arguments with getopts
-    ynh_handle_getopts_args "$@"
-    database="${database:-}"
+ynh_psql_execute_as_root() {
+	# Declare an array to define the options of this helper.
+	local legacy_args=sd
+	declare -Ar args_array=([s]=sql= [d]=database=)
+	local sql
+	local database
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+	database="${database:-}"
 
-    ynh_psql_connect_as --user="postgres" --password="$(sudo cat $PSQL_ROOT_PWD_FILE)" \
-        --database="$database" <<< "$sql"
+	ynh_psql_connect_as --user="postgres" --password="$(sudo cat $PSQL_ROOT_PWD_FILE)" \
+		--database="$database" <<<"$sql"
 }
 
 # Execute a command from a file as root user
@@ -50,17 +50,17 @@ ynh_psql_execute_as_root () {
 # | arg: -f, --file - the file containing SQL commands
 # | arg: -d, --database - the database to connect to
 ynh_psql_execute_file_as_root() {
-    # Declare an array to define the options of this helper.
-    local legacy_args=fd
-    declare -Ar args_array=( [f]=file= [d]=database= )
-    local file
-    local database
-    # Manage arguments with getopts
-    ynh_handle_getopts_args "$@"
-    database="${database:-}"
+	# Declare an array to define the options of this helper.
+	local legacy_args=fd
+	declare -Ar args_array=([f]=file= [d]=database=)
+	local file
+	local database
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+	database="${database:-}"
 
-    ynh_psql_connect_as --user="postgres" --password="$(sudo cat $PSQL_ROOT_PWD_FILE)" \
-        --database="$database" < "$file"
+	ynh_psql_connect_as --user="postgres" --password="$(sudo cat $PSQL_ROOT_PWD_FILE)" \
+		--database="$database" <"$file"
 }
 
 # Create a database and grant optionnaly privilegies to a user
@@ -72,10 +72,18 @@ ynh_psql_execute_file_as_root() {
 # | arg: user - the user to grant privilegies
 # | arg: pwd - the password to identify user by
 ynh_psql_create_db() {
-    local db=$1
+	local db=$1
 
-    ynh_psql_create_user "$user" "$pwd"
-	sudo --login --user=postgres createdb --owner="$user" "$db"
+    local sql="CREATE DATABASE ${db};"
+
+    # grant all privilegies to user
+	if [[ $# -gt 1 ]]; then
+		#ynh_psql_create_user "$user" "$pwd"
+		sql+=" GRANT ALL PRIVILEGES ON ${db} TO ${2}  WITH GRANT OPTION;"
+	fi
+
+	#sudo --login --user=postgres createdb --owner="$user" "$db"
+	ynh_psql_execute_as_root --sql="$sql"
 }
 
 # Drop a database
@@ -89,7 +97,7 @@ ynh_psql_create_db() {
 # | arg: db - the database name to drop
 ynh_psql_drop_db() {
 	local db=$1
-    sudo --login --user=postgres dropdb $db
+	sudo --login --user=postgres dropdb $db
 }
 
 # Dump a database
@@ -100,14 +108,14 @@ ynh_psql_drop_db() {
 # | arg: -d, --database - the database name to dump
 # | ret: the psqldump output
 ynh_psql_dump_db() {
-    # Declare an array to define the options of this helper.
-    local legacy_args=d
-    declare -Ar args_array=( [d]=database= )
-    local database
-    # Manage arguments with getopts
-    ynh_handle_getopts_args "$@"
+	# Declare an array to define the options of this helper.
+	local legacy_args=d
+	declare -Ar args_array=([d]=database=)
+	local database
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
 
-    sudo --login --user=postgres pg_dump "$database"
+	sudo --login --user=postgres pg_dump "$database"
 }
 
 # Create a user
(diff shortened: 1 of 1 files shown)


data/helpers.d/psql
```diff
@@ -221,7 +221,7 @@ ynh_psql_remove_db() {
 	ynh_handle_getopts_args "$@"
 
 	local psql_root_password=$(sudo cat $PSQL_ROOT_PWD_FILE)
-	if ynh_psql_database_exists "$db_name"; then # Check if the database exists
+	if $(ynh_psql_database_exists "$db_name"); then # Check if the database exists
```
alexAubin: Are you sure about this tho ? To me it means two different things ? o.O

data/helpers.d/psql
```diff
@@ -143,7 +143,7 @@ ynh_psql_user_exists() {
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
 
-	if [[ -n $(sudo --login --user=postgres PGUSER="postgres" PGPASSWORD="$(sudo cat $PSQL_ROOT_PWD_FILE)" psql -tAc "SELECT rolname FROM pg_roles WHERE rolname='$user';") ]]; then
+	if [[ -z $(sudo --login --user=postgres PGUSER="postgres" PGPASSWORD="$(sudo cat $PSQL_ROOT_PWD_FILE)" psql -tAc "SELECT rolname FROM pg_roles WHERE rolname='$user';") ]]; then
```
alexAubin: I guess you know what you're doing but you sure about this right ? xD

data/helpers.d/psql
```diff
@@ -128,7 +126,7 @@ ynh_psql_dump_db() {
 ynh_psql_create_user() {
 	local user=$1
 	local psql=$2
-	ynh_psql_execute_as_root --sql="CREATE USER $user WITH PASSWORD '$pwd'"
+	ynh_psql_execute_as_root --sql="CREATE USER $user WITH PASSWORD $pwd"
```
alexAubin: Uh what's this `psql` variable ? Shouldn't it be `pwd` ?

data/helpers.d/psql
```diff
@@ -202,6 +200,10 @@ ynh_psql_setup_db() {
 	# If $db_pwd is not given, use new_db_pwd instead for db_pwd
 	db_pwd="${db_pwd:-$new_db_pwd}"
 
+	if [ $(ynh_psql_user_exists --user=$db_user) ]; then
+		ynh_psql_create_user "$db_name" "$db_user" "$db_pwd"
```
alexAubin: There's 3 arguments given there ... but doesn't the helper only takes 2 ?

data/helpers.d/psql
maniackcrudelis: Indeed, with $() if will test the value of the string echoed by ynh_psql_database_exists. But there isn't any.
Without $(), if will test the exit code. So you probably don't want to add a sub shell here.

data/helpers.d/psql
maniackcrudelis: Same here, you would better try the return code I think.
And please don't use [[ ]] unless you really need them for their specific syntax.

data/helpers.d/psql
kay0u: This code is inspired by : https://github.com/YunoHost/yunohost/blob/stretch-unstable/data/helpers.d/mysql#L213

But ok, i'll change that.

data/helpers.d/psql
kay0u: Same here: https://github.com/YunoHost/yunohost/blob/stretch-unstable/data/helpers.d/mysql#L145

data/helpers.d/psql
kay0u: You're right :-)

data/helpers.d/psql
kay0u: I'll change that! Thx

kay0u:
Need another review now :-)

data/helpers.d/psql
```diff
@@ -1,148 +1,273 @@
...
-# Create a database and grant privilegies to a user
+# Create a database and grant optionnaly privilegies to a user
+#
+# [internal]
 #
 # usage: ynh_psql_create_db db [user [pwd]]
 # | arg: db - the database name to create
 # | arg: user - the user to grant privilegies
-# | arg: pwd  - the user password
+# | arg: pwd - the password to identify user by
```
maniackcrudelis: pwd doesn't seem to be used in that helper.
user is used as $2 instead $user.

Also, user is shown as optional but the code doesn't set a default value.

data/helpers.d/psql
```diff
@@ -1,148 +1,273 @@
...
-	local user="$2"
-	local pwd="$3"
-	ynh_psql_create_user "$user" "$pwd"
-	sudo --login --user=postgres createdb --owner="$user" "$db"
+	local db=$1
+
+    local sql="CREATE DATABASE ${db};"
+
+    # grant all privilegies to user
+	if [[ $# -gt 1 ]]; then
```
maniackcrudelis: You can use simple [] here

data/helpers.d/psql
```diff
@@ -1,148 +1,273 @@
...
 # usage: ynh_psql_create_user user pwd [host]
 # | arg: user - the user name to create
+# | arg: pwd - the password to identify user by
 ynh_psql_create_user() {
-	local user="$1"
-	local pwd="$2"
-	sudo --login --user=postgres psql -c"CREATE USER $user WITH PASSWORD '$pwd'" postgres
+	local user=$1
+	local pwd=$2
+	ynh_psql_execute_as_root --sql="CREATE USER $user WITH ENCRYPTED PASSWORD '$pwd'"
```
maniackcrudelis: Did you try it ?
Sure $pwd will be interpreted before going to ynh_psql_execute_as_root ?

data/helpers.d/psql
```diff
@@ -1,148 +1,273 @@
...
+		return 1
+	else
+		return 0
+	fi
+}
+
+# Check if a psql database exists
+#
+# usage: ynh_psql_database_exists --database=database
+# | arg: -d, --database - the database for which to check existence
```
maniackcrudelis: -d here, -u below

data/helpers.d/psql
kay0u: It was to keep the compatibility with mysql helpers. But you're probably right.

maniackcrudelis: "compatibility with mysql helpers" ?
What do you mean ?

data/helpers.d/psql
kay0u: Yep, tested. Here a log of my ynh-dev:

```
2019-02-20 23:32:06,069: DEBUG - + ynh_psql_create_user lstu w3i9QOs6zW1jPW1OUlsiyg0z
2019-02-20 23:32:06,069: DEBUG - + local user=lstu
2019-02-20 23:32:06,069: DEBUG - + local pwd=w3i9QOs6zW1jPW1OUlsiyg0z
2019-02-20 23:32:06,069: DEBUG - + ynh_psql_execute_as_root '--sql=CREATE USER lstu WITH ENCRYPTED PASSWORD '\''w3i9QOs6zW1jPW1OUlsiyg0z'\'''
```

maniackcrudelis: I trust you if you have tested it.

data/helpers.d/psql
```diff
@@ -67,18 +67,18 @@ ynh_psql_execute_file_as_root() {
...
 # [internal]
 #
-# usage: ynh_psql_create_db db [user [pwd]]
+# usage: ynh_psql_create_db db [user]
 # | arg: db - the database name to create
 # | arg: user - the user to grant privilegies
-# | arg: pwd - the password to identify user by
 ynh_psql_create_db() {
 	local db=$1
+	local user=$2
```
maniackcrudelis: If user is an optional argument, it should have a default value

kay0u: I can move this line after the `if [ $# -gt 1 ]; then`, or set a default argument. What do you think about it?

data/helpers.d/psql
kay0u: I mean that ynh_myql_create_db can take pwd in argument, then if a packager want to move from mysql to psql, he just have to change ynh_mysql_whatever to ynh_psql_whatever

data/helpers.d/psql
maniackcrudelis: Sorry, miss read, I though it was a helper with getopts.
Not sure though TRAP would not trig an empty $2. Did you try it ?
Otherwise, you can simply put `local user=${2:-}` to prevent a failure.

data/helpers.d/psql
kay0u: done

data/helpers.d/psql
kay0u: fixed

data/helpers.d/psql
maniackcrudelis: Ok, I understand. But don't you think that a packager who will not look at the code will then think that he can indeed use a password here ?
Except that the password won't work here.

I would rather not add that info if it's not actually used

kay0u: In fact, my argument is irrelevant because this function is flaged as "internal" and a packager should use `ynh_psql_setup_db` instead (helper which can be used the same way than ynh_mysql_setup_db I think)

data/helpers.d/psql
kay0u: You're right, this is a problem when we don't use this optional argument.

Josue-T:
Why did you change the order of the helper, it make the change really difficult to review :-/

kay0u:
Sorry, it was more convenient for me to compare with MySQL helper, but for reviewers it's a lot harder. I can rearrange the code as before

data/helpers.d/psql
alexAubin: Well yes indeed ... but in fact that looks like a bug in the mysql helper ? (@maniackcrudelis do you agree ?)

```
function foo()
{
    return 1
}

echo $(foo)
```

doesn't print "1" in the console ...

data/helpers.d/psql
alexAubin: Indeed ... to me this also looks like a bug in the mysql helper as well ... (@maniackcrudelis wdu think ? ;D)

data/helpers.d/psql
```diff
@@ -1,148 +1,273 @@
...
-	local pwd="$3"
-	ynh_psql_create_user "$user" "$pwd"
-	sudo --login --user=postgres createdb --owner="$user" "$db"
+	local db=$1
+	local user=${2:-}
+
+    local sql="CREATE DATABASE ${db};"
+
+    # grant all privilegies to user
+	if [ $# -gt 1 ]; then
```
alexAubin: Hm I guess this is meant to check if `$user` is not empty ? Why not use `[[ -n "$user" ]]` to be more explicit ? :sweat_smile:

data/helpers.d/psql
```diff
@@ -1,148 +1,273 @@
...
+		fi
+
+		systemctl start postgresql
+		sudo --login --user=postgres psql -c"ALTER user postgres WITH PASSWORD '$pgsql'" postgres
+
+		# force all user to connect to local database using passwords
+		# https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html#EXAMPLE-PG-HBA.CONF
+		# Note: we can't use peer since YunoHost create users with nologin
+		#  See: https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/user
+		sed -i '/local\s*all\s*all\s*peer/i \
```
kay0u: I should use the ynh_replace_string

data/helpers.d/psql
maniackcrudelis: Yes agree on the principle. Here you would ask yourself, what do you want, the return code or any echo in the function ?

data/helpers.d/psql
maniackcrudelis: Guess the hidden message is that I should have a look to the mysql helper ;)

data/helpers.d/psql
maniackcrudelis: Without [[ ]] !
Otherwise, it would work as well.

data/helpers.d/psql
maniackcrudelis: Indeed

alexAubin:
Sooo, are we good with this ? (Did anybody test those ?)

kay0u:
I tested everything but `ynh_psql_execute_file_as_root`.

You can install/restore lstu to test this helper

maniackcrudelis:
Would be interesting to have some feedback from people who are actually using that postgres in apps.
Otherwise, that's look good to me.
Anyway, can't be worse than the current version...

alexAubin:
Soooo flagging this as ready to merge, will merge in a few days if there's no complain since then

PR #659, opened 2019-02-21
[fix] Do not use colors if the terminal doesn't support them

maniackcrudelis:
## The problem

https://github.com/YunoHost/issues/issues/1309

## Solution

Fix https://github.com/YunoHost/issues/issues/1309
Check if the terminal can use colors.

## PR Status

Tested, can be reviewed.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/print
+++ b/data/helpers.d/print
@@ -54,6 +54,24 @@ ynh_print_log () {
   echo -e "${1}"
 }
 
+# Check if the current terminal supports colors
+#
+# [internal]
+#
+are_colors_supported () {
+	# Check if colors are available with this terminal
+	local colors=""
+	if colors=$(tput colors 2> /dev/null)
+	then
+		# And if there's at least 256 colors.
+		if [ $colors -ge 256 ]
+		then
+			return 0
+		fi
+	fi
+	return 1
+}
+
 # Print a warning on stderr
 #
 # usage: ynh_print_warn --message="Text to print"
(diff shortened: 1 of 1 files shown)


data/helpers.d/print
```diff
@@ -54,6 +54,24 @@ ynh_print_log () {
   echo -e "${1}"
 }
 
+# Check if the current terminal supports colors
+#
+# [internal]
+#
+are_colors_supported () {
```
alexAubin: Can't we optimize this a bit and check this once, globally, rather than every time we run a `ynh_print_*` ?

Psycojoker:
If it's only to fix the webadmin, we can fix that like I did in YunoRunner using this code https://github.com/YunoHost/yunorunner/blob/master/static/js/app.js#L13 and we'll even get pretty colours :yum:

data/helpers.d/print
maniackcrudelis: Not sure we would save any time.
That would means runs that helper at the beginning of any script. Or even when the file is sourced. Do we want to do that ?

maniackcrudelis:
That's a function that replace bash colors by colors supported by java script ?

Psycojoker:
(by html) yes.

Psycojoker:
That's how jobs are colorized on individual jobs pages.

maniackcrudelis:
Considering that my PR will remove colors for the admin JS. It could be indeed better to adapt colors instead

alexAubin:
So uh should we proceed with integrating https://github.com/YunoHost/yunorunner/blob/master/static/js/app.js#L13 instead ?

Psycojoker:
That would made more sens yes and also brings colors to the admin UI (hoping it won't make things unreadable)

maniackcrudelis:
Is it supposed to be solved since last year ?
Should we continue with that PR ?

alexAubin:
Closing because I agressively fixed this the other day directly in stretch-unstable because I got fed up with this ... imho we just don't need to colorize this output, it's already a bit confusing to see "Warning: [WARN]" or "Warning: [ERR]", but if "Warning" (or the background of the message, in the webadmin) is in yellow and "[ERR]" in red that's even more confusion .. if we want an error logging channel, we should implement one like we did for ynh_print_info

PR #660, opened 2019-02-22
[enh] Add the arch argument to ynh_install_nodejs

yalh76:
## The problem

When installing nodejs on arm64 and arch64, [N - Node.js version management](https://github.com/tj/n) handle the architecture as [AMD64](https://github.com/tj/n/pull/448) and the nodejs installation failed

## Solution

Use the N - Node.js version management argument --arch to specify an architecture

## PR Status

Done

## How to test

There is a package with the functionality setuped: `yunohost app install https://github.com/YunoHost-Apps/wikijs_ynh/tree/arm --debug`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -69,6 +69,7 @@ ynh_install_nodejs () {
 	local arch
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
+	arch=${arch:-}
 
 	# Create $n_install_dir
 	mkdir -p "$n_install_dir"
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -98,6 +98,7 @@ ynh_install_nodejs () {
 	test -x /usr/bin/npm_n && mv /usr/bin/npm_n /usr/bin/npm
 
 	# Install the requested version of nodejs
+	uname=$(uname -m)
 	if [[ $uname =~ aarch64 ||  $uname =~ arm64]]
 	then
 		n $nodejs_version --arch=arm64
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -64,8 +64,9 @@ ynh_install_nodejs () {
 
 	# Declare an array to define the options of this helper.
 	local legacy_args=n
-	declare -Ar args_array=( [n]=nodejs_version= )
+	declare -Ar args_array=( [n]=nodejs_version= [a]=arch=)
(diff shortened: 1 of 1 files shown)


alexAubin:
Uh technically this looks good, but shouldn't we try to detect the arch and feed it to the node command, instead of leaving this job to the app packager ?

yalh76:
> 
> 
> Uh technically this looks good, but shouldn't we try to detect the arch and feed it to the node command, instead of leaving this job to the app packager ?

Well, in N, there is already an arch detection: https://github.com/tj/n/blob/master/bin/n#L411, so we don't need to implement one in ynh_install_nodejs. The only issue is: for now N arch detection doesn't work for aarch64/arm64 and N maintainer doesn't seem to want to implement it even if there is already a [PR ](https://github.com/tj/n/pull/448) to solve it

alexAubin:
Hmokay, but what I mean is why don't we add, in this helper, something like 

```
if [[ $uname =~ aarch64 ]] then
    arch="arm64"
fi
```

and feed that to the node command ? I mean, otherwise the app packager will have to think about doing this him/herself, for each apps, and know how to do this and that doesn't look quite obvious ... ? Or am I missing something x_X

yalh76:
> 
> 
> Hmokay, but what I mean is why don't we add, in this helper, something like
> 
> ```
> if [[ $uname =~ aarch64 ]] then
>     arch="arm64"
> fi
> ```
> 
> and feed that to the node command ? I mean, otherwise the app packager will have to think about doing this him/herself, for each apps, and know how to do this and that doesn't look quite obvious ... ? Or am I missing something x_X

Good point, instead of adding an argument to ynh_install_nodejs, let him do the job... I will change the PR

data/helpers.d/nodejs
```diff
@@ -98,7 +98,12 @@ ynh_install_nodejs () {
 	test -x /usr/bin/npm_n && mv /usr/bin/npm_n /usr/bin/npm
 
 	# Install the requested version of nodejs
-	n $nodejs_version
+	if [[ $uname =~ aarch64 ||  $uname =~ arm64]]
```
alexAubin: I think you need to define `$uname` tho :P 

Probably `uname=$(uname -m)`

yalh76: of course ^^ I done it too quickly...

PR #662, opened 2019-02-24
[fix] Optimize dyndns requests

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1300 - unecessary madness which makes everything slower (dyndns update taking the lock every 2 minutes to do a full HTTPS request for nothing and therefore dynette is bombarded with those and sometimes aint able to answer)

## Solution

Don't make all those crazy request, not really necessary ... Also add some constrains to try to remove ambiguities

## PR Status

Yolocommited, not tested :|

## How to test

Try this on a server with a registered nohost.me/noho.st/... Try to subscribe multiple time. Try a case where subscribing failed to see if keys are indeed removed.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/dyndns.py
+++ b/src/yunohost/dyndns.py
@@ -157,12 +157,12 @@ def dyndns_subscribe(operation_logger, subscribe_host="dyndns.yunohost.org", dom
     try:
         r = requests.post('https://%s/key/%s?key_algo=hmac-sha512' % (subscribe_host, base64.b64encode(key)), data={'subdomain': domain}, timeout=30)
     except requests.ConnectionError:
-        os.system("rm %s" % private_file)
-        os.system("rm %s" % key_file)
+        os.system("rm -f %s" % private_file)
+        os.system("rm -f %s" % key_file)
         raise YunohostError('no_internet_connection')
     if r.status_code != 201:
-        os.system("rm %s" % private_file)
-        os.system("rm %s" % key_file)
+        os.system("rm -f %s" % private_file)
+        os.system("rm -f %s" % key_file)
         try:
             error = json.loads(r.text)['error']
         except:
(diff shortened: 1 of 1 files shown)


src/yunohost/dyndns.py
```diff
@@ -153,8 +157,12 @@ def dyndns_subscribe(operation_logger, subscribe_host="dyndns.yunohost.org", dom
     try:
         r = requests.post('https://%s/key/%s?key_algo=hmac-sha512' % (subscribe_host, base64.b64encode(key)), data={'subdomain': domain}, timeout=30)
     except requests.ConnectionError:
+        os.system("rm %s" % private_file)
+        os.system("rm %s" % key_file)
         raise YunohostError('no_internet_connection')
     if r.status_code != 201:
+        os.system("rm %s" % private_file)
+        os.system("rm %s" % key_file)
```
Psycojoker: You probably want a bunch of "-f" on those.

alexAubin: Indeed :+1:

alexAubin:
Tested the various pieces on a real-life server + in a dev machine, seem to work as expected, yolomergin'

Psycojoker:
I realized that this won't allow us to know when is the last time a dyndns domain tried to update itself. It shouldn't be a problem but this will close the possibility of automatically removed them (the tool old ones) if we try to do that one day.

I'm not expecting this to be a problem but we just have to keep that in mind.

alexAubin:
Hm yea I was wondering about that the other day ... still there could be some brutal way of doing this like removing all domains that points to an IP that doesnt ping (or doesnt answer to http(s) on /yunohost/sso ...) And if those were "actual" instance the dyndns should kick in a few minutes after and refresh the IP

Psycojoker:
Or we could just add a new mechanism in YunoHost to randomly ping once per week saying "I'm alive!" to dyndns.

PR #663, opened 2019-02-24
[enh] Optimize app setting helpers

alexAubin:
## The problem

https://github.com/YunoHost/issues/issues/1299

On an ARM board like RPi2, `yunohost app setting` takes 6\~7 second to complete... Considering that an app usually interacts with 5\~10 settings during an install, you may loose a whole damn minute just getting or setting settings... 

Instead, doing this with a "small" python snippet can divide this time by ~10.

## Solution

Have a small helper to manipulate the setting yaml files. Turns out it's not straightforward as there are some special cases (c.f. `redirected_urls` and `redirected_regex`) to handle to truly reproduce the `yunohost app setting` behavior.

## PR Status

Partially tested on my machine

## How to test...

Load and use the `ynh_app_setting_{get,set,delete}` in various case ;)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/setting
+++ b/data/helpers.d/setting
@@ -75,9 +75,11 @@ else:
     if action == "delete":
         if key in settings:
             del settings[key]
-    else:
+    elif action == "get":
         if key in ['redirected_urls', 'redirected_regex']:
             value = yaml.load(value)
+    else:
+        raise ValueError("action should either be get, set or delete")
     settings[key] = value
     with open(setting_file, "w") as f:
         yaml.safe_dump(settings, f, default_flow_style=False)
--- a/data/helpers.d/setting
+++ b/data/helpers.d/setting
@@ -75,12 +75,12 @@ else:
     if action == "delete":
         if key in settings:
             del settings[key]
-    elif action == "get":
+    elif action == "set":
         if key in ['redirected_urls', 'redirected_regex']:
             value = yaml.load(value)
+        settings[key] = value
     else:
         raise ValueError("action should either be get, set or delete")
-    settings[key] = value
     with open(setting_file, "w") as f:
         yaml.safe_dump(settings, f, default_flow_style=False)
 EOF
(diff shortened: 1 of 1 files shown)


keomabrun:
Just out of curuosity, why is the `yunohost app setting` command taking so long in the first place?

alexAubin:
That's a good question, but I'm pretty sure it's more about `yunohost` commands in general than just this one specifically... I guess loading the whole moulinette + yunohost (all doing a lot of parsing + module loading) is long ... Would be nice to run some profiling some day to see if there's anything obvious we could do

Psycojoker:
> Would be nice to run some profiling some day to see if there's anything obvious we could do

Already did and ... nothing really obvious came out of it :( (I remember doing some small optimisation but it wasn't dramatic)

Most of the time is spend in building the argument parsing tool with argparse (it's not even about the loading of the yaml/pickled file iirc) so I didn't end up finding something obvious to fix :[

Psycojoker:
I think the ideal solution would be to do a "yunolib" that extract auth independent business logic out of this code base and ship small CLI tools instead of inlining python into bash.

In the meantime that looks good enough I guess.

data/helpers.d/setting
```diff
@@ -50,5 +50,36 @@ ynh_app_setting_delete() {
...
+with open(setting_file) as f:
+    settings = yaml.load(f)
+if action == "get":
+    if key in settings:
+        print(settings[key])
+else:
+    if action == "delete":
+        if key in settings:
+            del settings[key]
+    else:
```
Psycojoker: What is this else branch? That's not really obvious from here :x

alexAubin: Should be fixed via 4a8563e

alexAubin:
Proposing to merge soon

PR #664, opened 2019-02-25
Add min version for all helpers

maniackcrudelis:
## The problem

When packaging, it can be difficult to evaluate the minimal version of YunoHost you should require.

## Solution

Add the minimal version required for each helpers.
Unfortunately, it will be hard to maintain...

## PR Status

Ready for reviews

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Josue-T:
It will quite difficult but I think that it's also important to know the change. By example for the nginx helper, we added the argument to pass some other variable to change. I think that this information is as important as the version when the helper was interoduced. For getops too.

Maybe we can write the required version since the last change...

alexAubin:
>Maybe we can write the required version since the last change...

Sooo do we implement this or merge as is ? Personally I'd say if those non-backward compatible changes are from before 3.0 we can ignore it :/

maniackcrudelis:
It's more changes after the version 3.
But I think it's already going to be hard to maintain those version numbers...

alexAubin:
Sooo uuuh, to merge or not to merge, that is the wondering ̣¿:o|

maniackcrudelis:
Let's merge in 3 days ;)

PR #667, opened 2019-02-26
[fix] adding a new test in case the info.json file exists but not the tar and vice versa

evilcel3ri:
## The problem

[this issue](https://github.com/YunoHost/issues/issues/1276)

## Solution

Adding one more test but it isn't the best one

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -2189,10 +2189,13 @@ def backup_info(name, with_details=False, human_readable=False):
 
     """
     archive_file = '%s/%s.tar.gz' % (ARCHIVES_PATH, name)
-
+    info_file = "%s/%s.info.json" % (ARCHIVES_PATH, name)
     # Check file exist (even if it's a broken symlink)
     if not os.path.lexists(archive_file):
         raise YunohostError('backup_archive_name_unknown', name=name)
+
+    if not (os.path.lexists(archive_file) and os.path.exists(info_file)):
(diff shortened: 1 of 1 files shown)


alexAubin:
Uh in fact I think that's not the real issue. It turns out that YunoHost is supposed to be able to work even if there's no `.info.json` file directly associated with the archive - because this file is actually *also* in the archive itself. This is what is handled here : https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/backup.py#L2213

I think the original issue lies in here : https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/backup.py#L2158-L2171

The original issue is in fact not trivial to reproduce but I think @e-lie somehow renamed an archive while keeping the `.tar.gz` in the name. So you can reproduce the issue like this : 
- creating a new backup using `yunohost backup create`
- going to `/home/yunohost.backup/archives`
- Rename the archive like this `mv your_archive.tar.gz your_archive.tar.gz.bkp` (it's important to keep `.tar.gz` in the archive name somewhere ;))
- Run `yunohost backup list`
- If it's still not clear what's happening, try to `touch foo.tar.gz.yolo.bar` and you should see it's trying to look for a `foo.tar.gz` archive because of that `f.rindex('.tar.gz')` :P

PR #668, opened 2019-03-01
Remove nistpxxx curve

Josue-T:
## The problem

The website https://safecurves.cr.yp.to/ say that the curve nistp256, and nistp384 is not secure. 

From some source, the NSA had backdoored Dual the curve NIST P-256.

## Solution

Remove all of theses curves

## PR Status

Not tested, but should work, with some other curves

## How to test

Just update your ssh config

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/conf_regen/03-ssh
+++ b/data/hooks/conf_regen/03-ssh
@@ -16,7 +16,7 @@ do_pre_regen() {
     # do not listen to IPv6 if unavailable
     [[ -f /proc/net/if_inet6 ]] && ipv6_enabled=true || ipv6_enabled=false
 
-    ssh_keys=$(ls /etc/ssh/ssh_host_{ed25519,rsa,ecdsa}_key 2>/dev/null || true)
+    ssh_keys=$(ls /etc/ssh/ssh_host_{ed25519,rsa}_key 2>/dev/null || true)
 
     # Support legacy setting (this setting might be disabled by a user during a migration)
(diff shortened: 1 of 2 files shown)


alexAubin:
Thanks for providing a source for this.

Doing some research on my side, I still find that this is all very much political speculation about those ...

c.f. https://wiki.archlinux.org/index.php/SSH_keys (ECDSA section) and https://news.ycombinator.com/item?id=7597653 which summarize the situation like this : 

> there are good reasons to favor Curve25519 and other implementation-friendly elliptic curves (namely, they are faster, and they are fewer ways of shooting yourself in the foot if you implement them), but "NIST curves are backdoored" is not a very serious one. 

So it looks a lot like "oh *somebody* says the NSA *might* have been involved at some point in the design of these so we shouldn't use them at all" ... But from what I read, the most compelling thing here is the difficulty of implementing the curves in a safe way ... and that still doesn't convince me that much, because I tend to think that if we don't trust the implementations in the lib we use, then we have bigger issues ...

And also clearly I'm still reluctant to have our own custom security cooking instead of sticking to the recommendation from a trustable third-party ... To me the weakest point in our SSH config right now is the fact that login happens using passwords instead of public key authentication. Which ultimately boils down to the UX issue of "how do we get people to set up public key auth". https://imgs.xkcd.com/comics/security.png

zamentur:
I close it, feel free to reopen this topic during the meetings if you really want this to be included

PR #669, opened 2019-03-04
[enh] allow 'display_text' ~fake~ argument in manifest.json

Psycojoker:
## The problem

A very simple idea I had since quite some time: allow to add a way in the argument list of an app to have an argument that simply display text and fetch no value as way to allow app packagers to display information for the user.

The usage is very simple, simply had this kind of argument in your manifest:

```json
            {
                "name": "some_text",
                "type": "display_text",
                "text": "This is some text that is going to be displayed.\n\nAnother line."
            },
```

## PR Status

Working. Need the equivalent for the admin UI once it's agreed on principle.

## How to test

Create an empty app and had this kind of argument to the manifest and had an empty shell script and run the installation.

Example:

2019-03-04-031132_714x243_escrotum

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -2202,6 +2202,11 @@ def _parse_action_args_in_yunohost_format(args, action_args, auth=None):
         if arg_type == 'boolean':
             arg_default = 1 if arg_default else 0
 
+        # do not print for webadmin
+        if arg_type == 'display_text' and msettings.get('interface') != 'api':
+            print(arg["text"])
+            continue
+
         # Attempt to retrieve argument value
         if arg_name in args:
             arg_value = args[arg_name]


Psycojoker:
> Out of curiosity, what's the difference between print and sys.stdout.write ?

I had the feeling that `sys.stdout.write` would interpret the "\n" (and not "print") but now thinking back about it I'm not sure it's actually needed... my brain probably farted ^^'

Psycojoker:
My brain did farted, I've removed it x)

Psycojoker:
Ping @YunoHost/apps for principle approval/review.

maniackcrudelis:
That's probably going to be useful!

alexAubin:
Yolomergin'

Psycojoker:
To be confirmed but we probably need to related PR for the webadmin before the release '^'

Psycojoker:
I can do that tomorrow I think.

PR #670, opened 2019-03-04
[enh] Check that required services are up before running app install and upgrade

alexAubin:
## The problem

see this kind of situation : https://forum.yunohost.org/t/nextcloud-erreur-502-et-mise-a-jour-impossible-apres-migration-yunohost-3-4-2-4/7204/11 - an app upgrade miserably failed trying to reload php7.0-fpm which was most probably down before the whole action even started and this could be avoided.

## Solution

Add a check that some services are up before running app install or app upgrades, according to the info given in the manifest. Unfortunately the manifest does not seem to be available during restore (or it's just not loaded/used ?) so it's not straigtforward to also add it here

## PR Status

Tested for app install

## How to test

E.g. stop php7.0-fpm and try installing wordpress

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -2574,6 +2576,31 @@ def unstable_apps():
...
+
+    # Some apps use php-fpm or php5-fpm which is now php7.0-fpm
+    def replace_alias(service):
+        if service in ["php-fpm", "php5-fpm"]:
+            return "php7.0-fpm"
+        else:
+            return service
+    services = [replace_alias(s) for s in services]
+
+    # We only check those, mostly to ignore "custom" services
```
Psycojoker: Shouldn't we only checked already known services only instead? I'm not sure. That will also filtered out not already installed services and, maybe?, could prevent other errors.

Psycojoker:
"Funny", post-installed failed on my vm because slapd wasn't running, maybe that could be a good place to do that check too, no?

src/yunohost/app.py
```diff
@@ -2574,6 +2576,31 @@ def unstable_apps():
...
 
 
+def _check_services_status_for_app(services):
+
+    logger.debug("Checking that required services are up and running...")
+
+    # Some apps use php-fpm or php5-fpm which is now php7.0-fpm
+    def replace_alias(service):
+        if service in ["php-fpm", "php5-fpm"]:
+            return "php7.0-fpm"
```
Josue-T: Not directly linked to this PR, but is it a good idea to hardcode the PHP version. We know that at every debian release the php version change, so maybe could we think about a universal solution about that ?

alexAubin: Meh yea but I can't really think of any simple universal solution ... But it's not like if grepping occurences of "php5" or "php7" and replacing those was the hard part :/

alexAubin:
So meh ... I dunno what to do with this ... the idea was mostly that I saw a few times happening that nginx or php was down before trying to install an app or upgrade one, and it miserably failed ...

We can try to generalize to all services but meh I really mainly care about php-fpm and mysql because those are the ones "likely" to be broken :|

alexAubin:
Ok guyz, I'm gonna merge this as is because I mainly want to stop getting those damn failed install / upgrade just because php-fpm ain't running ... we can still iterate later to refine this.

PR #671, opened 2019-03-07
[fix] Reject app password if they contains { or }

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1319

## Solution

Reject app passwords that contain { or } ...

## PR Status

Tested and working

## How to test

Try to install for example lstu_ynh and put { or } in password

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Yolomergin'

randomstuff:
As discussed in the bug entry, I'd day the proper way to fix this is to avoid shell command injections altogether.

PR #672, opened 2019-03-07
[fix] Backup fails because output directory not empty

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1308 which sometime breaks upgrade for silly reasons

## Solution

Implement the 'FIXME' by attempting to unmount / clean the directory.

While fixing this, I realized that the function `_recursive_umount` probably had a buggy behavior / return value ... Dunno if that explains why the original issue happened in the first place, but could be related.

## PR Status

Tested and works okay

## How to test

```bash
# Initialize dummy file
mkdir /var/toto
echo "toto" > /var/toto/toto

# Initialize dummy tmp directory
cd /home/yunohost.backup/tmp
mkdir yoloswag; cd yoloswag

# Mount dummy stuff in the tmp directory
mkdir toto
mount -o ro,bind /var/toto ./toto

# Try to create a backup
cd
yunohost backup create -n yoloswag --debug
```

Should see in the debug output (at the beginning) that it complains the directory ain't empty and attempts to clean it

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -326,10 +326,19 @@ class BackupManager():
         if not os.path.isdir(self.work_dir):
             filesystem.mkdir(self.work_dir, 0o750, parents=True, uid='admin')
         elif self.is_tmp_work_dir:
-            logger.debug("temporary directory for backup '%s' already exists",
+
+            logger.debug("temporary directory for backup '%s' already exists... attempting to clean it",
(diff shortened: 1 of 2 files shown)


alexAubin:
Yolomergin'

PR #673, opened 2019-03-08
Fix backup tests

alexAubin:
## The problem

Backup tests failed for two different reason ...
- a weird edge case about link creation during the backup ... files were actually on different device but there seem to be no way to detect this ... I suspect this is because of running on a weird LXC setup with encrypted hard drive inside some /dev/mapper stuff such that the real "logical" (?) devices are hidden ...
- some changes in the pytest version or idk, such that the way to access function / tests markers broke

## Solution

Fix those

## PR Status

Tested on my machine

## How to test

Run the backup/restore tests

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Planning to merge soon because that's mainly technical stuff that is only likely to affect people running tests...

PR #677, opened 2019-03-11
[enh] bash completion for yunohost cli

ChristopheVuillot:
Added a python script (yunohost_completion.py) which generates a bash completion file for the yunohost command based on yunohost.yml, in data/actionsmap

Added the output of the script in data/bash-completion.d/yunohost_completion

This is probably not the correct place for the script and the generation should
be done at some other time and place also.

## The problem
missing bash completion for yunohost command

## Solution
a python script which generates the file from yunohost.yml

## PR Status
[WIP]

## How to test
At least test that the generated file is syntactically correct ?

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost_completion.py
+++ b/data/actionsmap/yunohost_completion.py
@@ -1,0 +1,84 @@
+"""
+Simple automated generation of a bash_completion file
+for yunohost command from the actionsmap.
+
+Generates a bash completion file assuming the structure
+`yunohost domain action`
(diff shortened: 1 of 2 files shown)


ChristopheVuillot:
I think I PR to the wrong branch, I'm closing it for this reason

PR #678, opened 2019-03-11
[enh] bash completion for yunohost cli

ChristopheVuillot:
Added a python script (yunohost_completion.py) which generates a bash completion file for the yunohost command based on yunohost.yml, in data/actionsmap

Added the output of the script in data/bash-completion.d/yunohost_completion

This is probably not the correct place for the script and the generation should
be done at some other time and place also.

## The problem
missing bash completion for yunohost command

## Solution
a python script which generates the file from yunohost.yml

## PR Status
[WIP]

## How to test
Source the generated file:
`source yunohost_completion`
then you should be able to auto complete

To regenerate the file:
`python yunohost_completion.py`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
Again, thx a lot for that work :)

ChristopheVuillot:
Very happy to contribute :).

Thanks for adding the labels, it indeed needs reviews and tests. I don't know where the script should be put and how it should be run.

For tests, I guess, at least checking that the generated file is syntactically correct could be a start ?

alexAubin:
Hey there ! Finally taking the time to have a first look at this ! Thank so much for this, I think many people will find it pleasing to have auto completion !

I'm thinking maybe we can add something to be ran during debian's build to automatically generate the file ?

Also since you flagged this as "wip" is there anything else you were thinking about implementing ?

ChristopheVuillot:
Hey, thanks for having a look.
For the [wip] flag I wasn't sure if I needed to put it, I just put it there because this cannot be merged as is, but as you say something needs to be added to run the script. Although I don't know how debian's build works so would need some help on that.
As far as the script is concerned I don't intend to add anything else for now.

alexAubin:
Alrighty :wink: 

Indeed the whole Debian build thing ain't trivial at all ! I will have a look soon™

alexAubin:
Did a few tweaks, still need to test that running the script during debian builds does work

ChristopheVuillot:
Just fixed some names for coherence.

alexAubin:
Sooooo tested the debian build and it seems to work okay :ok_hand: 

That looks good for me, planning to merge soon

Note though that not all setups load the bash-completion thing (I believe some commands need to be ran inside the .bashrc of root/admin or something like this ...). There's no trivial solution for this tho ... But we could imagine something to add those command during the initial install.

ChristopheVuillot:
Thank you! I think that all the bash-completions are in principle loaded at boot, but otherwise one just has to source the thing once. So maybe putting it in .bashrc is a bit too much but just source the completion file during the initial install may be better ?

alexAubin:
Uuuh I doubt that bash completion is loaded "globally" ? To me it's something that must be sourced by the shell one way or another ...

On my machine, I see there's a `/etc/profile.d/bash_completion.sh` which is sourced by `/etc/profile`, which seems to be something like a default thing loaded by all shell when they start ... But my point was mainly that probably not all setups have this enabled by default ... Maybe it's just a matter of installing a bash-completion package, idk :s

ChristopheVuillot:
Oh yes you are right, sorry, I got confused. I don't know what is the default for debian actually. Probably people that want to use completion in general have already enabled it.

alexAubin:
So in fact I noticed that there is indeed a `bash-completion` package for debian which actually is in recommended packages for yunohost : 

```
> aptitude why bash-completion
i   yunohost Recommends bash-completion
```

So modulo the funny setup who do not install recommended packages (that should be solved, c.f. discussion on https://github.com/YunoHost/issues/issues/1320 ... ) the bash completion should be enabled on all setups I think o/

PR #680, opened 2019-03-13
[fix] Delete properly group as well as user

maniackcrudelis:
## The problem

https://github.com/YunoHost-Apps/wordpress_ynh/issues/57
That's not the first time we have this problem, a group isn't deleted because there's other users in that group.

## Solution

`delgroup` will remove a group, even if not empty. Also, having it's own condition, running the helper will remove the group even if the user doesn't exist anymore.

## PR Status

~Need to be tested...~ Tested.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/user
+++ b/data/helpers.d/user
@@ -145,7 +145,5 @@ ynh_system_user_delete () {
     then
 		echo "Remove the group $username" >&2
 		delgroup $username
-	else
-		echo "The group $username was not found" >&2
     fi
 }
--- a/data/helpers.d/user
+++ b/data/helpers.d/user
@@ -63,6 +63,21 @@ ynh_system_user_exists() {
     getent passwd "$username" &>/dev/null
 }
 
+# Check if a group exists on the system
+#
+# usage: ynh_system_group_exists --group=group
+# | arg: -g, --group - the group to check
+ynh_system_group_exists() {
+    # Declare an array to define the options of this helper.
+    local legacy_args=g
+    declare -Ar args_array=( [g]=group= )
+    local group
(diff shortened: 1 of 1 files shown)


data/helpers.d/user
```diff
@@ -116,11 +131,21 @@ ynh_system_user_delete () {
...
 		echo "The user $username was not found" >&2
     fi
+
+    # Check if the group exists on the system
+    if ynh_system_group_exists "$username"
+    then
+		echo "Remove the group $username" >&2
+		delgroup $username
+	else
+		echo "The group $username was not found" >&2
```
alexAubin: Do we really want to display this warning, considering that it should be okay that the group is not found (in the nominal case, it should have been deleted with the previous `deluser`)

maniackcrudelis: Nope, `deluser` will not remove the group anymore.
`userdel` would have done it ;)

alexAubin: Uhwokay x)

maniackcrudelis: From the man of `deluser`
> If called with one non-option argument and without the --group  option, deluser will remove a normal user.

deluser removes the user **and** the group...
You were right ;)

Anyway... if the group isn't empty, the issue at the origin of this PR, delgroup will be called and the group removed.
The user I added to this group wasn't removed, only the group. Everything's fine.

alexAubin: :+1:

PR #681, opened 2019-03-13
[enh] reload-or-restart instead of reload

maniackcrudelis:
## The problem

https://github.com/YunoHost-Apps/wordpress_ynh/issues/57
For unknown reasons, php7 can be sometime stopped, the reload fails then because the service is not active.

## Solution

Use reload-or-restart instead of reload, so it would try to reload, and if the service isn't running, it will restart instead.

## PR Status

Not tested, is it really needed ?

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/system
+++ b/data/helpers.d/system
@@ -59,7 +59,7 @@ ynh_get_debian_release () {
 # Start (or other actions) a service,  print a log in case of failure and optionnaly wait until the service is completely started
 #
 # usage: ynh_systemd_action [-n service_name] [-a action] [ [-l "line to match"] [-p log_path] [-t timeout] [-e length] ]
-# | arg: -n, --service_name= - Name of the service to reload. Default : $app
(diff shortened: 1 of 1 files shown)


kay0u:
The log expected by the argument `--line_match` could be differents if is it's a reload or a restart. What about a reload-or-restart? That could introduce an bug I think.

maniackcrudelis:
What do you mean ? That the line to match would be different if you restart or reload the service?
And so that it could be a problem if you ask for a reload and finally you have a restart ?

Well, first I don't think you would use --line_match with reload, reload isn't a long process.
Except in case you expect the reload to fail.
Anyway, in case that would happen, it will simply reach the time out. No big deal.

PR #683, opened 2019-03-16
Escape $ in getopts

maniackcrudelis:
## The problem

As found by JimboJoe here, https://github.com/YunoHost/yunohost/pull/646#issuecomment-473448470, $ have to be escaped as well as double quotes.

## PR Status

Ready to be review

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -203,7 +203,7 @@ ynh_handle_getopts_args () {
 				# Escape double quote to prevent any interpretation during the eval
 				arguments[$i]="${arguments[$i]//\"/\\\"}"
 				# Escape $ as well to prevent the string following it to be seen as a variable.
-				all_args[$i]="${all_args[$i]//$/\\\$}"
+				arguments[$i]="${arguments[$i]//$/\\\$}"
 
 				# Store each value given as argument in the corresponding variable
 				# The values will be stored in the same order than $args_array
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -152,6 +152,8 @@ ynh_handle_getopts_args () {
 								fi
 								# Escape double quote to prevent any interpretation during the eval
 								all_args[$i]="${all_args[$i]//\"/\\\"}"
+								# Escape $ as well to prevent the string following it to be seen as a variable.
+								all_args[$i]="${all_args[$i]//$/\\\$}"
 
 								eval ${option_var}+=\"${all_args[$i]}\"
 								shift_value=$(( shift_value + 1 ))
@@ -193,6 +195,8 @@ ynh_handle_getopts_args () {
 
 				# Escape double quote to prevent any interpretation during the eval
 				arguments[$i]="${arguments[$i]//\"/\\\"}"
+				# Escape $ as well to prevent the string following it to be seen as a variable.
+				all_args[$i]="${all_args[$i]//$/\\\$}"
 
 				# Store each value given as argument in the corresponding variable
 				# The values will be stored in the same order than $args_array
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -155,6 +155,13 @@ ynh_handle_getopts_args () {
 								# Escape $ as well to prevent the string following it to be seen as a variable.
 								all_args[$i]="${all_args[$i]//$/\\\$}"
 
+								# For the record.
+								# We're using eval here to get the content of the variable stored itself as simple text in $option_var...
+								# Other ways to get that content would be to use either ${!option_var} or declare -g ${option_var}
+								# But... ${!option_var} can't be used as left part of an assignation.
+								# declare -g ${option_var} will create a local variable (despite -g !) and will not be available for the helper itself.
+								# So... Stop fucking arguing each time that eval is evil... Go find an other working solution if you can find one!
+
 								eval ${option_var}+=\"${all_args[$i]}\"
 								shift_value=$(( shift_value + 1 ))
 							fi


maniackcrudelis:
If possible, this should be merged in the testing before an official stable release, as a fix for the current testing.

data/helpers.d/getopts
```diff
@@ -152,6 +152,15 @@ ynh_handle_getopts_args () {
...
 								all_args[$i]="${all_args[$i]//\"/\\\"}"
+								# Escape $ as well to prevent the string following it to be seen as a variable.
+								all_args[$i]="${all_args[$i]//$/\\\$}"
+
+								# For the record.
+								# We're using eval here to get the content of the variable stored itself as simple text in $option_var...
+								# Other ways to get that content would be to use either ${!option_var} or declare -g ${option_var}
+								# But... ${!option_var} can't be used as left part of an assignation.
+								# declare -g ${option_var} will create a local variable (despite -g !) and will not be available for the helper itself.
+								# So... Stop fucking arguing each time that eval is evil... Go find an other working solution if you can find one!
```
JimboJoe: I'm in favour of a disclaimer, but maybe avoid rude words...? :wink:

data/helpers.d/getopts
```diff
@@ -193,6 +202,8 @@ ynh_handle_getopts_args () {
 
 				# Escape double quote to prevent any interpretation during the eval
 				arguments[$i]="${arguments[$i]//\"/\\\"}"
+				# Escape $ as well to prevent the string following it to be seen as a variable.
+				all_args[$i]="${all_args[$i]//$/\\\$}"
```
JimboJoe: Shouldn't you escape `arguments` here...?

data/helpers.d/getopts
maniackcrudelis: I like things to be clear...
And I'm not usually smooth and politically correct...
However, ~I don't care...~ `[Politically mode ON]` *I'm not feeling absolutely concerned by what should be done with the text of this disclaimer and the way things should be said to be understood.*

data/helpers.d/getopts
maniackcrudelis: Oups indeed, bad copy paste !

data/helpers.d/getopts
randomstuff: Sill I find eval is evil :)

randomstuff:
See #685 for an alternative -- still using eval ;)

randomstuff:
As discussed, elsewhere backticks should probably be escaped as well.

maniackcrudelis:
https://github.com/YunoHost/yunohost/pull/646#issuecomment-473653473

randomstuff:
backslashes are not escaped as well.

maniackcrudelis:
'\' has been tested without problem.
Are you really sure it doesn't work !?
To be honest, you have proposed a lot of solutions so far, which were, for most of them, not working. So, I could believe that I missed something, but bring real examples of not working context.

randomstuff:
`\` can be passed verbatim in double quotes unless it follow some metacharacter. In this case, the backslash quotes the metacharacters:

```
+ ynh_debug -m '\$(ls)'
+ set +x
+ set +x
[DEBUG]ackend
debug
filesystem
getopts
ip
mysql
network
nodejs
package
print
psql
setting
string
system
user
util
```

maniackcrudelis:
???

```
ynh_debug -m '\$(ls)'
+ ynh_debug -m '\$(ls)'
+ set +x
+ set +x
[DEBUG] $(ls)
```
```
ynh_debug -m "\$(ls)"
+ ynh_debug -m '$(ls)'
+ set +x
+ set +x
[DEBUG] github
install_ynh-dev.sh
media
mount_directory.sh
move_to_test4.sh
move_to_test5.sh
package_check
upgradeVM.sh
use_ynh-dev.sh
```

I'm seriously fed up with this discussion...

randomstuff:
@maniackcrudelis, are you on 652edc20?

randomstuff:
The one works both on the branch and before your change:

```
$ ynh_debug -m '\";ls;\"'
backend  debug  filesystem  getopts  ip  mysql  network  nodejs  package  print  psql  setting  string  system  testit.sh  user  utils
bash: \ : commande introuvable
+ set +x
[DEBUG] \
```

maniackcrudelis:
We have many errors on the current testing because of that error.
Could we have at least one more review and merge it quickly.

randomstuff:
If you don't like the other option, manually escaping backtick, backslash, double quotes and dollar should be fine (I thinks that's all the characters which needs to be escaped):

```sh
all_args[$i]="${all_args[$i]//\\/\\\\}"
all_args[$i]="${all_args[$i]//\`/\\\`}"
all_args[$i]="${all_args[$i]//\"/\\\"}"
all_args[$i]="${all_args[$i]//$/\\\$}"
```

randomstuff:
I *think* it's actually much cleaner/simpler/easier to do it like this:

```sh
eval ${option_var}+='"${all_args[$i]}"'
```

(without any escape lines)

maniackcrudelis:
Some tests later, it doesn't looks like it breaks anything (this time...)
**But** as this fucking PR **WORKS AS IT IS**.
Let's merge it and not break anything more !

Then, as your code doesn't seem anymore to break anything, I could almost think you have tested it this time !
I think it could be merged in the next unstable, after the next stable release.
So it could be tested in all apps and be sure it really works.

randomstuff:
> Some tests later, it doesn't looks like it breaks anything (this time...) […] Then, as your code doesn't seem anymore to break anything, I could almost think you have tested it this time !

Well I tested the original version of the PR and it was working on all the tests I did. It's absolutely unsurprising if you know better than me how to properly test this: if you found something broken it would be a lot more constructive to actually point out what was broken.

This is actually the first time you claim something was broken in the previous version of my PR: until then you only said you were afraid something might break.

Your proposal of merging your quick fix first and take some time to test if the complete fix introduces any regression looks OK to me.

However, I don't think that ` ad748a75f62490386c7ca97b92191d3ee2d3fa0d` is any useful and I'd say it should be either removed or reworded in order to remove all the angriness.

maniackcrudelis:
Reading quickly all discussions so far, I explain myself sometime, some others I didn't.
Facts are, I spent a lot of time trying your "fixes" which wasn't fixing anything or even working.
Except the first version of #685, because I have real work to do instead of wasting my time trying pseudo fix.

You want to change the behavior of this line, you have to prove it worth it, and above all, you have to be sure that works.
I've wasted a lot of time with this PR, for a fucking simple "$", which was already fixed at the beginning, as soon as JimboJoe found it.

So yes, I'm angry, because I hate wasting my time for nothing. Also, bad luck, I'm the quick to anger guy in this community. I'm not politically correct, I don't want to be and I don't care.

PR #685, opened 2019-03-16
Avoid having to shell-escape arguments in ynh_handle_getopts_args

randomstuff:
## The problem

Shell characters are not all escaped in `eval` call.

## Solution

Escape shell metacharacters with a dedicated function.

Alternative to #683.

## PR Status

...

## How to test

```sh
#!/bin/sh

. ./getopts

function my_helper()
{
    declare -Ar args_array=( [a]=arg1= [b]=arg2= [c]=arg3 )
    local arg1
    local arg2
    local arg3
    ynh_handle_getopts_args "$@"

    echo "Inner"
    echo arg1=$arg1
    echo arg2=$arg2
    echo arg2=$arg3
    echo
}

my_helper --arg1 '$toto`ls`' -b "'" -c
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -152,19 +152,8 @@ ynh_handle_getopts_args () {
 									# If there's already another value for this option, add a ; before adding the new value
 									eval ${option_var}+="\;"
 								fi
-								# Escape double quote to prevent any interpretation during the eval
-								all_args[$i]="${all_args[$i]//\"/\\\"}"
-								# Escape $ as well to prevent the string following it to be seen as a variable.
-								all_args[$i]="${all_args[$i]//$/\\\$}"
 
-								# For the record.
-								# We're using eval here to get the content of the variable stored itself as simple text in $option_var...
-								# Other ways to get that content would be to use either ${!option_var} or declare -g ${option_var}
-								# But... ${!option_var} can't be used as left part of an assignation.
-								# declare -g ${option_var} will create a local variable (despite -g !) and will not be available for the helper itself.
-								# So... Stop fucking arguing each time that eval is evil... Go find an other working solution if you can find one!
-
-								eval ${option_var}+=\"${all_args[$i]}\"
+								eval ${option_var}+='"${all_args[$i]}"'
 								shift_value=$(( shift_value + 1 ))
 							fi
 						done
@@ -202,14 +191,7 @@ ynh_handle_getopts_args () {
 				# The variable name will be stored in 'option_var'
 				local option_var="${args_array[$option_flag]%=}"
 
-				# Escape double quote to prevent any interpretation during the eval
-				arguments[$i]="${arguments[$i]//\"/\\\"}"
-				# Escape $ as well to prevent the string following it to be seen as a variable.
-				arguments[$i]="${arguments[$i]//$/\\\$}"
-
-				# Store each value given as argument in the corresponding variable
-				# The values will be stored in the same order than $args_array
-				eval ${option_var}+=\"${arguments[$i]}\"
+				eval ${option_var}+='"${arguments[$i]}"'
 			done
 			unset legacy_args
 		else
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -153,6 +153,13 @@ ynh_handle_getopts_args () {
 									eval ${option_var}+="\;"
 								fi
 
+								# For the record.
+								# We're using eval here to get the content of the variable stored itself as simple text in $option_var...
+								# Other ways to get that content would be to use either ${!option_var} or declare -g ${option_var}
+								# But... ${!option_var} can't be used as left part of an assignation.
+								# declare -g ${option_var} will create a local variable (despite -g !) and will not be available for the helper itself.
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
I'm not in favor of this PR, the issue is fixed by https://github.com/YunoHost/yunohost/pull/683. We do not need to complicate this helper if not absolutely needed.

randomstuff:
I actually find it *simpler* than the alternative:

```sh
# Escape double quote to prevent any interpretation during the eval
all_args[$i]="${all_args[$i]//\"/\\\"}"
# Escape $ as well to prevent the string following it to be seen as a variable.
all_args[$i]="${all_args[$i]//$/\\\$}"

eval ${option_var}+=\"${all_args[$i]}\"
```

And it is more correct: the implementation in the snippet does not properly handles backticks and (AFAIU) backslashes which could have usability and security implications.

maniackcrudelis:
Calling another function means passing parameters again, with risk of interpretation, using simple quote could have consequences.
**It works**. I'm don't want to change again because maybe, in a very specific edge case not seen yet, it could be better.
This helper isn't intended to be used by admin, only by helpers used inside apps scripts.

randomstuff:
Summary of the previous episodes

The current version of `ynh_handle_getopts_args` does not handles backslashes and backticks correctly.

For reference, the variables are currently assigned values with:

```sh
all_args[$i]="${all_args[$i]//\"/\\\"}"
all_args[$i]="${all_args[$i]//$/\\\$}"
eval ${option_var}+=\"${all_args[$i]}\"
```

It's clear from the code that backticks and backslashes are not escaped properly, as examplified with:

```sh
ynh_debug -m '`ls`'
ynh_debug -m '\$(ls)'
 ynh_debug -m '\";ls;\"'
```

As a consequence, backslashes and backticks cannot be used reliably as argument for any function using `ynh_handle_getopts_args`: it might be a problem for passwords for example.

Moreover, this could be turn into a vulnerability issue: if any application calls any of these functions with untrusted input, it could be exploited to inject shell commands.

maniackcrudelis:
Eventually, a **real** issue...

Real examples would have been better though...
```
echo '`ls`' >&2
ynh_debug -m '`ls`'

echo '\$(ls)' >&2
ynh_debug -m '\$(ls)'

echo '\";ls;\"' >&2
ynh_debug -m '\";ls;\"'
```
With the current unstable version of getopts:
```
Warning: `ls`
Warning: [DEBUG] _common.sh
Warning: backup
Warning: change_url
Warning: install
Warning: remove
Warning: restore
Warning: upgrade
Warning: \$(ls)
Warning: [DEBUG] \_common.sh
Warning: backup
Warning: change_url
Warning: install
Warning: remove
Warning: restore
Warning: upgrade
Warning: \";ls;\"
Warning: /usr/share/yunohost/helpers.d/getopts: line 171: \: command not found
```

With ` and \ escaped as the other characters.
```
Warning: `ls`
Warning: [DEBUG] `ls`
Warning: \$(ls)
Warning: [DEBUG] \$(ls)
Warning: \";ls;\"
Warning: [DEBUG] \";ls;\"
```

With `eval ${option_var}+='"${all_args[$i]}"'` without escaping characters
```
Warning: `ls`
Warning: [DEBUG] `ls`
Warning: \$(ls)
Warning: [DEBUG] \$(ls)
Warning: \";ls;\"
Warning: [DEBUG] \";ls;\"
```

maniackcrudelis:
phpmyadmin is currently failing on unstable and testing because of this [line](https://github.com/YunoHost-Apps/phpmyadmin_ynh/blob/master/scripts/install#L75) where `*.*` are interpreted as a regex.
Using the current fix correct that issue for phpmyadmin.

Can't see the status of the unstable CI for the moment...
This fix should be tested on more apps.
So, we could merge in unstable to try it.

data/helpers.d/getopts
```diff
@@ -153,6 +153,13 @@ ynh_handle_getopts_args () {
 									eval ${option_var}+="\;"
 								fi
 
+								# For the record.
+								# We're using eval here to get the content of the variable stored itself as simple text in $option_var...
+								# Other ways to get that content would be to use either ${!option_var} or declare -g ${option_var}
+								# But... ${!option_var} can't be used as left part of an assignation.
+								# declare -g ${option_var} will create a local variable (despite -g !) and will not be available for the helper itself.
+								# So... Stop fucking arguing each time that eval is evil... Go find an other working solution if you can find one!
+
```
randomstuff: Still I find eval is evil :grin:

alexAubin:
Soooo, go for merge then ?

maniackcrudelis:
Rather wait for the CI to be available https://ci-apps-unstable.yunohost.org/ci/

maniackcrudelis:
Actually... It's a false start from the CI on testing.
An upgrade of dovecot has triggered the CI...

The CI will be available in a little while

alexAubin:
Following discussion in the chat : mergin' !

PR #687, opened 2019-03-17
Dump ynh log if an app script fails

maniackcrudelis:
## The problem

While playing with ynh_debug, I thought it's too bad to have to use that helper to know why an app script has crashed.

## Solution

In case of crash of an app script, print the last 30 lines of the log.

## PR Status

~It works only from the command line, because I use yunohost-cli.log.
I wonder if there's a variable with the log used, or anything that let you know the environment. I'll have a look to it...~ *If there's a cleaner way to find that info, would be better though.*
I tried to add the file and the line, like ynh_debug, but didn't find a way to do it only for a duplicated BASH_XTRACEFD. It either for all logs or for none.

## How to test

Add `false` wherever you want in a script to force it to fail.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps


zamentur:
The current log system is not enough ? 

As a reminder, this one displays the command or link to launch to get the end of the log file.

Many users were lost when the logs were displayed. As a result, we stopped displaying logs to users (now we have to use --debug) to prefer this log management system that invites the user to go and see. 

I understand in this PR that the objective is to help the debug, in this case this feature should only be activated in a development context, right?

Translated with www.DeepL.com/Translator

maniackcrudelis:
> I understand in this PR that the objective is to help the debug, in this case this feature should only be activated in a development context, right?

In development context, you can use --debug because you know the script is probably going to fail, or use ynh_debug because you know where the script will fail but want to know why.

This PR is more for "normal" user when a script fail. There's often simple solution, because a file wasn't removed before or other common issues. But the complete log is hard to read if you don't know what to look for.
The current log system, with the link to see the complete log, is really interesting to see what happened in the script. But it's mostly for us. Here I would like to show a small piece of log, just before the crash, so an admin could just read why its app has crashed and have a look to the forum or ask immediately for the problem he had.
I see often (even if I'm not on the support chat) error logs with simple errors, or already identified ones asked on the forum just because the user didn't read the log.
So my idea is more to show a trace of what happened just before the crash.

maniackcrudelis:
What the fuck ! I got utils from stretch-unstable !
Why do I have such a big diff...

maniackcrudelis:
I hate git...
Never mind...

After discussion with ljf, we decided that, since the log is mainly intended to be seen by power users, we could show it only in CLI.
For an execution from the web admin, a failure would anyway show a link to the log with the last lines of it.
A made also some aesthetic adjustments.
The final result is that:

data/helpers.d/utils
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Determine whether the scripts is executed from a terminal or the admin JS.
+#
+# [internal]
+ynh_is_term () {
+	# Return true if $TERM return xterm
+	# Which means the script is executed from a terminal, not from the admin JS
+	[[ "$TERM" =~ "xterm" ]]
```
alexAubin: Uuuuh tested it on my side but for me it contains (during the execution of the script) `rxvt-unicode-256color`

maniackcrudelis: Huh, the name of your terminal is `unicode-256color` !?

alexAubin: `rxvt-unicode-256color` yes, it's the whole `urxvt` thing ... but anyway it could be something else, e.g. I think gnome-terminal or other things have different values ?

maniackcrudelis: I don't think so actually, I think normal people using a normal terminal have xterm. Note that I'm not really using xterm, just the standard terminal of my distrib.

I'm going to try on my side as well, could you please try with `test -t` instead ?

maniackcrudelis: `test -t` says that the admin JS is a terminal...
I guess under the hood that's a virtual terminal which is being used...

Another solution is that apparently from the admin JS, I get `dumb` for $TERM, if you have the same, we can try with that.

alexAubin: So in https://github.com/YunoHost/yunohost/pull/687/commits/95d2c8f79940feb8bf757bec48422bcc484b9253 I propose to add a new YNH_INTERFACE variable which is equal to `cli` or `api` depending of the use case.

 (I also removed the `ynh_is_term` helper because imho `[ $YNH_INTERFACE == "cli" ]` is semantic enough but we can keep the dedicated helper if you want, idk)

maniackcrudelis: Well, yes why not.
Seems a little overkill, but I believe it works.

alexAubin:
During my tests, I added an `exit -1` in my install script basically near the very beginning of the script... But then the snippet of code detected the initial setup of `ynh_exit_properly` :  

```
3798 DEBUG + set -eu
3798 DEBUG + trap ynh_exit_properly EXIT
```

So the actual error was not shown (Instead, I got lines about helpers being loaded etc).

So to improve the accuracy, in https://github.com/YunoHost/yunohost/pull/687/commits/899d3d4f105e1627a0c10bd939ecca74035f9365 I propose to add a `+ ` such that this line gets caught instead : 

```
3802 DEBUG + ynh_exit_properly
```

maniackcrudelis:
Did you try it later in the script, with an exit at the end for example ?

alexAubin:
Well then it works yes ;)

maniackcrudelis:
Ok then

data/helpers.d/utils
```diff
@@ -28,6 +27,41 @@ ynh_exit_properly () {
...
+		# Remove the timestamp at the beginning of the line
+		# Remove "yunohost.hook..."
+		# Add DEBUG and color it at the beginning of each log line.
+		echo -e "$(tail --lines=30 "$ynh_log" \
+			| sed '1,/+ ynh_exit_properly/!d' \
+			| sed 's/^[[:digit:]: ,-]*//g' \
+			| sed 's/ *yunohost.hook.*\]/ -/g' \
+			| sed 's/^WARNING /&/g' \
+			| sed 's/^DEBUG /&  /g' \
+			| sed 's/^INFO /&   /g' \
```
alexAubin: So uh aren't the `DEBUG` supposed to be replaced ? (c.f. the screenshot)

maniackcrudelis: Those 3 lines, WARNING, DEBUG and INFO are only there to align the text after.
Those sed are only adding spaces.

alexAubin: Alrighty, flagging as resolved ;P

alexAubin:
Will merge in a few days

alexAubin:
Actually yolomergin' for 3.6 testing

PR #689, opened 2019-03-19
[enh] Nginx conf tweaks for theme

alexAubin:
## The problem

Currently, only ynhpanel.js is injected in app pages which integrate the yunohost portal button / overlay. This JS currently injects the CSS in the page. This is fine but for theme, we now want to inject two more files (the custom CSS and custom JS) ... and thisis a whole mess because those are then loaded ~after~ the inital "DOM loaded" event which makes some stuff harder to implement.

## Solution

Inject all those file directly, like ynhpanel.js was.

## PR Status

Tested and working

## How to test

Try to set a custom theme with https://github.com/YunoHost/SSOwat/pull/122

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/yunohost_panel.conf.inc
+++ b/data/templates/nginx/plain/yunohost_panel.conf.inc
@@ -4,5 +4,5 @@ sub_filter_once on;
 # Apply to other mime types than text/html
 sub_filter_types application/xhtml+xml;
 # Prevent YunoHost panel files from being blocked by specific app rules
-location ~ (ynh_portal.js|ynh_overlay.css) {
+location ~ (ynh_portal.js|ynh_overlay.css|userinfo.json) {
 }
--- a/data/templates/nginx/plain/yunohost_panel.conf.inc
+++ b/data/templates/nginx/plain/yunohost_panel.conf.inc
@@ -1,8 +1,8 @@
-# Insert YunoHost panel
-sub_filter </head> '<script type="text/javascript" src="/ynhpanel.js"></script><link type="text/css" rel="stylesheet" href="/ynhpanel.css"></link><script type="text/javascript" src="/ynhtheme/custom.js"></script><link type="text/css" rel="stylesheet" href="/ynhtheme/custom.css"></link></head>';
+# Insert YunoHost button + portal overlay
+sub_filter </head> '<script type="text/javascript" src="/ynh_portal.js"></script><link type="text/css" rel="stylesheet" href="/ynh_overlay.css"></link><script type="text/javascript" src="/ynhtheme/custom_portal.js"></script><link type="text/css" rel="stylesheet" href="/ynhtheme/custom_overlay.css"></link></head>';
 sub_filter_once on;
 # Apply to other mime types than text/html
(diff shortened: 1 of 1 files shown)


data/templates/nginx/plain/yunohost_panel.conf.inc
```diff
@@ -1,8 +1,8 @@
-# Insert YunoHost panel
-sub_filter </head> '<script type="text/javascript" src="/ynhpanel.js"></script></head>';
+# Insert YunoHost button + portal overlay
+sub_filter </head> '<script type="text/javascript" src="/ynh_portal.js"></script><link type="text/css" rel="stylesheet" href="/ynh_overlay.css"></link><script type="text/javascript" src="/ynhtheme/custom_portal.js"></script><link type="text/css" rel="stylesheet" href="/ynhtheme/custom_overlay.css"></link></head>';
 sub_filter_once on;
 # Apply to other mime types than text/html
 sub_filter_types application/xhtml+xml;
 # Prevent YunoHost panel files from being blocked by specific app rules
-location ~ ynhpanel\.(js|json|css) {
```
zamentur: ynhpanel.json disappear ?

alexAubin: Yes, I don't know what that stuff was supposed to be but it never was actually used ...

JimboJoe: Hum, I remember it to be necessary and to fix apps at [some point in time](https://github.com/YunoHost/yunohost/commit/bb11168c31064d6be6cba3b8f2c48bf3a70f02e8)... :sweat_smile:

alexAubin: @JimboJoe : for the JS and the CSS file, yes, here I just renamed then ... but what about `ynhpanel.json` ? It's never effectively requested anywhere ...

zamentur: It was used before here https://github.com/YunoHost/SSOwat/blob/4b85679471db9dc55e005a0a962c61da749e4c32/portal/assets/js/ynhpanel.js#L256

But now it's not used anymore.

zamentur: Note: this feature could be used to display an other page with app info. I aws thinking use that for menu_ynh, to create a default menu with the app of the user...

alexAubin: Alright, but I would like to avoid maintaining features that are not actually working / used ... This is not just about keeping `ynhpanel.json` in this file, since other piece of code referring to it were also deleted in SSOwat's PR for themes ... also for semantic it should be renamed to something like `userinfo.json` to reflect more its content :/.

zamentur: https://github.com/YunoHost/SSOwat/pull/133

alexAubin:
Planning to add this to the 3.5.x which I would like to release early next week ... so this should be merged in the next few days

PR #691, opened 2019-03-19
[enh] Avoid to send simultaneously too much emails

zamentur:
## The problem

It seems possible for a user to send thousand of emails simultaneously. It's pretty cool but a bit risky.

## Solution

Set up a rate delay of 12s to avoid to send simultaneously to the same destination server.

## PR Status

Ready, currently in test with one of my server

## How to test

```
service postfix reload
```
Next send a lot of emails and check delays.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


maniackcrudelis:
One email every 12s maximum ? What about scripts that send emails ?

zamentur:
One email every 12s to a same destination. But yes if a script send more than 1 message every 12s it could be a problem. On the other hand, sending 1msg per 12s with a script to the same destination could be dangerous or simply completely fill the inbox of the final recipient.

Have you a particular script in mind ?

maniackcrudelis:
I have especially a script that sends me emails as soon as my webcam see something moving in my flat.

But, between "thousand of emails simultaneously" and one per 12s, I hope there's a middle value. Like one per second.

alexAubin:
But is your webcam likely to send you several emails per second ? And is it a normal / wished behavior ? o.O

maniackcrudelis:
Not several per second, but several per 12s yes.
And that's just one of my use case. 2 or 3 emails during 12s does not seems like too much emails. Depending of what's happening at this time on the server.
For sure thousand of email at a time is a problem, but we shouldn't limit to 1/12s because **we** do not have that usage.

zamentur:
Note: if there is momentarily more than one email per 12s it is not a problem, the emails will be sent with a delay that's all.

But we can increase the frequency effectively.

About "1s", it seems quite strange to send to a same server 3600 emails in hour ! At this rate, your server could be blacklisted easily...

zamentur:
> For sure thousand of email at a time is a problem, but we shouldn't limit to 1/12s because **we** do not have that usage.

I think we can find a better rate. I was not aware of this kind of usecases Important: it's not a rate that put into trash bin set of emails sending in the same time, it is just distributed over time.

I understand specific use case with a lot of emails to send, but we need to think about ynh instances with RBL's issues or the free.fr filters... Big mail servers are more and more strict...

maniackcrudelis:
I'm not against a rule to prevent server to be used as spam relay.
But we should just be careful to not hind legit use cases.

alexAubin:
Soooo what do we do with this ? Would 1 email every 5 second would be a better rate ?

alexAubin:
So idk, arbitarily switched to 5s if that pleases more people ... Feel free to suggest other values ...

Otherwise planning to merge soonish to move forward

taziden:
Hi, as it's configured it's not really "1 mail every 5 seconds".
It's "we can't send more than 1 mail every 5 seconds to a particular recipient".
That doesn't prevent from sending a huge volume to one domaine, or several domains, every 5s.

PR #695, opened 2019-03-27
[fix] Rework tools update (only)

alexAubin:
## The problem

Many issues related to sources.list management these days ... Current code miserably crashes with "FetchFailedException" which doesn't tell anything about what's wrong.

## Solution

Rework `tools_update` to use `apt update` directly instead of `apt`'s python lib which is madness

Also properly shows warning/errors message occuring during the command, and detect the issue (and dump the sources.list lines to help bebugging)

## PR Status

Tested

## How to test

Run `yunohost tools update` in various conditions (e.g. voluntary break the sources.list)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Note that this contains commits from https://github.com/YunoHost/yunohost/pull/692

The motivation of this PR is that it should be smaller and less touchy as it doesn't affect `tools_upgrade` which is more complicated.

Ideally I'd like to include this in 3.5.x to avoid doing so much more boring support about FetchFailedException

src/yunohost/tools.py
```diff
@@ -474,23 +475,39 @@ def tools_update(ignore_apps=False, ignore_packages=False):
...
+        # LC_ALL=C is here to make sure the results are in english
+        command = "LC_ALL=C apt update"
+        # TODO : add @is_unit_operation to tools_update so that the
+        # debug output can be fetched when there's an issue...
+
+        # Filter boring message about "apt not having a stable CLI interface"
+        # Also keep track of wether or not we encountered a warning...
+        warnings = []
+        def is_legit_warning(m):
+            legit_warning = m.rstrip() and "apt does not have a stable CLI interface" not in m.rstrip()
```
Psycojoker: Shouldn't we use "apt-get" instead of "apt"? iirc apt is designed to be runed interactively not for scripting.

PR #698, opened 2019-04-02
[fix] Fix the dbus interface to get info for services

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1316

## Solution

Very intuitive™, one should use LoadUnit and not GetUnit ...

## PR Status

Tested and working ... though still need to check that this doesnt break webadmin

## How to test

Try to run `yunohost service status` with a disabled service. Also you can try to add in /etc/yunohost/services.yml a service that doesn't exist at all

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/service.py
+++ b/src/yunohost/service.py
@@ -325,7 +325,7 @@ def service_status(names=[]):
 
             result[name] = {
                 'status': str(status.get("SubState", "unknown")),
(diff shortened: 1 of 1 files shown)


Psycojoker:
Sometime I really have the feeling that the world is splited between people that want to make things intuitive and the dbus developpers.

PR #700, opened 2019-04-03
[fix] Nodejs helpers : use YNH_APP_INSTANCE_NAME instead of YNH_APP_ID

yalh76:
## The problem

f you install two time the same application, when removing one of them, nodejs is uninstalled

## Solution

YNH_APP_INSTANCE_NAME instead of YNH_APP_ID

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -123,7 +123,7 @@ ynh_install_nodejs () {
 	fi
 
 	# Store the ID of this app and the version of node requested for it
-	echo "$YNH_APP_ID:$nodejs_version" | tee --append "$n_install_dir/ynh_app_version"
+	echo "$YNH_APP_INSTANCE_NAME:$nodejs_version" | tee --append "$n_install_dir/ynh_app_version"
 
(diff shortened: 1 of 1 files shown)


alexAubin:
LGTM, Would be nice to merge this for 3.5.

Any confirmation from @YunoHost/apps that this is alright ?

kay0u:
I'm wondering, what about apps already using this help? :worried:

alexAubin:
Then I guess the line won't properly get removed and nodejs won't be uninstalled ... But in the nominal case it should be fine as the app id is usually the instance name .. only in multiinstance install you have foo instance "foo" vs. "foo__2"

So for existing install with multi-instance app, there's a small issue yes, but I think we can live with it ..?

kay0u:
Ok

yalh76:
> I'm wondering, what about apps already using this help?

At least:

- distbin
- mastodon
- peertube
- wekan
- wikijs

PR #701, opened 2019-04-04
[enh] WebSocket proxying

kay0u:
## The problem

Unable to use these lines in nginx conf:
```
	proxy_set_header Upgrade $http_upgrade;
	proxy_set_header Connection $connection_upgrade;
```
for WebSocket proxying.

I'm not an expert in Nginx but you can find doc here:
http://nginx.org/en/docs/http/websocket.html

## Solution

What I've done inside the file `server.tpl.conf`

## PR Status

...

## How to test

Add these lines in nginx conf of an app that use websocket
 ```
	proxy_set_header Upgrade $http_upgrade;
	proxy_set_header Connection $connection_upgrade;
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -1,6 +1,6 @@
 map $http_upgrade $connection_upgrade {
-        default upgrade;
-        ''      close;
+    default upgrade;
+    ''      close;
 }
 
 server {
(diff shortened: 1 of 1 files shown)


alexAubin:
Was wondering if multidomain would cause issue, but apparently it doesn't ô.o

PR #703, opened 2019-04-07
Add support in restore script of ynh_setup_source

Josue-T:
## The problem

Actually the path to the app.src file in the helper ynh_setup_source work only in install and upgrade script.

## Solution

Add the support for restore script by adding a specific path if the general path is not available.

## PR Status

Tested work on monitorix.

## How to test

Use the helper in a restore script of an app.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Merging right now because it's 15 days old with 3 approvals

PR #706, opened 2019-04-11
[enh] Bind OpenLDAP socket on localhost

rds13:
## The problem

OpenLDAP service open its socket on every network interfaces. This is not usefull as services are only connected to localhost.

## Solution

Change default OpenLDAP configuration to only bind on localhost.

## How to test

This command 
```
lsof -i :389 | grep LISTEN
```

should display binding on localhost only

```
slapd    3559 openldap    8u  IPv4 34292432      0t0  TCP localhost:ldap (LISTEN)
slapd    3559 openldap    9u  IPv6 34292433      0t0  TCP localhost:ldap (LISTEN)
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/slapd/slapd.default
+++ b/data/templates/slapd/slapd.default
@@ -21,7 +21,7 @@ SLAPD_PIDFILE=
 # sockets.
 # Example usage:
 # SLAPD_SERVICES="ldap://127.0.0.1:389/ ldaps:/// ldapi:///"
-SLAPD_SERVICES="ldap:/// ldapi:///"
+SLAPD_SERVICES="ldap://127.0.0.1:389/ ldap://[::1]:389/ ldapi:///"
 
 # If SLAPD_NO_START is set, the init script will not start or restart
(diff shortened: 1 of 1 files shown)


alexAubin:
To be discussed tho, but there's a advanced use case where people are interested in having access to the LDAP database from an external server ... So one could imagine having a global setting to enable / disable the exposure of LDAP on the outside. But imho that's okay to merge as is... Advanced users can still edit this file if they want to (though that increases the tweaking cost)

alexAubin:

Jaxom99:
> To be discussed tho, but there's a advanced use case where people are interested in having access to the LDAP database from an external server ... So one could imagine having a global setting to enable / disable the exposure of LDAP on the outside. But imho that's okay to merge as is... Advanced users can still edit this file if they want to (though that increases the tweaking cost)

yes indeed ! I keep this for further reference, as I may soon propose such a setup to a client of mine.

PR #707, opened 2019-04-12
[FIX] Config Panel

kay0u:
## The problem

[hook_exec](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/hook.py#L414) return the code + a json, that was not well handled

## Solution

Fix this issue

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Yolomergin'

PR #708, opened 2019-04-12
[FIX] app actions

kay0u:
## The problem

[admin panel use app_id](https://github.com/YunoHost/yunohost-admin/blob/stretch-unstable/src/views/app/app_actions.ms#L14)

## Solution

Add app_id

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1512,6 +1512,7 @@ def app_action_list(app):
     actions = os.path.join(APPS_SETTING_PATH, app, 'actions.json')
 
     return {
+        "app_id": app,
         "app": app,
         "app_name": app_info_dict["name"],
         "actions": read_json(actions) if os.path.exists(actions) else [],


kay0u:
I did that because of this: https://github.com/YunoHost/yunohost/blob/4e920c3040f9df7f0e95ed5fb59f4bcb88dade39/src/yunohost/app.py#L1596

It is duplicated for the config panel. (And I'm afraid of breaking something :P)

alexAubin:
I think you mean : https://github.com/YunoHost/yunohost-admin/blob/stretch-unstable/src/views/app/app_actions.ms#L14

so I think I understand the issue ;) But I would rather just change the app_actions.ms file (app_id -> app) rather than duplicate the info :s

kay0u:
No I mean that this information is already duplicated for the config panel, but I can change app_actions.ms, it would be clearer

https://github.com/YunoHost/yunohost/blob/4e920c3040f9df7f0e95ed5fb59f4bcb88dade39/src/yunohost/app.py#L1596-L1597

alexAubin:
Thanks !

PR #710, opened 2019-04-13
[fix] Fix extraction of weight value for ynh_script_progression

maniackcrudelis:
## The problem

The weight value will take everything after --weight=, especially --time or --last

## Solution

Fix the regex

## PR Status

Ready to be reviewed.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps


alexAubin:
Yolomerging because that looks like a small benign change ?

PR #711, opened 2019-04-13
Use ynh_systemd_action

maniackcrudelis:
## The problem

Use ynh_systemd_action helper in other helpers

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps


data/helpers.d/backend
```diff
@@ -442,7 +442,7 @@ EOF
   ynh_store_file_checksum "$finalfail2banjailconf"
   ynh_store_file_checksum "$finalfail2banfilterconf"
 
-  systemctl try-reload-or-restart fail2ban
+  ynh_systemd_action --service_name=fail2ban --action=reload
```
Josue-T: Well,

I think `reload` is not same than `try-reload-or-restart`.

And for this helper maybe we need to manage the error because actually some app install/upgrade fail juste because the fail2ban fail to reload (see https://github.com/YunoHost-Apps/synapse_ynh/issues/88#issuecomment-478808204)

data/helpers.d/psql
```diff
@@ -276,7 +276,8 @@ ynh_psql_test_if_first_run() {
 			ynh_die "postgresql shoud be 9.4 or 9.6"
 		fi
 
-		systemctl start postgresql
+		ynh_systemd_action --service_name=postgresql --action=start
```
Josue-T: Is it really usefull to use this helper in this case ?

data/helpers.d/backend
maniackcrudelis: https://github.com/YunoHost/yunohost/blob/stretch-unstable/data/helpers.d/system#L116-L119
https://www.freedesktop.org/software/systemd/man/systemctl.html#reload-or-restart%20PATTERN%E2%80%A6

data/helpers.d/psql
maniackcrudelis: If postgresql fails to start, the helper will give us a log.
So it seems to be useful.

Josue-T: ok

data/helpers.d/psql
```diff
@@ -289,6 +290,6 @@ ynh_psql_test_if_first_run() {
 		yunohost service add postgresql --log "$logfile"
 
 		systemctl enable postgresql
-		systemctl reload postgresql
+		ynh_systemd_action --service_name=postgresql --action=reload
```
alexAubin: Kinda unrelated but just to say that this stuff about the manipulating postgresql here is a bit weird / misleading : the "real" postgresql service seems to be `postgresql@9.6-main` (or something like this) compared to just `postgresql` service which is a different thing ... Dunno if that's an issue or not but would be nice if people are aware of this :P

maniackcrudelis:
Following this issue https://github.com/YunoHost-Apps/etherpad_mypads_ynh/issues/65, I suggest to restart instead of reload fail2ban.
Reload may not be enough if there's a lot of banned IP.

Josue-T:
> Following this issue YunoHost-Apps/etherpad_mypads_ynh#65, I suggest to restart instead of reload fail2ban.
> Reload may not be enough if there's a lot of banned IP.

Yes, but restart is sometime really really long one the slow architecture...

maniackcrudelis:
> Yes, but restart is sometime really really long one the slow architecture...

Ok then...
Anyway, that's an edge case and this behavior is fixed since version 0.10 with a reload that doesn't re-ban IP.

alexAubin:
Soooo are we good with this ?

maniackcrudelis:
I'm good with this PR.

alexAubin:
Will merge in the coming days except if there are change requests :wink:

PR #713, opened 2019-04-15
Force disconnection of all clients connected to the database

yalh76:
## The problem

Sometimes, when dropping a PostgreSQL database, there is still some connections to the database and the database refuses to drop

## Solution

Force disconnection of all clients to the database

## PR Status

Ready

## How to test

Don't know

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/psql
+++ b/data/helpers.d/psql
@@ -105,6 +105,9 @@ ynh_psql_create_db() {
 # Requires YunoHost version 3.?.? or higher.
 ynh_psql_drop_db() {
 	local db=$1
+	# First, force disconnection of all clients connected to the database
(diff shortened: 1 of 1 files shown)


yalh76:
> 
> 
> Not sure to know where that incantation comes from but trusting you I guess 🤔

Found that solution on various website:
- https://stackoverflow.com/questions/5408156/how-to-drop-a-postgresql-database-if-there-are-active-connections-to-it
- https://dba.stackexchange.com/questions/16426/how-to-drop-all-connections-to-a-specific-database-without-stopping-the-server

alexAubin:

PR #717, opened 2019-04-25
Reorganize helpers

alexAubin:
## The problem

Content of the various helper files is kind of a mess .. it's not clear where to find a given helper. The names are very ambiguous and kinda historical ... E.g. there is a dedicated file for `mysql` but `backend` contains stuff for systemd, php, fail2ban and logrotate all together :| 

## Solution

Reorganize the helpers and try to make everything a bit more semantic ... I've done various changes but basically we end up with : 

##### Individual files for interfaces with various technologies : 

- `nginx`
- `php`
- `apt`
- `mysql`
- `postgresql`
- `systemd`
- `nodejs`
- `fail2ban`
- `logrotate`

##### Yunohost-specific helpers so : 

- `logging` : print and exec helpers, related to debug/info/warn/critical levels
- `backup` : all helpers related to backup (including the checksum things)
- `getopts` : meta stuff for helper argument parsing
- `network` : stuff related to network (ip, port)
- `setting` : app settings, and also webpath_available/register which is kind of a setting ...
- `string` : helpers to help with string (password generation, string normalization..)
- `user` : manipulation of yunohost or system/unix users
- `utils` : misc stuff that don't really fit in other files (setup_source, secure_remove, exit_properly ...) though to be discussed

## PR Status

Not tested but uh, mainly to be discussed first

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
I know you really don't like that but we should really provide backward compatibility if we do that and it's not even that hard here: just replace function with `echo "depreciated warning blablabla"; call_to_new_function $@` and in 1 or 2 releases we can just send massive automatically generated PR to applications, that would way less break user space :/

maniackcrudelis:
@alexAubin is that PR only about moving stuff from files to another ones ?
Hard to read the diff since there's a lot of modifications.

About moving helpers to other files, I agree.
And I'm not worried about backward compatibility, all apps are using `/usr/share/yunohost/helpers` so content of files doesn't matter.

alexAubin:
Uh yea it's only about moving files (and I replaced some tabs with space because some stuff were a mess) so there shouldn't be any backward compatibility issue except for the initial sourcing of helpers ? (c.f. comment from Maniack)

maniackcrudelis:
Not for apps at least. All apps are using `/usr/share/yunohost/helpers`, not a specific helper file.

Psycojoker:
Oh ok, that's fine then :)

alexAubin:
Flagging as mergin soon then :stuck_out_tongue_winking_eye:

PR #719, opened 2019-05-04
Use local manifest

kay0u:
## The problem

https://github.com/YunoHost/issues/issues/1349

## Solution

Use the local manifest (I'm not sure I broke anything else)

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -260,6 +260,12 @@ def app_list(filter=None, raw=False, installed=False, with_backup=False):
 
     # Get app list from the app settings directory
     for app in os.listdir(APPS_SETTING_PATH):
+        if app not in app_dict:
+            app_id, app_instance_nb = _parse_app_instance_name(app)
+            # Handle multi-instance case
+            if app_instance_nb > 1:
+                app_dict[app] = app_dict[app_id]
+
         with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
             app_dict[app]['manifest'] = json.load(json_manifest)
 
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -261,17 +261,16 @@ def app_list(filter=None, raw=False, installed=False, with_backup=False):
...
-                # FIXME : What if it's not !?!?
+            app_id, app_instance_nb = _parse_app_instance_name(app)
+            # Handle multi-instance case
+            if app_instance_nb > 1:
+                app_dict[app] = app_dict[app_id]
 
-            with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
-                app_dict[app] = {"manifest": json.load(json_manifest)}
+        with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
+            app_dict[app]['manifest'] = json.load(json_manifest)
```
alexAubin: So if I understand correctly, the main fix comes from the unindent here ? Like, `app_dict[app]` is already supposed to be filled with the manifest from the applist but we override it with the one from the setting directory ?

I'm a bit afraid that this is not the right fix, as it means that for instance, for multi-instance app (e.g. installing a second wordpress) you would fill the manifest corresponding to the already-installed version of the wordpress ... Which may actually conflict with the version of the wordpress app that will actually be instsalled, and may create various issues ...

I think instead we should just explicitly add a new key somehow .... probably in the `status` section which also contains info like `installed_at` and `upgraded_at`.

kay0u: Oups, in fact it's not this part of app.py that used for the app_info.ms... Do I have to close and open a new PR to fix it? Or can I force-push?

kay0u:
I'll reopen a new PR with the right fix!

PR #723, opened 2019-05-15
Optimize ynh_script_progression

alexAubin:
## The problem

`ynh_script_progression` is awesome but I noticed that app logs are filled with hundreds of lines related to the computation of the progress bar. But some parts of it could be factorized to be only done once so that'll slightly simplify the debug logs

## Solution

Compute the stuff we can only once using "global"(?) variables

## PR Status

Tested and seems to be working, but I'm not an expert of this helper ;)

## How to test

Run an app install and check the detailed log

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/logging
+++ b/data/helpers.d/logging
@@ -198,6 +198,13 @@ ynh_print_ON () {
 # Requires YunoHost version 3.?.? or higher.
 increment_progression=0
 previous_weight=0
+max_progression=-1
+# Set the scale of the progression bar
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Too lazy to test, but read carefully the diff, looks ok.
Let's the CI told us if it's ok ;)

PR #724, opened 2019-05-15
Remove unecessary log messages

alexAubin:
## The problem

`ynh_script_progression` is awesome, but the info logging channel (what the regular user sees) is often kinda polluted with low-level technical messages, typically saying that `service X` is being reloaded...

E.g. :  

```text
$ yunohost app remove wordpress
Info: Removing application wordpress…
Info: [+...................] > Load settings
Info: [#+++++++............] > Remove dependencies
Info: [########+...........] > Remove the mysql database
Info: Removing database wordpress
Info: [#########+..........] > Remove app main directory
Info: [##########..........] > Remove nginx configuration
Info: Reload the service nginx
Info: [##########++........] > Remove php-fpm configuration
Info: /etc/php/7.0/fpm/conf.d/20-wordpress.ini wasn't deleted because it doesn't exist.
Info: Reload the service php7.0-fpm
Info: [############+++++...] > Remove fail2ban configuration
Info: Reload the service fail2ban
Info: [#################++.] > Remove the dedicated user
Info: Remove the user wordpress
Info: [####################] > Deletion completed
```

Imho those are duplicated / not necessary and the packager can see the detailed log if needed...

## Solution

Remove those info messages

## PR Status

Tested and working

## How to test

Install / remove wordpress for example

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/mysql
+++ b/data/helpers.d/mysql
@@ -225,7 +225,6 @@ ynh_mysql_remove_db () {
 
 	local mysql_root_password=$(sudo cat $MYSQL_ROOT_PWD_FILE)
 	if mysqlshow -u root -p$mysql_root_password | grep -q "^| $db_name"; then	# Check if the database exists
-		ynh_print_info --message="Removing database $db_name"
 		ynh_mysql_drop_db $db_name	# Remove the database	
 	else
 		ynh_print_warn --message="Database $db_name not found"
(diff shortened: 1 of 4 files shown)


kay0u:
I'm ok with it

maniackcrudelis:
Some of them could be useful, maybe instead of just removing them, we could turn them to simple echo.
So it would be into the log if needed.

alexAubin:
>we could turn them to simple echo. So it would be into the log if needed.

Indeed, but my understanding is that all the info is already in the debug log anyway (since it logs all the call to the helpers etc)

maniackcrudelis:
That's right indeed.
Don't know. For me reading the trace of a helper isn't a problem. Don't know for others.

alexAubin:
Sooo proposing to merge soon

PR #725, opened 2019-05-18
Set min version to 3.5.0 for helpers

maniackcrudelis:
## The problem

We forgot to set the minimum version required for helper during the last release.

## Solution

Set the last version to 3.5.0 for those helpers.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Merging as microdecision

PR #726, opened 2019-05-18
[fix] Stupid issue with files inside tar : foo is not the same as ./foo

alexAubin:
## The problem

https://forum.yunohost.org/t/restore-from-folder-rather-then-tar-gz/7885
https://cybre.space/@aleks/102117385695119433
https://serverfault.com/questions/457480/tar-leading-period-slash

## Solution

Check for both `info.json` and `./info.json` ... and similarly each time we try to extract a file.

## PR Status

Tested on my side and working

## How to test

You can reproduce the issue by untaring an archive and retaring it with : 

```
cd thearchive/
tar -zcvf ../buggyarchive.tar.gz ./
```

Then using `yunohost backup info buggyarchive` or trying to restore it.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -1850,11 +1850,19 @@ class TarBackupMethod(BackupMethod):
         # Mount the tarball
         logger.debug(m18n.n("restore_extracting"))
         tar = tarfile.open(self._archive_file, "r:gz")
-        tar.extract('info.json', path=self.work_dir)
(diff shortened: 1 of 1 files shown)


alexAubin:
Sooo proposing to merge soon

PR #730, opened 2019-05-25
[fix] current version in app_info

kay0u:
## The problem

Fix https://github.com/YunoHost/issues/issues/1349

## Solution

This PR have to be merged with: https://github.com/YunoHost/yunohost-admin/pull/246

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -362,6 +362,12 @@ def app_info(app, show_status=False, raw=False):
 
         ret['upgradable'] = upgradable
         ret['change_url'] = os.path.exists(os.path.join(app_setting_path, "scripts", "change_url"))
+        
+        with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
+            manifest = json.load(json_manifest)
+        
+        if 'version' in manifest:
+            ret['version'] = manifest.get('version', '-')
 
         return ret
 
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -362,6 +362,12 @@ def app_info(app, show_status=False, raw=False):
 
         ret['upgradable'] = upgradable
         ret['change_url'] = os.path.exists(os.path.join(app_setting_path, "scripts", "change_url"))
+        
+        with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
+            manifest = json.load(json_manifest)
+        
+        if 'version' in manifest:
+            ret['version'] = manifest.get('version', '-')
```
alexAubin: I think you can simplify those two by simply having the second line `ret['version'] = manifest.get('version', '-')` ? (`.get` already handle the case where the key doesn't exist, c.f. the second arg)

Otherwise in some cases there won't be a `version` key in `ret` which might make https://github.com/YunoHost/yunohost-admin/pull/246 buggy in its current form ?

kay0u: You're right!

PR #731, opened 2019-05-29
Add exception for data backup

maniackcrudelis:
## The problem

Some users were using the setting `backup_only_core` to not backup data during their backup.
As this setting wasn't meant to be used this way, I didn't care when changing the way it was handled.
https://forum.yunohost.org/t/yunohost-backup-create-ko/7688

## Solution

Implement a new setting `do_not_backup_data` to add to an app setting to avoid data backup.
This setting, as it's not part of the upgrade process, will not be removed by any script. So it could be used by an user to avoid to backup data for all backup script.

## PR Status

Ready to be reviewed.

## How to test

Set the setting for an app, `ynh_app_setting_set $app do_not_backup_data 1`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/backup
+++ b/data/helpers.d/backup
@@ -59,12 +59,18 @@ ynh_backup() {
     local not_mandatory="${not_mandatory:-0}"
 
     BACKUP_CORE_ONLY=${BACKUP_CORE_ONLY:-0}
+    test -n "${app:-}" && do_not_backup_data=$(ynh_app_setting_get $app do_not_backup_data)
 
     # If backing up core only (used by ynh_backup_before_upgrade),
     # don't backup big data items
-    if [ "$is_big" == "1" ] && [ "$BACKUP_CORE_ONLY" == "1" ] ; then
-      ynh_print_info --message="$src_path will not be saved, because backup_core_only is set."
-      return 0
+    if [ $is_big -eq 1 ] && ( [ ${do_not_backup_data:-0} -eq 1 ] || [ $BACKUP_CORE_ONLY -eq 1 ] ) 
+    then
+        if [ $BACKUP_CORE_ONLY -eq 1 ]; then
+            ynh_print_warn --message="$src_path will not be saved, because 'BACKUP_CORE_ONLY' is set."
+        else
+            ynh_print_warn --message="$src_path will not be saved, because 'do_not_backup_data' is set."
+        fi
+        return 0
     fi
 
     # ==============================================================================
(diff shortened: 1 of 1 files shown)


Josue-T:
Hello,

Since a long time I thought that using the environment variable to enable some option for yunohost is not really clean. Why not adding option to the yunohost command which enable the `BACKUP_CORE_ONLY` environment variable.

By example we can add a flag `-c` or `--core-only`.

So if we run the command `yunohost backup create --core-only --apps app1` the yunohost core get this argument and enable the `BACKUP_CORE_ONLY` environnement variable when the script is called.

The advantage of that is:
- Add the possibility to add this option in the admin
- Show cleanly this option in the help on the cli

And the best thing would be probably to do the same thing for the `NO_BACKUP_UPGRADE` variable.

maniackcrudelis:
`BACKUP_CORE_ONLY` is only used by `ynh_backup_before_upgrade`. I see no problem in using an environment variable to do so.
There's no need to document that thing (even though it is...) since that's only an internal process for the upgrade script.

maniackcrudelis:
But it isn't about any environment variable here !
By this PR, I add a setting which can be set for each app separately to inform the backup script to not backup data.

Josue-T:
Yes but my idea, was just to replace the settings by an option in the CLI which is cleaner.

Anyway do what you want, I don't mind this PR, it's just that I feel that we do many dirty things about that.

maniackcrudelis:
The idea is to have a setting for **each** app, not for the entire backup process.
That's simply to restore what users were doing with the previous setting `backup_core_only`. Except that this time, this is a setting design for that purpose.

alexAubin:
Just leaving a random idea, but we could also design something like this : 

```bash
# Backup app1 (only the core), app2 and app3
yunohost backup create --apps app1::core app2 app3

# Backup app 1 (core + data)
yunohost backup create --apps app1

# One could even imagine : 
yunohost backup create --apps app1::data
```

That's more an "ephemeral" solution (you specify what you want to backup at command line time) but also more flexible as it allows to avoid re-editing a setting everytime you want to backup stuff differently

maniackcrudelis:
It could be an idea for temporary choosing which part of a backup you want indeed. Why not.
But again, that's not the purpose of this PR.

data/helpers.d/backup
```diff
@@ -59,12 +59,18 @@ ynh_backup() {
     local not_mandatory="${not_mandatory:-0}"
 
     BACKUP_CORE_ONLY=${BACKUP_CORE_ONLY:-0}
+    test -n "${app:-}" && do_not_backup_data=$(ynh_app_setting_get $app do_not_backup_data)
```
kay0u:
```suggestion
    test -n "${app:-}" && do_not_backup_data=$(ynh_app_setting_get --app=$app --key=do_not_backup_data)
```

PR #733, opened 2019-06-01
Fix ynh_print_OFF when set -x is used in other helpers

maniackcrudelis:
## The problem

As reported by @Josue-T, ynh_print_OFF doesn't work when used before a helper using getopts, because the helper getopts use set -x.

## Solution

Redirect BASH_XTRACEFD, the file descriptor for xtrace, to /dev/null to get rid of the trace.
Then redirect back to stdout.

## PR Status

Tested on leed.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/logging
+++ b/data/helpers.d/logging
@@ -172,7 +172,7 @@ ynh_exec_fully_quiet () {
 #
 # Requires YunoHost version 3.2.0 or higher.
 ynh_print_OFF () {
-	set +x
+	exec {BASH_XTRACEFD}>/dev/null
 }
 
 # Restore the logging after ynh_print_OFF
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
;)
Not really black magic, just using BASH_XTRACEFD, which contains the file descriptor (usually 7) and redirect it to /dev/null instead of stdout.
You can see that when using `--debug`, the command line contains `BASH_XTRACEFD=7 [...] 7>&1'`

kay0u:
https://github.com/YunoHost/yunohost/blob/b3d29238c4bba040b7ad5afdbcc2be7c6a50058d/src/yunohost/hook.py#L360

PR #732, opened 2019-05-31
[enh] support config_panel in TOML format

Psycojoker:
## The problem

Writting a config_panel in json turns out to be unecessary complicated and error prone.

## Solution

After looking at a lot of different formats, toml turns out to be the most intuitive one and less error prone to use for not super-turbo-nerdy people which most app devs are (and even for turbo-nerdy writting json by hand sucks). App group people confirmed that it really looks like an improvement other json.

The solution to limit the modification of the code is to transform this new toml format in the current json format internally on load (which is easier to use for code)

Sidenode: I'm considering introducing support for `manifest.toml` and will certainly do `actions.toml` as it also really looks like an improvment: https://gist.github.com/Psycojoker/71aca1767738c4b254b6cd74e09c28e1#file-manifest-toml

In json it looks like that:

```json
{
	"name": "Unattended-upgrades configuration panel",
	"version": "0.1",
	"panel": [{
		"name": "Unattended-upgrades configuration",
		"id": "main",
		"sections": [{
			"name": "50unattended-upgrades configuration file",
			"id": "unattended_configuration",
			"options": [{
				"name": "Choose the sources of packages to automatically upgrade.",
				"help": "We can't use a choices field for now. In the meantime please choose between one of this values:<br>Security only, Security and updates.",
				"id": "upgrade_level",
				"type": "text",
				"//": "\"choices\" : [\"Security only\", \"Security and updates\"]",
				"default" : "Security only"
			},
			{
				"name": "Would you like to update YunoHost packages automatically ?",
				"id": "ynh_update",
				"type": "bool",
				"default": true
			},
			{
				"name": "Would you like to receive an email from Unattended-Upgrades ?",
				"help": "We can't use a choices field for now. In the meantime please choose between one of this values:<br>If an upgrade has been done, Only if there was an error, Never.",
				"id": "unattended_mail",
				"type": "text",
				"//": "\"choices\" : [\"If an upgrade has been done\", \"Only if there was an error\", \"Never\"]",
				"default" : "If an upgrade has been done"
			}]
		},
		{
			"name": "apticron cron file",
			"id": "apticron_configuration",
			"options": [{
				"name": "Would you like to receive an email to inform which upgrades need to be done ?",
				"id": "previous_apticron",
				"type": "bool",
				"default": true
			},
			{
				"name": "When do you want to receive this email ?",
				"help": "Choose an hour between 12 and 23.<br>",
				"id": "previous_apticron_hour",
				"type": "number",
				"default": 20
			},
			{
[...]

PR #734, opened 2019-06-04
use manifest.json/actions.json args format for the config_panel.toml

Psycojoker:
## The problem

WIP work on this weekend, it depends on the #732 toml PR.

THIS REQUIRES A LITTLE BIT OF CHANGES FOR THE FORMAT OF THE CONFIG PANEL since now it's the manifest/actions format that is used.

The modifications are to this format in the [options] section, so, from head:

- bool -> boolean
- name -> ask

Here is a working (format) config https://gist.github.com/Psycojoker/c9ea335e596c4684b042812b7b0804c3

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1658,7 +1658,7 @@ def app_config_show_panel(app):
                     option["default"] = args_dict[option["name"]]
                 else:
                     logger.debug("Variable '%s' is not declared by config script, using default", generated_name)
-                    option["default"] = option["default"]
+                    # do nothing, we'll use the default if present
 
     return {
         "app_id": app_id,
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1635,24 +1635,28 @@ def app_config_show_panel(app):
         for section in tab.get("sections", []):
             section_id = section["id"]
             for option in section.get("options", []):
-                option_id = option["id"]
-                generated_id = ("YNH_CONFIG_%s_%s_%s" % (tab_id, section_id, option_id)).upper()
-                option["id"] = generated_id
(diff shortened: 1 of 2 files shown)


src/yunohost/app.py
```diff
@@ -1672,42 +1675,43 @@ def app_config_apply(app, args):
...
-
     app_id, app_instance_nb = _parse_app_instance_name(app)
     env = {
         "YNH_APP_ID": app_id,
         "YNH_APP_INSTANCE_NAME": app,
         "YNH_APP_INSTANCE_NUMBER": str(app_instance_nb),
     }
     args = dict(urlparse.parse_qsl(args, keep_blank_values=True)) if args else {}
+    print(json.dumps(args, sort_keys=True, indent=4))
+    print("------------------------------------------")
```
alexAubin: Randomly passing by but isn't this an old debug line ? (+ another json.dumps later)

Psycojoker: Is it yes, the PR is WIP, I'll remove it once it's done

Psycojoker:
Okay, that should be good, I've tested most of the things and that works for me ™

As a reference here is the script (and modify all the time) and config_panel I've used to test this feature. It's lazy as feuk but does its purpose.

@maniackcrudelis you can test it now if you want :)

I think that's all on the todo list before moving out of experimental? I'm not 100% sure but we probably start advertising it to apps dev if everything is fine and gather feedback once we feel ready.

I need to add the `.toml` format to the actions and the manifest too I think.

Psycojoker:
Ah yeah, another important part is: how the hell are we going to do the CLI interface? (except if we spawn ncurse but that looks a bit overkill here) (while I really would love to put an urwid interface in yunohost that sucks for scripting)

PR #737, opened 2019-06-17
[enh] Allow localhost ips to connect as root

zamentur:
## The problem

https://forum.yunohost.org/t/cannot-edit-files-using-users-or-admin-account/8183

## Solution

I suggest to add authorization to connect to root user through ssh from 127.0.0.1/8 et ::1 (localhost) 

## PR Status

Need to be tested

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/ssh/sshd_config
+++ b/data/templates/ssh/sshd_config
@@ -81,5 +81,5 @@ Match User sftpusers
 # user admin can't be used...
 # If the server is a VPS, it's expected that the owner of the
 # server has access to a web console through which to log in.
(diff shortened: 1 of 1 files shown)


alexAubin:
@zamentur : so do you know if it's working ? The other day you were saying it wasn't ?

alexAubin:
Discussing this with @zamentur , apparently this does not solve the use case for the initial issue so let's close it ...

Psycojoker:
> Discussing this with @zamentur , apparently this does not solve the use case for the initial issue so let's close it ...

But it doesn't solve other problems? This addition looks legit from here.

PR #741, opened 2019-06-19
Use set +x / -x to hide the whole ynh_script_progression computation

alexAubin:
## The problem

`ynh_script_progression` is awesome, but it fills the output log with a huge bunch of boring lines related to the computation of the progress bar :[ 

## Solution

Add some set +x / -x to hide this computation

## PR Status

Tested and working on my machine

## How to test

Run an app install like wordpress and compare the full output

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/logging
+++ b/data/helpers.d/logging
@@ -208,6 +208,7 @@ progress_string0="...................."
 # Define base_time when the file is sourced
 base_time=$(date +%s)
 ynh_script_progression () {
+	set +x
 	# Declare an array to define the options of this helper.
 	local legacy_args=mwtl
 	declare -Ar args_array=( [m]=message= [w]=weight= [t]=time [l]=last )
@@ -217,6 +218,7 @@ ynh_script_progression () {
 	local last
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
+	set +x
 	weight=${weight:-1}
 	time=${time:-0}
 	last=${last:-0}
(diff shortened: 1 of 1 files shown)


data/helpers.d/logging
```diff
@@ -280,6 +282,7 @@ ynh_script_progression () {
 	fi
 
 	ynh_print_info "[$progression_bar] > ${message}${print_exec_time}"
+	set -x
```
maniackcrudelis: Globally agree with it, but don't you think this `set -x` should be just before the `ynh_print_info`, so we would have the echo into the log. Useful to know where the script is.
Didn't try though...

alexAubin: Uh but the echo will be in the log as set +x/-x only affects the debug output of Bash

In fact, there will automatically be a set -x *inside* ynh_print_info because it relies on getopts, which itself also uses set -x

So from what I remember, the only usefulness of putting this set -x here (rather than before) is to hide a few additional `local foo` at the beginning of `ynh_print_info`

maniackcrudelis: Huh, is there any use of that last `set -x` as you said indeed, getops will do it.

alexAubin:
So uh proposing to merge soonish ? Not sure what to do about the `set -x` ... (it's mainly there to be explicit instead of having to guess that the other helper will do it)

maniackcrudelis:
It won't harm anyway

PR #742, opened 2019-06-19
[enh] Redact secrets from logs

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1363

Note when working on this, I realized that the "python argument" case is already covered, c.f. these : 
https://github.com/YunoHost/yunohost/blob/stretch-testing/src/yunohost/log.py#L224
https://github.com/YunoHost/yunohost/blob/stretch-testing/src/yunohost/user.py#L274

## Solution

So here's an attempt to redact at least *most* of the sensitive / secret data in the logs. So far I was able to cover : 
- "static" secrets like `/etc/yunohost/mysql`
- secrets provided through the install form (password-type args)
- secrets generated or fetched(!) during the script execution, like mysql db password.

The last item is the least easiest and least "clean" one. It's all built on the way bash logs or not information ... So for example, we are somewhat lucky that [this line](https://github.com/YunoHost/yunohost/blob/stretch-unstable/data/helpers.d/mysql#L202) only results in showing the secret for the first time as : 

```
local new_db_pwd=the_super_secret
```

(and there's no weird intermediate output already showing the secret). Otherwise, it would be much more difficult to catch and we might end up having to *rewrite already written stuff in the file* which would complicates stuff (though not impossible but ugh)

So the current mechanism relies on catching matches for this regex : `(db_pwd|password)=(\S{3,})$`

So for example : `+ local new_db_pwd=pT0QepguMxajbmQE5Ad4ZPgd`

and therefore covers mysql and postgresql dbs created with the standard helpers. This also covers *getting* passwords from the setting file, as done with e.g. `upload_password=$(ynh_app_setting_get --app=$app --key=upload_password)` (as long as the variable name ends with `password` ...). So there are corner cases that might not be covered like if the variable is instead named `admin_pass`... To be discussed.

Hopefully that's a relatively understandable explanation :sweat_smile: 

## PR Status

Tested on the app install of my_webapp on my side ... seems to be kinda working

## How to test

Run some app install (and also remove, upgrade, ... any special case that could show a password or other info we would want to redact)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -25,6 +25,7 @@
 """
 
 import os
+import re
 import yaml
 import collections
 
@@ -297,9 +298,21 @@ class RedactingFormatter(Formatter):
 
     def format(self, record):
         msg = super(RedactingFormatter, self).format(record)
+        self.identify_data_to_redact(msg)
         for data in self.data_to_redact:
             msg = msg.replace(data, "**********")
         return msg
+
+    def identify_data_to_redact(self, record):
+
+        # Wrapping this in a try/except because we don't want this to
+        # break everything in case it fails miserably for some reason :s
+        try:
+            match = re.search(r'(db_pwd|password)=(\S{3,})$', record.strip())
+            if match and match.group(2) not in self.data_to_redact:
+                self.data_to_redact.append(match.group(2))
+        except Exception as e:
+            logger.warning("Failed to parse line to try to identify data to redact ... : %s" % e)
 
 
 class OperationLogger(object):
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -375,8 +375,11 @@ class OperationLogger(object):
         Write or rewrite the metadata file with all metadata known
         """
 
+        dump = yaml.safe_dump(self.metadata, default_flow_style=False)
+        for data in self.data_to_redact:
+            dump = dump.replace(data, "**********")
         with open(self.md_path, 'w') as outfile:
-            yaml.safe_dump(self.metadata, outfile, default_flow_style=False)
+            outfile.write(dump)
 
     @property
     def name(self):
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -308,7 +308,9 @@ class RedactingFormatter(Formatter):
         # Wrapping this in a try/except because we don't want this to
         # break everything in case it fails miserably for some reason :s
         try:
-            match = re.search(r'(db_pwd|password)=(\S{3,})$', record.strip())
+            # This matches stuff like db_pwd=the_secret or admin_password=other_secret
+            # (the secret part being at least 3 chars to avoid catching some lines like just "db_pwd=")
+            match = re.search(r'(pwd|pass|password)=(\S{3,})$', record.strip())
             if match and match.group(2) not in self.data_to_redact:
                 self.data_to_redact.append(match.group(2))
         except Exception as e:
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -289,6 +289,19 @@ def is_unit_operation(entities=['app', 'domain', 'service', 'user'],
     return decorate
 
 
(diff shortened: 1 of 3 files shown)


alexAubin:
(Had to change the output format of `_parse_args_from_manifest` to also fetch the info of arg type, which makes the whole thing a bit more complicated, but there was no obvious way to do this in a simple way :/ )

Josue-T:
Hello,

We should coordinate better because I also worked on this, so I think there are enough work to don't do two time the same things...

Anyway thanks for your work, it might be better than what I wanted to do.

src/yunohost/log.py
```diff
@@ -289,6 +290,31 @@ def func_wrapper(*args, **kwargs):
...
+        for data in self.data_to_redact:
+            msg = msg.replace(data, "**********")
+        return msg
+
+    def identify_data_to_redact(self, record):
+
+        # Wrapping this in a try/except because we don't want this to
+        # break everything in case it fails miserably for some reason :s
+        try:
+            match = re.search(r'(db_pwd|password)=(\S{3,})$', record.strip())
```
Josue-T: What about adding a more generic thing to be able to support all variable containg `pwd`.

Maybe we could also try to add some other key word like `pass` or `token`.

src/yunohost/log.py
```diff
@@ -309,6 +335,8 @@ def __init__(self, operation, related_to=None, **kwargs):
         self.ended_at = None
         self.logger = None
         self._name = None
+        self.data_to_redact = []
+        self.data_to_redact.append(open("/etc/yunohost/mysql", "r").read().strip())
```
Josue-T: What about to add also the postgresql password in `/etc/yunohost/psql` if it exist ?

src/yunohost/log.py
```diff
@@ -345,9 +373,12 @@ def _register_log(self):
         Register log with a handler connected on log system
         """
 
-        # TODO add a way to not save password on app installation
         self.file_handler = FileHandler(self.log_path)
-        self.file_handler.formatter = Formatter('%(asctime)s: %(levelname)s - %(message)s')
+        # We use a custom formatter that's able to redact all stuff in self.data_to_redact
+        # N.B. : the stubtle thing here is that the class will remember a pointer to the list,
```
decentral1se:
```suggestion
        # N.B. : the subtle thing here is that the class will remember a pointer to the list,
```

alexAubin:
>We should coordinate better because I also worked on this, so I think there are enough work to don't do two time the same things...

Yea sorry about that... In the end of the meeting it wasn't so clear who was taking care of it and I already had many ideas in mind ... But clearly should have focused on other todo's. We'll try to be more careful next times :/

Thanks for speaking out

Psycojoker:
Untested but from here it looks good, I did a quick review and that seems pretty much ok.

alexAubin:
So uh what do you think folks, shall we include this in the upcoming 3.6 testing release, or do you think if should be tested more deeply before being released ?

src/yunohost/log.py
alexAubin: Should be done via the lasts commits :+1:

decentral1se:
go go go go go go go go go go go go

alexAubin:
Small note before I forget : this PR covered the operation logs (which are the ones usually shared with yunopaste on the forum) but doesn't cover the yunohost-cli and yunohost-api logs which might contain sensitive data as well

Josue-T:
> Small note before I forget : this PR covered the operation logs (which are the ones usually shared with yunopaste on the forum) but doesn't cover the yunohost-cli and yunohost-api logs which might contain sensitive data as well

Well, for the yunohost-cli and yunohost-api, what would be the solution. Will we need to make an other PR ? I started a work which fix also that. This work is still useful ?

PR #744, opened 2019-06-30
[enh] allows actions.toml

Psycojoker:
## The problem

json is boring to write and we've discovered toml as a better replacement for humans so after doing this for config panel, allow this for actions.

Manifest is next to come.

ALTOUGH.

There is a problem charly.

Debian being debian, python-toml is version 0.9 here, this mean that this kind of super handy shorthand isn't parsed with the version 0.9 (but is with the version 0.10 which is available on pypi):

```toml
description.en = "le anglish"
description.fr = "the frounce"
```

And there is not python-toml version 0.10 in testing or unstable.

This is really annoying.

Ideally I guess we need to ship it ourselves but I really wasn't expecting this additional work.

We can still write things like although:

```toml
[description]
en = "le anglish"
fr = "the frounce"
```

But in subcategories it's a bit annoying:

```toml
[this_is_an_actions.arguments.argument_one]
    name = "plop" # stuff needs to be before there
    [this_is_an_actions.arguments.argument_one.ask]
    en = "le anglish"
    fr = "the frounce"
```
## PR Status

Tested and working, python tells me that json and toml versions are identic.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -363,10 +363,10 @@ def app_info(app, show_status=False, raw=False):
 
         ret['upgradable'] = upgradable
         ret['change_url'] = os.path.exists(os.path.join(app_setting_path, "scripts", "change_url"))
-        
+
         with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
             manifest = json.load(json_manifest)
-        
+
         ret['version'] = manifest.get('version', '-')
 
         return ret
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
@Psycojoker, I don't know how I'm suppose to build the toml format for an action, do you have an implementation example ?

PR #747, opened 2019-07-06
Python hooks reloaded

alexAubin:
## The problem

This is a rewrite of https://github.com/YunoHost/yunohost/pull/535 because there were too many changes ... : 

Python hooks are required for the new diagnosis system modularity / extendability, and more generally if we want to implement some hooks in python for some reason. This could be the case for example of the famous `app ssowatconf` command which should be a classical regen-conf thing but can't be so far because it has too many python calls and json stuff to be implemented in bash.

## Solution

Split the behavior of the `hook_exec` function depending on the nature of the hook. Python hooks are to be modules loaded by yunohost and expected to have a `main` function.

## PR Status

Tested and working

## How to test

Create some hook with a main function (e.g. just displaying `hello world`) and call `yunohost hook list your_hook_type` and `yunohost diagnosis callback your_hook_type`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -462,7 +462,13 @@ def _hook_exec_python(path, args, env, loggers):
 
     # TODO : We might want to check here that it's a tuple
     # containing an int + a dict ?
-    return module.main(args, env, loggers)
+    ret = module.main(args, env, loggers)
+    assert isinstance(ret, tuple) \
+            and len(ret) == 2 \
+            and isinstance(ret[0],int) \
+            and isinstance(ret[1],dict), \
+            "Module %s did not return a (int, dict) tuple !" % module
+    return ret
 
 
 def _extract_filename_parts(filename):
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -460,9 +460,8 @@ def _hook_exec_python(path, args, env, loggers):
         sys.path = [dir_] + sys.path
     module = import_module(name)
 
-    # TODO : We might want to check here that it's a tuple
-    # containing an int + a dict ?
     ret = module.main(args, env, loggers)
+    # # Assert that the return is a (int, dict) tuple
     assert isinstance(ret, tuple) \
             and len(ret) == 2 \
             and isinstance(ret[0],int) \
--- a/src/yunohost/hook.py
+++ b/src/yunohost/hook.py
@@ -27,6 +27,7 @@ import os
 import re
 import sys
 import tempfile
+import mimetypes
 from glob import iglob
 from importlib import import_module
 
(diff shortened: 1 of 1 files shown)


src/yunohost/hook.py
```diff
@@ -311,14 +313,56 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
+    # Check the type of the hook (bash by default)
+    hook_type = "bash"
+    # (non-bash hooks shall start with something like "#!/usr/bin/env language")
+    hook_ext = os.path.splitext(path)[1]
+    if hook_ext == ".py":
+        hook_type = "python"
+    else:
+        # TODO / FIXME : if needed in the future, implement support for other
+        # languages...
+        assert hook_ext == "", "hook_exec only supports bash and python hooks for now"
```
Josue-T: Maybe we could add the support for the `.sh` extension if we want to specify that it's a bash hook.

alexAubin:
Merging soonish

src/yunohost/hook.py
```diff
@@ -311,14 +313,56 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
     # Validate hook path
     if path[0] != '/':
         path = os.path.realpath(path)
     if not os.path.isfile(path):
         raise YunohostError('file_does_not_exist', path=path)
 
+    # Check the type of the hook (bash by default)
+    hook_type = "bash"
+    # (non-bash hooks shall start with something like "#!/usr/bin/env language")
+    hook_ext = os.path.splitext(path)[1]
```
Psycojoker: https://gist.github.com/LeMeteore/cac019562822388a5d11c0ead629848c#file-transform-py-L35-L37 you might want to use this kind of code for file type detection (that also handle python script without a ".py" at the end but a shebang)

src/yunohost/hook.py
```diff
@@ -427,13 +451,36 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
+def _hook_exec_python(path, args, env, loggers):
+
+    dir_ = os.path.dirname(path)
+    name = os.path.splitext(os.path.basename(path))[0]
+
+    if not dir_ in sys.path:
+        sys.path = [dir_] + sys.path
+    module = import_module(name)
+
+    # TODO : We might want to check here that it's a tuple
```
Psycojoker: Aren't you actually doing the check just bellow '-' ?

src/yunohost/hook.py
```diff
@@ -427,13 +451,36 @@ def hook_exec(path, args=None, raise_on_error=False, no_trace=False,
...
+    # containing an int + a dict ?
+    ret = module.main(args, env, loggers)
```
Psycojoker: Shouldn't we spawn a new python process that launch the python instead of calling it in the current python process? That would be better security and bug wise I think (but idk how hard would it be to do that)

alexAubin: Hm I kinda chose to do this because that's supposedly less ressource expensive, considering that we're likely to import code from yunohost's in each of those (e.g. 10~15ish hooks in the context of diagnosis), each likely to do stuff like importing `requests` or other libs that take hell of time to load on RPi for example

alexAubin:
Proposing to merge soonish in 3.7 (not really needed or anything but at least that becomes stable code upon which the migration mechanism can be built and possibly other things)

PR #749, opened 2019-07-07
[enh] Dpkg vendor set to YunoHost

zamentur:
## The problem

`dpkg-vendor --query Vendor` returns "debian"
see https://wiki.debian.org/Derivatives/Guidelines#Vendor

## Solution

Create /etc/dpkg/origins/yunohost and add a symlink on /etc/dpkg/origins/default

## PR Status

Ready (tested by putting file and symlink by hand on a yunohost instance)

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/other/dpkg-origins
+++ b/data/other/dpkg-origins
@@ -1,0 +1,4 @@
+Vendor: YunoHost
+Vendor-URL: https://yunohost.org/
+Bugs: https://github.com/YunoHost/issues/
+Parent: Debian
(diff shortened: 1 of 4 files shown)


Psycojoker:
I wonder if that could break scripts that except to be runed on debian and check like that? Maybe it's worth to have that in a PR :x

Psycojoker:
No one seems to be using that in any of their apps so merging.

PR #699, opened 2019-04-02
Add size of apps in backup_info result

kay0u:
## The problem

This PR is used by: https://github.com/YunoHost/yunohost-admin/pull/236

## Solution

Add the size of each app

## PR Status

...

## How to test

...

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #750, opened 2019-07-08
[fix] Can't delete a system backup with webadmin

zamentur:
## The problem

Backup info fails to display system backup. So the user can't delete it in Webadmin.
```
File “/usr/lib/python2.7/dist-packages/moulinette/interfaces/api.py”, line 439, in process
ret = self.actionsmap.process(arguments, timeout=30, route=_route)
File “/usr/lib/python2.7/dist-packages/moulinette/actionsmap.py”, line 523, in process
return func(**arguments)
File “/usr/lib/moulinette/yunohost/backup.py”, line 2292, in backup_info
key_info[“size”] = info[“size_details”][category][name]
TypeError: list indices must be integers, not str
```

https://forum.yunohost.org/t/impossible-de-supprimer-une-archive-de-sauvegarde/8393

## Solution

Transform the list of string of hook by an object like: 
```
{ 'hooks': ["..."], 'size': 542}
```

## PR Status
Ready
Tested on a yunohost with same bug by running `yunohost backup info`

## How to test

Try to get info on old backup with system hook include (not only with apps)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -2380,7 +2380,7 @@ def backup_info(name, with_details=False, human_readable=False):
                     if name in info["size_details"][category].keys():
                         key_info["size"] = info["size_details"][category][name]
                         if human_readable:
-                            key_info["size"] = binary_to_human(key_info['size']) + 'B'
+                            key_info["size"] = binary_to_human(key_info["size"]) + 'B'
                     else:
                         key_info["size"] = -1
                         if human_readable:
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -2372,13 +2372,18 @@ def backup_info(name, with_details=False, human_readable=False):
             for category in ["apps", "system"]:
                 for name, key_info in info[category].items():
                     if name in info["size_details"][category].keys():
-                        key_info["size"] = info["size_details"][category][name]
+                        size = info["size_details"][category][name]
                         if human_readable:
-                            key_info["size"] = binary_to_human(key_info["size"]) + 'B'
+                            size = binary_to_human(size) + 'B'
                     else:
-                        key_info["size"] = -1
+                        size = -1
                         if human_readable:
-                            key_info["size"] = "?"
+                            size = "?"
+
+                    if category == "system":
+                        info[category][name] = {'hooks': info[category][name], 'size': size}
+                    else:
+                        key_info['size'] = size
 
         result["apps"] = info["apps"]
         result["system"] = info[system_key]
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -2372,7 +2372,7 @@ def backup_info(name, with_details=False, human_readable=False):
             for category in ["apps", "system"]:
                 for name, key_info in info[category].items():
                     # Retrocompatibility with archives produced 
-                    # around january 2019
+                    # until 3.5.x
                     if not isinstance(key_info, dict):
                         key_info = {x: 'succeed' for x in key_info}
                         info[category][name] = key_info
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -2371,19 +2371,20 @@ def backup_info(name, with_details=False, human_readable=False):
         if "size_details" in info.keys():
             for category in ["apps", "system"]:
                 for name, key_info in info[category].items():
+                    # Retrocompatibility with archives produced 
+                    # around january 2019
+                    if not isinstance(key_info, dict):
+                        key_info = {x: 'succeed' for x in key_info}
+                        info[category][name] = key_info
+                        
                     if name in info["size_details"][category].keys():
-                        size = info["size_details"][category][name]
+                        key_info['size'] = info["size_details"][category][name]
                         if human_readable:
-                            size = binary_to_human(size) + 'B'
+                            key_info['size'] = binary_to_human(key_info['size']) + 'B'
                     else:
-                        size = -1
+                        key_info['size'] = -1
                         if human_readable:
-                            size = "?"
-
-                    if category == "system":
-                        info[category][name] = {'hooks': info[category][name], 'size': size}
-                    else:
-                        key_info['size'] = size
+                            key_info['size'] = "?"
 
         result["apps"] = info["apps"]
         result["system"] = info[system_key]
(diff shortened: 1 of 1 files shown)


zamentur:
I remove flag important, cause i was thinkig about case where the server could completely break due to this bug, but after think on it better it doesn't.

alexAubin:
Uuuuh do you have a step by step way to reproduce this ? I can't reproduce the issue nor with an old backup (`backup_system_from_2p4` used by pytests) or with a fresh backup ... (with and without the `--with-details` option)

zamentur:
Indeed i just check with a new backup, and format is different!

New format:
```
{"size_details": {"apps": {}, "system": {"conf_ldap": 4134}}, "description": "", "created_at": 1562680193, "apps": {}, "system": {"conf_ldap": {"/usr/share/yunohost/hooks/backup/05-conf_ldap": "succeed"}}, "size": 118861}
```

Old format generated on 2019/01/02 (probably with the testing at this date):
```
{"size_details": {"apps": {}, "system": {"conf_ldap": 3692}}, "description": "", "created_at": 1546422083, "apps": {}, "system": {"conf_ldap": ["/usr/share/yunohost/hooks/backup/05-conf_ldap"]}, "size": 116924}
```
I was not aware of this change.

So to not break system we need to convert properlly old backup info.

alexAubin:
But uh then is there a bug in any stable in the past, and if so how can someone reproduce it ? x_X

zamentur:
With this bug, all people with old backup (in this old format) have their webadmin broken and are obliged to remove the backup manually.

The change of format is due to this change i think:
https://github.com/YunoHost/yunohost/commit/16245f53d9591fcfeca70e5c54604ba3d26c2185

I don't know if this old format is a bug or not of the past stable. I think no. The bug seems created by this PR and the commit above: https://github.com/YunoHost/yunohost/pull/699 (merged in 3.6).

zamentur:
SO to reproduce, i think you just need to run:
On 3.5.x
```
yunohost backup create --system conf_ldap -n toto
```
Upgrade to 3.6.x
And next, try to get info (or to remove the backup from webadmin)
```
yunohost backup info toto
```

src/yunohost/backup.py
```diff
@@ -2371,6 +2371,12 @@ def backup_info(name, with_details=False, human_readable=False):
         if "size_details" in info.keys():
             for category in ["apps", "system"]:
                 for name, key_info in info[category].items():
+                    # Retrocompatibility with archives produced 
+                    # around january 2019
+                    if not isinstance(key_info, dict):
+                        key_info = {x: 'succeed' for x in key_info}
+                        info[category][name] = key_info
+                        
                     if name in info["size_details"][category].keys():
```
alexAubin: Hmyea in fact I don't know if it's so interesting to have all these `{'path1':'succeed', 'path2':'succeed'}` ... I'm not even sure that the value can be different than suceed ... So maybe having the dict is a buggy behavior (probably my fault ?) that wasn't noticed when implementing the whole `size_details` stuff and we should just stick with a list ?

zamentur: Indeed, only succeed are listed in info :
https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/backup.py#L613

zamentur:
Note if it was working with `backup_system_from_2p4` it was because YunoHost 2.4 backup info doesn't include 'size_details' key. So no sizes are added with this backup, in the webadmin in this case it display '0'.

src/yunohost/backup.py
zamentur: I try to create an history of this format:
2.4:
```
{"hooks": {"conf_ldap": ["/usr/share/yunohost/hooks/backup/05-conf_ldap"]}, "created_at": 1494198169, "apps": {}, "description": "", "size": 524892}
```

3.5, size_details and hooks is replaced by system
```
{"size_details": {"apps": {}, "system": {"conf_ldap": 3692}}, "description": "", "created_at": 1546422083, "apps": {}, "system": {"conf_ldap": ["/usr/share/yunohost/hooks/backup/05-conf_ldap"]}, "size": 116924}
```

Current 3.6, use a dict instead of a list
```
{"size_details": {"apps": {}, "system": {"conf_ldap": 4134}}, "description": "", "created_at": 1562680193, "apps": {}, "system": {"conf_ldap": {"/usr/share/yunohost/hooks/backup/05-conf_ldap": "succeed"}}, "size": 118861}
```

So now, if we change again, it will complexify the code. We could also decide to stop retrocompatibility of 2.4 format to simplify some code or decide to hotfix quickly 3.6 and ignore this new 3.6 format.

In a perfect world, and with the current features (not the future one's), i think the format should be:
```
{"description": "", "created_at": 1562680193, "apps": {}, "system": {"conf_ldap": {"hook": "/usr/share/yunohost/hooks/backup/05-conf_ldap", "size": 4134}}, "size": 118861}
```
So no size_details, and a dict to put some meta info about this "part of backup".

alexAubin: >In a perfect world, and with the current features (not the future one's), i think the format should be:
>
>`"system": {"conf_ldap": {"hook": "/usr/share/yunohost/hooks/backup/05-conf_ldap", "size": 4134}}`

We can't have this, because the value associated to `conf_ldap` could contain several paths ... so at least the value corresponding to `hook` should be a list (meh). That's why in 3.6 it's a dict of path ... but imho having a dict with just "succeed" as value is a bug of 3.6, and we should just revert to having simple list of paths ...

zamentur: Note, it could be a good idea to register in info, system_part and apps failed to backup... Indeed in webadmin it could be usefull to display a backup and see part in failures... So the idea of succeed/failed is not so bad.

And after some test, i discover it's possible to have several hooks attached to the same system part ! That's why a list was used and not just a string...

zamentur: > so at least the value corresponding to hook should be a list (meh)
It's exactly what the PR does

zamentur:
Are you ok to merge this ?

Psycojoker:
Don't know the details but ok to merge on principle.

src/yunohost/backup.py
alexAubin: (Follow-up : we discussed this IRL and this is not what the PR does. Ideally we should revert the thing to not store those useless "success" flags in the info.json, while still adding legacy code to support the buggy/legacy 3.5~3.6 format)

alexAubin:
Flagging this a 3.7, we really need a fix for this...

alexAubin:
Finally fixed it though 7d0119ad since I like finishing other people's PR so much...

PR #751, opened 2019-07-09
[enh] allow to have arguments in manifest for the remove script

Psycojoker:
## The problem

It's not possible to ask questions to user during the removal of an app, this cause problem for undefined behavior where we aren't sure what is the best course of action. The most common case is: should we keep nextcloud data?

## Solution

Allow to ask questions to user.

A serie of personal suggestions:

- I think that all apps should provide default value for all remove arguments
- the app groupe should discuss the general question of "should we keep app data?" default answer to uniformize this behavior amongs all apps, I'm expecting this to be the number of question that is going to pop out

## PR Status

Tested on my installation, works as excepted.

THE ADMIN INTERFACE PART IS NOT DONE. Mostly because bower fails without any reason for me x(

## How to test

You simply need to add a "remove" section in "arguments" exactely the same one than "install" in a random app.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -959,12 +959,13 @@ def app_install(operation_logger, app, label=None, args=None, no_remove_on_failu
 
 
 @is_unit_operation()
-def app_remove(operation_logger, app):
+def app_remove(operation_logger, app, args=None):
     """
     Remove app
 
     Keyword argument:
         app -- App(s) to delete
+        args -- Serialize arguments for app remove, if needed
 
     """
     from yunohost.utils.ldap import _get_ldap_interface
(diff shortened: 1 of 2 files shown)


src/yunohost/app.py
```diff
@@ -993,16 +994,41 @@ def app_remove(operation_logger, app):
...
 
-    env_dict = {}
+    args_dict = {} if not args else \
+        dict(urlparse.parse_qsl(args, keep_blank_values=True))
+
+    # Retrieve arguments list for remove script
+    if "remove" in manifest.get("arguments", {}) and manifest["arguments"]["remove"]:
+        args_odict = _parse_args_from_manifest(manifest, 'remove', args=args_dict)
+        args_list = [value[0] for value in args_odict.values()]
+        args_list.append(app)
```
alexAubin: Uh I have no idea what I'm talking about but shouldn't `app` be in first position so that it corresponds to `$1` ? (Well I guess anyway all recents apps likely to require argument for remove don't rely on `$1, $2, ...` anymore)

src/yunohost/app.py
```diff
@@ -993,16 +994,41 @@ def app_remove(operation_logger, app):
     os.system('chown -R admin: /tmp/yunohost_remove')
     os.system('chmod -R u+rX /tmp/yunohost_remove')
 
-    args_list = [app]
+    with open(os.path.join(APPS_SETTING_PATH, app, 'manifest.json')) as json_manifest:
+        manifest = json.load(json_manifest)
```
alexAubin: To be kept in mind that this should be replaced by the `get_manifest_for_app` thing from the other toml manifest PR

alexAubin:
Discussed in today's meeting /  not the right way to go, the only use case I can think of is the --purge option, which is handled through a generic core arg nowadays

PR #752, opened 2019-07-09
[fix] Fix values starting by - in getopts

maniackcrudelis:
## The problem

Getopts fails to parse a value if it's start by -. It's misread as another option in the argument list.
https://github.com/YunoHost-Apps/rainloop_ynh/issues/58

## Solution

Escape - if it's a value of an option.

## PR Status

Tested on rainloop.

## How to test

Try an value starting by -.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -76,6 +76,8 @@ ynh_handle_getopts_args () {
 			# ${#arguments[@]} is the size of the array
 			for arg in `seq 0 $(( ${#arguments[@]} - 1 ))`
 			do
+				# Escape options' values starting with -. Otherwise the - will be considered as another option.
+				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]}-/--${args_array[$option_flag]}\\-}"
 				# And replace long option (value of the option_flag) by the short option, the option_flag itself
 				# (e.g. for [u]=user, --user will be -u)
 				# Replace long option with =
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -77,7 +77,7 @@ ynh_handle_getopts_args () {
 			for arg in `seq 0 $(( ${#arguments[@]} - 1 ))`
 			do
 				# Escape options' values starting with -. Otherwise the - will be considered as another option.
-				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]}-/--${args_array[$option_flag]}\\-}"
+				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]}-/--${args_array[$option_flag]}\\TOBEREMOVED\\-}"
 				# And replace long option (value of the option_flag) by the short option, the option_flag itself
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
`${var/%Pattern/Replacement}` is only for the suffix, where `${var/#Pattern/Replacement}` is for the prefix. As `${var%Pattern}` and `${var#Pattern}`.
Here, `${var//Pattern/Replacement}` is the "Global replacement" (just before in the doc you linked), affecting the entire variable, not only the beginning or the end.

kay0u:
I thought you were trying to escape the `-` at the end. I should run it with debug option to well understand it.

maniackcrudelis:
No, I escape a potential `-` just after the argument and its `=`. For values starting by `-`, like it was for rainloop with a negative timezone.

kay0u:
I didn't well understand what was inside `arguments[arg]`, I get it now after read the code again!

maniackcrudelis:
`arguments` is an array that contains all arguments sent to the function, with each argument into a cell of the array.
`arguments[arg]` is the array read cell by cell. So that's just one argument at a time.

kay0u:
That's what I finally understood, thx! (And I didn't noticed that you've edited your post where you explain it clearly).

kay0u:
I found a bug /o\

With this:
```
ynh_app_setting_set --app=$app --key=wifi_device_id --value=-1
```

```
sudo yunohost app setting $app wifi_device_id
```
return: 
> \\-1

maniackcrudelis:
Not sure to get it.
Could you explain me what your were expecting and what you had ?

kay0u:
I would like to get a -1, but I got a \\-1

maniackcrudelis:
;) so far I got it.

`ynh_app_setting_set --app=$app --key=wifi_device_id --value=-1` gives you a \-1 into the settings.yml file?

kay0u:
In my settings.yml, I have this value:

> wifi_device_id: \\-1

maniackcrudelis:
!!!
That's really strange !
Like it would have removed the third parameter.

Could you provide a complete log of this part of the process ?

kay0u:
https://paste.yunohost.org/raw/lowuwetuni

code:

https://github.com/labriqueinternet/piratebox_ynh/blob/testing/scripts/install

maniackcrudelis:
I don't know why yet, but at least I do reproduce that error.

maniackcrudelis:
Guess the problem was, as I've tested with sed (which was the initial problem), it was interpreting \. Which isn't done by ynh_app_setting.

So, I found a work around by clearly escaping (to not miss interpret) and removing the escape character before storing the value into the variable.

Psycojoker:
Is it ready to merge then?

maniackcrudelis:
Yep, I think it is

Psycojoker:
Mergin' then :)

PR #753, opened 2019-07-10
[enh] Upgrade n version (also for compatibility with arm64)

yalh76:
## The problem

1. When installing node js using ynh_install_nodejs on aarch64, it's the x86-64 version of nodejs that is installed
2. N - Node.js version management included in YunoHost is version 2.1.7, one year old

## Solution

Updating N - Node.js version management version
since N version 3.0.0 aarch64 and arm64 are detected

## PR Status

Done

## How to test

Have and aarch64 or arm64 YunoHost, install a package that request nodejs like wekan: yunohost app install https://github.com/yalh76/wekan_ynh/tree/2.98 --debug

After the installation, just test someting like: `/opt/node_n/bin/node --version`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -16,8 +16,8 @@ ynh_install_n () {
 	ynh_print_info --message="Installation of N - Node.js version management"
 	# Build an app.src for n
 	mkdir -p "../conf"
-	echo "SOURCE_URL=https://github.com/tj/n/archive/v2.1.7.tar.gz
-SOURCE_SUM=2ba3c9d4dd3c7e38885b37e02337906a1ee91febe6d5c9159d89a9050f2eea8f" > "../conf/n.src"
+	echo "SOURCE_URL=https://github.com/tj/n/archive/v4.1.0.tar.gz
+SOURCE_SUM=3983fa3f00d4bf85ba8e21f1a590f6e28938093abe0bb950aeea52b1717471fc" > "../conf/n.src"
 	# Download and extract n
 	ynh_setup_source --dest_dir="$n_install_dir/git" --source_id=n
 	# Install n


yalh76:
Bump :)

Psycojoker:
LGTM by principle but I'm not really in the app group '-'

alexAubin:
Looks okay to me on the principle, but what about existing instance ? Is a migration required somehow or will it happen automagically when upgrading apps or something ?

yalh76:
> 
> 
> Looks okay to me on the principle, but what about existing instance ? Is a migration required somehow or will it happen automagically when upgrading apps or something ?

Maybe @maniackcrudelis can have a better response than mine as he seems to better know that helper.

After reading the code of this helper, n won't be upgraded automagically but as far as I understand the helper, a migration to the new version is not required

maniackcrudelis:
Haven't been noticed for this PR by github...
Maybe I should stop working on nuclear weapons :thinking: ...

N won't be upgraded, this function is called only if n isn't already install.
https://github.com/YunoHost/yunohost/blob/ce4fb552aee393fd9a782ce1cbfa5a743c9278fb/data/helpers.d/nodejs#L90-L94

alexAubin:
>N won't be upgraded, this function is called only if n isn't already install.

Uuuuh wokay, so shall we edit this helper to check that the version installed is the one expected by yunohost and upgrade it otherwise ?

maniackcrudelis:
Since the condition is testing the version, and that we now have an upgrade of this version.
Guess we could adapt the helper to update n in case the version is different.

yalh76:
After making a test, it seems that N is automagically updated....

alexAubin:
Weeell too lazy to test this, trusting you, that looks okay, yolomergin :+1:

PR #754, opened 2019-07-12
[fix] Path not displayed in log_corrupted_md_file

zamentur:
## The problem
An error is displayed without md_file var

"Le fichier yaml de metadata associé aux logs est corrompu : '{md_file}'"

## Solution

Use read_yaml moulinette helper to let us know on which file and error of yaml loading is triggered

## PR Status

Ready

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -106,7 +106,7 @@ def log_list(category=[], limit=None, with_details=False):
                     try:
                         metadata = yaml.safe_load(md_file)
                     except yaml.YAMLError:
-                        logger.warning(m18n.n('log_corrupted_md_file', file=md_path))
+                        logger.warning(m18n.n('log_corrupted_md_file', md_file=md_path))
 
                     entry["success"] = metadata.get("success", "?") if metadata else "?"
 
@@ -192,7 +192,7 @@ def log_display(path, number=50, share=False):
                 if 'log_path' in metadata:
                     log_path = metadata['log_path']
             except yaml.YAMLError:
-                error = m18n.n('log_corrupted_md_file', file=md_path)
+                error = m18n.n('log_corrupted_md_file', md_file=md_path)
                 if os.path.exists(log_path):
                     logger.warning(error)
                 else:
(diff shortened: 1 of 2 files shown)


src/yunohost/log.py
```diff
@@ -106,7 +106,7 @@ def log_list(category=[], limit=None, with_details=False):
                     try:
                         metadata = yaml.safe_load(md_file)
                     except yaml.YAMLError:
-                        logger.warning(m18n.n('log_corrupted_md_file', file=md_path))
+                        logger.warning(m18n.n('log_corrupted_md_file', md_file=md_path))
```
alexAubin: I will happily merge this and yolorelease it so we can investigate the bug, but can we also get the actual error message so we can know what's going on with those file ? (Since apparently the issue tends to disappear by itself as we saw with messages on the forum)

In fact, there's a moulinette helper called `read_yaml` that does exactly this ...

zamentur: Done

src/yunohost/log.py
```diff
@@ -192,7 +187,7 @@ def log_display(path, number=50, share=False):
                 if 'log_path' in metadata:
                     log_path = metadata['log_path']
             except yaml.YAMLError:
-                error = m18n.n('log_corrupted_md_file', file=md_path)
+                error = m18n.n('log_corrupted_md_file', md_file=md_path)
```
alexAubin: Well it would be nice to just use `read_yaml` here too ...

But I can't understand what's the intention in putting the `info['medata']` things inside the `try` block, nor what's the intention with displaying just a warning or raising an exception depending if `log_path` actually exists ... Can't we factorize out the loading of the yaml ?

Psycojoker:
Tried to finish the PR following the comment so we can merge.

Don't hesitate to modify if it's not ok.

alexAubin:
Cheers, will release this as a hotfix along with a few other hotfixes I can't remember

PR #757, opened 2019-07-20
[mod] on reset ldap password temp password display is not needed anymore

Psycojoker:
## The problem

Tested the command, it's not needed anymore to enter the password by hand. 

This also means that this script is not needed anymore in shell since "yunohost tools adminpw" don't ask for the password anymore, but in case someone still use it let's clean it.

@alexAubin I'm not micromerging because I want your validation but I think we can directly do it (and simplify much more this script but who knows we might need it?)

## PR Status

Working.

## How to test

Launch the script.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/sbin/yunohost-reset-ldap-password
+++ b/sbin/yunohost-reset-ldap-password
@@ -40,13 +40,6 @@ service slapd start
 #   Properly set new admin password   #
 #######################################
 
-# Display tmp password to user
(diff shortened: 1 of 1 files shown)


Psycojoker:
Aaaaand, I wasn't up to date with master >_>

PR #758, opened 2019-07-20
[fix] moulinette logs were never displayed #lol

Psycojoker:
Needed for https://github.com/YunoHost/moulinette/pull/205

For some reason logs from core of the moulinette were never displayed at all since 4 years :man_shrugging:

### DIFF
--- a/bin/yunohost
+++ b/bin/yunohost
@@ -145,7 +145,7 @@ def _init_moulinette(debug=False, quiet=False):
             },
             'moulinette': {
                 'level': level,
-                'handlers': [],
+                'handlers': handlers,
                 'propagate': True,
             },
             'moulinette.interface': {


Psycojoker:
Will merge in next days of no oppositions.

alexAubin:
Agreed on the principle but : 
- you might want to do the same thing on `yunohost-api` for consistency ?
- we might want to move the info message popping up every time you run a command about `running action foo_bar with args {"arg1":...}` to a debug level (and there are some other stuffs in Moulinette that we might want to consider moving to debug level)

Psycojoker:
> you might want to do the same thing on `yunohost-api` for consistency ?

I'm not sure, this is really a different context but maybe? Don't know how adapted it is here :x (I don't use the admin that much)

>     * we might want to move the info message popping up every time you run a command about `running action foo_bar with args {"arg1":...}` to a debug level (and there are some other stuffs in Moulinette that we might want to consider moving to debug level)

Yup, make total sens, there is probably some cleaning to do.

alexAubin:
>I'm not sure, this is really a different context but maybe? Don't know how adapted it is here :x (I don't use the admin that much)

Well eh, just that if the goal is to warn the user after lock taking a long time to acquire, you also want to have the warning being displayed in the webadmin and not just when using the cli :s

Psycojoker:
> > I'm not sure, this is really a different context but maybe? Don't know how adapted it is here :x (I don't use the admin that much)
> 
> Well eh, just that if the goal is to warn the user after lock taking a long time to acquire, you also want to have the warning being displayed in the webadmin and not just when using the cli :s

Ah yes, indeed! I need to check how it works in the admin.

Psycojoker:
> Ah yes, indeed! I need to check how it works in the admin.

So, I've took the time to check it and actually the admin is working with timeout and is raising this error message https://github.com/YunoHost/moulinette/blob/stretch-unstable/moulinette/core.py#L524 which turns out to totally sucks from an UX point of view, so I've proposed another wording here: https://github.com/YunoHost/moulinette/pull/222

We can also think about changing this mechanism in the admin but I don't know if we want to? It would also requires quite some more work and I don't know if we want to delay 3.7.x even more :/ (and there are other things to finish)

alexAubin:
Uuuh not sure I understand how this timeout thing relates to the fact that regular moulinette logs are not displayed ?

Psycojoker:
> Uuuh not sure I understand how this timeout thing relates to the fact that regular moulinette logs are not displayed ?

Replying to

> you also want to have the warning being displayed in the webadmin and not just when using the cli :s

The warning won't display in admin because it's not the same "blocking" mechanism that is using in admin, the admin uses a timeout to get the lock, the CLI will wait for ever.

I don't know if it's clearer :/? (sorry I'm really tired)

Psycojoker:
Propose to merge soon.

Psycojoker:
Delayed to 3.8, 3.7 is already too big.

alexAubin:
zzzzz, so merged this thinking "how bad can it be", and it's pretty bad. Apparently any stuff playing with regen conf displays a bunch of super weird messages for some reasons ...

2019-11-04-225444_1366x768_scrot

So for now this is reverted by 6a0959dd ....

Psycojoker:
Ah shit x( ...

PR #760, opened 2019-07-20
Add apache2 as conflicting package

alexAubin:
## The problem

In various situation, we end up with people voluntarily installing apache2, or various dependency stuff in apt/dpkg ending up willing to install apache2 because some packages have it as recommended or whatever ...

This ends up breaking instances because once the server restarts, apache may start before nginx

## Solution

Add apache2 as a damn conflict to end this madness. YunoHost is a packaged flagged as "Essential" in Debian control file, so YunoHost should win in a fight against apache2 (or at least, even trying to install apache2 explicitly manually will result in a "Are you really sure?" question from apt)

## PR Status

Yolocommited, not really tested but meh

## How to test

Well eh find a way to properly update the control file of an existing install and try to install apache2...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -43,6 +43,7 @@ Conflicts: iptables-persistent
  , yunohost-config-dovecot, yunohost-config-slapd
  , yunohost-config-nginx, yunohost-config-amavis
  , yunohost-config-mysql, yunohost-predepends
+ , apache2
 Replaces: moulinette-yunohost, yunohost-config
  , yunohost-config-others, yunohost-config-postfix
  , yunohost-config-dovecot, yunohost-config-slapd


Psycojoker:
I sadly know valid situations where people will change apache2 port and put it behind nginx so I'm not sure we can do that :s (which is a shame because preventing apache2 from being used is really a good idea for a lot of situations of uninformed users)

alexAubin:
Hmokay so agreed that this PR is not the right way to go ...

Alternatively : we can try to add small config stuff to prevent apache2 from being installed ... for example that kind of thing is already done to try to prevent networkmanager from being installed on ARM boards (because it completely messes up the whole /etc/resolv.conf) : 

https://github.com/YunoHost/arm-images/blob/master/userpatches/customize-image.sh#L23-L30

If we put this, apache2 shouldn't be inadvertendly installed ... but I think a user could still manage to install it (well in the worst case, only need to remove this piece of config in /etc/apt/preferences)

Psycojoker:
That would solve this problem on installation of YunoHost, right? I though to PR was for a running YunoHost 

Or do you mean this part? https://github.com/YunoHost/arm-images/blob/master/userpatches/customize-image.sh#L26-L30

alexAubin:
What I have in mind is adding some stuff to the regen-conf mechanism somehow to add pieces of config similar to [this](https://github.com/YunoHost/arm-images/blob/master/userpatches/customize-image.sh#L26-L30) in the appropriate file, so that would work both for existing instances and new instances

Still a random thought though, would need to actually test it to validate that it'd work and make sense

Psycojoker:
Ok, LGTM as a potential solution :+1:

alexAubin:
Closing, will work on another solution some day i feel like ranting again

PR #762, opened 2019-07-24
[enh] PostgreSQL password security

madtibo:
## The problem

PostgreSQL authentication are done using clear text password.

## Solution

Use an authentication using md-5 hashed password.
This should be transparent for all applications.

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/postgresql
+++ b/data/helpers.d/postgresql
@@ -283,11 +283,11 @@ ynh_psql_test_if_first_run() {
 
 		sudo --login --user=postgres psql -c"ALTER user postgres WITH PASSWORD '$psql_root_password'" postgres
 
-		# force all user to connect to local database using passwords
+		# force all user to connect to local databases using hashed passwords
 		# https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html#EXAMPLE-PG-HBA.CONF
 		# Note: we can't use peer since YunoHost create users with nologin
 		#  See: https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/user
-		ynh_replace_string --match_string="local\(\s*\)all\(\s*\)all\(\s*\)peer" --replace_string="local\1all\2all\3password" --target_file="$pg_hba"
+		ynh_replace_string --match_string="local\(\s*\)all\(\s*\)all\(\s*\)peer" --replace_string="local\1all\2all\3md5" --target_file="$pg_hba"
 
 		# Advertise service in admin panel
 		yunohost service add postgresql --log "$logfile"
--- a/locales/en.json
+++ b/locales/en.json
@@ -331,6 +331,7 @@
     "migration_description_0009_decouple_regenconf_from_services": "Decouple the regen-conf mechanism from services",
     "migration_description_0010_migrate_to_apps_json": "Remove deprecated appslists and use the new unified 'apps.json' list instead",
     "migration_description_0011_setup_group_permission": "Setup user group and setup permission for apps and services",
+    "migration_description_0012_postgresql_password_to_md5_authentication": "Force postgresql authentication to use md5 for local connections",
     "migration_0003_backward_impossible": "The stretch migration cannot be reverted.",
     "migration_0003_start": "Starting migration to Stretch. The logs will be available in {logfile}.",
     "migration_0003_patching_sources_list": "Patching the sources.lists…",
--- a/src/yunohost/data_migrations/0011_postgresql_password_to_md5_authentication.py
+++ b/src/yunohost/data_migrations/0011_postgresql_password_to_md5_authentication.py
@@ -1,0 +1,21 @@
+import glob
+import re
+from yunohost.tools import Migration
+from moulinette.utils.filesystem import chown
+
+
+class MyMigration(Migration):
+
+    "Force authentication in md5 for local connexions"
+
+    all_hba_files = glob.glob("/etc/postgresql/*/*/pg_hba.conf")
+
+    def forward(self):
+        for filename in self.all_hba_files:
+            pg_hba_in = read_file(filename)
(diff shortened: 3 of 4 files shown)


Psycojoker:
Hello,

I don't see any commit by you in https://github.com/YunoHost/yunohost/pull/762/commits, perhaps you've pushed the wrong branch?

Also today md5 is sadly close to being clear text :/ We should at least aim for a minimum of sha-256 and if possible as much as sha3-512 if it's supported.

Any, thanks a lot for your contribution :heart:

madtibo:
Oups, you are totally right!
Here is the commit  :-)

md-5 is the most secure solution for authentication in PostgreSQL 9.6.
There is a more secure option in version 10 (scram-sha-256).

Psycojoker:
We are moving to (postgresql version 11 in buster)[https://packages.debian.org/buster/postgresql], I guess it will really be the right time to update that code again at that time cc @alexAubin 

Since md5 is really really weak (it's so fast to compute that it's more or less plaintext from every security stuff I've read) I'm not sure it's worth it to upgrade to it but I guess it can't do any damage? I don't have any strong opinion on that.

madtibo:
In version 11, md5 will permit both md5 and scram-sha-256 authentication, depending on the way the password is generated. There should be no need to update the code.

Light security is always better than no security. Even if md5 is weak, it is much better than clear text authentication.

data/helpers.d/postgresql
```diff
@@ -282,11 +282,11 @@ ynh_psql_test_if_first_run() {
 
 		sudo --login --user=postgres psql -c"ALTER user postgres WITH PASSWORD '$psql_root_password'" postgres
 
-		# force all user to connect to local database using passwords
+		# force all user to connect to local databases using hashed passwords
 		# https://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html#EXAMPLE-PG-HBA.CONF
 		# Note: we can't use peer since YunoHost create users with nologin
 		#  See: https://github.com/YunoHost/yunohost/blob/unstable/data/helpers.d/user
-		ynh_replace_string --match_string="local\(\s*\)all\(\s*\)all\(\s*\)peer" --replace_string="local\1all\2all\3password" --target_file="$pg_hba"
+		ynh_replace_string --match_string="local\(\s*\)all\(\s*\)all\(\s*\)peer" --replace_string="local\1all\2all\3md5" --target_file="$pg_hba"
```
alexAubin: Hm so this will happen for instances installing postgresql for the first time, but what about existing instances ? Is this just a matter of replacing a line in a conf file ?

If so we can easily write a migration for existing instance using our migration mechanism. If you want to do it, there's a simple one here : https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/data_migrations/0001_change_cert_group_to_sslcert.py
(or we can take care of it if you don't feel confident writing / testing it ;))

alexAubin:
Just to understand better : here we're talking about how passwords are stored on the system right ?

madtibo:
> Just to understand better : here we're talking about how passwords are stored on the system right ?

No, the password are always stored hashed (either md5 or scram-sha-256 from version 10).
The pg_hba configuration set the way the authentication is done.
The documentation states:
```The method password sends the password in clear-text and is therefore vulnerable to password “sniffing” attacks. It should always be avoided if possible. If the connection is protected by SSL encryption then password can be used safely, though. (Though SSL certificate authentication might be a better choice if one is depending on using SSL). ```

Since we do not enforce SSL only connections, the password method should not be used.

alexAubin:
Uh wokay ? I'm not sure to understand how an auth mechanism can work other than sending the cleartext password or using asymetric keys :D Do you have a link on that ?

data/helpers.d/postgresql
madtibo: I created a new file `src/yunohost/data_migrations/0011_postgresql_password_to_md5_authentication.py`.
Does it do what you had in mind?

madtibo:
Here is a description from the documentation: https://www.postgresql.org/docs/9.6/auth-methods.html#AUTH-PASSWORD

alexAubin:
Okay I think I also found https://stackoverflow.com/a/34098797 so from what I understands, the server sends a nonce and the client is supposed to hash some info, and so the server side

Dunno much about how bad md5 is, but my understanding is that it's especially vulnerable to collision attacks where you generate a document B which has the same hash as A. But here since it's about generating a hash from a username:password:nonce combination so it doesn't sounds so bad to use md5 (still agree that sha-whatever seems much more robust in general - and possibly I have no idea what I'm talking about :D)

Cheers !

data/helpers.d/postgresql
alexAubin: Yes that looks very much like it !

I will spin some tests in the next few days once I'm done with boring work stuff from real life

alexAubin: Another question though : are we confident that applications can use the hash-auth mechanism transparently ? I suppose any reasonnable app uses a lib which I guess is supposed to be able to handle that ?

madtibo: Yes. md5 is a very very old authentication technique in PostgreSQL and will be handle by all tools.
Things can be different for scram-sha-256, but this is more a question about how we generate the password than a line in the pg_hba file, since `md5` method accepts scram-sha-256.

madtibo:
> Dunno much about how bad md5 is (...)

This method can be subject to man-in-the-middle attack, that is why scram-sha-256 now exists. There is a double secret exchange in scram-sha-256 that mitigate the risk.

alexAubin:
Alrighty thank very much for all the answers / explanation / commits ! Will get back to you during next week and we can merge sometimes soon™

madtibo:
Cool!

PR #763, opened 2019-07-28
Fix variable definition for permission in backup

Josue-T:
## The problem

Follow https://github.com/YunoHost/yunohost/pull/585#discussion_r307891952

## PR Status

Tested and it work

## How to test

- Install an app
- Make a backup of this app
- Remove manually the app (juste leave the data in LDAP)
- Try to restore the backup

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -1250,7 +1250,9 @@ class RestoreManager():
                                '(&(objectClass=permissionYnh)(!(cn=main.mail))(!(cn=main.metronome))(!(cn=main.sftp)))',
                                ['cn']):
             if not ldap.remove('cn=%s,ou=permission' % per['cn'][0]):
-                raise YunohostError('permission_deletion_failed', permission=permission, app=app)
+                raise YunohostError('permission_deletion_failed',
+                                    permission=per['cn'][0].split('.')[0],
+                                    app=per['cn'][0].split('.')[1])
 
         # Restore permission for the app which is installed
         for per in old_apps_permission:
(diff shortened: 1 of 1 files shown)


Josue-T:
After some test I've discovered that the way to manage the LDAP error (for the `add`, `update`, `remove` operation) is really bad. Most of time we do something like that:

```python
if ldap.add(...):
    # continue the operation
else:
    raise YunohostError(...)
```

But after seeing the code in the moulinette I'saw that:
- If the LDAP operation succeed, the method return `True`
- If the there are an error, the method raise `MoulinetteError`

In the case of an error it's really bad because the `YunohostError` will be never called...

So I purpose to replace everywhere the if/else by a try/except. In the except case we can catch the `MoulinetteError` and raise the `YunohostError` which is generally more explicit. This still a big work because there are aproximaly 23 place where we are in this case.

src/yunohost/backup.py
```diff
@@ -1250,7 +1250,9 @@ def _restore_system(self):
                                '(&(objectClass=permissionYnh)(!(cn=main.mail))(!(cn=main.metronome))(!(cn=main.sftp)))',
                                ['cn']):
             if not ldap.remove('cn=%s,ou=permission' % per['cn'][0]):
-                raise YunohostError('permission_deletion_failed', permission=permission, app=app)
+                raise YunohostError('permission_deletion_failed',
+                                    permission=per['cn'][0].split('.')[0],
```
decentral1se: We won't see any `KeyError` or `IndexError`s here?

Josue-T: Normally not. If it happens, we have really a big problem...

alexAubin:
>After some test I've discovered that the way to manage the LDAP error (for the add, update, remove operation) is really bad.

Hmmm so maybe can we open an issue to have a followup on this ?

alexAubin:
Yolomergin this in the meantime

PR #765, opened 2019-07-29
[mod] BREAKING (but should be fine) apps without path aren't considering using the domain anymore

Psycojoker:
## The problem

Following https://github.com/YunoHost/issues/issues/1384#event-2490152235 we stop considering that an app that only has a domain but not (http) path is having an HTTP api.

That might breaks some app (but I'm really expecting not, that looks too much like a bad practice.)

## PR Status

Tested and working for me.

## How to test

Have some apps, edit the settings of one (`/etc/yunohost/$app/settings.yml`) to remove the path and launch `yunohost app map`.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -420,6 +420,9 @@ def app_map(app=None, raw=False, user=None):
             continue
         if 'domain' not in app_settings:
             continue
+        if 'path' not in app_settings:
+            # we assume that an app that doesn't have a path doesn't have an HTTP api
+            continue
         if 'no_sso' in app_settings:  # I don't think we need to check for the value here
             continue
         if user is not None:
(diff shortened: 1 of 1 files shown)


Psycojoker:
@YunoHost/apps are you fine with that? Looking at the bug that triggers this PR I think so but in case of.

maniackcrudelis:
What does that change ?
I guess those apps won't go through the SSO anymore ?

Looks fine for me.

Psycojoker:
> What does that change ?
> I guess those apps won't go through the SSO anymore ?

Yes and won't be considered for conflicting urls.

Psycojoker:
Merging then.

PR #768, opened 2019-08-05
Rework migration system to have independent migrations

alexAubin:
## The problem

This follows discussions during meetings about https://github.com/YunoHost/yunohost/pull/585 which ships a critical migration (half of yunohost won't be able to ran as long as this migration ain't ran). 

The current migration system is too constraining as if a migration C or D fails (or has to be ran manually), then critical migration E won't be ran even though it's independent of C and D. 

So in the case of group permissions, the migration would / could not be ran just because some other pending / failed migration.

## Solution

The proposed solution is to rework the migration system to make migrations independent from each other (though we can still have dependencies between migration via `self.dependencies = [ ...list of required migrations to be done before this one ...]` inside the migration)

So if migrations C, D, and E are pending, running `yunohost tools migrations migrate` will attempt to run every of those pending migrations even if C or D fails.

This requires to change the way we store the state of migrations, as only the name/number of the last ran migrations was stored so far. Now instead, a yaml json containing the list of known migrations with a status `skipped`, `done` or `pending`.

In this PR, I also included many tweaks following past experiences with the migration mechanism, such as : 
- ability to re-run a migration even if it was already ran
- the `-t` option was confusing when you just want to revert the last migration (e.g. during development) so instead there's now a `--revert` option
- instead of json, yaml is probably better (i can remember many situation where I had to - or tell users to - manually edit the file for some reasons)

## PR Status

Tested on my side but should run more tests ...

Also this is blocking for 3.7 since it's a must have to be sure upgrades will happen smoothly even for people upgrading from old versions

## How to test

- Pull the branch
- If you run `yunohost tools migrations list` or `yunohost tools migrations state`, the old json should be transparently migrated to a new yaml file, you should be able to see this in `/etc/yunohost`
- Then you can try to revert some migrations such as `0007_ssh_conf_managed_by_yunohost_step1` and `0008_ssh_conf_managed_by_yunohost_step2` (though that last one you gotta do it by hand in the .yaml file because reverting aint supported for it)
- Then attempt to run just `0008_ssh_conf_managed_by_yunohost_step2`, should tell you that first you gotta run part1
- Then if you just run `yunohost tools migrations migrate` it should run both 0007 and 0008

... though there are many other options and use case to test, not sure I can list them all here é_è

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/tools.py
```diff
@@ -970,6 +970,43 @@ def tools_reboot(operation_logger, force=False):
...
+    logger.warn("The \033[1;34mldap\033[0m interface is available in this context")
+    try:
+        from IPython import embed
+        embed()
+    except ImportError:
+        logger.warn("You don't have IPython installed, consider installing it as it is way better than the standard shell.")
+        logger.warn("Falling back on the standard shell.")
+
+        import readline  # will allow Up/Down/History in the console
+        readline  # to please pyflakes
```
decentral1se: Can just use `# noqa` on the line above to avoid, I think.

src/yunohost/tools.py
```diff
@@ -982,208 +1019,206 @@ def tools_migrations_list(pending=False, done=False):
...
                    "name": migration.name,
                    "mode": migration.mode,
+                   "state": migration.state,
                    "description": migration.description,
                    "disclaimer": migration.disclaimer} for migration in migrations]
 
+    # If asked, filter pending or done migrations
+    if pending or done:
+        if done:
+            migrations = [m for m in migrations if m["state"] != "pending"]
```
decentral1se:
```suggestion
            migrations = [m for m in migrations if m["state"] != "done"]
```

?

src/yunohost/tools.py
```diff
@@ -982,208 +1019,206 @@ def tools_migrations_list(pending=False, done=False):
...
-    logger.debug(m18n.n('migrations_current_target', target))
+        # Otherwise, targets are all pending migrations
+        targets = [m for m in all_migrations if m.state == "pending"]
 
-    # no new migrations to run
-    if target == last_run_migration_number:
+    # If explicit targets are provided, we shall validate them
+    else:
+        targets = [get_matching_migration(t) for t in targets]
+        done = [t.id for t in targets if t.state != "pending"]
```
decentral1se:
```suggestion
        done = [t.id for t in targets if t.state != "done"]
```

?

src/yunohost/tools.py
```diff
@@ -982,208 +1019,206 @@ def tools_migrations_list(pending=False, done=False):
...
+    # Extract the list of migration ids
     try:
-        from IPython import embed
-        embed()
+        from . import data_migrations
     except ImportError:
-        logger.warn("You don't have IPython installed, consider installing it as it is way better than the standard shell.")
-        logger.warn("Falling back on the standard shell.")
+        # Well ugh what are we supposed to do if we can't do this >.>...
+        pass
```
decentral1se: Probably should explode here!

src/yunohost/tools.py
Psycojoker: Current code looks right, to have "done" target you just want exclude the "pending" ones (other states are "done" too)

Psycojoker:
A bit hard to do a good review with my current life schedule (and commits are quite mixed up :sweat_smile:) but it looks good from a quick review :)

src/yunohost/tools.py
zamentur: done and skip but i think we want to see done and skip if you put --done

zamentur: I think webadmin use --done, and we want to display skipped migrations as done

src/yunohost/tools.py
zamentur: like  that --done return  "done" and "skip" but i think webadmin want to display done and skipped migrations together.

src/yunohost/tools.py
zamentur: Personally i think we should remove the try except here

alexAubin: You mean we kind safely assume that we can perform the import because that would be really strange otherwise ? I kind of agree, e.g. we don't do `try ... except` for the usual imports at the top of the file either ..

src/yunohost/data_migrations/0003_migrate_to_stretch.py
```diff
@@ -29,11 +29,7 @@ class MyMigration(Migration):
 
     mode = "manual"
 
-    def backward(self):
-
-        raise YunohostError("migration_backward_impossible", name=self.name)
-
-    def migrate(self):
+    def forward(self):
```
Psycojoker: You've renamed everything to `forward` (which is not really part of this commit if we look at the commit message :/) but [it's still the `migrate` method that is called](https://github.com/YunoHost/yunohost/pull/768/files#diff-d1fe4f9d57b5045a5baa68a74c263423R1150) so this will break everything :zipper_mouth_face:

zamentur: I have renamed it just because some migration use forward and other migrate.

Migrate method is marked as to not override in the class Migration. this migrate method call forward, and i didn't find other part of the code calling this method. But i am currently doing an other pr derivated of this one with other stuff.

src/yunohost/tools.py
alexAubin: Marking as resolved (c.f. Bram's comment)

src/yunohost/tools.py
alexAubin: Marking as resolved (c.f. ljf's comment)

src/yunohost/tools.py
alexAubin: Eh it was just me moving some existing code around ... I can fix that but it's not really related to this PR, idk ...

alexAubin:
So proposing to merge soonish (aiming for a testing release at the end of the week)

PR #770, opened 2019-08-11
[mod] more permissive email regex

Psycojoker:
## The problem

Email regex is way too restrictive, doesn't allow "+" stuff etc...

Extracted from
https://github.com/YunoHost/yunohost/pull/176/files#diff-85bc3950f9df27a6e9d05ea5ff3c82fbR129
by Julien Malik

## Solution

Make simpler stuff, email for regex is fucked up after all, we can only assume to have a "@" somewhere.

## PR Status

Not tested but looks fine.

**We probably need to check if this doesn't allow ldap query injection**

## How to test

Create a new user with a fucked up email address.
## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -104,7 +104,7 @@ user:
                         ask: ask_email
                         required: True
                         pattern: &pattern_email
-                            - !!str ^[\w.-]+@([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+([^\W\d_]{2,})$
+                            - !!str ".+@.+"
                             - "pattern_email"
                 -p:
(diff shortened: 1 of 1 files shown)


zamentur:
In this precise case, we don't want an email with "+" ...
We discuss yesterday to remove this email question adn replace it by just asking for a domain, to avoid some user difficulties during thunderbird configuration.

Psycojoker:
Postponed then until we reach a conclusion on how to handle this email question.

zamentur:
I close this PR as #818 and the one about --domain and user create seems to fix the issue

PR #773, opened 2019-08-13
[enh] Allow admin to specify an smtp relay

zamentur:
## The problem

Some isp block 25 port others have ips listed in rbl and can't be unlisted. This allow admin to get update of postfix even if they specify a relay.

## Solution
Use settings and the regenconf.
Inspired by https://yunohost.org/#/isp_orange_fr

## PR Status

Ready, i have tested regen conf with and without account. I had tested with a yunohost relay, but i was not able to test with smtp orange relay bacause i have no orange account.

## How to test
```
yunohost settings set smtp.relay.host -v smtp.example.org
yunohost settings set smtp.relay.port -v 587
yunohost settings set smtp.relay.user -v john
yunohost settings set smtp.relay.password -v trololo1234
yunohost service regenconf postfix
echo 'toto' | mail -s 'toto' your@email.tld
```
Note you can use a yunohost as smtp relay, but you need to add the domain in yunohost and add to the user account all alias needed.
## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


zamentur:
Note this will be usefull in our infra to route all email through saperlipopette...

Psycojoker:
I really have no good knowledge of email but ok on the principle.

data/hooks/conf_regen/19-postfix
```diff
@@ -24,7 +24,17 @@ do_pre_regen() {
...
+  
+  # Add possibility to specify a relay
+  # Could be useful with some isp with no 25 port open or more complex setup
+  export relay_host="$(yunohost settings get 'smtp.relay.host')"
+  if [ ! -z "${relay_host}" ]; then
+    export relay_port="$(yunohost settings get 'smtp.relay.port')"
+    export relay_user="$(yunohost settings get 'smtp.relay.user')"
+    relay_password="$(yunohost settings get 'smtp.relay.password')"
+    echo "[${relay_host}]:${relay_port} ${relay_user}:${relay_password}" > /etc/postfix/sasl_passwd
+    postmap /etc/postfix/sasl_passwd
```
alexAubin: Ugh we really aren't supposed to edit the actual filesystem during the regenconf ... this should happen in `$pending_dir` (or using the `$postfix_dir` shortcut).

Also ideally we need to remove that file if the `relay_host` setting is removed later (I believe that's done by creating an empty file, c.f. the nginx regenconf)

data/templates/postfix/main.cf
```diff
@@ -167,4 +171,20 @@ default_destination_rate_delay = 5s
...
+# Relay email through an other smtp account
+# enable SASL authentication
+smtp_sasl_auth_enable = yes
+# disallow methods that allow anonymous authentication.
+smtp_sasl_security_options = noanonymous
+# where to find sasl_passwd
+smtp_sasl_password_maps = hash:/etc/postfix/sasl_passwd
+{% if relay_port == "587" %}
+# Enable STARTTLS encryption
+smtp_use_tls = yes
```
alexAubin: Is this what we want to do ? The documentation says

> With Postfix 2.3 and later use `smtp_tls_security_level` instead. 

 (we are using postfix 3.1 on stretch)

Also I'm confused by this setting, it sounds like it's really not related to just relaying ... Shouldn't we think about what setting we want in general ? (Default seems to be to send emails in cleartext between servers, maybe we want at least `smtp_tls_security_level = may` to encrypt when possible ?

src/yunohost/settings.py
```diff
@@ -46,6 +46,10 @@
         "choices": ["intermediate", "modern"]}),
     ("security.postfix.compatibility", {"type": "enum", "default": "intermediate",
         "choices": ["intermediate", "modern"]}),
+    ("smtp.relay.host", {"type": "string", "default": ""}),
+    ("smtp.relay.port", {"type": "int", "default": 587}),
+    ("smtp.relay.user", {"type": "string", "default": ""}),
+    ("smtp.relay.password", {"type": "string", "default": ""}),
```
alexAubin: I believe there are corresponding strings to be defined in `locales/en.json` for these.

Also maybe we want to rename this as `email.relay.host` (and so on) to anticipate the integration in the webadmin where we'll probably want something more friendly and general like "Email" category rather than the technical and specific "SMTP" acronym ?

data/templates/postfix/main.cf
zamentur: You are right smtp_tls_security_level = may replace this settings !

src/yunohost/settings.py
zamentur: I write some description. 

We have a smtp.allow_ipv6 in 3.8. What should we do ?

zamentur: So after discussion it seems we keep smtp

Aeris1One:
How about this PR ? 
Does this works by now ?

zamentur:
Yes it works, i think we could merge

khimaros:
Awesome! I'm excited to see this land!

alexAubin:
Yolomerging to move forward

PR #774, opened 2019-08-14
[enh] remove revert mechanism

zamentur:
## The problem

In fact we don't want revert mechanism we want migration to rollback automatically. Some dev code backward function that aren't relevant.

## Solution

To simplify the code this pr suggest to remove the revert mechanism.

## PR Status

In progress

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
While automatic revert is really really a good thing :+1:  you still want to allow the user to manually revert if needed, migrations might generates bug that we won't catch during their run.

src/yunohost/data_migrations/0007_ssh_conf_managed_by_yunohost_step1.py
```diff
@@ -59,15 +59,3 @@ def forward(self):
             copyfile(SSHD_CONF, '/etc/ssh/sshd_config.bkp')
             regen_conf(names=['ssh'], force=True)
             copyfile('/etc/ssh/sshd_config.bkp', SSHD_CONF)
-
-        # Restart ssh and backward if it fail
-        if not _run_service_command('restart', 'ssh'):
-            self.backward()
```
alexAubin: So don't we want to keep the backward code as a rollback mechanism here ?

alexAubin:
So what do we do with this ... I don't have a strong opinion on the topic, but https://github.com/YunoHost/yunohost/pull/768 is a show-stopper for 3.7 and the PRs are pilling up, I'd really like if we can manage to release a testing in the coming days ...

alexAubin:
So proposing to merge soon ...

Psycojoker:
I'm still not really convinced about removing the "--revert" part as it is a tool that could help for debugging (like I've added "--force" and "--dry-run") but that's not the biggest blocker ever (and that could be added again if needed) so I guess ... ok?

alexAubin:
Alrighty, sooo idk, merging to move forward for the 3.7

PR #777, opened 2019-08-15
[enh] Display other system logs

zamentur:
## The problem

Some system log are not available in web admin

## Solution

This PR add this log in Tools > Log like it was designed during the Log PR.
system_logs

## PR Status

Ready
This PR depends of an other on yunohost-admin https://github.com/YunoHost/yunohost-admin/pull/255

## How to test

We need to merge this in unstable to test completely, but you can test like this:

- Pull branch on yunohost and yunohost-admin
- use git on yunohost and yunohost-admin
- cp -R yunohost/data/other/logs/* /var/log/yunohost/categories/
- service yunohost-admin
- Try to use the webadmin to see the new log in screenshot

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/other/logs/package/alternatives.yml
```diff
@@ -0,0 +1 @@
+log_path: /var/log/alternatives.log
```
alexAubin: Uh wokay ... I'm not sure that this file really is any useful ? Can you elaborate on what kind of insight it provides ?

data/other/logs/package/dpkg.yml
```diff
@@ -0,0 +1 @@
+log_path: /var/log/dpkg.log
```
alexAubin: Same as alternatives.log, I'm not sure what we're expecting to find in this file :confused:

data/other/logs/package/history.yml
```diff
@@ -0,0 +1 @@
+log_path: /var/log/apt/history.log
```
alexAubin: Can we add also `term.log` which is really the actual log of apt operations ?

data/other/logs/system/kernel.yml
```diff
@@ -0,0 +1 @@
+log_path: /var/log/kern.log
```
alexAubin: It sounds like `kern.log` is just a subset of `messages` so I would choose one or the other (`messages.yml` apriori)

locales/en.json
```diff
@@ -301,6 +301,17 @@
...
+    "log_alternatives": "Change to default programs",
+    "log_auth": "Connection history",
+    "log_btmp": "Failed login attempts",
+    "log_dpkg": "History of installed debian packages",
+    "log_fail2ban": "Banning history",
+    "log_history": "History of installed debian packages with apt",
+    "log_kernel": "Information and events about the linux kernel",
+    "log_messages": "General system activities",
+    "log_root_bash_history": "History of commands ran by root",
+    "log_wtmp": "Login and logout history",
```
alexAubin: Apparently there's no `wtmp` configured...?

locales/en.json
```diff
@@ -301,6 +301,17 @@
     "log_tools_upgrade": "Upgrade system packages",
     "log_tools_shutdown": "Shutdown your server",
     "log_tools_reboot": "Reboot your server",
+    "log_admin_bash_history": "History of commands ran by admin",
+    "log_alternatives": "Change to default programs",
+    "log_auth": "Connection history",
+    "log_btmp": "Failed login attempts",
```
alexAubin: Apparently there's no `btmp` configured...?

locales/en.json
```diff
@@ -301,6 +301,17 @@
...
+    "log_dpkg": "History of installed debian packages",
+    "log_fail2ban": "Banning history",
+    "log_history": "History of installed debian packages with apt",
```
alexAubin: Well uh, is this for apt, or is this the category for the admin/root history ..? Shouldn't this be `log_package` ..?

Also I would really rename the `package` category to `apt_and_dpkg` or just `apt` or `system_packages` to make a clear distinction to apps or other packages

data/other/logs/package/alternatives.yml
zamentur: Could be usefull in some situation with malicious programs running as default programs.
Maybe also to debug some php version conflict see https://github.com/YunoHost-Apps/framaforms_ynh/blob/testing/scripts/install#L191 .

data/other/logs/package/dpkg.yml
zamentur: dpkg provide all manually dpkg packages installed, it could be usefull with some apps installing manually some dpkg.

data/other/logs/package/history.yml
zamentur: Good idea

locales/en.json
zamentur: yep i removed it because this file make a big 500 error in moulinette and is not so much relevant

zamentur:
About this PR. We discuss it and Aleks think it should be transform into conf files store into /etc/yunohost . So it make it quite more difficult to implement.

alexAubin:
Closed cf today's meeting:

>Aleks: Imho this really feels like some YAGNI ... There's no really any huge gain in terms of support to expose these in the webadmin -> To close ?
>ljf: yes cause it need to be redone AND because 777 is evil
>BUT i disagree with you. It could be helpfull to allow user to not need CLI at all to do the operation. However i think important event about those file could be reported by diagnosis system (oom_reaper, a lot of ssh attempt?, ...).

PR #778, opened 2019-08-17
[enh] Rebuild applist system and make it future proof

alexAubin:
## The problem

So ... been ranting about the applist system for a long time. Especially because : 
- The code is still a mess even though it has been rewritten ~2 years ago. And there's still a bunch of legacy code lying around.
- The ability to add / remove additional list is a big [YAGNI](https://www.monkeyuser.com/2019/yagni/?sc=true&dir=previous) leading to unecessary complex code, dozens of strings to internationalize, and regularly bugs (though now fixed)
- When fixing the bugs, one regularly had to edit the json file which is a mess to edit manually if you're not used to it
- Since I'm in the process of implementing the app categories, for this we need to change the format of `apps.json` ... which leads to *yet again* another migration to implement. And I'm pretty sure than in less than 6 months we will want to change the format *again*

## Solution

- Burn the existing code and rework everything
- Use a yaml for conf because it's much more readable and easy to edit manually
- 99.9% of our user base don't need to add custom list, so let's drop the API about adding/removing/listing the appslists. For power user than want to add additional lists, we'll simply tell them to edit the config file manually, which just isn't complex if it's a yaml file instead of json where you just need to put a name and a url. (It's not difficult to automatize either, you can just append the appropriate text to the conf file with an echo)
- Make the system future proof, meaning using a version number for the appslist API. So instead of fetching `app.yunohost.org/apps.json`, we fetch `app.yunohost.org/default/v{VERSION}/apps.json` where : 
   - `default` is the name of the list ... in case in future we want to support additional list (e.g. `dev` for special apps, or `non-free` though this is just an example, not really pushing for it...) with the same domain `app.yunohost.org`
   - `{VERSION}` is a number like 1, 2, 3, ... hardcoded in the code. Later in the future if we want to change the apps.json format. we just update the version in the code, expose the new json on `/default/v{NEW_VERSION}/apps.json` and the yunohost instance shall automatically see that the previous cache is outdated and update it.

## PR Status

Tested on my machine with the unit tests provided in the PR

## How to test

Read and run the unit tests ¯\\\_(ツ)\_/¯

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/data_migrations/0010_migrate_to_apps_json.py
```diff
@@ -1,44 +1,18 @@
...
-
-        appslists = _read_appslist_list()
-        for appslist, infos in appslists.items():
-            if infos["url"].split("//")[-1] in lists_to_remove:
-                app_removelist(name=appslist)
-
-        # Replace by apps.json list
-        app_fetchlist(name="yunohost",
-                      url="https://app.yunohost.org/apps.json")
+        logger.info("This is migration is oudated and doesn't do anything anymore. The migration 12 will handle this instead.")
```
decentral1se:
```suggestion
        logger.info("This migration is oudated and doesn't do anything anymore. The migration 12 will handle this instead.")
```

decentral1se:
> 99.9% of our user base don't need to add custom list, so let's drop the API about adding/removing/listing the appslists.

+1

> using a version number for the appslist API.

+1

src/yunohost/tools.py
```diff
@@ -510,11 +513,10 @@ def is_legit_warning(m):
 
     upgradable_apps = []
     if apps:
-        logger.info(m18n.n('updating_app_lists'))
         try:
-            app_fetchlist()
+            _update_appslist()
         except YunohostError as e:
-            logger.error(m18n.n('tools_update_failed_to_app_fetchlist'), error=e)
+            logger.error(str(e))
```
Psycojoker: We might want a better error here for later :sweat_smile:

alexAubin:
As for diagnosis, planning to merge soon, to be followed by another PR for the actual app categories integration

PR #769, opened 2019-08-07
Stop on failed app upgrade

Psycojoker:
## The problem

app_upgrade will try to upgrade all apps it receives, even if one upgrade fails. It turns out that in practice one app failed upgrade could break everything and make other app upgrades fail.

## Solution

Stop if the upgrade of one application failed.

## PR Status

Untested.

## How to test

I guess you need a list of fake apps to do that?

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #779, opened 2019-08-17
[enh] Add unit/functional tests for apps

alexAubin:
## The problem

We don't have tests for apps '-'

## Solution

Implement some ...

## PR Status

Tested on my machine, but still need to implement at least 3 of them

I think I spotted an important issue related to LDAP group permissions related to failed remove, gotta investigate

## How to test

Run the tests using ynh-dev or pytest manually

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -620,7 +620,10 @@ def app_upgrade(app=[], url=None, file=None):
 
         app_dict = app_info(app_instance_name, raw=True)
 
-        if file:
+        if file and isinstance(file, dict):
+            # We use this dirty hack to test chained upgrades in unit/functional tests
+            manifest, extracted_app_folder = _extract_app_from_file(file[app_instance_name])
+        elif file:
             manifest, extracted_app_folder = _extract_app_from_file(file)
         elif url:
             manifest, extracted_app_folder = _fetch_app_from_git(url)
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -667,7 +667,6 @@ def app_upgrade(app=[], url=None, file=None):
         # Execute App upgrade script
         os.system('chown -hR admin: %s' % INSTALL_TMP)
 
-
         try:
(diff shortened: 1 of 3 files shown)


Josue-T:
> I think I spotted an important issue related to LDAP group permissions related to failed remove, gotta investigate

Do you have any log. Maybe I could help to investigate...

alexAubin:
> Do you have any log. Maybe I could help to investigate...

It's not clear to me how to reproduce ... but basically some times I ended up with the check_ldap_db_consistency failing because some remaining permission existed for an app that was supposed  to be uninstalled ... That might be because I was tweaking a lot of code and the remove operation was sometime failing because before the part where it removes the thing from ldap.

So maybe I'll just wrap stuff in try/except to avoid that kind of things in the future even though that seems to happen mainly during dev.

>Documentation addition to the README on how to run these?
>They need to be run in the context of ynh-dev / live yunohost or?

Yes, in the context of a post-installed yunohost. I use `./ynh-dev test yunohost/apps` to run them, (c.f. [here](https://github.com/YunoHost/ynh-dev/blob/master/ynh-dev#L301-L340)) or you can run pytest by hand

alexAubin:
Proposing to merge soonish though I will implement the last tests for https://github.com/YunoHost/yunohost/pull/769

alexAubin:
So that was painful ... but I finally implemented and ran the remaining tests which allowed to find bugs in  https://github.com/YunoHost/yunohost/pull/769 and https://github.com/YunoHost/yunohost/pull/785 so we're good to go

The whole triptic of PR is a mess : this one is based on https://github.com/YunoHost/yunohost/pull/785 itself based on https://github.com/YunoHost/yunohost/pull/769  so i temporarily changed the base branch to this one is based on https://github.com/YunoHost/yunohost/pull/785 in case you want to read the relevant commit of this PR easily...

PR #780, opened 2019-08-17
[mod] Remove deprecated helpers while still supporting them through hacky patching

alexAubin:
## The problem

Stuff like `yunohost app initdb` has been deprecated for [more than 3 years](https://github.com/YunoHost/yunohost/commit/2f1bf014e8c8c0dee99e5c439ea662330ce44e88) now .. And I hate having legacy code all over the place with dozens of outdated strings :|

## Solution

Remove those unused / deprecated helpers ... well actually I implemented some sort of auto-patching to try to keep the compatibility ... It can be tested using the [apps unit tests](https://github.com/YunoHost/yunohost/pull/779) which runs on the [legacy test app](https://github.com/YunoHost/test_apps)

Should be noted that 90% of the apps that I found still using those helpers seem to be really outdated / notworking anyway ...

## PR Status

Tested and working on my machine

## How to test

Try to install some very outdated app still using stuff like `yunohost app initdb` or `yunohost app checkport`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -2951,3 +2856,77 @@ def _patch_php5(app_folder):
...
+
+        for helper, regexes in stuff_to_replace_compiled.items():
+            pattern, replace = regexes
+            # If helper is used, attempt to patch the file
+            if helper in content and pattern != "":
+                try:
+                    content = pattern.sub(replace, content)
+                    replaced_stuff = True
+                except Exception as e:
+                    import pdb; pdb.set_trace()
```
Psycojoker: Oupsi oupsi, debug code :sweat_smile: (btw "ipdb" or "pdbpp" are way better than pdb)

src/yunohost/app.py
```diff
@@ -2951,3 +2856,77 @@ def _patch_php5(app_folder):
...
+        content = read_file(filename)
+        replaced_stuff = False
+
+        for helper, regexes in stuff_to_replace_compiled.items():
+            pattern, replace = regexes
+            # If helper is used, attempt to patch the file
+            if helper in content and pattern != "":
+                try:
+                    content = pattern.sub(replace, content)
+                    replaced_stuff = True
```
Psycojoker: Misc: that one would be better in an "else:" statement of the try/except but that's really minor.

src/yunohost/app.py
```diff
@@ -2951,3 +2856,77 @@ def _patch_php5(app_folder):
...
+                except Exception as e:
+                    import pdb; pdb.set_trace()
+
+            # If we couldn't patch the deprecated helper, abort the install or whichever step is performed
+            if helper in content:
```
Psycojoker: Shouldn't that be an `elif` here :thinking: ? My understanding is that since it's an else and there is no `return`/`continue` previously it's going to be always matched.

alexAubin: Hmyea sorry I should maybe improve the comment here but basically : 
- if we detect that the helper is used, we attempt to patch it in the previous `if`
- *after* this, if the helper is still in `content`, that mean we couldn't effectively patch it, e.g. because the actual usage did not match the format defined by the regex or whatever

data/helpers.d/network
```diff
@@ -24,6 +24,31 @@ ynh_find_port () {
...
+# Requires YunoHost version 3.7.x or higher.
+ynh_port_available () {
+	# Declare an array to define the options of this helper.
+	local legacy_args=p
+	declare -Ar args_array=( [p]=port= )
+	local port
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+
+    if netcat -z 127.0.0.1 $port;
```
kay0u: You could use this:

https://github.com/YunoHost/yunohost/pull/827/files

And we could use this helper inside the function find port to avoid duplicated code.

Plus, I really think theses functions should have an argument to check port on ipv4,6 or both

PR #781, opened 2019-08-18
Small cleaning and update of the README.md

alexAubin:
Also found out about this cool summary of translation which imho is better to show than the old squary widget ¯\\\_(ツ)\_/¯

https://translate.yunohost.org/widgets/yunohost/-/multi-auto.svg

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,42 +1,45 @@
 # YunoHost core
 
-- [YunoHost project website](https://yunohost.org)
-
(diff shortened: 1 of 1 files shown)


Psycojoker:
Way better indeed :)

Psycojoker:
Thinking about it, that's definitively something that we can use in our communication to brings contributions!

PR #784, opened 2019-08-23
Propagate change on LDAP conf / authenticator

alexAubin:
## The problem

c.f. changes in https://github.com/YunoHost/moulinette/pull/216

## Solution

Propagate the change, use a simple dict instead of a bunch of tuples

## PR Status

Tested and working if you have the right moulinette PR checked out

## How to test

Try to launch `yunohost tools shell` and check that the `ldap` object if working properly

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/utils/ldap.py
+++ b/src/yunohost/utils/ldap.py
@@ -20,7 +20,7 @@
 """
 
 import atexit
-from moulinette.core import init_authenticator
+from moulinette.authenticators import ldap
 
 # We use a global variable to do some caching
 # to avoid re-authenticating in case we call _get_ldap_authenticator multiple times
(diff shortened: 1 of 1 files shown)


src/yunohost/utils/ldap.py
```diff
@@ -31,12 +31,16 @@ def _get_ldap_interface():
...
-        AUTH_PARAMETERS = {'uri': 'ldapi://%2Fvar%2Frun%2Fslapd%2Fldapi',
-                           'base_dn': 'dc=yunohost,dc=org',
-                           'user_rdn': 'gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth'}
-        _ldap_interface = init_authenticator(AUTH_IDENTIFIER, AUTH_PARAMETERS)
+
+        conf = { "vendor": "ldap",
+                 "name": "as-root",
+                 "parameters": { 'uri': 'ldapi://%2Fvar%2Frun%2Fslapd%2Fldapi',
+                                 'base_dn': 'dc=yunohost,dc=org',
+                                 'user_rdn': 'gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth' },
```
decentral1se:
```suggestion
                                 'user_rdn': 'gidNumber=0,uidNumber=0,cn=peercred,cn=external,cn=auth' },
```

I know it's the same as the original text but just checking - the `+` is not a typo, right!?

alexAubin: Eeeeh no idea :s (@Josue-T do you know about this ?)

Josue-T: Well, `+` is not same than `,`. `,` is the separation between each element of the LDAP tree. For `+` it's different, it's really specific, it's liked to the root authentication. Don't really know exactly what it mean but probably something like gidNumber=0 and uidNumber=0. Anyway `+` true.

PR #785, opened 2019-08-23
[enh] Iteration on sanity / anomaly checks for app operations

alexAubin:
## The problem

Currently we now do some checks that services requested by an app are up and running (and other anomaly checks) before some operations like install and upgrade ...Which is cool, but we can do better : 
- checking *before* and *after* the operations
- check nginx and fail2ban all the time even if not explicitly required by the app, because those are related to regular issues
- combine this with the dpkg check (and in the future we can add other basic sanity checks in that vein)

## Solution

Rework a few things to implement this

## PR Status

Tested and working using the PR about app unit tests ... Though maybe still need some work because not all tests are implemented and there are still weird issues in the test

## How to test

Eeeeh using the app unit test, or manually run stuff like app install with nginx down. Or try to install an app that break nginx or fail2ban.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -584,9 +584,6 @@ def app_upgrade(app=[], url=None, file=None):
         url -- Git url to fetch for upgrade
 
     """
-    if packages.dpkg_is_broken():
-        raise YunohostError("dpkg_is_broken")
-
     from yunohost.hook import hook_add, hook_remove, hook_exec, hook_callback
     from yunohost.permission import permission_sync_to_user
 
@@ -638,7 +635,7 @@ def app_upgrade(app=[], url=None, file=None):
 
         # Check requirements
         _check_manifest_requirements(manifest, app_instance_name=app_instance_name)
-        _check_services_status_for_app(manifest.get("services", []))
+        _assert_system_is_sane_for_app(manifest, "pre")
 
         app_setting_path = APPS_SETTING_PATH + '/' + app_instance_name
 
@@ -716,6 +713,7 @@ def app_upgrade(app=[], url=None, file=None):
             # So much win
             logger.success(m18n.n('app_upgraded', app=app_instance_name))
 
+            _assert_system_is_sane_for_app(manifest, "post")
             hook_callback('post_app_upgrade', args=args_list, env=env_dict)
             operation_logger.success()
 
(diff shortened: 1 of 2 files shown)


Psycojoker:
LGTM after a quick review. Maybe I'll have energy to do a better review later.

Also: super great to have this PR :heart:

alexAubin:
Hmso since there's some delay for 3.7, proposing to add this for 3.7 as well (also because I'd like to merge https://github.com/YunoHost/yunohost/pull/779 which is also helpful for https://github.com/YunoHost/yunohost/pull/769 ...)

alexAubin:
So proposing to merge soonish

alexAubin:
c.f. https://github.com/YunoHost/yunohost/pull/779#issuecomment-531524268 ... was painful but finally fixed a few things thanks to the tests. Rebased on top of https://github.com/YunoHost/yunohost/pull/769 for various reason. Temporarily changed the base branch in case you wanna read the diff.

src/yunohost/app.py
```diff
@@ -669,55 +666,81 @@ def app_upgrade(app=[], url=None, file=None):
...
+
+
+        try:
+            upgrade_retcode = hook_exec(extracted_app_folder + '/scripts/upgrade',
+                                        args=args_list, env=env_dict)[0]
+        except (KeyboardInterrupt, EOFError):
+            upgrade_retcode = -1
+        except Exception:
+            import traceback
+            logger.exception(m18n.n('unexpected_error', error=u"\n" + traceback.format_exc()))
```
Psycojoker: If you do a `, exc_info=1` you should normally get the traceback in the logger (so you don't need to use traceback directly)

src/yunohost/app.py
```diff
@@ -669,55 +666,81 @@ def app_upgrade(app=[], url=None, file=None):
...
+        finally:
+
+            # Did the script succeed ?
```
Psycojoker: Shouldn't we had a test for `-1` saying "the user stopped the installation"?

alexAubin: 875c570 should fix it :+1:

src/yunohost/app.py
alexAubin: I suppose, but I just copied that part from `app_install`, and turns out you're the author :sweat_smile: 
https://github.com/YunoHost/yunohost/commit/9e1cd734001c0fd6a7c18e95f08041c67826502b

src/yunohost/app.py
Psycojoker: Thx!

src/yunohost/app.py
Psycojoker: Oupsi :sweat_smile: (I probably wasn't aware of that at that time)

PR #789, opened 2019-09-03
Fix permission sychronization

Josue-T:
## The problem

https://github.com/YunoHost/issues/issues/1427

## Solution

Just consider that if the key `groupPermission` it's equivalent to en empty list. It avoid to do anything when we remove the last permission.

## PR Status

Tested and it work

## How to test

Just 
```
yunohost user permission clear mail
yunohost user permission add mail -u test toto
yunohost user permission remove mail -u test
yunohost user permission remove mail -u toto
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
@zamentur : does that look okay to you ?

PR #795, opened 2019-09-13
Polish group/permissions interface + simplify code

alexAubin:
## The problem

Sooooo this is a huge polishing iteration on the group and permission mechanisms. This is driven by some feedback from ljf who started implementing the admin views for this and we concluded that the general architecture was good, but the command interface was overly confusing and many implementation bits could be simplified...

## Solution

Here's an attempt to summarize the important bits : 

- A permission is now just designated by it's name like "xmpp" (implicitly "xmpp.main") or "wordpress.admin"
- Rework the interface to add/remove stuff to a group or permission and fix some naming stuff
   - Group members are added / removed using `user group update foo --add ... --remove ...`
   - Permission accesses are added / removed using `user permission update bar --add ... --remove ...`
   - Similar stuff for app helpers
   - remove is symetric of add, and delete is symetric of create
   - `permission_update` is now `permission_urls` to remove ambiguities with `user_permission_update`
- There was some very confusing messages / behavior (related to the legacy permission mechanism) when trying to edit a permission that just got created ... one add to first add a user/group to the permission, which would magically remove `all_users`... Changed this such that : 
    - you now need to explicitly remove `all_users` manually
    - if add a user/group without removing `all_users`, it displays a warning telling you that you probably want to remove `all_users` for stuff to make sense
    - `.main` permissions come with `all_users` by default, but custom (not `.main` perms) comes with nobody enabled by default
- Seeing primary (or eponym) groups and being able to add users into them is very confusing .. instead, we should not display them by default (in `yunohost user group list`) and refuse to add other users than the eponym user.
- Refactored / simplify quite a lot of code using more semantic variable names, list comprehensions, reusing other functions, etc...
- Fixed [the ldap error management madness which was found by Josue](c.f. https://github.com/YunoHost/issues/issues/1405)
- Tweak default options such that `user group list` and `user permission list` lists only the most important stuff by default. All other infos can be displayed using `--full`
- Backing up / restoring app permissions is now done using a simple yml file instead of ldif because that was overly complicated...
- Tried to regroup all i18n strings related to groups and permissions under the `group_` and `permission_` prefix to keep things tidy

The resulting interface is documented in [the corresponding PR draft in the documentation](https://github.com/YunoHost/doc/blob/group-and-permissions/groups_and_permissions.md)

## Stuff still to be answered and implemented at some point ...

It's not clear how precisely all of this is gonna be interfaced with SSOwat (~~and it looks to me that the whole group/permission mechanic remains largely useless for the end users and app packagers except for the mail/xmpp stacks ?~~ Edit : my bad, didn't realize this all happens in `app_map` !)

- How realistic is it to implement a system where, for example we have : 
    - `wordpress.main -> domain.tld/blog` with `group1` allowed
    - `wordpress.admin -> domain.tld/blog/admin` with `group2` allowed 
    - presumably `group2` is a subgroup of `group1`, but how do we really make sure of this and integrate this in the UX (e.g. what happens if the user configures permission with group2 being a larger group than group1 ?)
    - ~~how to tell SSOwat to handle the access in a consistent way, either with the existing mechanic (unprotected / protected / ... urls ) or new mechanics...~~
- What about url regexes : will we need them ? And how does that integrates in the rest of the picture ?
- For now, apps that define custom permission need to provide the full url (including the domain+path) which means that they also need to update it during `change_url` ... couldn't we use relative paths instead ?
- Shouldn't we integrate the `is_public` mechanism in this permission mechanics ? For example, we could have a special group called `visitors` which would mean that the url is configured as unprotected...

## PR Status

Tested on my machine, but would be nice if we can collectively validate the general design

## How to test

can be tested using `test_user-group`, `test_permissions` and `test_backuprestore`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Josue-T:
Hello,

Thanks for all your work. I'll try to find some time review this.

>  There was some very confusing messages / behavior (related to the legacy permission mechanism) when trying to edit a permission that just got created ... one add to first add a user/group to the permission, which would magically remove `all_users`... Changed this such that :

Yes, there was some discussion about that here https://github.com/YunoHost/issues/issues/1428 . 
For me it's ok with your purpose.

>   you now need to explicitly remove `all_users` manually

Note that we need to keep the retro-compatibility because many apps use the old mechanism. You can see here https://github.com/search?q=org%3AYunoHost-Apps+yunohost+app+addaccess&type=Code

>   Seeing primary (or eponym) groups and being able to add users into them is very confusing .. instead, we should not display them by default (in `yunohost user group list`) and refuse to add other users than the eponym user.

+1

> It's not clear how precisely all of this is gonna be interfaced with SSOwat (~and it looks to me that the whole group/permission mechanic remains largely useless for the end users and app packagers except for the mail/xmpp stacks ?~ Edit : my bad, didn't realize this all happens in `app_map` !)


>   How realistic is it to implement a system where, for example we have :
>       - `wordpress.main -> domain.tld/blog` with `group1` allowed
>       - `wordpress.admin -> domain.tld/blog/admin` with `group2` allowed
>       - presumably `group2` is a subgroup of `group1`, but how do we really make sure of this and integrate this in the UX (e.g. what happens if the user configures permission with group2 being a larger group than group1 ?)

Maybe we just can add a warning when a user is allowed in the specific permission but is not allowed in the main.

>       - ~how to tell SSOwat to handle the access in a consistent way, either with the existing mechanic (unprotected / protected / ... urls ) or new mechanics...~
>     - What about url regexes : will we need them ? And how does that integrates in the rest of the picture ?

Globally for SSOwat my idea was this:

Note that the next ssowat configuration will look like that:

```json
{
    ...

    "users": {
        "user1": {
            "domain.tld/wiki/": "Wiki",
        }, 
        "user2": {
            "domain.tld/wiki/": "Wiki",
        }
    }
    "permission": {
        "domain.tld/wiki/" : ["user1","user2"],
        "domain.tld/wiki/admin" : ["user1"],
        "domain.tld/wiki/diagnostique" : ["user1"]
    }
}
```

SSOwat should see in the permission section what is the longest url which match with the requested url. When this is found we can just check that the authenticated user is liseted in the matched path.

> For now, apps that define custom permission need to provide the full url (including the domain+path) which means that they also need to update it during `change_url` ... couldn't we use relative paths instead ?

I decided to use the full URL to make more easy to implement the multiple url support for an app see https://github.com/YunoHost/yunohost/pull/549.

> Shouldn't we integrate the `is_public` mechanism in this permission mechanics ? For example, we could have a special group called `visitors` which would mean that the url is configured as unprotected...

+1

src/yunohost/data_migrations/0011_setup_group_permission.py
```diff
@@ -85,14 +82,18 @@ def migrate_app_permission(self, app=None):
...
+                user_permission_update(app+".main", remove="all_users", add=allowed_group, sync_perm=False)
             app_setting(app, 'allowed_users', delete=True)
 
-
     def run(self):
+
+        # FIXME : what do we really want to do here ...
+        # Imho we should just force-regen the conf in all case, and maybe
+        # just display a warning if we detect that the conf was manually modified
+
```
Josue-T: Well, as I understand you want to force the regen-conf here ?

data/helpers.d/setting
```diff
@@ -230,70 +230,82 @@ ynh_webpath_register () {
...
     ynh_handle_getopts_args "$@"
-    if [[ -n ${defaultdisallow:-} ]]; then
-        defaultdisallow=",default_allow=False"
-    fi
 
     if [[ -n ${urls:-} ]]; then
         urls=",urls=['${urls//';'/"','"}']"
     fi
-    yunohost tools shell -c "from yunohost.permission import permission_add; permission_add('$app', '$permission' ${defaultdisallow:-} ${urls:-}, sync_perm=False)"
+    yunohost tools shell -c "from yunohost.permission import permission_create; permission_create('$app.$permission' ${urls:-}, sync_perm=False)"
```
Psycojoker: Shouldn't we use "add/remove" everywhere instead? Like "user create" is the only place for that and should probably want to get ride of that (or on the contrary user "create/remove" for domains/other things?)

zamentur:
I have read the documentation and i have nothing to say, it's like we discuss together, nice job!
LGTM

About problematic you pointed, you have open a draft about it on this ticket https://github.com/YunoHost/yunohost/pull/797

> For now, apps that define custom permission need to provide the full url (including the domain+path) which means that they also need to update it during change_url ... couldn't we use relative paths instead ?
Yes it seems possible to do that. Apps don't use the url info when they connect ldap, and the SSO conf is generated, so it could be done.

zamentur:
I will test your pr to adapt the yunohost-admin pr and we will see

alexAubin:
More notes / thoughts while implementing the visitor mechanism. Especially [this part](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/app.py#L439-L450) which does the stuff for the "main" permission, but we should extend this to handle the extra permission ...
 
- the `no_sso` option could now be implemented simply be removing the url associated to the main permission ...
- does it really make sense to have multiple urls associated to a given permission ? E.g. if you have 2 or more urls for the main permission : which one should be used for the sso tile ?
- I believe we should add an option like "tile_in_sso: true/false" so that app can declare if extra permission (e.g. some admin interface) should appear on the sso as an independent tile. I think there's a ticket opened somewhere for that kind of feature.
- and if extra permissions appear in the sso ... they need a label to be displayed. Which also makes sense from the webadmin perspective where it'd be nice to have a display name rather than just the technical name of the permission - though I believe in most case it would work (e.g. wordpress.admin can by default be transformed to Wordpress Admin)
- do we really need to handle the multiple domain use case ... Apparently, even in the PR for multiple domain/path, Bram was thinking about essentially *multiple paths* and not really multiple domains ... In fact, thanks to the permission mechanism, we can already handle those multiple domains and paths ! But the multiple domain scenario sounds like a big that add artificial complexity for something we ain't gonna need.
- in the long term, having a virtual user "visitors" in the ssowat configuration with the same format as what's in the `users` section (a dict `domain.tld/some_url: Some Fancy Label`) could allow to easily generate some sort of public homepage listing the publicly available services on a yunohost server, like @eauchat did iirc.

Josue-T:
> does it really make sense to have multiple urls associated to a given permission ? E.g. if you have 2 or more urls for the main permission : which one should be used for the sso tile ?

For me no.

> I believe we should add an option like "tile_in_sso: true/false" so that app can declare if extra permission (e.g. some admin interface) should appear on the sso as an independent tile. I think there's a ticket opened somewhere for that kind of feature.

> and if extra permissions appear in the sso ... they need a label to be displayed. Which also makes sense from the webadmin perspective where it'd be nice to have a display name rather than just the technical name of the permission - though I believe in most case it would work (e.g. wordpress.admin can by default be transformed to Wordpress Admin)

For me there are 2 solutions:

1. As you said. We add 2 new attribute for the permission objectclass:
- A boolean `tile_in_sso`
- A string `label` optional. If it's not defined we just try to get the name from the technical name.

2. We just add one attribute `label`. In the permission object we just can say:
- if the `label` attribute is set we show a tile in the SSO
- if no `label` attribute is defined we don't show the tile.

Psycojoker:
>     * Apparently, even in the PR for multiple domain/path, Bram was thinking about essentially _multiple paths_ and not really multiple domains ...

Ah? In my (bad) memories there was also the mutliple domains needs for some applications that needs several entry points (I have matrix/riot combinaison in mind but idk if it's the case but I remember some domains needed that). I think some app do requires multiple (sub?)domain although :thinking: ... like there is this xmpp stuff where it wants a domain to upload a file right? I don't know if we have a lot of example still but I'm definitively expecting annoying apps to needs several domains at some point >_<

src/yunohost/backup.py
```diff
@@ -1370,22 +1355,27 @@ def copytree(src, dst, symlinks=False, ignore=None):
...
+
+                permissions = read_yaml('%s/permissions.yml' % app_settings_new_path)
+                existing_groups = user_group_list()['groups']
+
+                for permission_name, permission_infos in permissions.items():
+
+                    permission_create(permission_name, urls=permission_infos.get("urls", []))
+
+                    if "allowed" not in permission_infos:
+                        logger.warning("'allowed' key corresponding to allowed groups for permission %s not found when restoring app %s ... You might need to reconfigure permissions yourself!" % (permission_name, app_instance_name))
```
comradekingu:
```suggestion
                        logger.warning("'allowed' key corresponding to allowed groups for permission %s not found when restoring app %s … You might have to reconfigure permissions yourself." % (permission_name, app_instance_name))
```

src/yunohost/data_migrations/0011_setup_group_permission.py
alexAubin: Yes, what are your thoughts on this ?

data/helpers.d/setting
alexAubin: Hmyeah we discussed this in length with ljf the other day ...

To me (and this is confirmed by [https://ux.stackexchange.com/questions/43174/update-vs-modify-vs-change-create-vs-add-delete-vs-remove]) : 
- creating/deleting is about creating something from scratch (or removing it from existence entirely)
- add/remove is about manipulating entities that already exist and adding/removing them to/from a system

So when talking about `user create` vs `user add`, it all depends wether you think of it as "creating an account" (the account doesn't exist yet, it is created) or "adding an (existing) human being to the system" (the human being already exist in real life). Personally I'm okay with the current `user create/remove` because I do think of it as creating an account. (The button on the webadmin says `+ New user` which is consistent with the idea of creating a new account as well.)  But I understand that other people might prefer thinking about it in term of "adding a human being" ... 

When talking about `domain create` vs `domain add`, it all depends wether you think of it as "creating a domain from scratch" (the domain doesn't exist yet anywhere on the internet) or "adding an (existing) domain name to the system" (the domain already exist on your registrar API). In the general case, I naively think more about "adding it" because I already created it on the registrar API ... but it's more ambiguous for the case of nohost.me / noho.st / ynh.fr which are created automatically when you add them ...

Similarly, you `create/delete` a backup archive (it didn't exist yet) and you `add/remove` a service (that you created in systemd first for example)

In the context of permission, to me the object really is `created/deleted` in that context (it doesn't exist yet), and then you `add/remove` groups / users to the permission and all the wording is consistent (and we really want to avoid `add` being ambiguously associated with the idea of "initializing" a new permission vs. allowing additional groups/members)

At least that's my thoughts on this ¯\\\_(ツ)\_/¯

src/yunohost/data_migrations/0011_setup_group_permission.py
Josue-T: As I remember I did this and I changed because we said that we can't drop the custom config of the user in case of a custom config.

data/helpers.d/setting
Josue-T: Thanks for the clarification. I'm too for the idea to be more precise about these both term. So yes, we should "create" a permission with the helper. And after we add a user/group in a specific permission.

alexAubin:
Alright folks, I dunno how to proceed with this big PR, but it would to know if there are people planning to provide feedback or that want specific items to be changed in this... But if the team agrees with the overall changes, we could merge this and keep on iterating on in https://github.com/YunoHost/yunohost/pull/797 or other additional PRs

(I'm especially worried that in the meantime translators will be translating strings that are reworked by this PR) (and also we need to move forward with all this)

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
+                            help: User(s) to remove in the group
                             nargs: "*"
                             metavar: USERNAME
                             extra:
                                 pattern: *pattern_username
 
                 ### user_group_info()
                 info:
-                    action_help: Get group information
+                    action_help: Get information for a specific group
```
comradekingu:
```suggestion
                    action_help: Get information about a specific group
```

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
                     api: GET /users/groups/<groupname>
                     arguments:
                         groupname:
-                            help: Groupname to get information
+                            help: Name of the group to get info about
```
comradekingu:
```suggestion
                            help: Name of the group to fetch info about
```

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
-                            nargs: "*"
-                            metavar: USER
-                        -g:
-                            full: --group
-                            help: Group name
-                            nargs: "*"
-                            metavar: GROUP
+                        -s:
+                            full: --short
+                            help: List only the names of permissions
```
comradekingu:
```suggestion
                            help: Only list permission names
```

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
+                            action: store_true
+                        -f:
+                            full: --full
+                            help: Display all informations known about each permissions, including the full list of users corresponding to allowed groups.
```
comradekingu:
```suggestion
                            help: Display all info known about each permission, including the full user list of each group it is granted to.
```
Not 100% sure what the original is conveying.

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
-                            extra:
-                                pattern: *pattern_username
 
-                ### user_permission_remove()
-                remove:
-                    action_help: Revoke access right to users and group
-                    api: PUT /users/permissions/<app>
+                ### user_permission_update()
+                update:
+                    action_help: Grant / remove permissions to groups or users
```
comradekingu:
```suggestion
                    action_help: Grant / remove group permissions, or users
```

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
-                            nargs: "*"
-                            metavar: PERMISSION
-                        -u:
-                            full: --username
-                            help: Username
+                        permission:
+                            help: Permission to manage (e.g. mail or nextcloud or wordpress.editors)
+                        -a:
+                            full: --add
+                            help: Group or user names to add to this permission
```
comradekingu:
```suggestion
                            help: Group or usernames to grant this permission
```

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
-                            metavar: USER
+                            metavar: GROUP_OR_USER
                             extra:
                                 pattern: *pattern_username
-                        -g:
-                            full: --group
-                            help: Group name
+                        -r:
+                            full: --remove
+                            help: Group or user names to remove from this permission
```
comradekingu:
```suggestion
                            help: Group or usernames revoke this permission from
```

data/helpers.d/setting
```diff
@@ -230,70 +230,82 @@ ynh_webpath_register () {
...
+    declare -Ar args_array=( [p]=permission= )
     local permission
     ynh_handle_getopts_args "$@"
 
-    yunohost tools shell -c "from yunohost.permission import permission_remove; permission_remove('$app', '$permission', sync_perm=False)"
+    yunohost tools shell -c "from yunohost.permission import permission_delete; permission_delete('$app.$permission', sync_perm=False)"
 }
 
-# Add a path managed by the SSO
+# Manage urls related to a permission
```
comradekingu:
```suggestion
# Manage URLs related to a permission
```

data/helpers.d/setting
```diff
@@ -230,70 +230,82 @@ ynh_webpath_register () {
...
 #
-# usage: ynh_permission_add_path --app "app" --permission "permission" --url "url" ["url" ...]
-# | arg: app - the application id
-# | arg: permission - the name for the permission
-# | arg: url - the FULL url for the the permission (ex domain.tld/apps/admin)
-ynh_permission_add_path() {
-    declare -Ar args_array=( [a]=app= [p]=permission= [u]=url= )
-    local app
+# usage: ynh_permission_urls --permission "permission" --add "url" ["url" ...] --remove "url" ["url" ...]
```
comradekingu:
```suggestion
# usage: ynh_permission_urls --permission "permission" --add "URL" ["URL" ...] --remove "URL" ["URL" ...]
```

data/helpers.d/setting
```diff
@@ -230,70 +230,82 @@ ynh_webpath_register () {
...
+# | arg: permission - the name for the permission (by default a permission named "main" is removed automatically when the app is removed)
```
comradekingu:
```suggestion
# | arg: permission - the permission name (by default a permission named "main" is removed automatically when the app is)
```

data/helpers.d/setting
```diff
@@ -230,70 +230,82 @@ ynh_webpath_register () {
...
+# | arg: add - (optional) a list of FULL urls to add to the permission (e.g. domain.tld/apps/admin)
```
comradekingu:
```suggestion
# | arg: add - (optional) a list of FULL URLs to add to the permission (e.g. domain.tld/apps/admin)
```

data/helpers.d/setting
```diff
@@ -230,70 +230,82 @@ ynh_webpath_register () {
...
+# | arg: remove - (optional) a list of FULL urls to remove from the permission (e.g. other.tld/apps/admin)
```
comradekingu:
```suggestion
# | arg: remove - (optional) a list of FULL URLs to remove from the permission (e.g. other.tld/apps/admin)
```

src/yunohost/data_migrations/0011_setup_group_permission.py
```diff
@@ -28,22 +28,42 @@ class MyMigration(Migration):
...
+        # ldap search will raise an exception if no corresponding object is found >.> ...
+        except Exception as e:
+            logger.debug("%s does not exist, no need to delete it" % target)
+            return
+
+        objects.reverse()
+        for o in objects:
+            for dn in o["dn"]:
+                dn = dn.replace(",dc=yunohost,dc=org", "")
+                logger.debug("Deleting old object %s ..." % dn)
```
comradekingu:
```suggestion
                logger.debug("Deleting old object '%s' …" % dn)
```

src/yunohost/data_migrations/0011_setup_group_permission.py
```diff
@@ -85,14 +103,18 @@ def migrate_app_permission(self, app=None):
...
             if permission:
                 allowed_group = permission.split(',')
-                user_permission_add([app], permission='main', group=allowed_group, sync_perm=False)
+                user_permission_update(app+".main", remove="all_users", add=allowed_group, sync_perm=False)
             app_setting(app, 'allowed_users', delete=True)
 
-
     def run(self):
+
+        # FIXME : what do we really want to do here ...
```
comradekingu:
```suggestion
        # FIXME : What do we really want to do here …
```

src/yunohost/data_migrations/0011_setup_group_permission.py
```diff
@@ -85,14 +103,18 @@ def migrate_app_permission(self, app=None):
...
+        # Imho we should just force-regen the conf in all case, and maybe
```
comradekingu:
```suggestion
        # IMHO we should just force-regen the conf in all case, and maybe
```

data/actionsmap/yunohost.yml
alexAubin: The idea was to keep "group" and "user" together since you can use both ... idk, or maybe let's rework the whole string with e.g. "Manage permissions accesses"

src/yunohost/data_migrations/0011_setup_group_permission.py
alexAubin: I appreciate the energy you put into this but please, we don't need such level of pedanticness for internal comments and debug messages which are not displayed to the regular user...

Josue-T:
> Alright folks, I dunno how to proceed with this big PR, but it would to know if there are people planning to provide feedback or that want specific items to be changed in this... But if the team agrees with the overall changes, we could merge this and keep on iterating on in #797 or other additional PRs
> 
> (I'm especially worried that in the meantime translators will be translating strings that are reworked by this PR) (and also we need to move forward with all this)

Well, I'm going to try to do a review a quick as possible.

data/actionsmap/yunohost.yml
```diff
@@ -240,132 +250,82 @@ user:
...
-                        app:
-                            help: Application to manage the permission
-                            nargs: "+"
-                        -p:
-                            full: --permission
-                            help: Name of permission (main by default)
-                            nargs: "*"
-                            metavar: PERMISSION
+                        permission:
+                            help: Permission to manage (e.g. mail or nextcloud or wordpress.editors)
```
Josue-T: Does we really want to share to user the concept separate the app name with the name of the permission with a dot ?

alexAubin: Uuuh not sure to understand what you mean ?

Josue-T: Before the technical name of the permission (app.permission) was not shared to the user. On the user side there was just an app with some permission.

src/yunohost/app.py
```diff
@@ -1092,15 +1078,15 @@ def app_addaccess(operation_logger, apps, users=[]):
...
     from yunohost.permission import user_permission_update
 
-    permission = user_permission_update(operation_logger, app=apps, permission="main", add_username=users)
+    output = {}
+    for app in apps:
+        permission = user_permission_update(app+".main", add=users, remove="all_users")
+        output[app] = permission["corresponding_users"]
 
-    result = {p : v['main']['allowed_users'] for p, v in permission['permissions'].items()}
+    return {'allowed_users': output}
```
Josue-T: Well are we really supporting the old concept of this "depreciated" function ?

As I understand we should add something like that:
if all_users group is allowed -> disallow it

data/actionsmap/yunohost.yml
alexAubin: Hmmm idk, our conclusion with ljf was that it was unecessarily complicated to separate the two pieces of the permission and we should just talk about `wordpress.editors` as "the name of the permission" and that's it. 

`user_permission_list` now displays the name of the permission (e.g. wordpress.editors) with associated information with something like : 

```
wordpress.main:
     allowed: all_users
wordpress.editors:
     allowed: volunteers
mail.main:
     allowed: all_users
    
```

and if you want to edit it, you "just" have to `yunohost user permission update wordpress.editors --add foo --remove bar`

(The notion of `app.permission` was also confusing (though not a huge deal) because there's also `mail`, `xmpp` and other things to come which are not really apps)

src/yunohost/backup.py
```diff
@@ -1232,46 +1231,35 @@ def _restore_system(self):
...
-            except:
-                logger.warning(m18n.n('permission_name_not_valid', permission=per['cn'][0]))
+        for permission_name, permission_infos in old_apps_permission.items():
+            app_name = permission_name.split(".")[0]
             if _is_installed(app_name):
-                if not ldap.add('cn=%s,ou=permission' % per['cn'][0], per):
-                    raise YunohostError('apps_permission_restoration_failed', permission=permission_name, app=app_name)
-
+                permission_create(permission_name, urls=permission_infos["urls"], sync_perm=False)
+                user_permission_update(permission_name, remove="all_users", add=permission_infos["allowed"])
```
Josue-T: In some case we really use a edge case for this. By example if the allowed group was `all_users`, we are in a state we the arg `remove` and `add` contains `all_users`. So is this a really good solution ?

src/yunohost/app.py
alexAubin: I think yes ? c.f. a few lines up there, it `remove=all_users` in all cases, so that should reproduce the old behavior I believe ? Though in the case were `all_users` isn't currently allowed it will display a warning and that's it, but that doesn't sound like a breaking change ?

src/yunohost/backup.py
alexAubin: Ah yes good catch ! We probably need to have an `if permissions["allowed"] != default_permissions` or something like this

src/yunohost/backup.py
```diff
@@ -1370,22 +1355,27 @@ def copytree(src, dst, symlinks=False, ignore=None):
...
+
+                    if "allowed" not in permission_infos:
+                        logger.warning("'allowed' key corresponding to allowed groups for permission %s not found when restoring app %s … You might have to reconfigure permissions yourself." % (permission_name, app_instance_name))
+                    else:
+                        should_be_allowed = [g for g in permission_infos["allowed"] if g in existing_groups]
+                        current_allowed = user_permission_list()["permissions"][permission_name]["allowed"]
+                        if should_be_allowed != current_allowed:
+                            user_permission_update(permission_name, remove=current_allowed, add=should_be_allowed)
+
+                os.remove('%s/permissions.yml' % app_settings_new_path)
```
Josue-T: Why do we need to remove this file here ?

alexAubin: No particular reason except that it's only meant for backup purpose and shouldn't later be confused by tech-savvy users as an actual permission config file

src/yunohost/data_migrations/0011_setup_group_permission.py
Josue-T: Anyway I'm for the idea to add this. We might avoid many issue if we force the regen-conf here I think.

src/yunohost/backup.py
Josue-T: Ok, because I was just thinking about a case when this file might be needed a other time for a restoration (by example if the restoration failed). It still just a question because I don't know exactly how work the backup here...

src/yunohost/permission.py
```diff
@@ -35,309 +36,247 @@
...
+#
+#
 
-    Keyword argument:
-        app        -- an application OR sftp, xmpp (metronome), mail
-        permission -- name of the permission ("main" by default)
-        username   -- Username to get informations
-        group      -- Groupname to get informations
 
+def user_permission_list(short=False, full=False, ignore_system_perms=False):
```
Josue-T: Are you sure that it's not usefull to be able to filter by permission, username or group ?

src/yunohost/permission.py
```diff
@@ -35,309 +36,247 @@
...
+        raise YunohostError('permission_not_found', permission=permission)
+
+    # Update permission with default (all_users)
+
+    operation_logger.related_to.append(('app', permission.split(".")[0]))
+    operation_logger.start()
 
     default_permission = {'groupPermission': ['cn=all_users,ou=groups,dc=yunohost,dc=org']}
+    try:
+        ldap.update('cn=%s,ou=permission' % permission, default_permission)
```
Josue-T: We probably should check that we change something in the permission. Because if before to call this function the only group which is allowed is already the group `all_users` LDAP will crash.

alexAubin: Indeed, we can add a test to validate the idempotence of the function

src/yunohost/permission.py
alexAubin: Hmmm idk, I didn't really need it ... Except maybe the "fetch the info for a specific permission" which could be useful in many places of the code. Dunno if we want to have a `user permission list foo.main` or if we want a `user permission info foo.main`  like we do elsewhere ...

Otherwise if you want to know which permissions are allowed for a specific group/user, you can just run `user group info the_user_or_group`

src/yunohost/permission.py
```diff
@@ -349,180 +288,175 @@ def permission_add(operation_logger, app, permission, urls=None, default_allow=T
...
+    logger.debug("The permission database has been resynchronized")
 
     app_ssowatconf()
 
     # Reload unscd, otherwise the group ain't propagated to the LDAP database
     os.system('nscd --invalidate=passwd')
     os.system('nscd --invalidate=group')
+
+
+def _normalize_url(url):
```
Josue-T: Maybe we can move this helper in the domain module if we need it an other time somewhere.

alexAubin: Well in fact in https://github.com/YunoHost/yunohost/pull/797 I removed it because I switched to relative URLs ...

src/yunohost/backup.py
alexAubin: A failed restoration is pretty much like a failed install, so all app files will be removed (including the setting folder). In such case, you can just re-try the restoration and this file will still be in the archive

Josue-T: Ok

src/yunohost/app.py
Josue-T: Ah yes didn't see this...

src/yunohost/permission.py
Josue-T: Ok, let's leave it here.

src/yunohost/data_migrations/0011_setup_group_permission.py
comradekingu: The extraneous effort was writing your comment instead of clicking yes ;)

Non-regular users can also be blind, and then you have to listen to "dot-dot-dot", which gets old. Having non-ellipses in the stringbase is a pain when trying to correct things. If errors are left in comments, they make their way into the other strings sooner or later.

data/actionsmap/yunohost.yml
comradekingu: I went half-and-half, accesses could mean a lot of other things.

Josue-T:
@alexAubin 

Are you going to fix theses fiew things or did you think that I do this ?

alexAubin:
I'll do it at some point but I have too many things going on ... But feel free to take care of it if you want

alexAubin:
Alright folks I propose to merge this as is to move forward and avoid translators translating strings that will be changed by this PR ...

Let's note the remaining items that needs to be adressed (and that will be adressed in follow ups PRs)

- [ ] Possibly force the regen-conf of slapd during migration (display a warning if we're overriding manual changes)
- [ ] We should clarify wether or not we want to allow multiple URLs per permission or not
- [ ] Check for messages that still need proper i18n implementation
- [ ] Fix https://github.com/YunoHost/yunohost/pull/795#discussion_r327328513
- [ ] Fix https://github.com/YunoHost/yunohost/pull/795#discussion_r327793604
- [ ] ???

PR #796, opened 2019-09-14
Maindomain ux feedback

Psycojoker:
## The problem

Small serie of UX feedback I've gathered seeing someone using YunoHost while being "a bit" angry :sweat_smile: 

There are a few more to come later but right now that already self contained :)

## Solution

List of things are:

* rename `--new-domain` of `yunohost tools maindomain` to `--new-main-domain` otherwise you can get the impression that you are adding a new domain
* move `yunohost tools maindomain` to `yunohost domain main-domain` with backward compatibility and useful deprecated warning
* when you can't change remove a domain because it's a main domain, provide much more useful error messages telling why you can't exactly and how to fix that account for the case of "you don't have other domains" and "you have other domains, here how to set another one as main"

## PR Status

Tested and working.

## How to test

Run `yunohost domain remove maindomain.com` with and without other domains
Run all those commands for deprecated warnings:
* yunohost tools maindomain
* yunohost domain maindomain
* yunohost domain main-domain

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/actionsmap/yunohost.yml
```diff
@@ -1584,7 +1599,7 @@ tools:
                 - PUT /domains/main
             arguments:
                 -n:
-                    full: --new-domain
+                    full: --new-main-domain
```
alexAubin: Probably need to propagate this on the controller in the webadmin : https://github.com/YunoHost/yunohost-admin/blob/stretch-unstable/src/js/yunohost/controllers/domains.js#L309

data/actionsmap/yunohost.yml
```diff
@@ -481,6 +481,21 @@ domain:
                             - !!str ^[0-9]+$
                             - "pattern_positive_number"
 
+        ### domain_maindomain()
+        main-domain:
+            action_help: Check the current main domain, or change it
+            deprecated_alias:
+                - maindomain
```
alexAubin: Will this really work considering that it's not in the same category ? Or am I mistaking what this does ?

src/yunohost/domain.py
```diff
@@ -229,6 +236,62 @@ def domain_dns_conf(domain, ttl=None):
...
+
+        os.symlink(new_ssl_key, ssl_key)
+        os.symlink(new_ssl_crt, ssl_crt)
+
+        _set_maindomain(new_main_domain)
+    except Exception as e:
+        logger.warning("%s" % e, exc_info=1)
+        raise YunohostError('maindomain_change_failed')
+
+    _set_hostname(new_main_domain)
```
alexAubin: Eh aren't you missing some import for this ?

src/yunohost/domain.py
```diff
@@ -229,6 +236,62 @@ def domain_dns_conf(domain, ttl=None):
...
+
+    # Generate SSOwat configuration file
+    app_ssowatconf()
```
alexAubin: Missing import I think ?

src/yunohost/tools.py
```diff
@@ -165,59 +165,10 @@ def tools_adminpw(new_password, check_strength=True):
...
-    try:
-        with open('/etc/yunohost/installed', 'r'):
-            regen_conf()
-    except IOError:
-        pass
-
-    logger.success(m18n.n('maindomain_changed'))
+def tools_maindomain(operation_logger, new_main_domain=None):
+    from yunohost.domain import domain_main_domain
+    logger.warning(m18n.g("deprecated_command_alias", prog="yunohost", old="tools maindomain", new="domain maindomain"))
```
alexAubin: Don't we want to point to `domain main-domain` instead of `domain maindomain` ?

src/yunohost/tools.py
```diff
@@ -165,59 +165,10 @@ def tools_adminpw(new_password, check_strength=True):
 
 
 @is_unit_operation()
```
alexAubin: I think we don't need this decorator here anymore since it got moved to the new function

src/yunohost/domain.py
Psycojoker: Oh damn x_x (flake8 was broken for me)

src/yunohost/tools.py
Psycojoker: Are you sure? The other command will take care of that?

src/yunohost/tools.py
Psycojoker: Indeed.

src/yunohost/tools.py
alexAubin: Hmyep since it has the decorator ?

data/actionsmap/yunohost.yml
Psycojoker: Ah yeah, I have code alredy done for that, forgot to push it :sweat:

Psycojoker: https://github.com/YunoHost/yunohost-admin/pull/260 done

data/actionsmap/yunohost.yml
Psycojoker: This one will work for when you write `yunohost domain maindomain`, for `yunohost tools maindomain` it's the `logger.warning` that will handle it.

src/yunohost/domain.py
Psycojoker: Fixed.

src/yunohost/tools.py
Psycojoker: I guess so, I've removed it then.

Psycojoker:
Ok, should be good for everything, sorry for the clunky code ^^'

src/yunohost/domain.py
```diff
@@ -229,6 +237,63 @@ def domain_dns_conf(domain, ttl=None):
     return result
 
 
+@is_unit_operation()
+def domain_main_domain(operation_logger, new_main_domain=None):
```
alexAubin: You might want to use this function now during the postinstall instead of `tools_maindomain` :wink: 

Also ideally we should rename the i18n key `log_tools_maindomain` to `log_domain_main_domain`

Psycojoker: > Also ideally we should rename the i18n key log_tools_maindomain to log_domain_main_domain

Isin't this one likely to "break" existing translations '-'?

Psycojoker: (done for the other one)

alexAubin: >Isin't this one likely to "break" existing translations '-'?

Meh yes I suppose ... but is it really worth maintaining a legacy i18n string for an operation that happens basically only one or two times per installation é_è

Psycojoker: (done for the other one)

> > Isin't this one likely to "break" existing translations '-'?
> 
> Meh yes I suppose ... but is it really worth maintaining a legacy i18n string for an operation that happens basically only one or two times per installation é_è

Actually I think that should be ok if I change everything thinking about it, so let's try it :)

Psycojoker: Should be good :)

comradekingu: Break existing translations how?

Psycojoker: By changing the keys but I don't think it will actually break it.

Psycojoker:
Rebased. Will merge in some days.

Psycojoker:
Delayed to 3.8, 3.7 is already too big.

PR #797, opened 2019-09-15
'visitors' mechanics (replace the old public/private mechanic) + integrate the permission system with SSOwat

alexAubin:
## The problem

The whole `is_public` and `unprotected_uris` mechanic should be integrated in the permission system to have a consistent permission management mechanic

## Solution

So that's a completely untested implementation of the visitors group and a mechanism to replace the weird "unprotected_uris = /" to allow public access.

This way, we get a consistent permission mechanism where the admin can choose for a given app to :
- make it "public", i.e. allow it for the group visitors
- make it "private", i.e. allow it for the group all (registered) users
- make it only available to specific groups and/or users

No need for apps to save and manage some `is_public` state outside the install script. And no need for admins to reinstall an app just to tweak the public/private status ...

Admins and packagers shall be able to reconfigure the publicness/privateness of the app using something like :

```
# cli 
yunohost user permission update wordpress --remove all_users --add visitors

# app helpers
ynh_permission_update --permission main --remove all_users --add visitors
```

... or through the webadmin once it gets implemented

This can easily be extended to subparts of the app (e.g. admin interface). For example eauchat recently was in a situation where he wanted to have the mailman admin interface accessible to visitors because he didn't feel like creating yunohost accounts just for this.

Soooo yup, at least that's the theory, 'cause the hard part ends up managing the legacy `unprotected_uris` (and all the variants protected vs. unprotected and uris vs. regexes). And still need to think about how this correlates with relative or absolute urls and regexes

## PR Status

Not tested at all lol

## How to test

Well uh there are some unit tests drafts for now

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Digging into this, I noticed that many app use `ynh_app_setting_set "$app" skipped_uris "/"` instead of unprotected ... :/ 

Some do this to temporarily allow access to access the setup interface for the app, but some other do use it to make the app public ... dunno if that's just some copypasta they found or if there's a real motivation compared to unprotected

alexAubin:
Soooo, made some progress on this and now I'm stuck ...

- I encountered [this huge conceptual issue](https://github.com/YunoHost/yunohost/blob/7102c5d0ca1ce36a3ff90e3fcb688d4a5be0d221/src/yunohost/app.py#L463-L502) which can't be dealt with the current paradigm of SSOwat. TL;DR : we can't properly handle permissions with multiple urls or regexes right now. Though I really wonder if we really need to allow having multiple urls for one permission ...

- We should add an option `tile_in_ssowat` to the permission system to know if a given permission should be displayed on ssowat... Though (c.f. the previous issue), in the current SSOwat paradigm, having access implies to have the tile displayed and viceversa ...

- And last but not least (the biggest issue imho which really makes the whole thing stuck). Consider the following generated ssowat configuration : 

```
    "protected_urls": [
        "yolo.test/urlpermissionapp", 
        "yolo.test/urlpermissionapp/admin"
    ], 
    "users": {
        "alice": {
            "yolo.test/urlpermissionapp": "Permissions Test", 
            "yolo.test/urlpermissionapp/admin": "Permissions Test (Admin)"
        }, 
        "bob": {
            "yolo.test/urlpermissionapp": "Permissions Test"
        }
    }

```

    - `alice` and `bob` do have access to `/urlpermissionapp` as expected.
    - `visitors` don't have access, as defined in `protected_urls`.
    - `alice` has access to `/admin` as expected
    - ... but `bob` does also have access to `/admin`, because [this function in SSOwat](https://github.com/YunoHost/SSOwat/blob/stretch-unstable/helpers.lua#L235) will only checked if the requested uri starts with an authorized uri for this user.

This is in fact not a bug but a feature on which several apps rely on (though now that I actually dig into it, not so many ...) : lutim, bozon, lektor, dotclear2, lufi, and lstu

So one quick solution (apart rewriting the whole SSOwat paradigm and I really don't want to do that yet...) is to tweak the `has_access` function mentionned earlier to check for a longer match in the list of `protected_urls/regex` and deny access if so.

That will break the apps mentionned in a conservative way (meaning : even authentified user won't be able to access those uris) until they implement the new permission system but I guess we can live with this as they don't seem to be the most critical app ever ?

Josue-T:
> We should add an option tile_in_ssowat to the permission system to know if a given permission should be displayed on ssowat... Though (c.f. the previous issue), in the current SSOwat paradigm, having access implies to have the tile displayed and viceversa ...

I could push on this branch or on an other one a commit to add an attribute in LDAP for this if you want.

Josue-T:
> So one quick solution (apart rewriting the whole SSOwat paradigm and I really don't want to do that yet...) is to tweak the has_access function mentionned earlier to check for a longer match in the list of protected_urls/regex and deny access if so.

As I understand with your solution we need to do some change of SSOwat. So the minimum is:
- Add the `protected_url` key support
- tweak the `has_access` function to check for a longer match in the list of protected_urls/regex and deny access if so.

And as I see your example we have a duplication of information (between each user and the `protected_url`). I don't mind but maybe there are a cleaner way to do this :smile: 

So my personal preference is to change this concept but it still my point of view.

alexAubin:
>So my personal preference is to change this concept but it still my point of view.

Well yes I do agree that in the ideal world we should just rewrite the whole format of the ssowat config and handling of the un/protected uris/regexes and the user ACL thingies. But as always, things are less about agreeing and the principle and more about what are the practical implications. In this case, the fact that we gotta keep supporting the damn legacy, and that it's going to be a huge mess if we do this right now (and I don't really know when will be a good time actually, maybe in ~6months or 1 year when most of the app will have moved to the new permission system or idk...) But anyway this whole permission thing is already huge as fuck, so if we can find some pragmatic middle ground that'll be better :/

Josue-T:
Ideally my purpose for the next SSOwat config for now would be by example:
```json
{
    "/route1": {
        "permission": ["visitors", "user1", "user2", ...],   # Public route
        "label": "App1"
    },
    "/route2_with_a_regex$": {
        "permission": ["user1", "user2"],       # Private route
        "label": None,
   },
    "/route3": {
        "permission": "skipped",                   # Skipped route idk
        "label": None,
    },
    "/route4": {
        "permission": "unprotected",           # Unprotected route idk
        "label": "App3"
    }
}
```

But yes I know, implementing this on the SSOwat side is a work.

data/helpers.d/setting
```diff
@@ -176,6 +176,8 @@ else:
     elif action == "set":
         if key in ['redirected_urls', 'redirected_regex']:
             value = yaml.load(value)
+        if key in ["unprotected_uris", "unprotected_regex", "protected_uris", "protected_regex"]:
+           logger.warning("/!\\ Packagers ! This app is using the legacy permission system. Please delete these legacy settings and use the new helpers ynh_permission_{create,urls,update,delete} and the 'visitors' group to manage public/private access.")
```
comradekingu:
```suggestion
           logger.warning("/!\\ Packagers! This app is using the legacy permission system. Please delete these legacy settings and use the new helpers 'ynh_permission_{create,urls,update,delete}' and the 'visitors' group to manage public/private access.")
```

data/helpers.d/setting
```diff
@@ -230,11 +232,24 @@ ynh_webpath_register () {
...
 # | arg: urls - (optional) a list of FULL urls for the permission (e.g. domain.tld/apps/admin)
+# | arg: urls - (optional) a list of urls to specify for the permission.
+#
+#  Urls are assumed to be relative to the app domain/path if they start with '/'.
+#  For example:
+#     /                             -> domain.tld/app
+#     /admin                        -> domain.tld/app/admin
+#     domain.tld/app/api            -> domain.tld/app/api
+#
+#  Urls can be treated as regexes when they start with "re:".
```
comradekingu:
```suggestion
#  URLs can be treated as regexes when they start with "re:".
```

comradekingu:
```suggestion
#  URLs can be treated as regexes when they start with "re:".
```

data/helpers.d/setting
```diff
@@ -230,11 +232,24 @@ ynh_webpath_register () {
...
 # Create a new permission for the app
 #
+# example: ynh_permission_create --permission admin --urls /admin
+#
 # usage: ynh_permission_create --permission "permission" [--urls "url" ["url" ...]]
 # | arg: permission - the name for the permission (by default a permission named "main" already exist)
 # | arg: urls - (optional) a list of FULL urls for the permission (e.g. domain.tld/apps/admin)
+# | arg: urls - (optional) a list of urls to specify for the permission.
+#
+#  Urls are assumed to be relative to the app domain/path if they start with '/'.
```
comradekingu:
```suggestion
#  URLs are assumed to be relative to the app domain/path if they start with '/'.
```

data/helpers.d/setting
```diff
@@ -230,11 +232,24 @@ ynh_webpath_register () {
 
 # Create a new permission for the app
 #
+# example: ynh_permission_create --permission admin --urls /admin
+#
 # usage: ynh_permission_create --permission "permission" [--urls "url" ["url" ...]]
 # | arg: permission - the name for the permission (by default a permission named "main" already exist)
 # | arg: urls - (optional) a list of FULL urls for the permission (e.g. domain.tld/apps/admin)
+# | arg: urls - (optional) a list of urls to specify for the permission.
```
comradekingu:
```suggestion
# | arg: urls - (optional) a list of URLs to specify for the permission.
```

locales/en.json
```diff
@@ -230,6 +230,9 @@
     "group_already_exist_on_system": "Group {group} already exists in the system group",
     "group_created": "Group '{group}' successfully created",
     "group_creation_failed": "Failed to create group {group}: {error}",
+    "group_cannot_edit_all_users": "The group 'all_users' cannot be edited manually. It is a special group meant to contain all users registered in Yunohost",
```
comradekingu:
```suggestion
    "group_cannot_edit_all_users": "The group 'all_users' cannot be edited manually. It is a special group meant to contain all users registered in YunoHost",
```

src/yunohost/permission.py
```diff
@@ -142,10 +142,15 @@ def user_permission_update(operation_logger, permission, add=None, remove=None,
...
 
-    if len(new_allowed_groups) > 1 and "all_users" in new_allowed_groups:
-        # FIXME : i18n
-        # FIXME : write a better explanation ?
-        logger.warning("This permission is currently enabled for all users in addition to other groups. You probably want to either remove the 'all_users' permission or remove the specific groups currently allowed.")
+    if len(new_allowed_groups) > 1:
+        if "all_users" in new_allowed_groups:
+            # FIXME : i18n
+            # FIXME : write a better explanation ?
+            logger.warning("This permission is currently enabled for all users in addition to other groups. You probably want to either remove the 'all_users' permission or remove the other groups currently allowed.")
```
comradekingu:
```suggestion
            logger.warning("This permission is currently granted to all users in addition to other groups. You probably want to either remove the 'all_users' permission or remove the other groups it is currently granted to.")
```

src/yunohost/permission.py
```diff
@@ -142,10 +142,15 @@ def user_permission_update(operation_logger, permission, add=None, remove=None,
...
+        if "visitors" in new_allowed_groups:
+            # FIXME : i18n
+            # FIXME : write a better explanation ?
+            logger.warning("This permission is currently enabled for visitors in addition to other groups. You probably want to either remove the 'visitors' permission or remove the other groups currently allowed.")
```
comradekingu:
```suggestion
            logger.warning("This permission is currently granted to visitors in addition to other groups. You probably want to either remove the 'visitors' permission or remove the other groups it is currently granted to.")
```

src/yunohost/permission.py
```diff
@@ -263,7 +268,18 @@ def permission_create(operation_logger, permission, urls=None, sync_perm=True):
 
     Keyword argument:
         permission -- Name of the permission (e.g. mail or nextcloud or wordpress.editors)
-        urls       -- list of urls to specify for the permission
+        urls       -- list of urls to specify for the permission.
```
comradekingu:
```suggestion
        urls       -- list of URLs to specify for the permission.
```

src/yunohost/permission.py
```diff
@@ -263,7 +268,18 @@ def permission_create(operation_logger, permission, urls=None, sync_perm=True):
...
+
+    Urls are assumed to be relative to the app domain/path if they start with '/'.
+    For example:
+       /                             -> domain.tld/app
+       /admin                        -> domain.tld/app/admin
+       domain.tld/app/api            -> domain.tld/app/api
+
+    Urls can be later treated as regexes when they start with "re:".
```
comradekingu:
```suggestion
    URLs can be later treated as regexes when they start with "re:".
```

src/yunohost/permission.py
```diff
@@ -321,8 +337,8 @@ def permission_urls(operation_logger, permission, add=None, remove=None, sync_pe
 
     Keyword argument:
         permission -- Name of the permission (e.g. mail or nextcloud or wordpress.editors)
-        add        -- List of urls to add
-        remove     -- List of urls to remove
+        add        -- List of urls to add (c.f. permission_create for documentation about their format)
```
comradekingu:
```suggestion
        add        -- List of URLs to add (c.f. permission_create for documentation about their format)
```

src/yunohost/permission.py
```diff
@@ -321,8 +337,8 @@ def permission_urls(operation_logger, permission, add=None, remove=None, sync_pe
...
+        remove     -- List of urls to remove (c.f. permission_create for documentation about their format)
```
comradekingu:
```suggestion
        remove     -- List of URLs to remove (c.f. permission_create for documentation about their format)
```

src/yunohost/tests/test_permission.py
```diff
@@ -156,6 +158,31 @@ def check_permission_for_apps():
...
+                         data={"user": logged_as,
+                               "password": dummy_password},
+                         headers={"Referer": sso_url,
+                                  "Content-Type": "application/x-www-form-urlencoded"},
+                         verify=False)
+            # We should have some cookies related to authentication now
+            assert session.cookies
+            r = session.get(webpath, verify=False)
+
+    # If we can't access it, we got redirected to the sso
```
comradekingu:
```suggestion
    # If we can't access it, we got redirected to the SSO
```

data/helpers.d/setting
```diff
@@ -230,11 +232,24 @@ ynh_webpath_register () {
...
+#
+#  Urls are assumed to be relative to the app domain/path if they start with '/'.
+#  For example:
+#     /                             -> domain.tld/app
+#     /admin                        -> domain.tld/app/admin
+#     domain.tld/app/api            -> domain.tld/app/api
```
comradekingu: Should that be "API"?

alexAubin: uuuuuh idk, it's really just an example to explain what's going on, the case doesn't really matter ...

src/yunohost/app.py
```diff
@@ -1366,34 +1451,66 @@ def _get_setting(settings, name):
...
+
+        # Redirected
+        redirected_urls.update(app_settings.get('redirected_urls', {}))
+        redirected_regex.update(app_settings.get('redirected_regex', {}))
+
+        # Legacy permission system using (un)protected_uris and _regex managed in app settings...
+        unprotected_urls += [_sanitized_absolute_url(uri) for uri in _get_setting(app_settings, 'unprotected_uris')]
+        protected_urls += [_sanitized_absolute_url(uri) for uri in _get_setting(app_settings, 'protected_uris')]
+        unprotected_regex += _get_setting(app_settings, 'unprotected_regex')
+        protected_regex += _get_setting(app_settings, 'protected_regex')
```
Josue-T: Are you sure that you can't have multiple `protected_regex` and `unprotected_regex` ?

src/yunohost/app.py
```diff
@@ -1366,34 +1451,66 @@ def _get_setting(settings, name):
...
-            for item in _get_setting(app_settings, 'protected_regex'):
-                protected_regex.append(item)
-            if 'redirected_urls' in app_settings:
-                redirected_urls.update(app_settings['redirected_urls'])
-            if 'redirected_regex' in app_settings:
-                redirected_regex.update(app_settings['redirected_regex'])
+                protected_urls += urls
+
+                # Legacy stuff : we remove now unprotected-urls that might have been declared as protected earlier...
+                unprotected_urls = [u for u in unprotected_urls if u not in urls]
```
Josue-T: Are you sure of that ? :-/ At line 1506 OK, but here I don't understand why...

src/yunohost/data_migrations/0011_setup_group_permission.py
```diff
@@ -62,14 +62,20 @@ def migrate_LDAP_db(self):
         try:
             self.remove_if_exists("cn=sftpusers,ou=groups")
             self.remove_if_exists("ou=permission")
-            self.remove_if_exists('cn=all_users,ou=groups')
+            self.remove_if_exists('ou=groups')
```
Josue-T: In this case I think we could remove the line 63.

src/yunohost/app.py
```diff
@@ -1366,34 +1451,66 @@ def _get_setting(settings, name):
...
+
+        # Skipped
+        skipped_urls += [_sanitized_absolute_url(uri) for uri in _get_setting(app_settings, 'skipped_uris')]
+        skipped_regex += _get_setting(app_settings, 'skipped_regex')
+
+        # Redirected
+        redirected_urls.update(app_settings.get('redirected_urls', {}))
+        redirected_regex.update(app_settings.get('redirected_regex', {}))
+
+        # Legacy permission system using (un)protected_uris and _regex managed in app settings...
```
Josue-T: Don't really understand why maintain the legacy permission here if you migrate the old `unprotected_uris` in the visitor group (in the migration).

src/yunohost/user.py
```diff
@@ -670,13 +670,18 @@ def user_group_update(operation_logger, groupname, add=None, remove=None, force=
...
     # Those kind of group should only ever contain the user (e.g. sam) and only this one.
     # We also can't edit "all_users" without the force option because that's a special group...
-    existing_users = user_list()['users'].keys()
-    uneditable_groups = existing_users + ["all_users", "admins"]
-    if groupname in uneditable_groups and not force:
-        raise YunohostError('group_cannot_be_edited', group=groupname)
+    if not force:
+        if groupname == "all_users":
+            raise YunohostError('group_cannot_edit_all_users')
+        elif groupname == "all_users":
```
Josue-T:
```suggestion
        elif groupname == "visitors":
```

src/yunohost/app.py
alexAubin: These are already lists, and `+=` "extends" the list, it doesn't just append one value. 

e.g.

```
L = [1,2]
L += [3,4]
```

results in `L == [1,2,3,4]`

src/yunohost/app.py
alexAubin: Hmmm, idk, don't remember exactly, but basically if add an url as protected, it shouldn't be "unprotected" by legacy stuff ... E.g. say that you install an app still using the legacy stuff as public : 
- it will have declared `unprotected_urls = /` in its setting
- now the user use the permission system to declare that it shouldn't be public anymore
- so now the new permission system will set `/` as protected
- but without the line we're discussing, `/` will also be in the unprotected urls ...

src/yunohost/app.py
Josue-T: Oh, yes. I didn't see that `_get_setting` a list...

src/yunohost/app.py
alexAubin: Because the migration only covers the case where unprotected_uris is /, which is the classic use case, but there are many others that we can't migrate automagically where apps defined specific unprotected_uris and protected_uris

src/yunohost/app.py
Josue-T: Ok, but why we don't cleanup the settings and remove all old permission things ?

Josue-T: Well reading your other answer I understand...

Josue-T: > Hmmm, idk, don't remember exactly, but basically if add an url as protected, it shouldn't be "unprotected" by legacy stuff ... E.g. say that you install an app still using the legacy stuff as public :
>    - it will have declared `unprotected_urls = /` in its setting
>    - now the user use the permission system to declare that it shouldn't be public anymore
>    - so now the new permission system will set `/` as protected
>    - but without the line we're discussing, `/` will also be in the unprotected urls ...

For me there are 2 cases:
- The permission was migrated automatically. In this case we can use the permission to change the access to the app. And in this case we don't need this line.
- The permission wasn't migrated because it was too customized. in this case the user is not able to change the permission (because it was not migrated). So we don't have any permission for this custom case in LDAP. So this line in this case is also not needed.

alexAubin: >The permission was migrated automatically. In this case we can use the permission to change the access to the app. And in this case we don't need this line.

It's fuzzy to me, but I think that we still need to keep the line, because as said in the previous comment, if the app install declares `unprotected_urls = /` (so the app is public) then the user changes the permission to private using the new permission system, then the line `unprotected_urls = /` is **still there** so we need to remove it from the "compiled" ssowat configuration if there's a conflict with the new permission system ...

alexAubin:
Sooo, implemented the remaining items from https://github.com/YunoHost/yunohost/pull/795#issuecomment-538499638 . Not that I finally chose to drop the "multiple urls per permission" thing because I don't see any usecase where that'd be useful and that add unecessary complexity

I also fixed the inteface with SSOwat in https://github.com/YunoHost/SSOwat/pull/147

Validated everything using unit tests etc.

So that PR and the permission system is finally pretty much complete to me :+1:

alexAubin:
Soooo planning merge this like during Tuesday's meeting to move forward and possibly have an alpha-stage release

alexAubin:
Sooo, merging, moving on with heavy testing using the unstable app CI to validate the 3.7 release..

PR #803, opened 2019-09-19
Get rid of etckeeper

alexAubin:
## The problem

etckeeper has been added with no actual motivation and we still don't really use it.

But it does create several issues with user encountering cases where etckeeper complains because no identity is defined, or it display huge boring log about files being versionned when you run apt, or the git index ends up being broken for weird reason which leads to other issues, or somehow the git history ends up being several Go's big.

## Solution

This was a big YAGNI. Get rid of it.

## PR Status

Yolo commited

## How to test

Test a new install with this, and check if stuff goes okay on previous instances

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


taziden:
I understand the reasons for removing it. That's too bad because it can be so very useful when something messes up your system config files and you have no backup (which is the case for most users unfortunately) :(

Psycojoker:
> etckeeper has been added with no actual motivation 

That's not really true, it has been added at the beginning of the integration of regen-conf with the idea of having an history of configuration modification and being able to rollback on broken configuration (also at that time regen-conf had broken some system in some situation.)

It ended up not being correctly integrated but it's very far away from being a YAGNI and I still think that having both history and rollback of configuration is a good feature for being able to debug things, but yeah, the current state isn't really great.

It would probably be easier and more robust to move to custom git/something else commands only on the files touched by the regen-conf to get those features and not use the full etckeeper and have an integration in core YunoHost.

PR #793, opened 2019-09-10
[fix] Stale / undefined i18n strings

alexAubin:
## The problem

There are many unused i18n strings which creates artificial work for translators ...

Also, some strings which are used in the code are not defined anywhere.

## Solution

I use this small old script in `tests/_test_m18nkeys.py` to parse the code in a very dirty way to identify what strings are used inside `m18n.n` and `YunohostError`

I fixed a few things, but in the these strings are still not defined : 

```
   - apps_already_up_to_date
   - bind_mouting_disable
   - error_reading_file
   - operation_interrupted
   - permission_update_failed_clear
   - permission_update_failed_populate
   - premission_update_failed
```

## PR Status

Should implement / fix those missing strins somehow

## How to test

Uh idk :[ I guess mostly the thing to test is the small refactor about testing the requirements

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #805, opened 2019-09-20
Simplify madness code about checking requirements

alexAubin:
## The problem

Code is unecessarily complex and define strings which are basically unused

(This PR was previously in https://github.com/YunoHost/yunohost/pull/793 )

## Solution

Simplify code

## PR Status

I don't remember if I really tested this, maybe adding some small unit tests could be nice

## How to test

Try to install an app while playing with requirements in the manifest

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Flagging as merging soon

PR #811, opened 2019-09-26
Better handling of app install and remove failures

alexAubin:
## The problem

While trying to package an app, I messed up my environment because the app install failed, then the app auto removal was hanging forever. So I Ctrl+C, but it did not clean the permission and other things.

Working on this, I again realized that the handling of the install failure was not so clear either (both the code and the messages) so fixed it ...

## Solution

- When we remove an app (after a failed install or a regular removal), handle the KeyboardInterrupt exceptions (which are in fact not caught by `Exception`)
- Simplify the handling of failed install (script failure, manual interrupt, exception in yunohost's code, or app that broke the system)
- Improve messaging

## PR Status

Ongoing tests while I'm packaging an app

## How to test

Try to install buggy apps with buggy install or buggy remove (e.g. a `sleep 3000`)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -908,77 +908,101 @@ def confirm_install(confirm):
...
             broke_the_system = False
             _assert_system_is_sane_for_app(manifest, "post")
         except Exception as e:
             broke_the_system = True
-            error_msg = operation_logger.error(str(e))
+            logger.exception(str(e))
+            operation_logger.error(str(e))
+
+        # If the install failed or broke the system, we remove it
+        if install_failed or broke_the_system:
```
Psycojoker: From an UX pov I think we should tell the user which case is selected, like : "we found out that this app broke the system" vs "the installation script failed".

It might be covered earlier with all the logger.exception but I'm not sure :s 

Or at least at the end with the exception raised?

src/yunohost/app.py
```diff
@@ -908,77 +908,101 @@ def confirm_install(confirm):
...
+            if remove_retcode != 0:
+                msg = m18n.n('app_not_properly_removed',
+                             app=app_instance_name)
+                logger.warning(msg)
+                operation_logger_remove.error(msg)
+            else:
+                try:
+                    _assert_system_is_sane_for_app(manifest, "post")
+                except Exception as e:
+                    operation_logger_remove.error(e)
```
Psycojoker: Again, is the error message good UX here? 

Like informing the user about a list of things that needs to be fixed seems like a good idea '-'

src/yunohost/app.py
```diff
@@ -1057,11 +1081,24 @@ def app_remove(operation_logger, app):
...
+    except (KeyboardInterrupt, EOFError, Exception):
+        ret = -1
+        import traceback
+        logger.exception(m18n.n('unexpected_error', error=u"\n" + traceback.format_exc()))
 
+    if ret == 0:
+        logger.success(m18n.n('app_removed', app=app))
         hook_callback('post_app_remove', args=args_list, env=env_dict)
+    else:
+        logger.warning(m18n.n('app_not_properly_removed', app=app))
```
Psycojoker: We don't do `_assert_system_is_sane_for_app(manifest, "post")` on remove operation? Dunno if that would be a good idea.

src/yunohost/app.py
```diff
@@ -908,77 +908,101 @@ def confirm_install(confirm):
...
     # Execute the app install script
-    install_retcode = 1
+    install_failed = True
     try:
         install_retcode = hook_exec(
             os.path.join(extracted_app_folder, 'scripts/install'),
             args=args_list, env=env_dict
         )[0]
+        # "Common" app install failure : the script failed and returned exit code != 0
+        install_failed = (install_retcode != 0)
```
decentral1se:
```suggestion
        install_failed = True if install_retcode != 0 else False
```

src/yunohost/app.py
alexAubin: Yes it's covered by the earlier `logger.exception` ... (dunno what it really does compared to `logger.error` but anyway the message is displayed)

src/yunohost/app.py
alexAubin: Yes we just reuse the message from https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/app.py#L2982

src/yunohost/app.py
Psycojoker: (it's just a different level of importance of exceptions, like exception > error, but I don't think we really handle them differently)

src/yunohost/app.py
alexAubin: Dunno either ... more generally I wish we could just find a way to reuse the regular `app_remove` instead of duplicating this code which generates inconsistencies when we make it evolve and forget to propagate stuff on both parts (not to mention the similar piece of code in `backup_restore`)

src/yunohost/app.py
```diff
@@ -908,77 +908,101 @@ def confirm_install(confirm):
...
-                permission_list = [p['cn'][0] for p in result]
-                for l in permission_list:
-                    permission_remove(app_instance_name, l.split('.')[0], force=True)
-
-                if remove_retcode != 0:
-                    msg = m18n.n('app_not_properly_removed',
-                                 app=app_instance_name)
-                    logger.warning(msg)
-                    operation_logger_remove.error(msg)
+            # Here again, calling hook_exec could failed miserably, or get
```
decentral1se:
```suggestion
            # Here again, calling hook_exec could fail miserably, or get
```

src/yunohost/app.py
```diff
@@ -1057,11 +1081,24 @@ def app_remove(operation_logger, app):
...
     operation_logger.flush()
 
-    if hook_exec('/tmp/yunohost_remove/scripts/remove', args=args_list,
-                 env=env_dict)[0] == 0:
-        logger.success(m18n.n('app_removed', app=app))
+    try:
+        ret = hook_exec('/tmp/yunohost_remove/scripts/remove',
+                        args=args_list,
+                        env=env_dict)[0]
+    # Here again, calling hook_exec could failed miserably, or get
```
decentral1se:
```suggestion
    # Here again, calling hook_exec could fail miserably, or get
```

PR #812, opened 2019-09-26
[enh] Add some advice about a strange locale problem with postgresql

zamentur:
## The problem

When locale are badly configured (should not be the case cause the install script manage this case, but it might be) postgresql is not correctly installed ! the dir /etc/postgresql is empty

## Solution

I suggest to add a message to redirect on the stackoverflow issue, just in case

## PR Status

Ready

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
It is not possible to more accurately detect this situation and have a more precise error message?

autra:
We could check that the values in `LC_*` vars are in the output of `localectl list-locales`. If not, either we generate them, or we throw an error.

PR #813, opened 2019-09-27
[fix] Psql user should own the database

zamentur:
## The problem

I found a bug in odoo 9 where the bug is due to the fact the psql user don't own the db.
I think it should.

## Solution

Change the db owner

## PR Status

Ready

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/postgresql
+++ b/data/helpers.d/postgresql
@@ -87,7 +87,7 @@ ynh_psql_create_db() {
 
 	# grant all privilegies to user
 	if [ -n "$user" ]; then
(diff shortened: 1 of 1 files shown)


decentral1se:
> I found a bug in odoo 9

What was the bug?

Do you forsee any issues with this affecting other installations?

alexAubin:
Yolomerging to move forward.

PR #814, opened 2019-09-27
Spookier warnings for dangerous app install

alexAubin:
## The problem

Currently the warning message about notworking / thirdparty apps is displayed and user can bypass by simply entering `Y`.

When we are able to detect that the app is explicitly not working (or is third party) it should be much spookier and we should tell the user to type something like `Yes, I understand` to incetivise much more to read and think about what they're doing.

Moreover, READMEs like [this one](https://github.com/YunoHost-Apps/haste_ynh) will suggest to type `sudo yunohost app install -l Haste https://github.com/YunoHost-Apps/haste_ynh` and we should make the code a bit more smart and able to check the state of app `haste` in the appslist

## Solution

- Red / yellow warnings instead of blue
- Spookier messages saying that no support will be provided if installing a dangerous app breaks the system or doesnt work
- Ask to type `Yes, I undestand` for broken / thirdparty apps (also to incentivise people to actually [add their app in the damn application list](https://dash.yunohost.org/appsobservatory/unlisted))
- Decent quality apps should now be at least level 5 (to be consistent with the webadmin)

Screenshot : 

2019-09-27-174400_1366x768_scrot

## PR Status

Works on my machine™

## How to test

Try for example to `sudo yunohost app install -l Haste https://github.com/YunoHost-Apps/haste_ynh`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


kay0u:
It will be a mess for the app packager (although I understand the advantages). Is there a way to bypass this message with ynh-dev, with an env var for example

src/yunohost/app.py
```diff
@@ -789,20 +789,41 @@ def confirm_install(confirm):
...
     if app in raw_app_list or ('@' in app) or ('http://' in app) or ('https://' in app):
+
+        # If we got an app name directly (e.g. just "wordpress"), we gonna test this name
         if app in raw_app_list:
-            state = raw_app_list[app].get("state", "notworking")
-            level = raw_app_list[app].get("level", None)
+            app_name_to_test = app
+        # If we got an url like "https://github.com/foo/bar_ynh, we want to
+        # extract "bar" and test if we know this app
+        elif ('http://' in app) or ('https://' in app):
```
decentral1se: https://docs.python.org/2/library/urlparse.html?

Psycojoker: `app.lstrip().startswith(('http://', 'https://'))` is you really want minors improvement (but that really doesn't change anything here)

alexAubin: I was just reusing the same test done in the `if` like 3 lines earlier ¯\\\_(ツ)\_/¯

PR #815, opened 2019-09-27
Allow system users to auth on the mail stack and send emails

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/947

This is not easy to configure an app to send emails. While packaging an app (transpay), it explicitly required to be able to auth on the mail server. I started creating a system user ... but realized it could not authenticated (because only LDAP accounts are able to with the current conf). 

## Solution

Thanks to the help of @taziden I was able to configure dovecot to accept system users to authenticate. For this, a new passwd-like file `/etc/postfix/app_senders_login_maps` is added and filled with the appropriate app users.

```
foobar:{CRYPT}$6$SivRDh9NDh3w5.zs$Flq1MfFWq/RfVq3VvssWPwR2pMDWlTN87hretdavCUZlTwkROY1vmpqPxdEpeVg8AafmpDrGfbnyEwe7/5NX1/::::::allow_nets=127.0.0.1/24
```

But postfix also needs to be tweak to allow `foobar` to send emails with From equal to `foobar@somedomain.tld`. For this, I added a new file `/etc/postfix/app_senders_login_maps` to the `smtpd_sender_login_maps`. It is automatically filled with the appropriate user map

```
foobar@somedomain.tld foobar
```

On the packaging side, one only need to have:

```toml
    [resources.system_user]
    allow_email = true
```

which will automatically provision a `mail_pwd` setting

Then configure the app to use the app id as login (NB: the app id, NOT appid@domain.tld) and `$mail_pwd`/`__MAIL_PWD__`  as password

## PR Status

Tested and working

### DIFF
--- a/hooks/conf_regen/25-dovecot
+++ b/hooks/conf_regen/25-dovecot
@@ -53,6 +53,8 @@ do_post_regen() {
     chown root:mail /var/mail
     chmod 1775 /var/mail
 
+    python3 -c 'from yunohost.app import regen_mail_app_user_config_for_dovecot_and_postfix as r; r(only="dovecot")'
+
     [ -z "$regen_conf_files" ] && exit 0
 
     # compile sieve script
--- a/conf/dovecot/dovecot.conf
+++ b/conf/dovecot/dovecot.conf
@@ -44,20 +44,20 @@ passdb {
   driver = ldap
 }
 
-# Internally, allow authentication from system user
-# who might want to send emails (e.g. from apps)
+# Internally, allow authentication from apps system user who have "enable_email = true"
 passdb {
-  override_fields = allow_nets=127.0.0.1/32,::1/64
-  driver = pam
+  driver = passwd-file
+  args = /etc/dovecot/app-senders-passwd
 }
 
 userdb {
-  args = /etc/dovecot/dovecot-ldap.conf
   driver = ldap
+  args = /etc/dovecot/dovecot-ldap.conf
 }
 
 userdb {
-  driver = passwd
+  driver = passwd-file
+  args = /etc/dovecot/app-senders-passwd
 }
 
 protocol imap {
--- a/conf/dovecot/dovecot.conf
+++ b/conf/dovecot/dovecot.conf
@@ -38,14 +38,26 @@ ssl_prefer_server_ciphers = no
 ###############################################################################
 
 
+# Regular Yunohost accounts
 passdb {
   args = /etc/dovecot/dovecot-ldap.conf
   driver = ldap
 }
 
+# Internally, allow authentication from system user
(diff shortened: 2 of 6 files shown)


Josue-T:
Hello,

The other solution would be to create a "special" user in LDAP for the app in, by example,  `ou=apps,dc=yunohost,dc=org`. The idea is to create a helper which create this specific user. The other adventage is that we also solve the issue said by @yalh76, about the authentication of some apps to LDAP with a specific user and password.

alexAubin:
>The other adventage is that we also solve the issue said by @yalh76, about the authentication of some apps to LDAP with a specific user and password.

Uh can you elaborate on this ?

Josue-T:
So, @yalh76 said at one meeting that some apps (like mastodon) need a user and password to access to LDAP, even if the access is free. So one solution was to create a yunohost user and use this user to be able to authenticate to LDAP. But this solution is not really clean because an app is not really a user. So my purpose to solve this  is (and I think @yalh76 said the same thing) to create a user but in a other place in LDAP, so the real user is separated to the "app" user. So to the solution is by example to put the real user in LDAP in `ou=users,dc=yunohost,dc=org` and the user for the apps in `ou=apps,dc=yunohost,dc=org`. The other advantage of that is that you can also use this user for email by example, so by this you can have more flexibility.

If I would implement this (I've written  this in my todo list but I have just no many things, so it might be never done...), I would create a function in the core based on the function for the user, but maybe with just some customization more adapted on to the apps.

zamentur:
So what's the status? is this PR needed ? Or https://github.com/YunoHost/yunohost/pull/977 is enough to fix the issue ?

Gredin67:
up @zamentur @alexAubin see https://github.com/YunoHost-Apps/synapse_ynh/issues/362

hooks/conf_regen/25-dovecot
```diff
@@ -53,6 +53,8 @@ do_post_regen() {
     chown root:mail /var/mail
     chmod 1775 /var/mail
 
+    python3 -c 'from yunohost.app import regen_mail_app_user_config_for_dovecot_and_postfix as r; r(only="dovecot")'
```
alexAubin, line 56: Note that we have these weird lines in dovecot and postfix regenconf because ... I need a regenconf-like mechanism but also wanted to trigger the regeconf from the resource reconciliation python snippet ... And the code that generate these files was easier to write in Python ...

alexAubin:
Discussed today : 
- we need a mechanism (maybe an app setting ?) to be able to override the user part
- same for the domain part

alexAubin:
Remaining items are done, planning to merge in 11.2

PR #818, opened 2019-10-05
Update the email regex so it accept the '+' sign.

jershon:
## The problem

Fix https://github.com/YunoHost/issues/issues/1333.

## Solution

Add \+ in the first part of the regex.

## PR Status

Done

## How to test

Add a new email address containing the + symbol.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Psycojoker:
We don't especially really want that, see discussion here https://github.com/YunoHost/yunohost/pull/770

alexAubin:
In fact, the initial issue is about *email forward*, not about the email of the user account ...

Turns out that these two are probably different format because : 
- it doesn't make sense that your "raw" email contains a `+` since `+` is to create a sort of on-the-fly "sub"-email ?
- but it does make sense to forward to an email address containing a `+`, for example to automatically sort it ...

So instead of editing the mail argument during create, this PR should be tweaking the pattern for the `--add-mailforward` arg in `user update` I suppose ?

cyxae:
As said in the first post of yunohost/issues#1333 , someone could choose an email address like nickname+whatever@serv.er as its primary (or "raw") address so I think that jershon's commit is relevant in this case.

alexAubin:
@cyxae : uh but I still don't understand ... This question during user creation is about *creating a mailbox*. Think about it like creating a yahoo or gmail account (yes, bad evil google etc whatever) : you can't create an account with "foo+bar@gmail.com" because that assumes the mailbox "foo@gmail.com" already exists but it doesnt... It does make sense to forward email to "foo+bar" (which'll endup in mailbox foo), but you can't create a mailbox named "foo+bar".

zamentur:
Sorry i made my change in rebase conflict resolving...

So now the + is accepted with forward but not with main mailbox and alias.

zamentur (approved):
LGTM tested

alexAubin (approved):

alexAubin:
(Btw I realize now that maybe we also have to tweak the SSO code if we want to allow people to add forward emails with + themselves from the user portal as well ..)

PR #819, opened 2019-10-06
Use str instead of strerror

decentral1se:
See https://forum.yunohost.org/t/cant-create-a-user-after-post-intsallation/9190.

## The problem

https://forum.yunohost.org/t/cant-create-a-user-after-post-intsallation/9190

## Solution

Use `str`.

## PR Status

Untested.

## How to test

Do something weird in your `/etc/ssowat/conf.json.persistent` file and see that this error is reported.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Yolomerging because already saw that kind of issue and applied the same fix in other contexts

PR #827, opened 2019-10-25
[FIX] find udp port

kay0u:
## The problem

- *The actual helper don't catch the use of udp port*

## Solution

- *Use netstat instead of netcat*

The current way is to try to open this port, if it fail inc the port and try again...

My way just checks if it used, if it is inc the port and try again.

## PR Status

I can add arguments to this helper to force the check in ipv4 or ipv6 if needed (check both by default)

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/network
+++ b/data/helpers.d/network
@@ -17,7 +17,7 @@ ynh_find_port () {
 	ynh_handle_getopts_args "$@"
 
 	test -n "$port" || ynh_die --message="The argument of ynh_find_port must be a valid port."
-	while netstat -ltu | grep -q -w :$port       # Check if the port is free
+	while netstat -nltu | grep -q -w :$port       # Check if the port is free
 	do
 		port=$((port+1))	# Else, pass to next port
 	done
--- a/data/helpers.d/network
+++ b/data/helpers.d/network
@@ -17,7 +17,7 @@ ynh_find_port () {
 	ynh_handle_getopts_args "$@"
 
 	test -n "$port" || ynh_die --message="The argument of ynh_find_port must be a valid port."
-	while netcat -z 127.0.0.1 $port       # Check if the port is free
+	while netstat -ltu | grep -q -w :$port       # Check if the port is free
 	do
 		port=$((port+1))	# Else, pass to next port
 	done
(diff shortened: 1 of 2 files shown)


data/helpers.d/network
```diff
@@ -17,7 +17,7 @@ ynh_find_port () {
 	ynh_handle_getopts_args "$@"
 
 	test -n "$port" || ynh_die --message="The argument of ynh_find_port must be a valid port."
-	while netcat -z 127.0.0.1 $port       # Check if the port is free
+	while netstat -ltu | grep -q -w :$port       # Check if the port is free
```
alexAubin:
```suggestion
	while netstat -nltu | grep -q -w :$port       # Check if the port is free
```

Gotta add `-n`, otherwise it tries to replace well-known ports like `443` to `https` or `5222` to `xmpp-client` and so on

kay0u: Yes indeed!

Rafi594:
netstat seems not installed by default

taziden:
We can use ss (part of the iproute2 package) instead of netstat. I think it's installed by default.

Rafi594:
https://github.com/kay0u/yunohost/pull/1/files

kay0u:
Uuh, I was not sure. `netcat-openbsd` is a [requirement](https://github.com/YunoHost/yunohost/blob/stretch-unstable/debian/control) for yunohost (as well as iproute), `netcat-openbsd` doesn't contain netstat? I can apply you PR @Rafi594  anyway, what do you think about it @alexAubin?

Rafi594:
netstat is a part of net-tools https://packages.debian.org/stretch/net-tools

taziden:
> Uuh, I was not sure. `netcat-openbsd` is a [requirement](https://github.com/YunoHost/yunohost/blob/stretch-unstable/debian/control) for yunohost (as well as iproute), `netcat-openbsd` doesn't contain netstat? I can apply you PR @Rafi594 anyway, what do you think about it @alexAubin?

netstat is part of the net-tools package. It's quite different from netcat btw.

alexAubin:
Yolo mergin' bruh

PR #829, opened 2019-10-26
iproute2 instead of iproute

kay0u:
## The problem

https://packages.debian.org/stretch/iproute

> This is a transitional dummy package to get upgrading systems to install the iproute2 package. It can safely be removed once no other package depends on it. 

## Solution

Install iproute2

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
Yolomergin', it's also needed for the buster upgrade anyway :+1:

PR #831, opened 2019-10-30
XMPP http upload

pitchum:
## The problem

Sharing files with XMPP's http upload mechanism is currently impossible.
This PR is an attempt to address issue [#1278](https://github.com/YunoHost/issues/issues/1278 )

Metronome's configuration is ready for http upload but port 5290 is not reachable.

## Solution

http upload requires a dedicated subdomain (I have chosen *jabber.thedomain.net* instead of upload.thedomain.net to avoid possible conflicts with possible other "upload" things).

This subdomain should be reachable via HTTPS so we need:
- [X] add a DNS entry for that subdomain
- [X] transparently include *jabber.thedomain.net* as a SAN in the same certificate for *thedomain.net*
- [X] explicitly define the storage path in metronome's config (the same will be configured in nginx)
- [X] create an nginx config for that subdomain
- [x] silently recreate existing Letsencrypt certificates

## PR Status
The PR ready for review.

## How to test

Create 2 accounts : alice and bob
Install Dino and configure those accounts.
Happily share pictures between alice and bob.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/conf_regen/12-metronome
+++ b/data/hooks/conf_regen/12-metronome
@@ -47,6 +47,10 @@ do_post_regen() {
   # create metronome directories for domains
   for domain in $domain_list; do
     mkdir -p "/var/lib/metronome/${domain//./%2e}/pep"
+    # http_upload directory must be writable by metronome and readable by nginx
+    mkdir -p  "/var/www/xmpp-upload.${domain}/upload"
+    chmod g+s "/var/www/xmpp-upload.${domain}/upload"
+    chown -R metronome:www-data "/var/www/xmpp-upload.${domain}"
   done
 
   # fix some permissions
--- a/data/hooks/conf_regen/12-metronome
+++ b/data/hooks/conf_regen/12-metronome
@@ -42,16 +42,18 @@ do_post_regen() {
   regen_conf_files=$1
 
   # retrieve variables
+  main_domain=$(cat /etc/yunohost/current_host)
   domain_list=$(yunohost domain list --output-as plain --quiet)
 
   # create metronome directories for domains
   for domain in $domain_list; do
     mkdir -p "/var/lib/metronome/${domain//./%2e}/pep"
-    # http_upload directory must be writable by metronome and readable by nginx
-    mkdir -p  "/var/xmpp-upload/${domain}/upload"
-    chmod g+s "/var/xmpp-upload/${domain}/upload"
-    chown -R metronome:www-data "/var/xmpp-upload/${domain}"
   done
+  # http_upload directory must be writable by metronome and readable by nginx
+  mkdir -p  "/var/xmpp-upload/${main_domain}/upload"
+  chmod g+s "/var/xmpp-upload/${main_domain}/upload"
+  chown -R metronome:www-data "/var/xmpp-upload/${main_domain}"
+
 
   # fix some permissions
   chown -R metronome: /var/lib/metronome/
--- a/data/hooks/conf_regen/12-metronome
+++ b/data/hooks/conf_regen/12-metronome
@@ -48,9 +48,9 @@ do_post_regen() {
   for domain in $domain_list; do
     mkdir -p "/var/lib/metronome/${domain//./%2e}/pep"
     # http_upload directory must be writable by metronome and readable by nginx
-    mkdir -p  "/var/www/xmpp-upload.${domain}/upload"
-    chmod g+s "/var/www/xmpp-upload.${domain}/upload"
-    chown -R metronome:www-data "/var/www/xmpp-upload.${domain}"
+    mkdir -p  "/var/xmpp-upload/${domain}/upload"
+    chmod g+s "/var/xmpp-upload/${domain}/upload"
+    chown -R metronome:www-data "/var/xmpp-upload/${domain}"
   done
 
   # fix some permissions
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -68,8 +68,9 @@ server {
     listen 443 ssl http2;
     listen [::]:443 ssl http2;
     server_name xmpp-upload.{{ domain }};
+    root /dev/null;
 
-    location /upload {
+    location /upload/ {
         alias /var/xmpp-upload/{{ domain }}/upload;
         # Pass all requests to metronome, except for GET and HEAD requests.
         limit_except GET HEAD {
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -6,7 +6,7 @@ map $http_upgrade $connection_upgrade {
 server {
     listen 80;
     listen [::]:80;
-    server_name {{ domain }};
+    server_name {{ domain }} xmpp-upload.{{ domain }};
 
     access_by_lua_file /usr/share/ssowat/access.lua;
 
@@ -96,4 +96,85 @@ server {
 
     access_log /var/log/nginx/{{ domain }}-access.log;
     error_log /var/log/nginx/{{ domain }}-error.log;
+}
+
+# vhost dedicated to XMPP http_upload
+server {
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
+    server_name xmpp-upload.{{ domain }};
+
+    location /upload {
+        alias /var/www/xmpp-upload.{{ domain }}/upload;
+        # Pass all requests to metronome, except for GET and HEAD requests.
+        limit_except GET HEAD {
+          proxy_pass http://localhost:5290;
+        }
+
+        include proxy_params;
+        add_header 'Access-Control-Allow-Origin' '*';
+        add_header 'Access-Control-Allow-Methods' 'HEAD, GET, PUT, OPTIONS';
+        add_header 'Access-Control-Allow-Headers' 'Authorization';
+        add_header 'Access-Control-Allow-Credentials' 'true';
+        client_max_body_size 105M; # Choose a value a bit higher than the max upload configured in XMPP server
+    }
+
+    ssl_certificate /etc/yunohost/certs/{{ domain }}/crt.pem;
+    ssl_certificate_key /etc/yunohost/certs/{{ domain }}/key.pem;
+    ssl_session_timeout 5m;
+    ssl_session_cache shared:SSL:50m;
+
+    {% if compatibility == "modern" %}
+    # Ciphers with modern compatibility
+    # https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.6.2&openssl=1.0.1t&hsts=yes&profile=modern
+    # The following configuration use modern ciphers, but remove compatibility with some old clients (android < 5.0, Internet Explorer < 10, ...)
+    ssl_protocols TLSv1.2;
+    ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256';
+    ssl_prefer_server_ciphers on;
+    {% else %}
+    # As suggested by Mozilla : https://wiki.mozilla.org/Security/Server_Side_TLS and https://en.wikipedia.org/wiki/Curve25519
+    ssl_ecdh_curve secp521r1:secp384r1:prime256v1;
+    ssl_prefer_server_ciphers on;
+
+    # Ciphers with intermediate compatibility
+    # https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.6.2&openssl=1.0.1t&hsts=yes&profile=intermediate
+    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+    ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
+
+    # Uncomment the following directive after DH generation
+    # > openssl dhparam -out /etc/ssl/private/dh2048.pem -outform PEM -2 2048
+    #ssl_dhparam /etc/ssl/private/dh2048.pem;
+    {% endif %}
+
+    # Follows the Web Security Directives from the Mozilla Dev Lab and the Mozilla Obervatory + Partners
+    # https://wiki.mozilla.org/Security/Guidelines/Web_Security
+    # https://observatory.mozilla.org/
+    {% if domain_cert_ca != "Self-signed" %}
+    more_set_headers "Strict-Transport-Security : max-age=63072000; includeSubDomains; preload";
+    {% endif %}
+    more_set_headers "Content-Security-Policy : upgrade-insecure-requests";
+    more_set_headers "Content-Security-Policy-Report-Only : default-src https: data: 'unsafe-inline' 'unsafe-eval'";
+    more_set_headers "X-Content-Type-Options : nosniff";
+    more_set_headers "X-XSS-Protection : 1; mode=block";
+    more_set_headers "X-Download-Options : noopen";
+    more_set_headers "X-Permitted-Cross-Domain-Policies : none";
+    more_set_headers "X-Frame-Options : SAMEORIGIN";
+
+    {% if domain_cert_ca == "Let's Encrypt" %}
+    # OCSP settings
+    ssl_stapling on;
+    ssl_stapling_verify on;
+    ssl_trusted_certificate /etc/yunohost/certs/{{ domain }}/crt.pem;
+    resolver 127.0.0.1 127.0.1.1 valid=300s;
+    resolver_timeout 5s;
+    {% endif %}
+
+    # Disable gzip to protect against BREACH
+    # Read https://trac.nginx.org/nginx/ticket/1720 (text/html cannot be disabled!)
+    gzip off;
+
+#    access_by_lua_file /usr/share/ssowat/access.lua;
+
+    access_log /var/log/nginx/xmpp-upload.{{ domain }}-access.log;
+    error_log /var/log/nginx/xmpp-upload.{{ domain }}-error.log;
 }
--- a/data/templates/nginx/server.tpl.conf
+++ b/data/templates/nginx/server.tpl.conf
@@ -38,42 +38,11 @@ server {
     ssl_session_timeout 5m;
     ssl_session_cache shared:SSL:50m;
 
-    {% if compatibility == "modern" %}
-    # Ciphers with modern compatibility
-    # https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.6.2&openssl=1.0.1t&hsts=yes&profile=modern
-    # The following configuration use modern ciphers, but remove compatibility with some old clients (android < 5.0, Internet Explorer < 10, ...)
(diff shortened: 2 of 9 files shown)


data/templates/metronome/domain.tpl.cfg.lua
```diff
@@ -1,15 +1,96 @@
...
+      namefield = "cn",
+    },
   }
+
+
+-- Discovery items
+disco_items = {
+  { "muc.{{ domain }}" },
+  { "pubsub.{{ domain }}" },
+  { "upload.{{ domain }}" },
```
cricriiiiii: does this line need to be "jabber.{{ domain }}" ?

pitchum: absolutely

zamentur:
What do you think to replace jabber by upload-xmpp ? Is jabber a common name for upload http ?

data/templates/metronome/domain.tpl.cfg.lua
```diff
@@ -1,15 +1,96 @@
...
+  http_file_size_limit = 6*1024*1024
+  http_file_quota = 60*1024*1024
+  http_upload_file_size_limit = 100 * 1024 * 1024 -- bytes
+  http_upload_quota = 10 * 1024 * 1024 * 1024 -- bytes
+  http_upload_path = "/var/lib/metronome/jabber.{{ domain }}"
+  http_external_url = "https://jabber.{{ domain }}/"
+
+---Set up a VJUD service
+Component "vjud.{{ domain }}" "vjud"
+  ud_disco_name = "{{ domain }} User Directory"
```
M5oul: You are moving global configuration to the domain specific configuration template. There will be duplicate configurations due to multiple domains configurations. Why did you do that?

pitchum: There won't be any duplicate since each component name contains the {{ domain }} variable.
Or did I miss something?

Just to be clear, this change (moving global configuration to the virtualhost config) is not related to the main goal of that PR (make http upload working). But it's a prerequisite.
I consider this is a long-standing bug in yunohost: only the main yunohost domain can host a full-featured XMPP server, because there are no components defined for the other domains.

colmoneill:
Hi all, just here to show my support, I've been using the XMPP & metronome features of my ynh instance, I would have never been able to configure these myself so wanted to say thanks and looking forwards to the possibility of uploads via ynh xmpp too. Cheers !

alexAubin:
(Flagging as work needed / test needed until you give the green light :stuck_out_tongue_winking_eye: )

pitchum:
Status of this PR: Proof of concept validated
It works for me, in a VM using ynh-dev.

Some things may need improvement:
- find better location and permissions on the directory shared by metronome and nginx
- cleanup hard-coded values ("jabber." + domain)
- rename jabber.thedomain.net to xmpp-upload.thedomain.net
- check if disable bosh module does not break something else

data/hooks/conf_regen/12-metronome
```diff
@@ -36,6 +36,10 @@ do_pre_regen() {
     [[ $domain_list =~ $domain ]] \
       || touch "${metronome_conf_dir}/${file}"
   done
+
+  # create director for http_upload
```
M5oul: directory.

data/templates/metronome/domain.tpl.cfg.lua
M5oul: We would get a `muc` (multi-user-channel) `pubsub`, `xmpp-upload`, and `vjud` per domain? Why for? Does anyone would see a benefit in full-featured XMPP on all of their domains? Would that open new potential flaws?

---

Why not creating two PRs? One to fix this `http-upload` and the other to add components to all domains? Wouldn’t that be easier to review? The title and the description of this PR is about fixing this `http-upload` not adding components to all domains. This is very confusing. Let’s make things KISS when possible. We do not need this complexity here.

pitchum: I too was thinking about making two PR. It will be a bit more work for me but I it'll make things easier for reviewers.

That said, I still consider that each domain should have its own xmpp components. Otherwise users of domainB and domainC would have to use chatrooms of domainA for example. That would make Yunohost unsuitable for hosters like [CHATONS](https://chatons.org/en) who would like to offer full-featured XMPP services for multiple different groups of persons.

maranda:
I'd suggest using mod_http_upload_external instead of mod_http_upload, although configuration extra care in regards of nginx configuration should be adopted since the recent discovering of vulnerabilities in its FastCGI implementation.

maranda:
That should remove all the limitations of narrow file sizes, and also meddling too much with forwarding ports as all requests will be handled by nginx directly.

pitchum:
I've simplified my changes to make this PR easier to review.
I'll propose other PR later.

One important thing I don't know how to handle properly is the renewal of existing letsencrypt certificates. We should find a way to silently drop existing certificates and create a new one including SAN for xmpp-upload.thedomain.net

pitchum:
> I'd suggest using mod_http_upload_external instead of mod_http_upload

@maranda using *mod_http_upload_external* would need more extra work (implementing a new service handling PUT requests). I have currently no idea how to implement this easily and I don't have enough time to work on this.
My solution is simpler: nginx handles GET requests and forwards PUT requests to metronome.

maranda:
> @maranda using _mod_http_upload_external_ would need more extra work (implementing a new service handling PUT requests). I have currently no idea how to implement this easily and I don't have enough time to work on this.
> My solution is simpler: nginx handles GET requests and forwards PUT requests to metronome.

@pitchum you don't have to work on a PUT service because the php script to perform the uploads is *already* included into Metronome it just has to be configured and it's much less work than messing with forwarding all together.

pitchum:
I still don't see why mod_http_external_upload would be useful here and I won't have time to work on this.
So I suggest you create a separate PR when this one is merged.

I'd like to have http_upload feature working in yunohost ASAP.

maranda:
@pitchum all it takes is adding the relative configuration to nginx, replacing paths into the supplied share.php... That's a few minutes work, but otherwise whatever suits your boat.

colmoneill:
Hey all, just checking back in to show support. If there is any specific testing a non technically aware person like myself could do, I can spin up a dev RPI for these purposes ?
Let me know how I can help and thanks for pushing this fwd !

pitchum:
@colmoneill I still have to implement a transparent migration of existing certificates. When it's done I'll describe a testing procedure. I hope to have time to work on this tomorrow.

pitchum:
@colmoneill I did not have time to implement automatic certificate renewal but I've build a DEB package for testing http_upload though.

To test it connect to your yunohost using SSH and type this:
```
wget https://gramaton.org/~pitchum/tmp/ynh/yunohost_3.6.5.3+pitchum1_all.deb
dpkg -i yunohost_3.6.5.3+pitchum1_all.deb
```
Then:
1. make sure you have added `xmpp-upload 3600 IN CNAME @` to your DNS zone
2. enforce renewal of certificate for the corresponding domain
3. use some modern XMPP client (Conversations, Gajim or Dino for example) and try to share an image with contacts or in a chatroom

colmoneill:
Amazing @pitchum thanks so much !

Would you say this is safe to try on my 'production' YNH machine ?
If so, I will try it straight away, if not, I will have to set up a new dev env.

Thanks again, looking forwards to trying this !

All the best !

pitchum:
@colmoneill
> Would you say this is safe to try on my 'production' YNH machine ?

**No, do not try this on your production YNH**, you are the first one to test this (after myself on a testing VM)

data/templates/nginx/server.tpl.conf
```diff
@@ -93,3 +93,84 @@ server {
     access_log /var/log/nginx/{{ domain }}-access.log;
     error_log /var/log/nginx/{{ domain }}-error.log;
 }
+
```
M5oul: This file is a template changed, replicated and used by every domains of the instance.
Do we want the `xmpp-upload` domain to be specified multiple times?
I would suggest to include a kind of mechanism in YunoHost to always create this domain name. That’s it.

---
Thanks for what you are doing. I am looking forward to use this feature.

src/yunohost/certificate.py
```diff
@@ -639,6 +639,9 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
     # Set the domain
     csr.get_subject().CN = domain
 
+    # Include xmpp-upload subdomain as subject alternate names
+    csr.add_extensions([crypto.X509Extension("subjectAltName", False, "DNS:xmpp-upload." + domain)])
+
```
M5oul: I am not sure you why have to take care of this one and the upper one.
Why do you need to touch these files?
LE certificates are working for all domains.
These files are here to handle the the CA (certificate authority) or this kind of stuff.

data/templates/nginx/server.tpl.conf
```diff
@@ -93,3 +93,84 @@ server {
...
+# vhost dedicated to XMPP http_upload
+server {
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
+    server_name xmpp-upload.{{ domain }};
+
+    location /upload {
+        alias /var/www/xmpp-upload.{{ domain }}/upload;
```
M5oul: Is this path generated by Metronome? Is metronome storing the uploaded files in there?

M5oul:
I manage to have it working! \o/
I also had to add a rule in SSOwat configuration in `etc/ssowat/conf.json.persistent`:

```json
{
    "skipped_urls": [
        "xmpp-upload.domain.tld"
    ]
}
```

We should find a clean and permanent rule in SSOwat for that

colmoneill:
Great to hear this progress, sorry I haven't had time to do any tests myself, might wait for your changes to be merged @M5oul ?

Thanks all !

data/templates/nginx/server.tpl.conf
pitchum: As I said earlier, I really think that each yunohost domain should have his own XMPP components. I've created [issue #1516](https://github.com/YunoHost/issues/issues/1516 ) to discuss about this.

src/yunohost/certificate.py
pitchum: This change does not affect the CA itself, it only affects every self-signed certificates generated with this CA.
Although let's encrypt certificates are much preferable, this change is still needed for people who don't use Let's encrypt yet.

pitchum:
> I also had to add a rule in SSOwat configuration in `etc/ssowat/conf.json.persistent`:

That's weird. In my tests I didn't have to change anything in SSOWat to have it work, which is quite surprising now that I think of it.
Did you try with the [DEB package ](https://gramaton.org/~pitchum/tmp/ynh/yunohost_3.6.5.3+pitchum1_all.deb ) I linked earlier? It actually contains a backport of this PR to the stable version of yunohost. I did that because the current unstable branch was not usable for me.

src/yunohost/certificate.py
M5oul: Would people have to install a LE certificat on `xmpp-upload` sub-domain to have the upload working. Is this change supposed to take care of this? Is `xmpp-upload` a default domain created for every domain on the instance? Or should it be the admin to create this domain name to have this feature? What’s the plan of this PR?

M5oul:
> Did you try with the DEB package I linked earlier?

I did apply the changes manually till having something working. I might be wrong with this SSOWat change. We have to check.

src/yunohost/certificate.py
pitchum: It's all automatic (except for existing certificates which all need to be forcibly renewed). Every LE certificate will be valid for both "thedomain.tld" and "xmpp-upload.thedomain.tld". The same for self-signed certificates.
The only condition for LE is that the name xmpp-upload.thedomain.tld has a valid DNS record. But normally this is the case because yunohost's suggested configuration contains a wildcard DNS record. Nevertheless, I've added a CNAME record in the suggestion dns-config to make it more explicit.

colmoneill:
Hi all, 
What would be the next steps to get this PR accepted ?
I'm really eager to have file sharing on my instance of YNH, it will be a great convincer for lots of people around me.
If there is any testing I can do, let me know ?

Cheers !

ei8fdb:
Hi @colmoneill !

Do you need more testing done? I'm happy to try it also.

pitchum:
*Thanks for the reminder. I've been away from that PR for too long now. Let's get back to work.*

Tests have already proven that it works on a fresh yunohost. It also works on an running instance, after some manual actions. Now I need to implement a solution to make the transition as smooth as possible on existing yunohost installations.

## The transitioning problem

Currently the admin has to manually force a certificate renewal in order to include the new subdomain (*xmpp-upload.domain.net*). We should find a way to avoid human intervention. Many people using yunohost probably don't care about XMPP, so we should not bother them.

And the certificate renewal is only possible if the DNS subdomain resolves to the same address as the base domain. In most cases it should be fine because the default DNS configuration suggested by yunohost includes a wildcard entry by default.
But corner cases probably exist and we should take them into account.

So we need to automatically detect if a forced renewal is needed and if it is needed, check first that the DNS name is OK.

My questions are:
1. when should we check the DNS subdomain?
    - on each certificate renewal?
    - only the first time? then how do we do that?
2. what do we do if the DNS name is missing or wrong?
    - in this case an human intervention is required to update the DNS zone. How do we notify the admin?

I'm considering writing a migration script. It looks like the best solution to answer the two questions. But then I have to update the PR so that everything works event if the migration did not occur yet.

I someone has a better idea...

src/yunohost/certificate.py
```diff
@@ -639,6 +639,9 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
     # Set the domain
     csr.get_subject().CN = domain
 
+    # Include xmpp-upload subdomain as subject alternate names
+    csr.add_extensions([crypto.X509Extension("subjectAltName", False, "DNS:xmpp-upload." + domain)])
```
alexAubin: To answer your previous post, and if I understood correctly, naively I would just have something here that checks that there's a good expected DNS record for xmpp-upload. In pseudo-code:

```text
if [DNS record for wildcard exist] OR ([DNS record for xmpp-upload.domain.tld exists] AND [the value of that DNS record match the value for @])
then
    add xmpp-upload to the csr extensions
fi
```

pitchum: Actually, I was reluctant to perform those extra DNS queries each time a certificate renewal attempt is made. But maybe I was too thrifty on this. After all, renewal attempt only occur once a day.
Let's do that, and possibly improve this later.

alexAubin: hmyea if you want there's also a possibility to interface with the diagnosis system from 3.8 for this

alexAubin: But nvm about that. Renewal attempts do indeed only occur once a day, and *actual* renewal (which is what we are concerned about here) is supposedly only happening like every 2.5 months or so

colmoneill:
Hey all, thanks for this and the last push to get things together.
It's really interesting to follow the process for someone like myself, with low skill levels. I know it will be another while before this makes it's way into an YNH update, but it will be great when we have it.

Will this process of adding a new subdomain work for people using DynDNS systems ? My production setup is using a NoIP domain name which is already a subdomain similar to the trio made available by noho.st etc

Just asking as it would be a good prep step for me to get a specific domain if this was going to be a requirement for XMPP file sharing to work.

Thanks again !

pitchum:
I think this PR is ready to be merged.
Here is how to test (**DON'T DO THIS ON A PRODUCTION INSTANCE**):
```
apt update
apt dist-upgrade
wget https://gramaton.org/~pitchum/tmp/ynh/pr831/moulinette_3.7.0.1%2Bnmu1_all.deb
wget https://gramaton.org/~pitchum/tmp/ynh/pr831/yunohost_3.8.0~alpha%2Bnmu1_all.deb
dpkg -i moulinette_3.7.0.1+nmu1_all.deb
dpkg -i yunohost_3.8.0~alpha+nmu1_all.deb
```

Some errors during migrations will appear. Run `yunohost diagnosis show --issues` to see them all.
In my case it was only "normal" errors:
- "inconsistent version numbers" : I realize that my package versions are weird but it's not important
- missing DNS recors: that was intended, I've added missing records later
- "Domain xxxx.yyy is unreachable through HTTP from outside": this one is wrong, don't know why

data/hooks/conf_regen/12-metronome
```diff
@@ -47,6 +47,10 @@ do_post_regen() {
   # create metronome directories for domains
   for domain in $domain_list; do
```
alexAubin: Same as other comment : we probably want to do this only for the main domain ?

data/templates/nginx/server.tpl.conf
```diff
@@ -97,3 +97,84 @@ server {
...
     error_log /var/log/nginx/{{ domain }}-error.log;
 }
+
+# vhost dedicated to XMPP http_upload
+server {
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
+    server_name xmpp-upload.{{ domain }};
+
+    location /upload {
```
alexAubin: Gotta check if we're vulnerable to path-traversal issue here. We might want to add a trailing `/` I think.

src/yunohost/certificate.py
```diff
@@ -639,6 +639,15 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
     # Set the domain
     csr.get_subject().CN = domain
 
+    # Include xmpp-upload subdomain in subject alternate names
+    subdomain="xmpp-upload." + domain
+    try:
+        _check_domain_is_ready_for_ACME(subdomain)
```
alexAubin: Here I think we just want to check the IP with something like 

```
if not _dns_ip_match_public_ip(get_public_ip(), subdomain):
    [...]
```

reason for this is that the other check that `_check_domain_is_ready_for_ACME` (`_domain_is_accessible_through_HTTP`) sometimes have false-negative in the context of hairpinning, which therefore would trigger an exception here. But anyway if we're in this function (`_prepare_certificate_signing_request`), we're already in a context where the domain is ready for ACME and we just want to check the DNS record for xmpp-upload.

src/yunohost/certificate.py
```diff
@@ -639,6 +639,15 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
...
+        logger.info("Subdmain {} is ready for ACME and will be included in the certificate.".format(subdomain))
+        csr.add_extensions([crypto.X509Extension("subjectAltName", False, "DNS:" + subdomain)])
+    except YunohostError:
+        logger.warning(m18n.n('certmanager_warning_subdomain_dns_record', subdomain=subdomain, domain=domain))
```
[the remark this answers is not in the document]
pitchum: done!
But be warned that I will reverse this change when I'll take care of issue [1516](https://github.com/YunoHost/issues/issues/1516) :)

data/templates/nginx/server.tpl.conf
pitchum: done but untested yet.

src/yunohost/certificate.py
pitchum: done!

pitchum:
I followed @alexAubin's suggestions but I haven't tested those changes yet.

kay0u:
I was about to test it, but at almost every command, I get this error:

> Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/moulinette/actionsmap.py", line 555, in process
    fromlist=[func_name],
  File "/usr/lib/moulinette/yunohost/tools.py", line 41, in <module>
    from yunohost.domain import domain_add, domain_list
  File "/usr/lib/moulinette/yunohost/domain.py", line 35, in <module>
    import yunohost.certificate
  File "/usr/lib/moulinette/yunohost/certificate.py", line 46, in <module>
    from yunohost.domain import _get_maindomain
ImportError: cannot import name _get_maindomain

the `_get_maindomain` function exists... any hint?

src/yunohost/certificate.py
```diff
@@ -639,6 +639,16 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
     # Set the domain
     csr.get_subject().CN = domain
 
+    from yunohost.domain import _get_maindomain
+    if domain == _get_maindomain():
+        # Include xmpp-upload subdomain in subject alternate names
+        subdomain="xmpp-upload." + domain
+        if _dns_ip_match_public_ip(get_public_ip(), subdomain):
+            logger.info("Subdmain {} is ready for ACME and will be included in the certificate.".format(subdomain))
```
alexAubin: This string should be internationalized with `m18n` like a few lines below. But idk, personally I'd just vote to remove the message, but that's up to you.

alexAubin:
Anyway, let's try to move forward with this.

I'm proposing to merge this on Monday afternoon and it'll still be possible to do more tests and fixes after merge anyway

pitchum:
I've made some tests on a VM and everything looks fine. I think this PR is ready to be merged right now, but I'm OK to wait monday :)

@kay0u the import error is caused by the line ["import yunohost.certificate" in domain.py ](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/domain.py#L35).
That's why we can't import something from *domain.py* in *certificate.py*... some kind of infinite loop occurs.
My workaround is to [lazy-import](https://github.com/pitchum/yunohost/commit/094cb15b0aaafa8b96d7c3e86f9490352e0d8a1a) the function.

@alexAubin maybe I'm wrong but it looks like the code concerning certificates in *domain.py* is dead code and can be removed.
I would remove [line 35](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/domain.py#L35) and [lines 300-311](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/domain.py#L300-L311) and insert a specific import [line 109](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/domain.py#L109).

MaStr:
I just wanted to implement  manually this and found this PR.
Please do a proper announcement, if you are rolling out the changes for this, because people issuing their own LE certs may need to adjust their stuff. (I'm affected, because need to run a reverse proxy (having multiple webservers behind nat)).

alexAubin:
Merging naow, congratz everybody!

---

>@alexAubin maybe I'm wrong but it looks like the code concerning certificates in domain.py is dead code and can be removed.
I would remove line 35 and lines 300-311 and insert a specific import line 109.

Eh pretty sure that lines 300-311 are not dead, they are what makes the link between the CLI "yunohost domain cert-status" (and install, renew) and the actual functions ... This was implemented before the subcategory system, otherwise probably would have been directly "yunohost domain certificate status" (though anyway there are still naming/import trickery to do if you want to have the subcategory code in a separate file ...)

But you're right about the lazy loading which might improve performances a bit, I'll do a commit after the merge

PR #834, opened 2019-11-04
Get rid of app's status.json

alexAubin:
## The problem

`status.json` never got actually used - except for storing the currently installed revision ... Meanwhile it increases the code complexity.

## Solution

Getting rid of it to simplify the code / remove tech debt.

## PR Status

Tested on my side

## How to test

Run the migration, validate that there's no status.json after it and that all stuff like `app_list` and `app_info` and `tools_update(apps=True)` still works as expected

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -338,13 +334,12 @@ def app_list(filter=None, raw=False, installed=False, with_backup=False):
     return {'apps': list_dict} if not raw else list_dict
 
 
-def app_info(app, show_status=False, raw=False):
```
kay0u: You should remove the `-s` in this file:

https://github.com/YunoHost/yunohost/blob/cb247a8140d9ee8a71c663a2da9b3d106ef969bf/data/actionsmap/yunohost.yml#L591-L594

alexAubin: Ah yes nice catch, thanks ! :+1:

Psycojoker:
I know some applications like weblate started using it because they needed information in it like which commit is actually installed so you might need to do some check before to see if that isn't going to break them :/ (iirc the commit information is now in the app settings)

PR #840, opened 2019-11-10
[enh] Improve permission helpers

kay0u:
## The problem

- No requirements
- No helper to check if a permission exists

## Solution

- Add requirements
- Add `ynh_permission_exists`

## PR Status

...

## How to test

...

## Validation

- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/setting
+++ b/data/helpers.d/setting
@@ -271,20 +271,11 @@ ynh_permission_create() {
         url="None"
     fi
 
-    # Check if permission already exists
-    if ynh_permission_exists --permission $permission; then
-        # If permission exits, update it
-        if [[ -n ${allowed:-} ]]; then
-            allowed="--add ${allowed//';'/" "}"
-        fi
-        ynh_exec_warn_less ynh_permission_update --permission $permission ${allowed:-}
-    else
-        # If not, create it
-        if [[ -n ${allowed:-} ]]; then
-            allowed=",allowed=['${allowed//';'/"','"}']"
-        fi
-        yunohost tools shell -c "from yunohost.permission import permission_create; permission_create('$app.$permission', url=$url ${allowed:-} , sync_perm=False)"
+    if [[ -n ${allowed:-} ]]; then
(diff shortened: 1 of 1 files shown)


data/helpers.d/setting
```diff
@@ -274,7 +275,18 @@ ynh_permission_create() {
         allowed=",allowed=['${allowed//';'/"','"}']"
```
kay0u: Should I move this line in the else part line `287`?

data/helpers.d/setting
```diff
@@ -274,7 +275,18 @@ ynh_permission_create() {
...
     fi
 
-    yunohost tools shell -c "from yunohost.permission import permission_create; permission_create('$app.$permission', url=$url ${allowed:-} , sync_perm=False)"
+    # Check if permission already exists
+    if ynh_permission_exists --permission $permission; then
+        # If permission exits, update it
+        local add
+        if [[ -n ${add:-} ]]; then
+            add="--add ${allowed//';'/" "}"
```
Josue-T: Are you sure of that. The `allowed` variable already contains the good value

data/helpers.d/setting
Josue-T: Yes

data/helpers.d/setting
kay0u: If `allowed` is empty, then the helper is called like that: `ynh_permission_update --permission $permission --add` and getops don't like it.

But the condition above is wrong

Josue-T: Well, I don't know if it's really a good idea to support the case of a permission already exist for this helper. Because if we support this we need to support all situation. Like when we allow a new user but one user is already allowed. Or when we allow nobody but somebody is already allowed...

Josue-T: I think we could just think that it's the work of the packager to check that the permission exist. And on the core side we just return an error when we try to create a permission which already exist.

kay0u: You probably right, adding the function `ynh_permission_exists` is enough

PR #845, opened 2019-11-19
[fix] add all users in visitors

kay0u:
## The problem

https://github.com/YunoHost/issues/issues/1485
https://github.com/YunoHost/issues/issues/1484

## Solution

Add all users in the visitors group (Should I create a new migration?)

## PR Status

- Install nextcloud with new permission system:
```
yunohost app install https://github.com/YunoHost-Apps/nextcloud_ynh/tree/new-permissions-system
```
- Add `visitors` permission.
- Try to login
 
## How to test

...

## Validation

- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -379,6 +379,7 @@
     "migration_description_0012_postgresql_password_to_md5_authentication": "Force PostgreSQL authentication to use MD5 for local connections",
     "migration_description_0013_futureproof_apps_catalog_system": "Migrate to the new future-proof apps catalog system",
     "migration_description_0014_remove_app_status_json": "Remove legacy status.json app files",
+    "migration_description_0015_add_all_users_in_visitors": "Add all users in visitors group",
     "migration_0003_start": "Starting migration to Stretch. The logs will be available in {logfile}.",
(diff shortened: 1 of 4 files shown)


Josue-T:
> (Should I create a new migration?)

Well if we think that we need to fix this also for the actual testing Yes.

kay0u:
Done!

alexAubin:
Zblerg I'm trying to think what's the best option here ...

This fix probably works but I'm afraid that it breaks the semantic and I'd like to avoid having to re-shift all the migrations in unstable to add this new one just because we already released a testing T.T

I'm wondering about another way which would be that : everytime you add `visitors`, it implicitly adds the `all_users` group (if not already there), and forbidding to remove the `all_users` permission if `visitors` is enabled... Which also simplify stuff for app packagers who only have to do `yunohost user permission update --add/remove visitors` to make an app public/private instead of having to remove `all_users`. Still, theoretically that would also require a migration for people already in testing ... or we can make a small script to be ran, that adds `all_users` to all existing perms with `visitors` 

idk :|

kay0u:
Your solution looks better indeed. I'll try to implement it

PR #848, opened 2019-11-20
[fix] metronome domain folder permissions

mdeabreu:
## The problem

When adding new domains to Yunohost a directory for each newly added domain is created in `/var/lib/metronome` unfortunately since the directory is created with `sudo mkdir` that means `root:root` owns the directory. Metronome will now fail to write to the directory.

## Solution

After creating the directory `chown` it.

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/conf_regen/12-metronome
+++ b/data/hooks/conf_regen/12-metronome
@@ -51,6 +51,7 @@ do_post_regen() {
   # create metronome directories for domains
   for domain in $domain_list; do
     sudo mkdir -p "/var/lib/metronome/${domain//./%2e}/pep"
+    sudo chown -R metronome: /var/lib/metronome/${domain//./%2e}/
   done
 
   [[ -z "$regen_conf_files" ]] \
--- a/data/hooks/conf_regen/12-metronome
+++ b/data/hooks/conf_regen/12-metronome
@@ -41,18 +41,17 @@ do_pre_regen() {
 do_post_regen() {
   regen_conf_files=$1
 
-  # fix some permissions
-  sudo chown -R metronome: /var/lib/metronome/
-  sudo chown -R metronome: /etc/metronome/conf.d/
-
   # retrieve variables
   domain_list=$(sudo yunohost domain list --output-as plain --quiet)
 
   # create metronome directories for domains
   for domain in $domain_list; do
     sudo mkdir -p "/var/lib/metronome/${domain//./%2e}/pep"
-    sudo chown -R metronome: /var/lib/metronome/${domain//./%2e}/
   done
+
+  # fix some permissions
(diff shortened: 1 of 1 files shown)


alexAubin:
Oh thanks, that looks like an easy useful fix ! But uh, out of curiosity can you elaborate on why / when does metronome needs to write in that directory ?

M5oul:
All my `var/lib/metronome/{domain}` have the `metronome:` owner.
Isn’t the following enough?
https://github.com/YunoHost/yunohost/blob/7c09982d3d3861b02bd9ff3dbbcc73a9aa7d518c/data/hooks/conf_regen/12-metronome#L45

mdeabreu:
> Oh thanks, that looks like an easy useful fix ! But uh, out of curiosity can you elaborate on why / when does metronome needs to write in that directory ?

Here is an example error message that shows what and where it is trying to write.

mdeabreu:
> All my `var/lib/metronome/{domain}` have the `metronome:` owner.
> Isn’t the following enough?
> https://github.com/YunoHost/yunohost/blob/7c09982d3d3861b02bd9ff3dbbcc73a9aa7d518c/data/hooks/conf_regen/12-metronome#L45
> 
> Did you do something custom which gave you an other owner?

You'll notice that the regen-conf steps for metronome occur in this order when a new domain.tld is added:
1. sudo chown -R metronome: /var/lib/metronome/ 
2. sudo mkdir -p "/var/lib/metronome/domain%2etld/pep"

If nothing else triggers a regen-conf then the newly created domain%2etld/pep directory will have an owner of root (since mkdir was run through sudo). Should something trigger a regen-conf the first chown will fire and fix the permissions issue. But if nothing triggers another regen-conf the most recently created directory for a given domain will be owned by root and unable to be written to by the metronome process.

M5oul:
Ok, then may be moving…
https://github.com/YunoHost/yunohost/blob/7c09982d3d3861b02bd9ff3dbbcc73a9aa7d518c/data/hooks/conf_regen/12-metronome#L44-L46
…afterwards in order to have one command changing the owner.

mdeabreu:
> Ok, then may be moving…
> https://github.com/YunoHost/yunohost/blob/7c09982d3d3861b02bd9ff3dbbcc73a9aa7d518c/data/hooks/conf_regen/12-metronome#L44-L46
> 
> 
> …afterwards in order to have one command changing the owner.

I was initially hesitant but looking a bit closer I think that's a better path forward. Updated changeset to reflect this.

alexAubin:
LGTM, thanks !

PR #851, opened 2019-11-23
[fix] Use ID instead of name for generated_option

Josue-T:
## The problem

Using the "name" for each element in the config panel is bad because it could contains a space.

## Solution

Use the ID.

## PR Status

Tested with synapse on [this branch](https://github.com/YunoHost-Apps/synapse_ynh/pull/158).

## How to test

Install an app with a config panel. Try the config panel and check that each change is applied.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1525,28 +1525,28 @@ def app_config_show_panel(operation_logger, app):
         for section in tab.get("sections", []):
             section_id = section["id"]
             for option in section.get("options", []):
-                option_name = option["name"]
-                generated_name = ("YNH_CONFIG_%s_%s_%s" % (tab_id, section_id, option_name)).upper()
-                option["name"] = generated_name
-                logger.debug(" * '%s'.'%s'.'%s' -> %s", tab.get("name"), section.get("name"), option.get("name"), generated_name)
+                option_id = option["id"]
+                generated_id = ("YNH_CONFIG_%s_%s_%s" % (tab_id, section_id, option_id)).upper()
+                option["id"] = generated_id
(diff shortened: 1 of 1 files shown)


alexAubin:
Sounds legit I guess, but @Psycojoker will know better than I do

Psycojoker:
Looking at this I really don't understand how I could choose to use "name" instead of "id" when writing this code frankly...

Looking at [this commit](https://github.com/YunoHost/yunohost/commit/8027e9268373410cb11989cf21c6e37624742471) it's really weird that I've decided to change from it to name for variable generation... It's probably due to some shitty property of yunohost args format and was probably done for a good reason but 6 month later I have to admit that I really don't remember why :<

I need to dig into that.

Psycojoker:
Ok I got it, the short answer is: you are out of date with the recent modifications.

Long answer:
* config panel and actions have been modified to use the same format than manifest.json/toml
* manifest.json/toml is "weird" and use "name" instead of "id" for actual id, see https://github.com/YunoHost-Apps/unattended_upgrades_ynh/blob/master/manifest.json#L24
* "id", iirc, doesn't have any sens anymore in config_panel.json/toml
* to replace the old name you need to use "ask" instead
* all the details are here on the format you are supposed to respect https://github.com/YunoHost/yunohost/pull/734

Also see: https://gist.github.com/Psycojoker/c9ea335e596c4684b042812b7b0804c3#file-config_panel-toml-L11

So except if we want to switch back to the old format, we don't want this modification so I'm closing this PR (re-open if needed :))

PR #855, opened 2019-11-29
Visitors permission needs All Users

kay0u:
## The problem

YunoHost/issues#1485
YunoHost/issues#1484

## Solution

To add `visitors`, we need to add `all_users` first
Close #845 
To be merged with: https://github.com/YunoHost/test_apps/pull/4

## PR Status

Can be tested, I got an issue during tests:
> E           ConnectionError: HTTPSConnectionPool(host='ynh.dev', port=443): Max retries exceeded with url: /urlpermissionapp (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7ff47b4af210>: Failed to establish a new connection: [Errno 111] Connection refused',))

For tests `test_permission_app_propagation_on_ssowat` and `test_permission_legacy_app_propagation_on_ssowat`, but seems not to be correlated with my PR

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/permission.py
```diff
@@ -140,12 +140,14 @@ def user_permission_update(operation_logger, permission, add=None, remove=None,
...
+    
+    if "all_users" in new_allowed_groups:
+        if len(new_allowed_groups) > 1 and "visitors" not in new_allowed_groups or len(new_allowed_groups) > 2:
             logger.warning(m18n.n("permission_currently_allowed_for_all_users"))
-        if "visitors" in new_allowed_groups:
-            logger.warning(m18n.n("permission_currently_allowed_for_visitors"))
+    
+    # If visitors are allowed, but not all users, it can break some applications, so we prohibit it.
+    if "visitors" in new_allowed_groups and "all_users" not in new_allowed_groups:
+        raise YunohostError('permission_allowed_for_visitors_but_not_for_all_users')
```
alexAubin: Here I don't know what to do in terms of design / UX ... Shouldn't we implictly add `all_users` and display a warning that we did add it automatically ?

That way running `yunohost permission update foo.main --add visitors` will work whatever the current state we're in

kay0u: I don't know if it's really clear for the user (my current implementation isn't). What to do in case of removing `all_users`? Should we remove `visitors` too?

alexAubin: Well let's go for this : 
- when adding `visitors`, it implicitly adds `all_users` if needed
- when removing `all_users`, it raises an exception if `visitors` is allowed, advising the user to remove it first.

PR #857, opened 2019-11-30
[fix] prevent firefox to mix CA and server certificate

autra:
Fixes https://github.com/YunoHost/issues/issues/1479: 

## The problem

yunohost was using the exact same Distinguished Name for
the CA certificate and the main domain server certificate. When creating
alternate domain name, firefox thought the CA for this second domain was
the server certificate for the first domain. As the key mismatches,
Firefox raised a bad key usage error, which is not bypassable.

## Solution

To fix this, we "simply" need to make sure the DN for the CA is
distinct for any other DN. I did so by adding a Organization to it, and
I decided to just remove the last part of the domain and use that as an
organization name. It is certainly possible to do something else, as
long as we end up having a distinct DN. So yolo.test gives a yolo
organization for instance.

More info here https://bugzilla.mozilla.org/show_bug.cgi?id=1590217

## How to test

Run postinstall with yolo.test, create a new domain with yolo2.test and install an app in it. Before the fix, yolo2.test was in error, after the fix, we can make firefox accept the certificate.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -321,7 +321,7 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
         'touch %s/index.txt' % ssl_dir,
         'cp %s/openssl.cnf %s/openssl.ca.cnf' % (ssl_dir, ssl_dir),
         'sed -i s/yunohost.org/%s/g %s/openssl.ca.cnf ' % (domain, ssl_dir),
-        'openssl req -x509 -new -config %s/openssl.ca.cnf -days 3650 -out %s/ca/cacert.pem -keyout %s/ca/cakey.pem -nodes -batch' % (ssl_dir, ssl_dir, ssl_dir),
+        'openssl req -x509 -new -config %s/openssl.ca.cnf -days 3650 -out %s/ca/cacert.pem -keyout %s/ca/cakey.pem -nodes -batch -subj /CN=%s/O=%s' % (ssl_dir, ssl_dir, ssl_dir, domain, os.path.splitext(domain)[0]),
         'cp %s/ca/cacert.pem /etc/ssl/certs/ca-yunohost_crt.pem' % ssl_dir,
         'update-ca-certificates'
     ]


alexAubin:
Uh wokay that sounds a bit cryptic and I'm too lazy to research this, but LGTM and trusting you I guess ? :sweat_smile:

autra:
> trusting you I guess ? sweat_smile

Well if you feel like it :smile: 

But maybe someone else than me can test this just to check I'm not breaking anything? @hugopoi maybe, as he is the reporter of https://github.com/YunoHost/issues/issues/1479?

HugoPoi:
I will test this by the end of the week !

kay0u:
I have this issue, I tried to test your fix but I don't want to rerun a postinstall. How can I fix my actual config? (A migration may be useful here?).
(I have no idea of what you have done here ;-))

HugoPoi:
I'm not sure a migration will re-generate a valid CA certificate

autra:
> How can I fix my actual config?

Replace <your_domain> and <your_domain_without_the_last_part> by the relevant string (if your domain is foo.fr, your_domain_without_the_last_part would be foo). Warning, untested :-)

```bash
openssl req -x509 -new -config /usr/share/yunohost/yunohost-config/ssl/yunoCA/openssl.ca.cnf -days 3650 -out /usr/share/yunohost/yunohost-config/ssl/yunoCA/ca/cacert.pem -keyout /usr/share/yunohost/yunohost-config/ssl/yunoCA/ca/cakey.pem -nodes -batch -subj /CN=<your_domain>/O=<your_domain_without_the_last_part>'
cp /usr/share/yunohost/yunohost-config/ssl/yunoCA/ca/cacert.pem /etc/ssl/certs/ca-yunohost_crt.pem'
update-ca-certificates
```

> (A migration may be useful here?).

Actually, yes! I'll see if I can come up with something.

kay0u:
That's exactly what I've done (in my edit part). It fixed the problem.

HugoPoi:
I have tested this PR, it fix the problem with Firefox with our self-signed certificate.

I think we can merge this.

autra:
Thanks @HugoPoi. Before merging, I just need to find 30 min somewhere in my life to find how to tell yunohost to regenerate certificates on next update :laughing:  I might have time next week.

autra:
So I had a look at it. Indeed I can easily create a migration that will:
- regenerate the correct cacert.pem and cakey.pem
- eventually regenerates self-signed certificates

Although I'm not quite sure it is worth the effort, are self-signed certificates really useful in prod setup ? If it is only for dev, then we don't really need a migration... 

Anyway I'm willing to do it if needed, so it's your call @alexAubin :-)

alexAubin:
Meh, do we really need to have a migration for this ? My understanding is that new installs won't trigger this issue, which should be good enough for the issue that is addressed ?... I don't see the benefit of migrating old certs :s

autra:
> My understanding is that new installs won't trigger this issue, which should be good enough for the issue that is addressed ?

I agree :-)

alexAubin:
Yolo then

PR #861, opened 2019-12-22
Extends permissions features

Josue-T:
## The problem

- There are no way to protect a permission.
- There are no way to set multiple uris for the same permission.
- There are no clean way to manage multiple uris for the same application.
- https://github.com/YunoHost/issues/issues/1486
- https://forum.yunohost.org/t/yunohost-3-7-spooky-testing-call-for-feedback/9385/31?u=josue
- https://github.com/YunoHost/issues/issues/1420
- https://github.com/YunoHost/issues/issues/1534

## Solution

- Add an attribute in LDAP to know if this permission is protected (solve ttps://github.com/YunoHost/issues/issues/1486)
- Add an attribute in LDAP to know if we need to set the auth HEADER or not to the application (solve https://github.com/YunoHost/issues/issues/1420)
- Add the possibility to have multiple uris for the same permission. Note that we will have one main uris for each permission an optionally we will have a list of additional uris.
- Add an attribute in LDAP to have the label for each permissions and add on other attribute to know if we will show a tile in the SSO for this permission (solve https://github.com/YunoHost/issues/issues/1534)
- Rework the ssowatconf.json to pass all this informations to SSOwat.

Here is an example of how will look like the new ssowatconf.json.
```jsonc
{
    "additional_headers": {
        "Auth-User": "uid", 
        "Email": "mail", 
        "Name": "cn", 
        "Remote-User": "uid"
    }, 
    "domains": [
        "domainA.tld", 
        "domainB.tld"
    ], 
    "permissions": {
        "sogo.main": { // Equals to actual protected_uris
            "users": [
                "alice",
                "bob"
            ],
            "label": "SOGo", // Mandatory
            "show_tile": true,
            "auth_header": true,
            "protected": true,
            "uris": [   // Not mandatory, note that the first entry will be the main uris, the others will be the additional uris.
                "domainA.tld/SOGo",
                "domainB.tld/something",
                "domainA.tld/someotherthing"
            ]
        },
        "gitea.main": { // Equals to actual skipped_uris
            "users": [
                "alice",
                "jack"
            ],
            "label": "Gitea",
            "show_tile": true,
            "auth_header": false,
            "protected": false,
            "uris": [
                "domainB.tld/gitea"
            ]
        },
        "gitea.admin": { // Equals to actual unprotected_uris
            "users": [
                "alice"
            ],
            "label": "Gitea admin",
            "show_tile": false,
            "auth_header": true,
            "protected": false,
            "uris": [
                "domainB.tld/gitea/admin"
            ]
        },
        "regex-example.main": { // Equals to actual unprotected_regex
            "users": [
                "alice"
            ],
            "label": "Regex example",
            "show_tile": true,
            "auth_header": true,
            "protected": false,
            "uris": [
                "re:domainB.tld/test[a-z]+/example"
            ]
        },
        "other-example.main": { // Equals to actual protected but with no auth header
            "users": [
                "alice"
            ],
            "label": "Other example B",
            "show_tile": false,
            "auth_header": false,
            "protected": true,
            "uris": [
                "re:domainB.tld/test[1-9]+/other-example"
            ]
        },
        "core": { // Core permission
            "users": [], // Have no sens here, so just put nothing...
            "label": "Core permissions",
            "show_tile": false,
            "auth_header": false,
            "protected": false,
            "uris": [
                "^[^/]*/%.well%-known/acme%-challenge/.*$", 
                "^[^/]*/%.well%-known/autoconfig/mail/config%-v1%.1%.xml.*$"
            ]
        }
    }, 
    "portal_domain": "domain.tld", 
    "portal_path": "/yunohost/sso/", 
    "redirected_regex": {
        "domain.tld/yunohost[\\/]?$": "https://domain.tld/yunohost/sso/"
    }, 
    "redirected_urls": {}, 
}
```

## PR Status

- [x] Implement a migration
- [x] Manage backup/restore
- [x] Implement the unit test
- [x] Test the code (done by the unit test)
- [x] Try to remove the big legacy code in `app_ssowatconf`. https://github.com/YunoHost/yunohost/pull/935
- [x] Update webadmin with new feature (attribute : `label`, `show_tile`). 
https://github.com/YunoHost/yunohost-admin/pull/297
- [x] Update documentation https://github.com/YunoHost/doc/pull/1263
- [x] Test everything :smile:  Done with theses apps (maybe we need to test with some other tricky app, but I don't know which one):
    - wordpress
    - nextcloud
    - gitea
    - etherpad_mypads
    - leed
- [ ] Fix the test (maybe we need to run the migration before to launch the test). Locally all test pass.

## How to test

- Create a protected permission (in a app) or just use the mail permission (which should be protected).
- Try to add or remove the visitors group in this permission.
...

Note that to test this with the unit test you need this branch : https://github.com/YunoHost/test_apps/pull/5

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/helpers.d/setting
```diff
@@ -259,23 +261,30 @@ ynh_webpath_register () {
...
     else
         url="None"
     fi
-
     if [[ -n ${allowed:-} ]]; then
         allowed=",allowed=['${allowed//';'/"','"}']"
     fi
+    if [ -n ${is_protected} ]; then
+        if [ $is_protected == "true" ]; then
+            is_protected=",is_protected=$is_protected=True"
```
kay0u: I didn't tested, but

```suggestion
            is_protected=",is_protected=True"
```
That sounds more like what you want to do here? (Or I'm missing something)

Same for `false`, and in `ynh_permission_update()`

Josue-T: Yes :+1:

src/yunohost/data_migrations/0015_add_permission_protection.py
```diff
@@ -0,0 +1,43 @@
...
+
+        ldap = _get_ldap_interface()
+
+        permission_list = user_permission_list(short=True)["permissions"]
+
+        for permission in permission_list:
+            if permission.split('.')[0] in SYSTEM_PERMS:
+                ldap.update('cn=%s,ou=permission' % permission, {'isProtected': "TRUE"})
+            elif permission.endswith(".main"):
+                ldap.update('cn=%s,ou=permission' % permission, {'isProtected': "FALSE"})
```
kay0u: Some apps should be protected by default (for example phpmyadmin? pi-hole...) What is happening if isProtected is not defined?

Josue-T: Well, here we just define if a user is allowed to add/remove the visitors group to this permission. So why for phpmyadmin need to protected ?

> What is happening if isProtected is not defined?

As I defined this attribute as mandatory slapd won't be happy when you try to update this object if this attribute is not defined.

kay0u: Perhaps phpmyadmin was not a good example, anyway, there i no magic solution here, and the packager should update the permission if needed.

> As I defined this attribute as mandatory slapd won't be happy when you try to update this object if this attribute is not defined.

Ok, no problem then!

Josue-T:
Hello,

Started to work on this. The actual state is:

- [x] Update core permission code
- [x] Update LDAP schema
- [x] Update helper 
- [ ] Rework `app_ssowatconf`, `app_map`, `ynh_webpath_available`
- [ ] Improve management of uris conflict
- [ ] Depreciate old command and helper `yunohost app change-label`
- [ ] Update migrations
- [ ] Update unit test
- [ ] Adapt ssowat for the new [config file](https://github.com/YunoHost/yunohost/pull/861#issue-356092601) and new concept (@kay0u maybe ?)

kay0u:
```jsonc
"regex-example.main": { // Equals to actual unprotected_regex
            "users": [
                "alice"
            ],
            "label": "Regex example",
            "show_tile": true,
            "auth_header": true,
            "protected": false,
            "uris": [
                "re:domainB.tld/test[a-z]+/example"
            ]
        },
```

How can we display a tile in the sso with a regex url? We should either force `show_tile` as false, either have a non regex uris first

Josue-T:
> How can we display a tile in the sso with a regex url? We should either force `show_tile` as false, either have a non regex uris first

Well, yes it's a really bad example. My idea was if somebody try this, we will show a warning and `show_tile` will be at false in all case when the first uris is a regex.

kay0u:
One more thing (sorry), at the beginning the field `protected` wasn't designed to forbid user to add the group visitors? (Or this value have different meanings on whether we are in the SSO or in Yunohost?)

Josue-T:
> One more thing (sorry), at the beginning the field `protected` wasn't designed to forbid user to add the group visitors? (Or this value have different meanings on whether we are in the SSO or in Yunohost?)

Well, good question because actually it could be confusing, if you find some other term I would be happy change this.

In the sso configuration `protected` mean accessible by the public or not. On the other part of code (like LDAP schema, permission.py, helper), `protected` mean that the user can't add/remove the visitor group for this permission.

kay0u:
> In the sso configuration `protected` mean accessible by the public or not. On the other part of code (like LDAP schema, permission.py, helper), `protected` mean that the user can't add/remove the visitor group for this permission.

Ok :+1: 
 
> Well, good question because actually it could be confusing, if you find some other term I would be happy change this.

What about `isPublic` or `public` ? At least it will be easier to understand both sides of the code

Josue-T:
> What about `isPublic` or `public` ? At least it will be easier to understand both sides of the code

Why not, `protected` is like `protected_uris`, but we can take out this term.

Josue-T:
Yes we can go forward with the term `public`

alexAubin:
>Yes we can go forward with the term public.

What about 'allow_public' ?

(Though btw if we're talking about the ssowat conf here, does it really make sense to put this info in it considering it's supposedly only a setting meant to be used in the internal core to refuse to make an app accessible to visitors ? I'm randomly jumping in so maybe i'm completely offtopic :sweat_smile: )

Josue-T:
> (Though btw if we're talking about the ssowat conf here, does it really make sense to put this info in it considering it's supposedly only a setting meant to be used in the internal core to refuse to make an app accessible to visitors ? I'm randomly jumping in so maybe i'm completely offtopic sweat_smile )

Well, don't really understand what you mean, the main question is to solve [this](https://github.com/YunoHost/yunohost/pull/861#issuecomment-606702286)

Josue-T:
Hello,

Did some work on this. The actual state is:

- [x] Update core permission code
- [x] Update LDAP schema
- [x] Update helper 
- [x] Rework `app_ssowatconf`, `app_map`
- [x] Depreciate command `yunohost app change-label`
- [x] Check url/regex validity before to add in LDAP
- [ ] Improve management of uris conflict and `ynh_webpath_available`, `domain_url_available`, `_get_conflicting_apps`
- [ ] Update migrations
- [ ] Try to remove the big legacy code in `app_ssowatconf`.
- [ ] Update unit test
- [x] Adapt ssowat for the new [config file](https://github.com/YunoHost/SSOwat/pull/161)
- [ ] Test everything :smile:

src/yunohost/app.py
```diff
@@ -1265,60 +1245,91 @@ def _sanitized_absolute_url(perm_url):
...
-                if url not in unprotected_urls:
-                    unprotected_urls.append(url)
-
                 # Legacy stuff : we remove now protected-urls that might have been declared as unprotected earlier...
                 protected_urls = [u for u in protected_urls if u != url]
             else:
-                if url not in protected_urls:
-                    protected_urls.append(url)
-
                 # Legacy stuff : we remove now unprotected-urls / skipped-urls that might have been declared as protected earlier...
```
Josue-T, line 1274: Don't really understand why we do this. @alexAubin still it necessary ?

@alexAubin, do know if would be possible to migrate all `unprotected_urls`, `skipped_urls` and `protected_urls` to some permission. By example we could create something like that:
permissions:
- `app.legacy_skipped_uris`: list of all `skipped_uris` and `skipped_regex`
- `app.legacy_unprotected_urls`: list of all `unprotected_urls` and `unprotected_regex`
- `app.legacy_protected_urls`: list of all `protected_urls` and `protected_regex`

alexAubin: If I remember correctly, we do this because : 
- some legacy app can be installed and for example set (un)protected_uri to /
- a main permission is automatically created with the appropriate public/private (visitor allowed or not) state
- later, the user might change is visitors are allowed or not
- but this is not propagated on the (un)protected setting
- so right here we have to remove "some stuff that was (un)protected when looking at the legacy ... but should actually be different once we load the new permission system"

Of course you will say "that's batshit crazy, let's just migrate the legacy settings once and for all?" and you are right. But of course the complex questions start happening when you think about the fact that : 
- the settings might be more complex than just `/`
- legacy apps may re-write the (un)protected setting during upgrades/restore ...
- ??? other shitty case that I may have forgot about ...

alexAubin: index

Josue-T: Thanks for you answer.

> * the settings might be more complex than just `/`

For this I don't know how to manage this. Do you have an example.

> * legacy apps may re-write the (un)protected setting during upgrades/restore ...

Maybe we can do something in the helper for this ?

kay0u: We could replace all (not exactly the syntax but it's for the idea):
`ynh_app_settings_set $app uprotected_uris "url"`
by
`ynh_permission_url "main" url` and `ynh_permission_update "main" "visitors"`?

alexAubin: >For this I don't know how to manage this. Do you have an example.

https://github.com/YunoHost-Apps/etherpad_mypads_ynh/blob/testing/scripts/install#L314

or

https://github.com/YunoHost-Apps/ttrss_ynh/blob/testing/scripts/install#L163

(But granted that the vast majority of apps just play with '/')

>Maybe we can do something in the helper for this ?

Hmyea indeed, we could try to something like `if (current_script != install) and (there's already a permission that corresponds to the url) then [ ignore the change ]` 

(again, not sure about all the detail and edge cases...)

alexAubin: >We could replace all (not exactly the syntax but it's for the idea):
ynh_app_settings_set $app uprotected_uris "url"
by
ynh_permission_url "main" url and ynh_permission_update "main" "visitors"?

Yes and no ... For example, if url = "/admin", you don't want to define this as the url of the main permission ... A new permission has to be created, and it feels weird to automatically do this for the packager.

And if url = "a list" you are screwed...

Josue-T: Tried to migrate all of theses things here https://github.com/YunoHost/yunohost/pull/861/commits/9cdf00b0315daaebc332ad7cd28554a97d712cd1 ...

Hope that I won't break everything...

alexAubin:
@Josue-T and @kay0u be like

3v39tm

Josue-T:
>   - Are these legacy permission gonna get end up displayed in the webadmin and sso ? :/

Yes in the webadmin, but it's protected. No in the sso because `show_tile` is set to false.

>  - Shouldn't we adapt things for the special case `urls = ['/']` that should corresponds to the main permission ? (and actually that should be the most common case we encouter, if you `grep "ynh_app_setting_set" | grep uri` in all apps scripts)

Yes :+1: , did it in the helper, but forgot to manage this in the migration.

>  - Ideally for cases where `urls = ` someting like `['/admin']` we could a bit smarter and try to create a permission called `Admin` right away ...

We, we can but it might be really complicated to manage it after. By example if the packager update his app and remove the settings we won't really know the name of the permission...

Josue-T:
> Note that in the vast majority of cases, packagers seem to be using `ynh_app_setting_set` (and sometimes `yunohost app setting` (yes ...)) so imho all that legacy logic should be taken care in it ?

>> (In fact I cannot find any app using those ... Imho we should just get rid of them, that's just redundant with ynh_app_setting)

Thought the same things...

Josue-T:
> @Josue-T and @kay0u be like
> 
> 3v39tm

Yes, you are true, but I get ride of 100 line of legacy code in the ssowatconf :smile:

Josue-T:
Hello,

Thinking about how to manage the old skipped, unprotected, protected uris I realized, that some feature still miss in the permission.

By example I've an app with this:
- `/`: protected
- `/api`: unprotected

So, in this case I need to create 2 permission (by example `main` and `api`), but I want that a for theses 2 permission, I have the same allowed user. Actually it's not possible because we need to create 2 different permission.

Thinking about that I've some idea:

- add 2 other attribute in the permission for a new feature:
  - linked_to: list of some other permission which the list of all allowed group will be sychronized with the other one except the visitor group.
  - hidden: can give the possibility to hide this permission in the webadmin and in the cli (if `--full` is not enabled).
- Add a concept of master/slave permission. The idea is that allowed group of the slave permission contains the same allowed group to the master permission (except for the visitors group). For the slave permission the allowed group won't be editable (we need to edit the master group). We can also think about to hide the slave permission in some case.

Don't really know if it's totally an edge case and we decide to don't support this. Anyway if we want to take out all skipped, unprotected, protected uris settings it might be easier to extends a new time the permission feature.

What do you think about this ?

Josue-T:
@alexAubin, @kay0u do you think that my idea is completely overkill ?

kay0u:
I'm not sure I understand what you mean here, because if `/api` is `unprotected`, then it will be `public` after the migration... all users will have access to it (in my current implementation in ssowat).

I'm worried about helper after all these possibilities

Josue-T:
> I'm not sure I understand what you mean here, because if /api is unprotected, then it will be public after the migration... all users will have access to it (in my current implementation in ssowat).

Yes but in my last example we could have this: 
`all_users` contains alice, bob, jack.
but only alice and bob is allowed to access to this app.

In protected case it's easy alice and bob will be allowed to access and jack will be refused. In unprotected case as said in this PR https://github.com/YunoHost/SSOwat/pull/158, we need to set the header only for alice and bob. Jack will be allowed to access to the app but won't have the authentication header.

Note that by example with sogo, if the user have the authentication header but is not allowed to authenticate (by the LDAP filter), all access will be blocked (by SOGo). This could be problematic in case when alice share with a special link a calendar to jack.

> I'm worried about helper after all these possibilities

Me too, it's why I'm asking here.

alexAubin:
Ugh I don't really have a clear view to really have a personal conclusion ... Naively to me it sounds like a really specific case that could be taken care of using the post_add/remove_access hooks .. Granted that it has a hackish feel; but to me that's prefferable to adding 2 whole new booleans with super specific meaning... But overall that depends of how recurrent that need is (is there only 1 app that would need such a thing ? Or 3 ? Or 10 ?).

---

Kinda unrelated, I'm taking the risk to share a concern about where we going with all this permission system. I'm happy about the specification draft and work so far (though didn't take a deep look) but I'm worried about the increased tight-coupling with LDAP (though of course i'm biased by the fact that LDAP is obscure to me even though less than before the 3.7).

Of course, having some structure (user, group, permission, with associated memberUid and so on so that it behaves as "unix-like" stuff) is mandatory to have a unified auth system across the different software pieces and apps.

But what we don't really need is to store all the infos like URLs, labels, and other flags about auth_header, protected and so on which are in fact only relevant for SSOwat. We could just have this in a flat yml file for example. You might say "but we were storing all this in settings.yml before, why are you talking about reverting this !?" but the changes with the new permission systems were more about changing the semantic and opening up more complex possibilities that about moving to LDAP (apart from the necessity and hard work to include LDAP attributes for the group <> user <> permission relationships).

If you think about it, we put these infos in LDAP ... just to later generate a json that will be fed to SSOwat (again, apart from the necessary LDAP group <> user <> permission relationships). So why not have these info directly as json (or yaml). In fact, we could even imagine that apps have their `permissions.yml` (or maybe `routes.yml` would be a more accurate name but whatever) sitting in `/etc/yunohost/apps/{appname}/` and SSOwat just auto-including `/etc/yunohost/apps/*/permissions.yml` ... No need to regenerate the ssowat conf anymore ! (Though what we would still need to do is to sync the yml and LDAP - instead of the other way around)

Clearly that implies to multiply the work by an unknown factor :sweat_smile: and in fact I didn't really dig the topic to be sure that wouldn't cause more troubles that what it solves...  But imho that's worth discussing because as good as LDAP can be, it increases the fragileness (if the service stops, everything explodes) and complexity to maintain (c.f. having to ldap.update() and ldap.search() all over the place ... and the regen-conf / migrations "just" to add new bools/strings args, and the way to add them in the scheme is really obscure ...).

Josue-T:
Soo,

I removed for now the migration of the old settings on this branch. I created  an other branch named `remove_legacy_settings` with this. You can see the diff [here](https://github.com/YunoHost/yunohost/compare/permission_protection...remove_legacy_settings?expand=1).

Josue-T:
Hello,

Did some work on this. The actual state is:

- [x] Update core permission code
- [x] Update LDAP schema
- [x] Update helper 
- [x] Rework `app_ssowatconf`, `app_map`
- [x] Depreciate command `yunohost app change-label`
- [x] Check url/regex validity before to add in LDAP
- [x] Improve management of uris conflict and `ynh_webpath_available`, `domain_url_available`, `_get_conflicting_apps`
- [x] Update migrations
- [x] Try to remove the big legacy code in `app_ssowatconf`. (https://github.com/YunoHost/yunohost/pull/861#discussion_r402252529), done by on branch [remove_legacy_settings](https://github.com/YunoHost/yunohost/compare/permission_protection...remove_legacy_settings?expand=1)
- [x] Manage backup/restore from old version (should be ok, need test)
- [x] Adapt ssowat for the new [config file](https://github.com/YunoHost/SSOwat/pull/161)
- [x] Add warning in case of regex in main url and `show_tile=true` and disable `show_tile` to false
- [ ] Fix unit test
- [ ] Add more test
- [ ] Update webadmin with new feature (attribute : `label`, `show_tile`).
- [ ] Test everything :smile: 
- [ ] Update documentation

Josue-T:
Hello,

I've fixed all tests. We still need to understand why we have some issue with ssowat.

src/yunohost/data_migrations/0015_extends_permissions_features_1.py
```diff
@@ -0,0 +1,106 @@
...
+                    ldap.update('cn=%s,ou=permission' % permission, {
+                        'authHeader': ["TRUE"],
+                        'label': [label],
+                        'showTile': ["TRUE"],
+                        'isProtected': ["FALSE"]
+                    })
+                else:
+                    ldap.update('cn=%s,ou=permission' % permission, {
+                        'authHeader': ["TRUE"],
+                        'label': ["%s (%s)" (label, permission.split('.')[1])],
```
kay0u, line 61:
```suggestion
                        'label': ["%s (%s)" % (label, permission.split('.')[1])],
```

Josue-T: Thanks

src/yunohost/data_migrations/0015_extends_permissions_features_1.py
```diff
@@ -0,0 +1,106 @@
...
+                        'showTile': ["FALSE"],
+                        'isProtected': ["TRUE"]
+                    })
+                app_setting(permission.split('.')[0], 'label', delete=True)
```
kay0u, line 65: An application had 2 permissions (main and admin) the label on the second permission is dead because it was removed here during the migration of the first one

kay0u: Should be fixed with your last modifications

src/yunohost/app.py
```diff
@@ -1219,14 +1186,25 @@ def app_ssowatconf():
...
+            "show_tile": False,
+            "auth_header": False,
+            "public": False,
+            "uris": [
+                [domain + '/yunohost/admin' for domain in domains] + \
+                [domain + '/yunohost/api' for domain in domains] + [
+                    "re:^[^/]*/%.well%-known/ynh%-diagnosis/.*$",
+                    "re:^[^/]*/%.well%-known/acme%-challenge/.*$",
+                    "re:^[^/]*/%.well%-known/autoconfig/mail/config%-v1%.1%.xml.*$"
+                ]
```
kay0u, line 1204: There is a double array here
On my `ynh-dev`:
```
"uris": [
                [
                    "domain.tld/yunohost/admin", 
                    "domain.tld/yunohost/api", 
                    "re:^[^/]*/%.well%-known/ynh%-diagnosis/.*$", 
                    "re:^[^/]*/%.well%-known/acme%-challenge/.*$", 
                    "re:^[^/]*/%.well%-known/autoconfig/mail/config%-v1%.1%.xml.*$"
                ]
]
```

src/yunohost/app.py
```diff
@@ -1219,14 +1186,25 @@ def app_ssowatconf():
...
-    protected_regex = []
+    all_permissions = user_permission_list(full=True, ignore_system_perms=True, full_path=True)['permissions']
+
+    permissions = {
+        'core_skipped': {
+            "users": [],
+            "label": "Core permissions - skipped",
+            "show_tile": False,
+            "auth_header": False,
+            "public": False,
```
kay0u, line 1197: Should be true here I guess

src/yunohost/data_migrations/0015_extends_permissions_features_1.py
```diff
@@ -0,0 +1,106 @@
...
+                        'authHeader': ["TRUE"],
+                        'label': [label],
+                        'showTile': ["TRUE"],
+                        'isProtected': ["FALSE"]
+                    })
+                else:
+                    ldap.update('cn=%s,ou=permission' % permission, {
+                        'authHeader': ["TRUE"],
+                        'label': ["%s (%s)" % (label, permission.split('.')[1])],
+                        'showTile': ["FALSE"],
```
kay0u, line 62: I don't see my permission "myapp.admin"

src/yunohost/app.py
```diff
@@ -1265,60 +1245,91 @@ def _sanitized_absolute_url(perm_url):
...
-    for domain in domains:
-        skipped_urls.extend([domain + '/yunohost/admin', domain + '/yunohost/api'])
+        # Create special permission for legacy apps
+        if skipped_urls != []:
+            permissions[app + ".legacy_skipped_urls"] = {
+                "users": [],
+                "label": "Legacy permission - skipped_urls for app :" + app,
+                "show_tile": False,
+                "auth_header": False,
+                "public": False,
```
kay0u, line 1285: Should be True

src/yunohost/app.py
```diff
@@ -1265,60 +1245,91 @@ def _sanitized_absolute_url(perm_url):
...
+                "uris": skipped_urls
+            }
+        if unprotected_urls != []:
+            permissions[app + ".legacy_unprotected_urls"] =  {
+                "users": all_permissions[app + '.main']['corresponding_users'],
+                "label": "Legacy permission - unprotected_urls for app :" + app,
+                "show_tile": False,
+                "auth_header": True,
+                "public": False,
```
kay0u, line 1294: Should be True

src/yunohost/app.py
```diff
@@ -1265,60 +1245,91 @@ def _sanitized_absolute_url(perm_url):
...
+                "uris": unprotected_urls
+            }
+        if protected_urls != []:
+            permissions[app + ".legacy_protected_urls"] = {
+                "users": all_permissions[app + '.main']['corresponding_users'],
+                "label": "Legacy permission - protected_urls for app :" + app,
+                "show_tile": False,
+                "auth_header": True,
+                "public": True,
```
kay0u, line 1303: Should be False

src/yunohost/data_migrations/0015_extends_permissions_features_1.py
Josue-T: Well, not sure to understand what you mean.

By default the tile won't be shown for the other permission than main. But you still can change this with `yunohost user permssion myapp.admin --show_tile True`.

src/yunohost/app.py
Josue-T: Should be fixed now

src/yunohost/app.py
Josue-T: Thanks, should be fixed now

src/yunohost/app.py
Josue-T: Should be fixed now, thanks

src/yunohost/app.py
Josue-T: Should be fixed now.

src/yunohost/app.py
Josue-T: Thanks, maybe it's way I have some issue with ssowat. Should be fixed now

src/yunohost/data_migrations/0015_extends_permissions_features_1.py
Josue-T: Should be also fixed

Josue-T:
Hello,

Did some work on this. The actual state is:

- [x] Try to remove the big legacy code in `app_ssowatconf`. Done here https://github.com/YunoHost/yunohost/pull/935
- [ ] Fix the test (maybe we need to run the migration before to launch the test)
- [ ] Update webadmin with new feature (attribute : `label`, `show_tile`).
- [ ] Update documentation
- [ ] Test everything :smile:

Josue-T:
Hello,

Made some tests with this following apps:

    wordpress
    nextcloud
    gitea
    etherpad_mypads
    leed

I found some regression and some bug which are now fixed. So I think it should be OK to go forward with this.

Josue-T:
Hello, it's should be finish on my side now.

- [x] Try to remove the big legacy code in `app_ssowatconf`. https://github.com/YunoHost/yunohost/pull/935
- [x] Update webadmin with new feature (attribute : `label`, `show_tile`). 
https://github.com/YunoHost/yunohost-admin/pull/297
- [x] Update documentation https://github.com/YunoHost/doc/pull/1263
- [x] Test everything :smile:  Done with theses apps (maybe we need to test with some other tricky app, but I don't know which one):
    - wordpress
    - nextcloud
    - gitea
    - etherpad_mypads
    - leed
- [ ] Fix the test (maybe we need to run the migration before to launch the test). Locally all test pass.

data/helpers.d/setting
```diff
@@ -158,12 +158,21 @@ ynh_webpath_register () {
...
 # | arg: permission - the name for the permission (by default a permission named "main" already exist)
 # | arg: url - (optional) URL for which access will be allowed/forbidden
+# | arg: additional_urls - (optional) List of additional URL for which access will be allowed/forbidden
+# | arg: auth_header - (optional) Define for the URL of this permission, if SSOwat pass the authentication header to the application. Default is true
 # | arg: allowed - (optional) A list of group/user to allow for the permission
+# | arg: label - (optional) Define a name for the permission. This label will be shown on the SSO and in the admin. Default is "APP_LABEL (permission name)".
+# | arg: show_tile - (optional) Define if a tile will be shown in the SSO
+# | arg: protected - (optional) Define if this permission is protected. If it is protected the administrator
+# |                             won't be able to add or remove the visitors group of this permission.
+# |                             By default it's 'true' (for the permission different than 'main').
```
maniackcrudelis, line 175: Considering a PR is ongoing to standardize the helpers https://github.com/YunoHost/yunohost/pull/947
Would be better to clean that one before merging it.

data/helpers.d/setting
```diff
@@ -178,25 +187,62 @@ ynh_webpath_register () {
...
     ynh_handle_getopts_args "$@"
 
     if [[ -n ${url:-} ]]; then
-        url="'$url'"
-    else
-        url="None"
+        url=",url='$url'"
+    fi
+
+    if [[ -n ${additional_urls:-} ]]; then
```
maniackcrudelis, line 207: Same here
And probably other places as well

data/helpers.d/setting
```diff
@@ -178,25 +187,62 @@ ynh_webpath_register () {
...
+    if [[ -n ${protected:-} ]]; then
+        if [ $protected == "true" ]; then
+            protected=",protected=True"
+        else
+            protected=",protected=False"
+        fi
+    fi
     
-    yunohost tools shell -c "from yunohost.permission import permission_create; permission_create('$app.$permission', url=$url ${allowed:-} , sync_perm=False)"
+    yunohost tools shell -c "from yunohost.permission import permission_create; permission_create('$app.$permission' ${url:-} ${additional_urls:-} ${auth_header:-} ${allowed:-} ${label:-} ${show_tile:-} ${protected:-} , sync_perm=False)"
```
maniackcrudelis, line 245: Helpers are **shell** functions.
To maintain those helpers, helpers are to stay **bash** functions.

A yunohost command should be used here.

alexAubin: If I remember correctly, it was designed this way to avoid making the permission creationg/deletion/update/.. API/CLI public because we want to avoid users messing with permissions manually ... But clearly that helper code is becoming quite funky.

Maybe a middle-ground to have a special helper that's able to call yunohost functions directly such that we can focus on the bash logic inside the actual helpers but idk how realistic or better this would be.

maniackcrudelis: Yet users should be able to deal with permission on their own if they want to.
Especially considering a CLI tool will refrain crazy usage.

What's for sure is that python command inside a bash helper can be maintain easily.

alexAubin: >Yet users should be able to deal with permission on their own if they want to.

Yes if that's about updating the permission like adding/removing users, but here's we're talking about things like creating/deleting permission which can make everything explode if you start deleting/creating "main" permissions yourself ... Or any permission in fact ... unless you know exactly what you're doing ...

maniackcrudelis: Didn't go througt that permission enough to say if needed or not.
But I feel like sooner or later, we'll want to have a CLI tool for that.

Anyway, that's not really the problem here.
The problem is that bash helper should not call python directly into their code.

maniackcrudelis: Oh damn I missed that !!!
Especially when that python code is actually full of bash substitutions !

data/helpers.d/setting
```diff
@@ -179,27 +189,72 @@ ynh_webpath_register () {
...
     then
-        url="'$url'"
-    else
-        url="None"
+        url=",url='$url'"
+    fi
+
+    if [[ -n $additional_urls ]]
+    then
+        additional_urls=",additional_urls=['${additional_urls//';'/"','"}']"
```
maniackcrudelis, line 218: The example use `--additional_urls 'domain.tld/otherurl /superadmin'`, only one value.
The variable is set with `additional_urls=${additional_urls:-}` which will lose the second path...
And here `;` are replaced by ','

Are the separate urls supposed to be splitted by `;` ? No handling of multiple urls anywhere.

And I have to admit that the comment, just before that line is hard to read...

If the purpose (I guess...?) was to use the multiple value feature of getopts, as the **comment** explains https://github.com/YunoHost/yunohost/blob/stretch-unstable/data/helpers.d/getopts#L40
First do not put quotes on the options during the call for the helper, otherwise it become only one value.
And then, replace `;` not `';'`
And finally... are you sure the $additional_urls is interpreted between simple quotes ?

data/helpers.d/setting
```diff
@@ -179,27 +189,72 @@ ynh_webpath_register () {
...
+        if [ $auth_header == "true" ]
+        then
+            auth_header=",auth_header=True"
+        else
+            auth_header=",auth_header=False"
+        fi
     fi
 
     if [[ -n $allowed ]]; then
         allowed=",allowed=['${allowed//';'/"','"}']"
```
maniackcrudelis, line 232: Same here
Except that this time, the example does not use quotes

data/helpers.d/setting
```diff
@@ -179,27 +189,72 @@ ynh_webpath_register () {
...
+
+    if [[ -n $auth_header ]]
+    then
+        if [ $auth_header == "true" ]
+        then
+            auth_header=",auth_header=True"
+        else
+            auth_header=",auth_header=False"
+        fi
     fi
```
maniackcrudelis, line 229: Considering the final value is a boolean.
What's the behavior if auth_header is empty ?

Same for other variable used the same way

Josue-T: We use the default value define [here](https://github.com/YunoHost/yunohost/blob/permission_protection/src/yunohost/permission.py#L260-L263)

maniackcrudelis: Then comment and explain.
When using the helper, before using true or false, the packager should know what are the default values.

data/helpers.d/setting
Josue-T: Well,
Yes the first comment (`--additional_urls 'domain.tld/otherurl /superadmin'`) is false, I did probably some change and forgot to change also here...

But if you use correctly this helper it work. I've tested in the test app. cf https://github.com/YunoHost/test_apps/blob/permission_protection/permissions_app_ynh/scripts/install#L29

Yes I admin that we need some comment. Maybe a global comment could do this because it's everywhere the same (tricky) thing.

So do you prefer this ?
```suggestion
        additional_urls=",additional_urls=['${additional_urls//;/"','"}']"
```

data/helpers.d/setting
Josue-T: The default value are documented here https://github.com/YunoHost/yunohost/pull/861/files#diff-bb250d2955b70f7920149b8c2942ec59R169

data/helpers.d/setting
maniackcrudelis: I "do prefer" something that actually works...
So my question, does it work with that code !?

Damn is that really so difficult to explain your code !!!???

Is the expected result is that ?
`,additional_urls=['url1','url2','url3']`
If so, I do prefer a code easy to read (as usual...) like
`additional_urls=",additional_urls=['${additional_urls//;/\',\'}']"`
Except the first and final quote, the adding of quote around the comma is clear.
Which does not spare a comment to explain.

data/helpers.d/setting
maniackcrudelis: What about `show_tile` ?
Add a comment around those lines, the behavior should be clear.

data/helpers.d/setting
```diff
@@ -158,12 +158,22 @@ ynh_webpath_register () {
...
+# | arg: -u, url=               - (optional) URL for which access will be allowed/forbidden
+# | arg: -A, additional_urls=   - (optional) List of additional URL for which access will be allowed/forbidden
+# | arg: -h, auth_header=       - (optional) Define for the URL of this permission, if SSOwat pass the authentication header to the application. Default is true
+# | arg: -a, allowed=           - (optional) A list of group/user to allow for the permission
+# | arg: -l, label=             - (optional) Define a name for the permission. This label will be shown on the SSO and in the admin.
+# |                                     Default is "APP_LABEL (permission name)".
+# | arg: -t, show_tile=         - (optional) Define if a tile will be shown in the SSO
+# | arg: -P, protected=         - (optional) Define if this permission is protected. If it is protected the administrator
+# |                             won't be able to add or remove the visitors group of this permission.
+# |                             By default it's 'true' (for the permission different than 'main').
```
maniackcrudelis, line 176: > By default it's 'true'

Does that mean any permission added by an app will be unmovable by an admin by default ?

data/helpers.d/setting
maniackcrudelis: > Yes I admit that we need some comment. Maybe a global comment could do this because it's everywhere the same (tricky) thing.

Are comments especially expensive in your IDE ?
Comment everywhere anyone could wonder about your code.

data/helpers.d/setting
Josue-T: Yes

maniackcrudelis: It may sound completely silly for sure, but when taking some decision that are actually completely app related, it sounds a little bit legit to talk with the apps group...

And actually, it does not make any sense to forbid users to modify the permission by default for all apps.
For sure we can have some specific situations, and probably that option may be handy.
But to not forbid by default make just no sense !

Josue-T: Well,

Note that it's protect only to add/remove the visitor on this permission. Also the main permission is not protected by default.

I forbid by default because I thought that probably In some case the packager will create a permission to unprotected by example the API. For the permission without uris it don't make sens to not protect it as playing with the visitor group on a permission without uris has no effect.

But yes we should discuss.

maniackcrudelis: > probably In some case

Which sound clearly up to the packager to decide whether or not !

Unless I missed something this is not about a permission without uris since a few lines before there's `-u, url=`
If you do consider that a permission without uris (does not make any sense for me right now...) or an API should be protected, then add a comment explaining that.
To be honest, the comments in the header of that helper, which looks like a very important helper, are really not helpful to actually understand what to do with the different options it has.
- `-u, url=` should give clear examples of what you can do with it.
- `additional_urls=` makes no sense for me. Why there's a second argument ?
- Considering `url=` is optional !? What other usage you could have ? This is about the web access !
- `auth_header=` Yet I know how the SSO work, this is still not completely clear for me. And I know the mechanism... What to expect for other ?
I guess this is actually skipped_uris. Which means it didn't work... But anyway... -_-
- `label=` and `show_tile=` Those 2 are partially linked and that should be explicitly explained with concrete example.
- `protected=` Somehow clear on what it does but not why you would use it. Example again should be given.

But `protected`, I have no doubt we can have use cases, even if I have none in mind. But that's clearly does not sound like the default behavior a packager would want for a permission other than main.
Currently most use case we have for extra permissions are generated links for downloading files or images. I see no reason for an admin to not want to have them private for a restricted circle of users.
And API indeed, which again could totally be turn private by an user not using them and seeing (which is quite right) a potential security breach.
And yet, such users may change their mind and turn it public again afterward.

Josue-T: > But `protected`, I have no doubt we can have use cases, even if I have none in mind. But that's clearly does not sound like the default behavior a packager would want for a permission other than main.
> Currently most use case we have for extra permissions are generated links for downloading files or images. I see no reason for an admin to not want to have them private for a restricted circle of users.
> And API indeed, which again could totally be turn private by an user not using them and seeing (which is quite right) a potential security breach.
> And yet, such users may change their mind and turn it public again afterward.

Why I added this:
https://github.com/YunoHost/issues/issues/1486
https://forum.yunohost.org/t/yunohost-3-7-spooky-testing-call-for-feedback/9385/30

maniackcrudelis: Huh yes...
Well I read
> It would be nice to have a system to forbid adding a permission to "visitors"

I do agree, it's nice to have that possibility, but I'm not saying that's a bad idea to protect a permission, I'm not saying that argument shouldn't exist.
I'm saying that it should not be forbidden by default.
"To be nice to" doesn't mean, let's impose that to everyone !

Josue-T: Ok, so I'll disable it by default.

data/helpers.d/setting
```diff
@@ -158,48 +158,155 @@ ynh_webpath_register () {
...
 #
-# example: ynh_permission_create --permission=admin --url=/admin --allowed="alice bob"
-#
-# usage: ynh_permission_create --permission "permission" [--url=url] [--allowed="group1 group2"]
-# | arg: -p, --permission=  - the name for the permission (by default a permission named "main" already exist)
-# | arg: -u, --url=         - (optional) URL for which access will be allowed/forbidden
-# | arg: -a, --allowed=     - (optional) A list of group/user to allow for the permission
-#
-# If provided, 'url' is assumed to be relative to the app domain/path if they
+# example 1: ynh_permission_create --permission=admin --url=/admin --additional_urls=domain.tld/admin /superadmin --allowed=alice bob \
```
maniackcrudelis, line 161: Thanks for the comment header.
The purpose and usages of this helper is clearer now.

Just a few questions though
Why here --additional_urls=**domain.tld**/admin if the path is relative ?
You explain later that the path is relative, and yet you add again the domain. Is there a reason to add the domain before the path, and isn't there a risk that the domain is concatenated to the real domain ?

data/helpers.d/setting
```diff
@@ -158,48 +158,155 @@ ynh_webpath_register () {
...
+# |                             By default it's 'true' (for the permission different than 'main').
+#
+# If provided, 'url' or 'additional_urls' is assumed to be relative to the app domain/path if they
 # start with '/'.  For example:
 #    /                             -> domain.tld/app
 #    /admin                        -> domain.tld/app/admin
 #    domain.tld/app/api            -> domain.tld/app/api
 #
-# 'url' can be later treated as a regex if it starts with "re:".
+# 'url' or 'additional_urls' can be later treated as a regex if it starts with "re:".
```
maniackcrudelis, line 204: What do you mean as "can be later treated as a regex" ?
Is the feature not yet implemented ?

Josue-T: Normally when the [PR in ssowat](https://github.com/YunoHost/SSOwat/pull/161) is merged, regex should be now available.

data/helpers.d/setting
Josue-T: > Why here --additional_urls=domain.tld/admin if the path is relative ?

No it's full path.

Maybe it's not really clear. But the idea is here to give the possibility to have multiple path and also multiple domain for the same app. An app have a "main" domain and a "main" path which is set in the settings. You can define a permission with a path relative to this main path. In this case it start with `/`. If you need more flexibility you need to use a full path with the domain specified.

maniackcrudelis: I'm not sure to have fully get your answer.

Line 198, the comment says clearly that the path is "assumed to be relative"
And in the example below, you show a domain.tld.

So, it's relative only if it starts by a /, otherwise it's not ?

And regex can be relative as well ?
Line 206, the comment let think that a relative regex will be concatenated to the domain anyway. what would happen with a domain with a `-` in the middle ?
Those are known to failed the regex into the SSO.

Josue-T: > Line 198, the comment says clearly that the path is "assumed to be relative"

Only if it start with `/` cf line 199

> And in the example below, you show a domain.tld.
> 
> So, it's relative only if it starts by a /, otherwise it's not ?

Yes

> And regex can be relative as well ?

Yes (line 206)

> Line 206, the comment let think that a relative regex will be concatenated to the domain anyway. what would happen with a domain with a - in the middle ?

Oh yes, it's not specified. The regex need to be a PCRE regex not lua regex. Note that the '.' in the domain is automatically replaced by `\.` when it's a relative regex.

maniackcrudelis: If a regex is relative, then is concatenated to the domain ? Really important because it currently does not work and grab all other domains matching the regex.

> what would happen with a domain with a - in the middle ?

Not a `.`, a `-`. That currently fail the regex.

The comment you've added says `PCRE (lua)`. Did you mean `PCRE (not lua)` ?

Josue-T: > If a regex is relative, then is concatenated to the domain ? Really important because it currently does not work and grab all other domains matching the regex.

Yes it's concatenated with the domain and the app path.

> > what would happen with a domain with a - in the middle ?

As I know `-` is not considered a special shar with PCRE regex.

> The comment you've added says `PCRE (lua)`. Did you mean `PCRE (not lua)` ?

Oh yes, my bad.

maniackcrudelis: Maybe it was just a problem with the lua "regex". Hope so...
`-` is used to separate a fields of value.

kay0u:
I tried to install wordpress with this branch, and I had this problem at the very beginning:
> 4021 ERROR Could not create permission 'wordpress__2.main': error during LDAP add operation with: rdn='cn=wordpress__2.main,ou=permission', attr_dict={'cn': 'wordpress__2.main', 'objectClass': ['top', 'permissionYnh', 'posixGroup'], 'isProtected': ['FALSE'], 'showTile': ['FALSE'], 'gidNumber': '76762', 'label': ['Wordpress__2'], 'authHeader': ['TRUE']} and exception {'info': 'isProtected: attribute type undefined', 'desc': 'Undefined attribute type'}

Fun fact: the app is considered as "installed" but the install script is never launched because of this error

PS: Could you please resolve the conflicts? (I'm too afraid to break something)

Josue-T:
> I tried to install wordpress with this branch, and I had this problem at the very beginning:
> 
> > 4021 ERROR Could not create permission 'wordpress__2.main': error during LDAP add operation with: rdn='cn=wordpress__2.main,ou=permission', attr_dict={'cn': 'wordpress__2.main', 'objectClass': ['top', 'permissionYnh', 'posixGroup'], 'isProtected': ['FALSE'], 'showTile': ['FALSE'], 'gidNumber': '76762', 'label': ['Wordpress__2'], 'authHeader': ['TRUE']} and exception {'info': 'isProtected: attribute type undefined', 'desc': 'Undefined attribute type'}
> 
> Fun fact: the app is considered as "installed" but the install script is never launched because of this error

Did you run the migration before to try the install ?

> PS: Could you please resolve the conflicts? (I'm too afraid to break something)

No problem, I'll do this.

kay0u:
> Did you run the migration before to try the install ?

Yes, my bad sorry

SirFerdek:
Hi, is this change abandoned?

Josue-T:
> I'm ready to continue this PR and [YunoHost/SSOwat#161](https://github.com/YunoHost/SSOwat/pull/161) since 4.0 is out, can you resolve conflicts :-D?

Done, but I haven't tested...

kay0u:
I tried to create a new lxc, with this and ssowat branch to test permission, when I try to start migration, it fails with this:

> yunohost tools migrations migrate --debug

https://paste.yunohost.org/raw/lojinujute

Any clue?

alexAubin:
(Same here :

```
Error: error during LDAP update operation with: rdn='cn=xmpp.main,ou=permission', attr_dict={'showTile': ['FALSE'], 'isProtected': ['TRUE'], 'label': 'XMPP', 'authHeader': ['FALSE']}, new_rdn=False and exception: {'info': u'label: multiple values provided', 'desc': u'Constraint violation'}
```

_

data/helpers.d/setting
```diff
@@ -158,48 +158,155 @@ ynh_webpath_register () {
...
+        # allowed=['alice', 'bob']
+        allowed=",allowed=['${allowed//;/\',\'}']"
+    fi
+
+    if [[ -n ${label:-} ]]; then
+        label=",label='$label'"
     else
-        url="None"
+        label=",label='$YNH_APP_LABEL ($permission)'"
     fi
```
alexAubin, line 287: Do we foresee ways for people to edit "sub-permissions" labels ?

Otherwise it sounds to me like the current implementation means that the label of a subpermission is pretty much permanent (no mean for the admin to change it) which may be confusing if : 

- main app label is e.g. `Wordpress`
- there's a sub-permission for the admin, e.g. `Wordpress (admin)`
- the admin changes the label of the app to `My Blog`
- .. but the wordpress's admin tile remains named "Wordpress (admin)`
- now imagine if you have 3+ wordpress installed with 3 tiles all named "Wordpress (admin)" and no way to change it ? :s 

So imho, when creating a subpermission, we should only define the label of that permission, but in the end the displayed text should always be reconstructed as `<the app label> (<the subperm label>)` ?

locales/en.json
```diff
@@ -483,7 +494,7 @@
     "permission_not_found": "Permission '{permission:s}' not found",
     "permission_update_failed": "Could not update permission '{permission}': {error}",
     "permission_updated": "Permission '{permission:s}' updated",
-    "permission_update_nothing_to_do": "No permissions to update",
+    "permission_protected": "Permission {permission} is protected. You cannott add or remove the visitors group to/from this permission.",
```
kay0u, line 497:
```suggestion
    "permission_protected": "Permission {permission} is protected. You cannot add or remove the visitors group to/from this permission.",
```

kay0u:
Hi

I found a little bug: Apps don't have a label
image
image

And during an action (app install, backup...) there are no logs (top bar doesn't appear)

data/helpers.d/setting
Josue-T: We can do this when we rebuild the SSOwat configuration.

locales/en.json
```diff
@@ -25,6 +27,7 @@
     "app_install_failed": "Could not install {app}: {error}",
```
ericgaspar, line 27:
```suggestion
    "app_install_failed": "Unable to install {app}: {error}",
```

locales/en.json
```diff
@@ -1,6 +1,8 @@
 {
     "aborting": "Aborting.",
     "action_invalid": "Invalid action '{action:s}'",
+    "additional_urls_already_added": "Additionnal url '{url:s}' already added in the additional URL for permission '{permission:s}'",
+    "additional_urls_already_removed": "Additionnal url '{url:s}' already removed in the additional URL for permission '{permission:s}'",
     "admin_password": "Administration password",
     "admin_password_change_failed": "Cannot change password",
```
ericgaspar, line 7:
```suggestion
    "admin_password_change_failed": "Unable to change password",
```

ericgaspar (changes requested):
Small typos

locales/en.json
```diff
@@ -274,6 +276,7 @@
     "domain_exists": "The domain already exists",
     "domain_hostname_failed": "Could not set new hostname. This might cause an issue later (it might be fine).",
```
ericgaspar, line 277:
```suggestion
    "domain_hostname_failed": "Unable to set new hostname. This might cause an issue later (it might be fine).",
```

locales/en.json
```diff
@@ -408,12 +412,13 @@
...
     "migration_description_0018_xtable_to_nftable": "Migrate old network traffic rules to the new nftable system",
-    "migration_0011_create_group": "Creating a group for each user…",
-    "migration_0011_LDAP_update_failed": "Could not update LDAP. Error: {error:s}",
-    "migration_0011_migrate_permission": "Migrating permissions from apps settings to LDAP...",
-    "migration_0011_update_LDAP_database": "Updating LDAP database...",
-    "migration_0011_update_LDAP_schema": "Updating LDAP schema...",
-    "migration_0011_failed_to_remove_stale_object": "Could not remove stale object {dn}: {error}",
+    "migration_description_0019_extends_permissions_features_1": "Extend/rework the app permission management system",
+    "migration_0019_create_group": "Creating a group for each user…",
+    "migration_0019_LDAP_update_failed": "Could not update LDAP. Error: {error:s}",
```
ericgaspar, line 417:
```suggestion
    "migration_0019_LDAP_update_failed": "Unable to update LDAP. Error: {error:s}",
```

locales/en.json
```diff
@@ -408,12 +412,13 @@
...
+    "migration_0019_migrate_permission": "Migrating permissions from apps settings to LDAP...",
+    "migration_0019_update_LDAP_database": "Updating LDAP database...",
+    "migration_0019_update_LDAP_schema": "Updating LDAP schema...",
+    "migration_0019_failed_to_remove_stale_object": "Could not remove stale object {dn}: {error}",
```
ericgaspar, line 421:
```suggestion
    "migration_0019_failed_to_remove_stale_object": "Unable to remove stale object {dn}: {error}",
```

locales/en.json
```diff
@@ -563,6 +577,8 @@
     "service_stop_failed": "Could not stop the service '{service:s}'\n\nRecent service logs:{logs:s}",
```
ericgaspar, line 577:
```suggestion
    "service_stop_failed": "Unable to stop the service '{service:s}'\n\nRecent service logs:{logs:s}",
```

locales/en.json
```diff
@@ -579,6 +595,7 @@
     "tools_upgrade_special_packages_completed": "YunoHost package upgrade completed.\nPress [Enter] to get the command line back",
     "unbackup_app": "{app:s} will not be saved",
     "unexpected_error": "Something unexpected went wrong: {error}",
+    "unknown_main_domain_path": "Unknown domain or path for app '{app}'. You need to specify a domain and a path to be able to specify a url for permission.",
     "unlimit": "No quota",
     "unrestore_app": "{app:s} will not be restored",
     "update_apt_cache_failed": "Could not to update the cache of APT (Debian's package manager). Here is a dump of the sources.list lines, which might help identify problematic lines: \n{sourceslist}",
```
ericgaspar, line 601:
```suggestion
    "update_apt_cache_failed": "Unable to update the cache of APT (Debian's package manager). Here is a dump of the sources.list lines, which might help identify problematic lines: \n{sourceslist}",
```

locales/en.json
```diff
@@ -25,6 +27,7 @@
     "app_install_failed": "Could not install {app}: {error}",
     "app_install_script_failed": "An error occurred inside the app installation script",
     "app_make_default_location_already_used": "Can't make the app '{app}' the default on the domain, '{domain}' is already in use by the other app '{other_app}'",
```
ericgaspar, line 29:
```suggestion
    "app_make_default_location_already_used": "Unable to make '{app}' the default app on the domain, '{domain}' is already in use by '{other_app}'",
```

ericgaspar (changes requested):
- Small typos
- *Unable to* instead of *Cannot*

alexAubin:
Follow-up of yesterday's discussion about non-trivial legacy permissions. Here's a dump of what exists in the wild (only apps declared as working) : 

https://drop.devloprog.org/r/kep83fHuY_#kPT86CnRL7Zf7eqqyvNZflfT4DHRgO2Q/GmKC71do7k=

2020-09-16-184735_1366x768_scrot

kay0u:
> Follow-up of yesterday's discussion about non-trivial legacy permissions. Here's a dump of what exists in the wild (only apps declared as working) :
> 
> https://drop.devloprog.org/r/kep83fHuY_#kPT86CnRL7Zf7eqqyvNZflfT4DHRgO2Q/GmKC71do7k=
> 
> 2020-09-16-184735_1366x768_scrot

More related to this [PR](https://github.com/YunoHost/yunohost/pull/935). Many thanks to you for this

alexAubin (approved):
Completed my first pass :+1: 

Didn't run much tests yet, but here are some minor~ish comments

src/yunohost/app.py
```diff
@@ -963,12 +928,13 @@ def confirm_install(confirm):
...
-    # If an app doesn't have at least a domain and a path, assume it's not a webapp and remove the default "/" permission
+    # If the app haven't set the url of the main permission and domain and path is set set / as main url
     app_settings = _get_app_settings(app_instance_name)
     domain = app_settings.get('domain', None)
     path = app_settings.get('path', None)
-    if not (domain and path):
-        permission_url(app_instance_name + ".main", url=None, sync_perm=False)
+    if domain and path and user_permission_list(full=True, full_path=False)['permissions'][app_instance_name + '.main']['url'] is None:
+        permission_url(app_instance_name + ".main", url='/', sync_perm=False)
+    user_permission_update(app_instance_name + ".main", show_tile=True, sync_perm=False)
```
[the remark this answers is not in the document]
Josue-T: Oh yes :+1:

src/yunohost/permission.py
```diff
@@ -44,65 +44,94 @@
...
+            perm["label"] = infos.get("label", [None])[0]
+            perm["show_tile"] = infos.get("showTile", [False])[0] == "TRUE"
+            perm["protected"] = infos.get("isProtected", [False])[0] == "TRUE"
+            perm["url"] = infos.get("URL", [None])[0]
+            perm["additional_urls"] = infos.get("additionalUrls", [])
+
+            if full_path:
+                app_base_path = apps_base_path[app] if app in apps_base_path else "" # Meh in some situation where the app is currently installed/removed, this function may be called and we still need to act as if the corresponding permission indeed exists ... dunno if that's really the right way to proceed but okay.
+                perm["url"] = _get_absolute_url(perm["url"], app_base_path)
+                perm["additional_urls"] = [_get_absolute_url(url, app_base_path) for url in perm["additional_urls"]]
```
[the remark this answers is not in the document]
Josue-T: By example here: https://github.com/YunoHost/yunohost/pull/861/files#diff-64307ee11522399bf4f7c6ce1e3c22f4R390
If we pass `True` we convert all relative path to full path and we want to do that

src/yunohost/permission.py
```diff
@@ -145,16 +173,22 @@ def user_permission_update(operation_logger, permission, add=None, remove=None,
         if "visitors" not in new_allowed_groups or len(new_allowed_groups) >= 3:
             logger.warning(m18n.n("permission_currently_allowed_for_all_users"))
 
-    # Don't update LDAP if we update exactly the same values
-    if set(new_allowed_groups) == set(current_allowed_groups):
-        logger.warning(m18n.n("permission_already_up_to_date"))
-        return existing_permission
```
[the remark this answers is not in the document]
Josue-T: Well, I probably removed this as this situation now is managed the moulinette. But we can keep it to just show the warning. But we can't return here as we have some other things to update with the new implementation.

src/yunohost/permission.py
```diff
@@ -174,7 +174,7 @@ def user_permission_update(operation_logger, permission, add=None, remove=None,
         if "visitors" not in new_allowed_groups or len(new_allowed_groups) >= 3:
             logger.warning(m18n.n("permission_currently_allowed_for_all_users"))
 
-    # Note that we can get is argument as string we it come from the CLI
+    # Note that we can get this argument as string if we it come from the CLI
```
ericgaspar, line 177:
```suggestion
    # Note that we can get this argument as string if it comes from the CLI
```

ericgaspar (changes requested):

src/yunohost/app.py
```diff
@@ -1470,13 +1458,12 @@ def _sanitized_absolute_url(perm_url):
...
 def app_change_label(app, new_label):
+    from permission import user_permission_update
     installed = _is_installed(app)
     if not installed:
         raise YunohostError('app_not_installed', app=app, all_apps=_get_all_installed_apps_id())
-
-    app_setting(app, "label", value=new_label)
-
-    app_ssowatconf()
+    logger.warning(m18n.n('app_label_deprecated'))
```
alexAubin, line 1465: Re-reading this, I don't know what to think about this deprecation ... on one hand, I understand the will to have only one consistent way to manage all aspects of permissions ... on the other hand, 'yunohost user permission update' ain't trivial to use and weird when what you want is just to change the label of an app (anyway the main way to do this should be using the webadmin and that should be transparent for the admin, but zbleh)

src/yunohost/data_migrations/0019_extends_permissions_features_1.py
```diff
@@ -0,0 +1,163 @@
...
+                                  auth_header=False, label=legacy_permission_label(app, "skipped"),
+                                  show_tile=False, allowed='visitors', protected=True, sync_perm=False)
+            if unprotected_urls != []:
+                permission_create(app + ".legacy_unprotected_uris", additional_urls=unprotected_urls,
+                                  auth_header=True, label=legacy_permission_label(app, "unprotected"),
+                                  show_tile=False, allowed='visitors', protected=True, sync_perm=False)
+            if protected_urls != []:
+                permission_create(app + ".legacy_protected_uris", additional_urls=protected_urls,
+                                  auth_header=True, label=legacy_permission_label(app, "protected"),
+                                  show_tile=False, allowed=permission_list()['permissions'][app + ".main"]['allowed'],
```
alexAubin, line 115: @Josue-T : Here I added a supposedly missing `[app+'.main']` otherwise that was probably miserably crashing ... can you confirm this was the initial intention ?

src/yunohost/data_migrations/0019_extends_permissions_features_1.py
```diff
@@ -0,0 +1,163 @@
...
+            s = app_setting(app, name)
+            return s.split(',') if s else []
+
+        for app in apps:
+            skipped_urls = [uri for uri in _get_setting(app, 'skipped_uris') if uri != '/']
+            skipped_urls += ['re:' + regex for regex in _get_setting(app, 'skipped_regex')]
+            unprotected_urls = [uri for uri in _get_setting(app, 'unprotected_uris') if uri != '/']
+            unprotected_urls += ['re:' + regex for regex in _get_setting(app, 'unprotected_regex')]
+            protected_urls = [uri for uri in _get_setting(app, 'protected_uris') if uri != '/']
+            protected_urls += ['re:' + regex for regex in _get_setting(app, 'protected_regex')]
```
alexAubin, line 102: For these I removed an undefined `_sanitized_absolute_url` which was I believe coming from some old copypasta from app_map / app_ssowat ... My understanding is that in that new context it's not needed anymore (we're not generating the ssowat conf but filling some permission properties) ... but not 100% sure ?

Josue-T: Well, I think it was not a copy pas. The idea was probably to be sure that all url that we put in the permission a already sanitized. And as some really old instance still have some "dirty" I think that it's good to run this here to correct anything everywhere.

src/yunohost/data_migrations/0019_extends_permissions_features_1.py
Josue-T: Yes probably :+1:

src/yunohost/permission.py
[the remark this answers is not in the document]
alexAubin: When you say it's now managed by the moulinette, you mean it doesn't miserably crash if there's nothing to update anymore ? (Maybe I forgot about some PR that did this :thinking: )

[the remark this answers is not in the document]
Josue-T: Yes,
This PR fix the issue: https://github.com/YunoHost/moulinette/pull/231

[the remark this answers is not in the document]
alexAubin: Alrighty, thanks for the reminder ! :+1:

src/yunohost/app.py
alexAubin: (Flagging this comment as resolved, feeling a bit lazy to re-read the whole thing, I think it was addressed by Josue's PR about the legacy permission PR thingy that got merged recently)

alexAubin:
So finally took some time and courage to try to run a big-ass as close to a "real" context by setting up a server with these apps: 

```
docker.yolo.test/: Portainer
yolo.test/Microsoft-Server-ActiveSync: Z-Push
yolo.test/blog: WordPress
yolo.test/date: OpenSondage
yolo.test/dokuwiki: Dokuwiki
yolo.test/dotclear2: DotClear 2
yolo.test/dotclear2/admin: DotClear 2 (Admin)
yolo.test/hextris: Hextris
yolo.test/lufi: Lufi
yolo.test/nextcloud: Nextcloud
```

Then ran the migration. It went surprisingly well despite the number of apps installed ... the only surprise is that there was a huge wall of warnings about the "This app uses legacy permission settings etc..." probably because the migration calls app_setting() or something like that.

The trick about hard-coded legacy permission naming seems to do the trick kinda well

*But* I found a bigger issue, which can be triggered by installing lufi in private mode. During such an install, lufi does that kind of thing (not the only app though) : 

```
ynh_app_setting_set --app=$app --key=protected_regex --value="$domain_regex$path_url/stats$","$domain_regex$path_url/manifest.webapp$","$domain_regex$path_url/$","$domain_regex$path_url/d/.*$","$domain_regex$path_url/m/.*$"
```

Which later leads to this kind of output in `yunohost app map`

```
> yunohost app map
docker.yolo.test/: Portainer
re:yolo.test/lufi/stats$: Lufi (user interface)
yolo.test/Microsoft-Server-ActiveSync: Z-Push
yolo.test/blog: Wordpress
yolo.test/date: Opensondage
yolo.test/dokuwiki: Dokuwiki
yolo.test/dotclear2: Dotclear 2
yolo.test/dotclear2/admin: Dotclear 2 (admin)
yolo.test/hextris: Hextris
yolo.test/lufi: Lufi
yolo.test/lufi/$: Lufi (user interface)
yolo.test/lufi/d/.*$: Lufi (user interface)
yolo.test/lufi/m/.*$: Lufi (user interface)
yolo.test/lufi/manifest.webapp$: Lufi (user interface)
yolo.test/nextcloud: Nextcloud
```

Or similarly in `yunohost user permission info lufi.legacy_protected_uris` : 

```
additional_urls: 
  - yolo.test/lufi/manifest.webapp$
  - yolo.test/lufi/$
  - yolo.test/lufi/d/.*$
  - re:yolo.test/lufi/stats$
  - yolo.test/lufi/m/.*$
allowed: all_users
auth_header: True
corresponding_users: johndoe
label: Lufi (user interface)
protected: True
show_tile: False
url: None
```

Not sure how to handle this though ... the buggy `re:` can be handled easily I suppose, but not sure what to do in terms of displays in `yunohost app map`, etc... Dunno if that's even a big issue or not

alexAubin:
(Another small issue I see in my test is that no legacy permission seems to have been created out of the skipped_regex from Nextcloud ?)

src/yunohost/data_migrations/0019_extends_permissions_features_1.py
alexAubin: Re-investigating this ... they are later automatically sanitized inside the permission_create (or rather permission_url) so no need to pre-sanitize them I believe

src/yunohost/app.py
```diff
@@ -1253,31 +1200,109 @@ def app_setting(app, key, value=None, delete=False):
...
+
+        # GET
+        if value is None and not delete:
+            # FIXME FIXME FIXME : what about the main url ...?
+            return ','.join(permission['additional_urls']) if permission else None
+
+        # DELETE
+        if delete:
+            # If 'is_public' setting still exists, we interpret this as
+            # coming from a legacy app (because new apps should manage the
```
kay0u, line 1225:
```suggestion
            # coming from a legacy app (because new apps shouldn't manage the
```

alexAubin (approved):
Alrighty folks, pushed a bunch of commits to fix a few issues that were mentioned previously or that I encountered while re-running tests

Putting my stamp of approval, PR is ready to merge imho :+1:

PR #863, opened 2019-12-29
Fix unit test and improve test stability

Josue-T:
## The problem

- Some unit test fail (settings and app catalog)
- It's sometime really complicated to pass all test because of bad DNS registry and dirty app cleanup

## Solution

- Fix the settings implementation
- Fix the app catalog test
- Improve the app cleanup after each tests
- Force the DNS to make all request locally

## PR Status

Now all test pass

Note that this PR also depends on https://github.com/YunoHost/moulinette/pull/231 to fix everything.

## How to test

Just launch the unit test (with `./ynh-dev test yunohost`)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tests/test_apps.py
+++ b/src/yunohost/tests/test_apps.py
@@ -13,6 +13,7 @@ from yunohost.app import app_install, app_remove, app_ssowatconf, _is_installed,
 from yunohost.domain import _get_maindomain, domain_add, domain_remove, domain_list
 from yunohost.utils.error import YunohostError
 from yunohost.tests.test_permission import check_LDAP_db_integrity, check_permission_for_apps
+from yunohost.permission import user_permission_list, permission_delete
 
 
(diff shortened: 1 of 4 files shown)


src/yunohost/tests/test_apps.py
```diff
@@ -60,6 +61,33 @@ def clean():
     os.system("systemctl reset-failed nginx")  # Reset failed quota for service to avoid running into start-limit rate ?
     os.system("systemctl start nginx")
 
+    # Clean permission
```
kay0u: I can't suggest in the right place but I think we should make sure that legacy_app__2 is removed too (multi instance tests):

line 41:
```
    if _is_installed("legacy_app__2"):
        app_remove("legacy_app__2")
```

Issue #33, opened 2014-08-11
Can’t upgrade all apps when a custom app is installed


M5oul:
It’s requerided to upgrade app one by one.

M5oul:
Fixed since a long time.

PR #864, opened 2019-12-30
Implement `upgrade_only_if_version_changes` in manifest

Josue-T:
## The problem

- https://github.com/YunoHost/yunohost/pull/756#issuecomment-518417991
- There are actually not a simple way to detect if we need to do in action in the upgrade script only for some specific old package version. Actually we try to find somewhere a difference and test this difference.

## Solution

Implement in the core this concept

Add helper `ynh_compare_package_version`.

## PR Status

Work done

## How to test

- Install an app.
- Add `"upgrade_only_if_version_changes":true` in the manifest
- Upgrade an app in debug mode and see that you will have environment variable named `YNH_APP_UPGRADE_TYPE` which will give you the type of upgrade. It could be:
  - `UPGRADE_FORCED` when there are nothing to update but the user forced the update
  - `UNKNOWN` for the app which don't use the correct syntax for the app version management.
  - `UPGRADE_FULL` when the upstream and the package need an update.
  - `UPGRADE_APP` when the upstream need to be updated.
  - `UPGRADE_PACKAGE` when the package need to be updated.
- You will see that no upgrade will be done if there are nothing to update but you still can fore this with `-F`.
- Try the helper `ynh_compare_package_version` in the upgrade of an app.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -483,6 +512,9 @@ def app_upgrade(app=[], url=None, file=None):
         env_dict["YNH_APP_ID"] = app_id
         env_dict["YNH_APP_INSTANCE_NAME"] = app_instance_name
         env_dict["YNH_APP_INSTANCE_NUMBER"] = str(app_instance_nb)
+        env_dict["YNH_APP_UPGRADE_TYPE"] = upgrade_type
+        env_dict["YNH_APP_MANIFEST_VERSION"] = app_new_version
+        env_dict["YNH_APP_OLD_VERSION"] = app_actual_version
```
maniackcrudelis, line 517:
```suggestion
        env_dict["YNH_APP_CURRENT_VERSION"] = app_actual_version
```
OLD isn't really clear about what we're talking about. Stumble on it on the helper above.

data/helpers.d/utils
```diff
@@ -487,3 +485,52 @@ ynh_check_app_version_changed () {
...
+
+    local version
+    local comparaison
+    local old_version=$YNH_APP_OLD_VERSION
+    # Manage arguments with getopts
+    ynh_handle_getopts_args "$@"
+
+    if [[ ! $version =~ '~ynh' ]] || [[ ! $old_version =~ '~ynh' ]]; then
+        ynh_print_warn "Invalid agument for version."
+        return 1
```
maniackcrudelis, line 512: Shouldn't it be ynh_die instead of a return here ?

data/helpers.d/utils
```diff
@@ -487,3 +485,62 @@ ynh_check_app_version_changed () {
...
+# This example will check if the installed version is greater than (gt) the version 2.3.2~ynh1
+#
+# usage: ynh_compare_package_version --comparison lt|gt|le|ge
+# | arg: --comparison - Comparison type. Could be : le (lower than), gt (greater than), le (lower or equal), ge (greater or equal)
+# | arg: --version - The version to compare. Need to be a version in the yunohost package version type (like 2.3.1~ynh4)
+#
+# Return 0 if the evaluation is true. 1 if false.
+#
+# Requires YunoHost version 3.8.0 or higher.
+ynh_compare_package_version() {
```
maniackcrudelis, line 502: Considering the way the helper work
`ynh_compare_current_package_version` sound more clear to me.
I had a hard time figure out what was the expecting result of this helper.

maniackcrudelis:
And globally, add fucking comments !!!
We shouldn't have to try to figure out how it works !

data/helpers.d/utils
Josue-T: We can discuss about this. My point of view is that if the packager has not used the official way to define the version (by example in a really old package version) we avoid to fail completely the upgrade.

maniackcrudelis: But this helper should be used only on purpose ? No ?
I mean we won't have such situation where an old package will use it without knowing it ?

Also, to have a return 1 will act like the condition is false, that's all.

data/helpers.d/utils
```diff
@@ -492,6 +492,12 @@ ynh_check_app_version_changed () {
 # example: ynh_compare_package_version --comparison gt --version 2.3.2~ynh1
 # This example will check if the installed version is greater than (gt) the version 2.3.2~ynh1
 #
+# Generally you might probably use it as follow in the upgrade script
+#
+# if ynh_compare_package_version --comparaison gt --version 2.3.2~ynh1; then
+#     # Do something that is needed for the package version older than 2.3.2~ynh1
```
maniackcrudelis, line 499: Huh I may have not fully understand in which way it works...
If the current version is greater than 2.3.2\~ynh1 (Yet I don't see any situation I would do that, rather the opposite), you say it does something for package older than 2.3.2\~ynh1...

My brain is in knots...

My new package version is 2.0
I want to migrate from 1.0 to 2.0
if ynh_compare_package_version --comparison lt --version 2.0~ynh1
then
  I do my migration
fi
Because my current version is 1.0, so my version is lower than 2.0

Is that it ?

data/helpers.d/utils
```diff
@@ -487,3 +485,69 @@ ynh_check_app_version_changed () {
...
+    then
+        if [ $(printf "$version\n$current_version" | sort --version-sort | tail -n 1) == $current_version ]; then
+            return 0
+        else
+            return 1
+        fi
+
+    # Else if the current version is lower than the one given as argument
+    else
+        if [ $(printf "$version\n$current_version" | sort --version-sort | tail -n 1) == $version ]; then
```
alexAubin, line 547: I'm randomly jumping in but wouldn't it be simpler to just use `dpkg --compare-versions` ?

maniackcrudelis: Indeed http://man7.org/linux/man-pages/man1/dpkg.1.html
Probably more secure

data/helpers.d/utils
Josue-T: Oh yes it's the opposite.

maniackcrudelis: Maybe a clear explanation then

data/helpers.d/utils
Josue-T: Oh yes. Sow this but I thought that wasn't useable as a simple tools. Now it should be better

src/yunohost/app.py
```diff
@@ -459,12 +460,40 @@ def app_upgrade(app=[], url=None, file=None):
...
+        upgrade_type = "UNKNOWN"
+        # Get actual_version and new version
+        app_new_version = manifest.get("version", "?")
+        app_actual_version = app_dict.get("version", "?")
+
+        if manifest.get("upgrade_only_if_version_changes", None) is True:
+
+            # do only the upgrade if there are a change
+            if app_actual_version == app_new_version and not force:
+                logger.success(m18n.n('app_already_up_to_date', app=app_instance_name))
```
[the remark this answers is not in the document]
Josue-T: Now it should be better

data/helpers.d/utils
yalh76: A French explanation from https://debian-handbook.info/browse/fr-FR/stable/sect.manipulating-packages-with-dpkg.html

dpkg étant le programme de référence pour manipuler les paquets Debian, il fournit également l'implémentation de référence de la logique de comparaison des numéros de versions. C'est pourquoi il dispose d'une option --compare-versions utilisable par des programmes externes (et notamment les scripts de configuration exécutés par dpkg lui-même). Cette option requiert trois paramètres : un numéro de version, un opérateur de comparaison et un second numéro de version. Les différents opérateurs possibles sont lt (strictement plus petit que — lower than), le (plus petit ou égal à — lower or equal), eq (égal à — equal), ne (différent de — not equal), ge (plus grand ou égal à — greater or equal) et gt (strictement plus grand que — greater than). Si la comparaison est avérée, dpkg renvoie le code de retour 0 (succès) ; sinon il renvoie une valeur non nulle (indiquant un échec).

And the most important is the last line:
Return 0 if it's true
Return not 0 if it's false....

maniackcrudelis: @yalh76 you do realize that I'm not asking an explanation for me.
My concern is that when a packager will use that helper, the expected behavior was clear.

Considering that I can read the code and yet I ask if I correctly understood how the helper works. I expect packagers to ask and not be sure about this helper.

So, if when reading the helper, I ask for explanation, I think this is not clear enough.

yalh76: Yes I know that is not clear, and you don't need it.

I just though that adding some french explanations for dpkg --compare-versions could help to improve the documentation in the helper....

yalh76: dpkg --compare-versions is implemented in wekan_ynh... and it took me several reads to understand it

yalh76: the english translation if it could help.. or not:
https://debian-handbook.info/browse/stable/sect.manipulating-packages-with-dpkg.html

Josue-T: Soo, as you are better than me to write the documentation, what would you want that write.

Personally I don't think that it would be good to write in French as it's also dedicated to the English packager (or maybe I didn't understand what you mean).

yalh76: well... as I'm not in the core group, just proposing adjustments in https://github.com/YunoHost/yunohost/pull/934

I think too that french is bad...

src/yunohost/app.py
```diff
@@ -156,10 +156,18 @@ def app_info(app, full=False):
 
 
 def _app_upgradable(app_infos):
+    from packaging import version
```
alexAubin, line 159: Maybe we should add "python-packaging" in debian/control ?

src/yunohost/app.py
```diff
@@ -156,10 +156,18 @@ def app_info(app, full=False):
...
 
     # Determine upgradability
     # In case there is neither update_time nor install_time, we assume the app can/has to be upgraded
 
+    # Firstly use the version to know if an upgrade is available
+    if app_infos["version"] != "-" and app_infos["from_catalog"]["manifest"].get("version", None):
```
alexAubin, line 165: Are we sure that it's gonna be a "good" version if it's not "-" though ? Couldn't it be something messy that's not "-" ?

Also I was initially thinking about checking if "upgrade_only_if_version_changes" is set before doing this check ... But that's debatable

src/yunohost/app.py
```diff
@@ -459,12 +469,42 @@ def app_upgrade(app=[], url=None, file=None):
...
+                upgrade_type = "UPGRADE_FORCED"
+            elif "~ynh" in app_current_version and "~ynh" in app_new_version:
+                app_current_version_upstream, app_current_version_pkg = app_current_version.split("~ynh")
+                app_new_version_upstream, app_new_version_pkg = app_new_version.split("~ynh")
+                if app_current_version_upstream == app_new_version_upstream:
+                    upgrade_type = "UPGRADE_PACKAGE"
+                elif app_current_version_pkg == app_new_version_pkg:
+                    upgrade_type = "UPGRADE_APP"
+                else:
+                    upgrade_type = "UPGRADE_FULL"
```
alexAubin, line 506: Uh it's not clear what happens in the remaining "else" case here ? Sounds like if the old or new app doesn't have an `~ynh` prefix, the upgrade_type is gonna stay to unknown which is a bit weird ?

src/yunohost/app.py
Josue-T: I think it's not mandatory because the package `python-pkg-resources` also contains this module. 
`yunohost` depends of ` python-jinja2` and ` python-jinja2` depends of `python-pkg-resources` so it should be installed on all yunohost instances.

```
root@ynh-dev1:~# apt-cache rdepends --installed python-pkg-resources
python-pkg-resources
Reverse Depends:
  python-setuptools
  python-jinja2
  python-chardet
root@ynh-dev1:~# apt-cache rdepends --installed python-jinja2
python-jinja2
Reverse Depends:
  yunohost
  yunohost
```

src/yunohost/app.py
Josue-T: > Also I was initially thinking about checking if "upgrade_only_if_version_changes" is set before doing this check ... But that's debatable

Yes it's better to check this before.

Josue-T: Should be down

src/yunohost/app.py
Josue-T: Yes, after reading a new time this code it's confusing. Now it should be better.

src/yunohost/app.py
```diff
@@ -162,7 +162,8 @@ def _app_upgradable(app_infos):
     # In case there is neither update_time nor install_time, we assume the app can/has to be upgraded
 
     # Firstly use the version to know if an upgrade is available
-    if app_infos["version"] != "-" and app_infos["from_catalog"]["manifest"].get("version", None):
+    if app_infos["manifest"].get('integration', {}).get("upgrade_only_if_version_changes", None) is True and \
+       '~ynh' in app_infos["version"] and app_infos["from_catalog"]["manifest"].get("version", None):
         if version.parse(app_infos["version"]) < version.parse(app_infos["from_catalog"]["manifest"].get("version", "-")):
```
alexAubin: I'm not sure to understand why we would need to check if "~ynh" is in the current installed version ... (Well after reading the next commit I might understand)

Modulo the handling of presence/absence of ~ynh, we could drastically improve readability using intermediate variables that can be also used in the other lines below the new code  : 

```suggestion
    app_is_in_catalog = bool(app_infos.get("from_catalog"))
    upgrade_only_if_version_changes = app_infos["manifest"].get('integration', {}).get("upgrade_only_if_version_changes", None) is True
    installed_version = version.parse(app_infos["version"])
    version_in_catalog = version.parse(app_infos.get("from_catalog", {}).get("manifest", {}).get("version", 0))
    
    if app_is_in_catalog \
      and upgrade_only_if_version_changes \
      and installed_version < version_in_catalog:
```

src/yunohost/app.py
```diff
@@ -483,28 +483,30 @@ def app_upgrade(app=[], url=None, file=None, force=False):
...
+                    app_current_version_upstream, app_current_version_pkg = app_current_version.split("~ynh")
+                    app_new_version_upstream, app_new_version_pkg = app_new_version.split("~ynh")
+                    if app_current_version_upstream == app_new_version_upstream:
+                        upgrade_type = "UPGRADE_PACKAGE"
+                    elif app_current_version_pkg == app_new_version_pkg:
+                        upgrade_type = "UPGRADE_APP"
+                    else:
+                        upgrade_type = "UPGRADE_FULL"
+            else:
+                logger.warning("/!\\ Packagers ! You have enabled the setting 'upgrade_only_if_version_changes' but you haven't used the official way to define the package version")
```
alexAubin, line 509: Uh wokay but it's not clear to me where that hard constrains comes from ? Granted that it's better to follow the convention, but `version.parse` is able to parse correctly a version with and without ~ynh...

After testing it, I do noticed that it will return some weird stuff for example here : 

```
>>> version.parse("1.2~ynh3") < version.parse("1.4")
True
>>> version.parse("1.2") < version.parse("1.4~ynh1")
False
```

(the second one ain't right...). Might be related to the fact that it parses it as "Legacy version" ?

But anyway, couldn't we hack it and concatenate `~ynh0` if there's no `~ynh` ? (And maybe display a warning if so)

Josue-T: > (the second one ain't right...). Might be related to the fact that it parses it as "Legacy version" ?
> 
> But anyway, couldn't we hack it and concatenate `~ynh0` if there's no `~ynh` ? (And maybe display a warning if so)

My point of view is that, the packager need to enable explicitly this feature so for me in all case it's won't be enabled for old legacy package. So for me I don't think that it's really a problem to put this restriction. I think trying to manage some legacy package is not really a good idea as it might make issue than it solve.

alexAubin: Meh okay :/ I can anticipate that some people will just yoloenable the flag "upgrade_only_if_version_changes" without paying attention to their version format because it's not intuitive that this format is required to support it...

https://paste.yunohost.org/raw/uzuvuqojol

But I'll at least add a check in package_linter about it ...

Josue-T: > Meh okay :/ I can anticipate that some people will just yoloenable the flag "upgrade_only_if_version_changes" without paying attention to their version format because it's not intuitive that this format is required to support it...

It's why I added a warning.

The other problem is that (except for the `-ynh` instead of `~ynh`), don't think that it's really a good idea to enable this if there are not a package version in the version. If we do this we will go in a state when the app is not upgraded when the upstream version is not changed even if there are some change in the package version.

src/yunohost/app.py
```diff
@@ -156,10 +156,23 @@ def app_info(app, full=False):
...
 def _app_upgradable(app_infos):
+    from packaging import version
 
     # Determine upgradability
     # In case there is neither update_time nor install_time, we assume the app can/has to be upgraded
 
+    # Firstly use the version to know if an upgrade is available
+    app_is_in_catalog = bool(app_infos.get("from_catalog"))
+    upgrade_only_if_version_changes = app_infos["manifest"].get('integration', {}).get("upgrade_only_if_version_changes", None) is True
+    installed_version = version.parse(app_infos["version"])
```
alexAubin, line 167: In fact I was a bit too optimist about the fact that apps do provide a version key ... Many of them do not ... 

```suggestion
    installed_version = version.parse(app_infos.get("version", "0~ynh0"))
```

src/yunohost/app.py
```diff
@@ -156,10 +156,23 @@ def app_info(app, full=False):
...
+    version_in_catalog = version.parse(app_infos.get("from_catalog", {}).get("manifest", {}).get("version", "0~ynh0"))
+    
+    if app_is_in_catalog and '~ynh' in app_infos["version"]:
```
alexAubin, line 170:
```suggestion
    if app_is_in_catalog and '~ynh' in str(installed_version) and '~ynh' in str(version_in_catalog):
```

src/yunohost/app.py
```diff
@@ -156,10 +156,23 @@ def app_info(app, full=False):
...
+        if upgrade_only_if_version_changes and installed_version < version_in_catalog:
+            return "yes"
+        else:
+            return "no"
+
     if not app_infos.get("from_catalog", None):
```
alexAubin, line 176:
```suggestion
    if not app_is_in_catalog:
```

src/yunohost/app.py
```diff
@@ -459,12 +474,44 @@ def app_upgrade(app=[], url=None, file=None):
...
+                    app_current_version_upstream, app_current_version_pkg = app_current_version.split("~ynh")
+                    app_new_version_upstream, app_new_version_pkg = app_new_version.split("~ynh")
+                    if app_current_version_upstream == app_new_version_upstream:
+                        upgrade_type = "UPGRADE_PACKAGE"
+                    elif app_current_version_pkg == app_new_version_pkg:
+                        upgrade_type = "UPGRADE_APP"
+                    else:
+                        upgrade_type = "UPGRADE_FULL"
+            else:
+                logger.warning("/!\\ Packagers ! You have enabled the setting 'upgrade_only_if_version_changes' but you haven't used the official way to define the package version")
```
alexAubin, line 513:
```suggestion
        # Manage upgrade type and avoid any upgrade if there is nothing to do
        upgrade_type = "UNKNOWN"
        upgrade_only_if_version_changes = manifest.get('integration', {}).get("upgrade_only_if_version_changes", None) is True
        # Get current_version and new version
        app_new_version = version.parse(manifest.get("version", "?"))
        app_current_version = version.parse(app_dict.get("version", "?"))

        if "~ynh" not in str(app_current_version) or "~ynh" not in str(app_new_version):
            logger.warning("/!\\ Packagers ! You have enabled the setting 'upgrade_only_if_version_changes' but you haven't used the official way to define the package version")
            upgrade_only_if_version_changes = False

        if upgrade_only_if_version_changes:
            if app_current_version >= app_new_version and not force:
                # In fact it sounds like we should never get in there
                # because we check app_dict["upgradable"] before this ...?

                # No new version available
                logger.success(m18n.n('app_already_up_to_date', app=app_instance_name))
                # Save update time
                now = int(time.time())
                app_setting(app_instance_name, 'update_time', now)
                app_setting(app_instance_name, 'current_revision', manifest.get('remote', {}).get('revision', "?"))
                continue
            elif app_current_version > app_new_version:
                upgrade_type = "DOWNGRADE_FORCED"
            elif app_current_version == app_new_version:
                upgrade_type = "UPGRADE_FORCED"
            else:
                app_current_version_upstream, app_current_version_pkg = str(app_current_version).split("~ynh")
                app_new_version_upstream, app_new_version_pkg = str(app_new_version).split("~ynh")
                if app_current_version_upstream == app_new_version_upstream:
                    upgrade_type = "UPGRADE_PACKAGE"
                elif app_current_version_pkg == app_new_version_pkg:
                    upgrade_type = "UPGRADE_APP"
                else:
                    upgrade_type = "UPGRADE_FULL"

```

Sorry, another big suggestion, but should really be my last one :sweat_smile: Essentially just simplify indentation level and improve readability though that was already pretty good

src/yunohost/app.py
alexAubin: Alrighty let's roll with this ! I added a check in packager_linter (and in the same process realizing that so many apps don't even have "version" defined ... :/ )

src/yunohost/app.py
alexAubin: (or maybe I missed some reason to not check the `version_in_catalog` here as well ?)

Josue-T: I thought that it's enought to check on on one side because if the official version is used on one side it will probably be the same on the other side. But yes it's more robust to check on booth side.

src/yunohost/app.py
Josue-T:
```suggestion
# Manage upgrade type and avoid any upgrade if there is nothing to do
        upgrade_type = "UNKNOWN"
        upgrade_only_if_version_changes = manifest.get('integration', {}).get("upgrade_only_if_version_changes", None) is True
        # Get current_version and new version
        app_new_version = version.parse(manifest.get("version", "?"))
        app_current_version = version.parse(app_dict.get("version", "?"))
        if "~ynh" not in str(app_current_version) or "~ynh" not in str(app_new_version):
            logger.warning("/!\\ Packagers ! You have enabled the setting 'upgrade_only_if_version_changes' but you haven't used the official way to define the package version")
            upgrade_only_if_version_changes = False
        if upgrade_only_if_version_changes:
            if app_current_version >= app_new_version and not force:
                # In case of upgrade from file or custom repository
                # No new version available
                logger.success(m18n.n('app_already_up_to_date', app=app_instance_name))
                # Save update time
                now = int(time.time())
                app_setting(app_instance_name, 'update_time', now)
                app_setting(app_instance_name, 'current_revision', manifest.get('remote', {}).get('revision', "?"))
                continue
            elif app_current_version > app_new_version:
                upgrade_type = "DOWNGRADE_FORCED"
            elif app_current_version == app_new_version:
                upgrade_type = "UPGRADE_FORCED"
            else:
                app_current_version_upstream, app_current_version_pkg = str(app_current_version).split("~ynh")
                app_new_version_upstream, app_new_version_pkg = str(app_new_version).split("~ynh")
                if app_current_version_upstream == app_new_version_upstream:
                    upgrade_type = "UPGRADE_PACKAGE"
                elif app_current_version_pkg == app_new_version_pkg:
                    upgrade_type = "UPGRADE_APP"
                else:
                    upgrade_type = "UPGRADE_FULL"

```

I just specified that we can go in the case of downgrade if we upgrade from a custom repository or from file.

SilverViper:
Fixes #33 ?

src/yunohost/app.py
```diff
@@ -156,11 +156,24 @@ def app_info(app, full=False):
 
 
 def _app_upgradable(app_infos):
+    from packaging import version
```
alexAubin, line 159: Well in fact ... not sure what's happening on my LXC but `python-packaging` is not installed despite `python-pkg-resources` being installed ... Anyway it's good to explicitly state dependencies in `debian/control` even though it implicitly depends on other dependency X just in case we drop X for some reason...

alexAubin: Well I see your commit https://github.com/YunoHost/yunohost/pull/864/commits/5c6b4118b7d41f2fb2b25034eb92fb825d4d255c but let me reiterate ... 
- I'm trying to `import packaging` on my RPi2 which runs yunohost, and it does **not** works even though `python-pkg-resources` is installed...
- Also, `apt show python-jinja2` will show that `python-pkg-resources` is **not** a Depends of `python-jinja2` but only a Recommends, so you have no absolute guarantee that it's gonna be installed
- `python-pkg-resources` does **not** depends either on `python-packaging` as a deb package but it apparently ships packaging in `/usr/lib/python2.7/dist-packages/pkg_resources/_vendor/` but that's not loaded with the default `$PYTHONPATH`

TL;DR : let's just add `python-packaging` to the damn debian/control because anyway, it's free, and it's more robust and transparent to have an explicit dependency relationship than an implicit one ...

Josue-T: Sorry I didn't really understood what you want ? So now it should be better.

data/actionsmap/yunohost.yml
```diff
@@ -660,6 +660,10 @@ app:
                 -f:
                     full: --file
                     help: Folder or tarball for upgrade
+                -F:
+                    full: --force
+                    help: Force the update, even though the app is up to date
+                    action: store_true
```
zamentur, line 666: I suggest a break change:

-f is used in install operation as in other part of the cli. I suggest to define -F for --file and -f for --force.

This change concerns packagers, and i prefer a break change for packager, than an inconsistency for final users

alexAubin: N.B. : it's used here : https://github.com/YunoHost/package_check/blob/master/sub_scripts/testing_process.sh#L965 so gotta be careful

I'm okay with the change but maybe we can do this in a separate PR to move forward ...

src/yunohost/app.py
```diff
@@ -489,12 +501,41 @@ def app_upgrade(app=[], url=None, file=None):
...
+            if app_current_version >= app_new_version and not force:
+                # In case of upgrade from file or custom repository
+                # No new version available
+                logger.success(m18n.n('app_already_up_to_date', app=app_instance_name))
+                # Save update time
+                now = int(time.time())
+                app_setting(app_instance_name, 'update_time', now)
+                app_setting(app_instance_name, 'current_revision', manifest.get('remote', {}).get('revision', "?"))
+                continue
+            elif app_current_version > app_new_version:
```
kay0u, line 525: Here `app_current_version` and `app_new_version` should be of the format `X~ynhY`? How is this kind of comparison possible?

src/yunohost/app.py
```diff
@@ -489,12 +501,41 @@ def app_upgrade(app=[], url=None, file=None):
...
+                upgrade_type = "DOWNGRADE_FORCED"
+            elif app_current_version == app_new_version:
```
kay0u, line 527: Same here

src/yunohost/app.py
```diff
@@ -489,12 +501,41 @@ def app_upgrade(app=[], url=None, file=None):
...
             logger.success(m18n.n('app_already_up_to_date', app=app_instance_name))
             continue
 
+        # Manage upgrade type and avoid any upgrade if there is nothing to do
+        upgrade_type = "UNKNOWN"
+        # Get current_version and new version
+        app_new_version = version.parse(manifest.get("version", "?"))
+        app_current_version = version.parse(app_dict.get("version", "?"))
+        if "~ynh" in str(app_current_version) and "~ynh" in str(app_new_version):
+            if app_current_version >= app_new_version and not force:
```
kay0u, line 516: And here

src/yunohost/app.py
kay0u: Because `app_current_version` and `app_new_version` are not string, but `Version`, sowwy for the noise

kay0u (approved):

PR #865, opened 2019-12-30
fix journalctl --since

yalh76:
## The problem

Difficult to explain, as it's more sensations and difficult to reproduce...

Sometime you try to install a package and even if the service is started and the `line_match` can be found in journalctl, `ynh_systemd_action` doesn't detect it. 

Same during package_check with services fail to starts but start during manual installation

## Solution

Diving in ynh_systemd_action and in [journalctl arguments](https://www.commandlinux.com/man-page/man1/journalctl.1.html). It seems that `--since=-0` is documented nowhere

## PR Status

Ready

## How to test

Make some install and check what happens :/

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/systemd
+++ b/data/helpers.d/systemd
@@ -105,7 +105,7 @@ ynh_systemd_action() {
         # Following the starting of the app in its log
         if [ "$log_path" == "systemd" ] ; then
             # Read the systemd journal
-            journalctl --unit=$service_name --follow --since=-0 --quiet > "$templog" &
+            journalctl --unit=$service_name --follow --since="now" --quiet > "$templog" &
             # Get the PID of the journalctl command
             local pid_tail=$!
         else


Josue-T:
Hello,

I also detected sometime that this helper is sometime a little bit buggy. Mainly with [gitea](https://framagit.org/YunoHost-Apps/gitea_ynh/) so without systemd.

Did you see that your patch fix your issue ?

maniackcrudelis:
`--since=-0` is since minus 0 seconds.
Just tried the command, it works the exact same way with `-0` and `now`

Are you sure that's the issue you had ?

yalh76:
I'm like 95% of the issue I had but no proof, no logs, no screenshot, nothing....

I can just tell the conditions...

I was working on https://github.com/YunoHost-Apps/distbin_ynh/pull/52 => applying last example_ynh on distbin_ynh

As I worked like a barbarian, change_url script wasn't working.... And as I didn't took time to read all logs, I decided to test manually change_url on distbin_ynh

I applied a snapshot and started a virtual machine, launch three ssh windows with a command for each one:

1. `$ yunohost app install distbin --debug`
2. `$ journalctl -u distbin -fe`
3. `$ journalctl -u nginx-fe`

like that:
image

At the end of the install, ynh_systemd_action was waiting distbin service start... but in the `$ journalctl -u distbin -fe` windows I seen that the disbin service was really started. So I launch a fourth ssh to execute `$ service distbin restart` and the package installation ended well.

I have seen that a few times on different packages but wasn't able to reproduce systematically...

I will try to make additional tests. and let you know.

maniackcrudelis:
The thing is, using `--since=-0` as well as `--since="now"` will just start to the follow log with no lines before.

The behavior is the same, as said before, between now and 0s before now.
So I really think the problem is elsewhere.

You should probably try to follow `$templog` as well in a separate terminal to see if the helper really miss something.

Also, instead of a sleep 60, you can use the timeout option of ynh_systemd_action.

yalh76:
I wasn't able to reproduce systematically. So closing for now

PR #867, opened 2020-01-08
adaptation for various actions

yalh76:
## The problem

Logs speak about started even if the action was stop

## Solution

using `${action}`

## PR Status

Done

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


maniackcrudelis:
Clearly better.
Don't you think though that it would be even better with a correct verb, "stopped", "started", and so on ?

yalh76:
> Don't you think though that it would be even better with a correct verb, "stopped", "started", and so on ?

I agree with you but it will add some complexity, I tried to make it simple :)

PR #868, opened 2020-01-15
fix legacy permission management

kay0u:
## The problem

The variable `unprotected_url` or `skipped_url` used by all our apps will be deprecated after the 3.7. We don't want kill all our apps after the 3.7 upgrade.

## Solution

`add` or `remove` `visitors` to mirror what those variables did before 3.7

## PR Status

...

## How to test

Install an app with a `config` script that can change the `is_public` variable ([leed](https://github.com/YunoHost-Apps/leed_ynh) for example), and play with the config panel/actions to change the value of `is_public`:
```bash
yunohost app action run leed public_private
```

Then check permissions with:
```bash
yunohost user permission list
```

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/setting
+++ b/data/helpers.d/setting
@@ -158,6 +158,11 @@ ynh_add_protected_uris() {
 #
 ynh_app_setting()
 {
+    if [[ "$1" == "delete" ]] && [[ "$3" =~ ^(unprotected|skipped)_ ]] 
+    then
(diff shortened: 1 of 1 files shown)


kay0u:
Done

PR #872, opened 2020-01-25
[FIX] Diagnosis bad response from the server

kay0u:
## The problem

If the diagnosis.yunohost.org server crashes (omg, this NEVER happens ;-)), we're getting a strange error message like: "Impossible to parse json blablabla"

## Solution

Manage whether the answer is correct or not

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/diagnosis/14-ports.py
+++ b/data/hooks/diagnosis/14-ports.py
@@ -26,16 +26,20 @@ class PortsDiagnoser(Diagnoser):
                 ports[port] = service
 
         try:
-            r = requests.post('https://diagnosis.yunohost.org/check-ports', json={'ports': ports.keys()}, timeout=30).json()
-            if "status" not in r.keys():
-                raise Exception("Bad syntax for response ? Raw json: %s" % str(r))
-            elif r["status"] == "error":
-                if "content" in r.keys():
-                    raise Exception(r["content"])
-                else:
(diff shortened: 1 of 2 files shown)


data/hooks/diagnosis/14-ports.py
```diff
@@ -26,16 +26,20 @@ def run(self):
...
-                raise Exception("Bad syntax for response ? Raw json: %s" % str(r))
+                elif r["status"] == "error":
+                    if "content" in r.keys():
+                        raise Exception(r["content"])
+                    else:
+                        raise Exception("Bad syntax for response ? Raw json: %s" % str(r))
+                elif r["status"] != "ok" or "ports" not in r.keys() or not isinstance(r["ports"], dict):
+                    raise Exception("Bad syntax for response ? Raw json: %s" % str(r))
+            else:
+                raise Exception("Bad response from the server https://diagnosis.yunohost.org : %s" % str(r.status_code))
```
Psycojoker: You probably wants to put the full url with the path here, to ease diagnosis (and maybe display r.content in the error).

Same for after.

alexAubin: Should be resolved by https://github.com/YunoHost/yunohost/pull/872/commits/4399c9b740b5c4901de76f5b2245df4b349eae89

PR #874, opened 2020-01-26
[FIX] config panel ask and actions description

kay0u:
## The problem

The description in actions could be not displayed if we define `description` but not `description.en`
Same for config panel: `ask` and `help`

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1779,7 +1779,7 @@ def _get_app_actions(app_id):
 
             arguments = []
             
-            if not isinstance(action["description"], list) or "en" not in action["description"]:
+            if "description" in action and (not isinstance(action["description"], list) or "en" not in action["description"]):
                 action["description"] = {"en": action["description"]}
 
             for argument_name, argument in value.get("arguments", {}).items():
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1778,6 +1778,10 @@ def _get_app_actions(app_id):
             action["id"] = key
 
             arguments = []
+            
+            if not isinstance(action["description"], list) or "en" not in action["description"]:
+                action["description"] = {"en": action["description"]}
+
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -1660,6 +1660,10 @@ def _get_app_actions(app_id):
             action["id"] = key
 
             arguments = []
+            
+            if not isinstance(action["description"], list) or "en" not in action["description"]:
```
kay0u, line 1664:
```suggestion
            if action["description"] and (not isinstance(action["description"], list) or "en" not in action["description"]):
```

alexAubin: LGTM ?

Psycojoker:
Right now I think we are always in need of using the "en" key in every yunohost-args-format, either it's config panel or it's manifest or whatever :/

That's a more global change that a fix here fyi

Psycojoker:
Note that I think this might also need to adjustment on the `list_builder.py` part regarding including translation in case someone also starts to do this for manifests in applications '^'

Psycojoker:
And I've read too fast, it's for actions >.>

kay0u:
What to do with this fix? 3.8, not 3.8?

alexAubin:
Zblerg I'm really not familiar with the whole action system so I don't understand what Bram comment is about xD But otherwise as long as it's relevant, this sounds like a small easy fix so LGTM to merge in 3.8 ...

zamentur:
I tried to test this change but i got:

bug

With the original code, i have not this error...

Is this PR still relevant and could you explain exactly the change ?

kay0u:
> Is this PR still relevant 

Probably not :-)

alexAubin:
Alrighty closing then !

PR #876, opened 2020-01-29
[FIX] if permission main url is none

kay0u:
## The problem

If the `app.main` permission url is `None`, the app is not displayed in the sso 

## Solution

Use the default url if url is `None`

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -236,13 +236,13 @@ def app_map(app=None, raw=False, user=None):
 
             return perm_domain, perm_path
 
-        this_app_perms = {p: i for p, i in permissions.items() if p.startswith(app_id + ".") and i["url"]}
+        this_app_perms = {p: i for p, i in permissions.items() if p.startswith(app_id + ".")}
         for perm_name, perm_info in this_app_perms.items():
             # If we're building the map for a specific user, check the user
             # actually is allowed for this specific perm
             if user and user not in perm_info["corresponding_users"] and "visitors" not in perm_info["allowed"]:
                 continue
-            if perm_info["url"].startswith("re:"):
+            if perm_info["url"] and perm_info["url"].startswith("re:"):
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -236,13 +236,13 @@ def _sanitized_absolute_url(perm_url):
 
             return perm_domain, perm_path
 
-        this_app_perms = {p: i for p, i in permissions.items() if p.startswith(app_id + ".") and i["url"]}
+        this_app_perms = {p: i for p, i in permissions.items() if p.startswith(app_id + ".")}
```
alexAubin: Are you sure it's a bug and not a feature ? :/

iirc this was done this way to cover non-webapps like for example the mumbleserver app, for which it doesn't make sense to display a tile because there's no web page to display 

(and also : iirc the `app_map` is what is used to detect url conflicts, so with this new code, for example installing mumbleserver on a domain mean you can't install any other app on that domain)

kay0u: Related to: https://github.com/YunoHost/yunohost/pull/877 and more precisely, app like https://github.com/YunoHost-Apps/lutim_ynh/pull/44.

The situation is as follows:

We have an app that protects url via a regex:
https://github.com/YunoHost-Apps/lutim_ynh/blob/b117e3d5547ef2f5265e59d95f8d88d23b0c998a/scripts/install#L205
(same for [opensondage](https://github.com/YunoHost-Apps/opensondage_ynh/))
To make the application private, we only want to protect the urls in the regex, not the whole application.

We can indeed create a new permission, without url called "public images" or whatever, and do the same job with it :-)

kay0u:
I close this PR because it's not a bug :-)

PR #877, opened 2020-01-29
more informations in hooks permission

kay0u:
## The problem

When an app protects his paths with regex, we can't use the current permission system.

## Solution

One possible solution is to deal with hooks.

Con: We can only manage add and and remove of `visitors`.

Needed by: https://github.com/YunoHost-Apps/lutim_ynh/pull/44

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


kay0u:
Can we merge it?

PR #879, opened 2020-02-08
Migrate to php7.3 into ynh_add_fpm_config

maniackcrudelis:
## The problem

- Buster uses php7.3 instead of php7.0
- Apps have to migrate to this version unless they use a specific version of php.

## Solution

- Set a default version to php7.3, and migrate automatically.
- Work would have to be done as well into apps scripts as well. https://github.com/YunoHost/example_ynh/pull/110
- Add the env variable $YNH_DEFAULT_PHP_VERSION to define the default php version to use. Which would prevent any further modifications into apps.
- Define the env variable $YNH_PHP_VERSION for packaging willing to use a custom version of php.

## PR Status

Not tested yet.
But should work as expected.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


kay0u:
We should merge https://github.com/YunoHost/yunohost/pull/880, propagate it to `buster-unstable` branch, then re apply this PR to have the same base IMO

src/yunohost/app.py
```diff
@@ -2771,16 +2771,16 @@ def _assert_system_is_sane_for_app(manifest, when):
 
     # Some apps use php-fpm or php5-fpm which is now php7.0-fpm
```
nathanael-h: ` # Some apps use php-fpm or php5-fpm which is now php7.0-fpm`

Souldn't this be `"[...] which is now php7.3-fpm"` 😉

data/helpers.d/php
```diff
@@ -14,16 +18,18 @@ ynh_add_fpm_config () {
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
 
-	# Configure PHP-FPM 7.3 by default
-	phpversion="${phpversion:-7.3}"
+	# Set the default PHP-FPM version by default
+	phpversion="${phpversion:-$YNH_PHP_VERSION}"
```
Josue-T: Why this variable is not set as local ?

src/yunohost/app.py
```diff
@@ -59,6 +59,7 @@
 APPS_CATALOG_CRON_PATH = "/etc/cron.daily/yunohost-fetch-apps-catalog"
 APPS_CATALOG_API_VERSION = 2
 APPS_CATALOG_DEFAULT_URL = "https://app.yunohost.org/default"
+APPS_DEFAULT_PHP_VERSION = "7.3"
```
Josue-T: Is it really a good idea to set the default php version to `7.3` when we are on debian stretch. I thought that the default php version would be the official php version provided by debian so on stretch it's `7.0`.

I don't really mind this decision but the question is more to launch the debate about this because it still a important decision particularly sine we any many issues with sury.

data/helpers.d/php
maniackcrudelis: Just a few lines above ;)

src/yunohost/app.py
maniackcrudelis: This PR will be merged into buster-unstable branch

Josue-T: Ok in this case

PR #880, opened 2020-02-09
Pave the way to migration to php7.3 and following

maniackcrudelis:
## The problem

Php version will change for buster, and so on for next debian release.

## Solution

Following https://github.com/YunoHost/yunohost/pull/879, prepare the migration by not using a fixed version of php into apps, but the default version set by $YNH_DEFAULT_PHP_VERSION 

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/nginx
+++ b/data/helpers.d/nginx
@@ -12,6 +12,7 @@
 #   __PORT__      by  $port
 #   __NAME__      by  $app
 #   __FINALPATH__ by  $final_path
+#   __PHPVERSION__ by  $YNH_PHP_VERSION ($YNH_PHP_VERSION is either the default php version or the version defined for the app)
 #
 #  And dynamic variables (from the last example) :
 #   __PATH_2__    by $path_2
@@ -44,6 +45,7 @@ ynh_add_nginx_config () {
 	if test -n "${final_path:-}"; then
 		ynh_replace_string --match_string="__FINALPATH__" --replace_string="$final_path" --target_file="$finalnginxconf"
 	fi
+	ynh_replace_string --match_string="__PHPVERSION__" --replace_string="$YNH_PHP_VERSION" --target_file="$finalnginxconf"
 
 	# Replace all other variable given as arguments
 	for var_to_replace in $others_var
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# Declare the actual php version to use.
+# A packager willing to use another version of php can override the variable into its _common.sh.
+YNH_PHP_VERSION=${YNH_PHP_VERSION:-$YNH_DEFAULT_PHP_VERSION}
+
 # Create a dedicated php-fpm config
 #
 # usage: ynh_add_fpm_config [--phpversion=7.X]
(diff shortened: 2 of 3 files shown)


maniackcrudelis:
> I think we should add this default php version in the backup script.

Don't know -_-

kay0u:
There is an issue in case of backup if this version is not defined. Moreover, I think the PHP help should always call the helper `ynh_install_php $YNH_PHP_VERSION` to ensure it well installed.

If the default version change, the old default version may be removed, and the backup using the old version crash because his version is not installed.

maniackcrudelis:
You mean in case of restore ?
If the restore happens on a system where the specific version (or old default) is not installed?
It may indeed doesn't like to not have the service running for its version of php...

As I can see, you did think of this issue into your implementation for nextcloud, https://github.com/YunoHost-Apps/nextcloud_ynh/pull/229/files#diff-68321f788a23b70a33d17888e31bf3efR98

I feel like the best thing to do, into the app scripts, would be to run the helper php instead of restoring the previous config.

kay0u:
> I feel like the best thing to do, into the app scripts, would be to run the helper php instead of restoring the previous config.

The rule is:
When we use the default php version, we don't have to do anything, when we use a specific php version (even if that version is the current default version) the app has to install it via the helper. This way, we should not have any problems during install, upgrade or restore.
We don't need to backup the php fpm file anymore, and during the restore we run the php fpm helper.

Sounds good to me!

kay0u:
But we still need to add the env var default php version inside the backup script. Do you see any reason to not do it?

maniackcrudelis:
> But we still need to add the env var default php version inside the backup script. Do you see any reason to not do it?

What do you mean ?
What I think though about this backup thing is that we should still make a backup, just in case an user would have tweak its config. So it wouldn't be just deleted without backup.

kay0u:
> What do you mean ?

I think we should add `YNH_DEFAULT_PHP_VERSION` inside backup.py to be able to use php helpers

> What I think though about this backup thing is that we should still make a backup, just in case an user would have tweak its config. So it wouldn't be just deleted without backup.

[...]

> I feel like the best thing to do, into the app scripts, would be to run the helper php instead of restoring the previous config.

So we make a backup of the php fpm file, but we don't use it during the restore?

kay0u:
Any thoughts about the backup/restore script? What should we do?

maniackcrudelis:
Did I forget this ? Nooo ! I wouldn't dare ! :grin:

So to remind the potential issue (If I do correctly)
What we worry about is that during a restore script, if it happens that the app uses a specific version of php, it may not have this version available.
And obviously restoring the fpm config would failed.

But, are we talking only about app with specific version of php ?
Or also about app that currently are in 7.0 (stretch), restored in a buster version ?
I guess for that last situation we should have an ugly magic trick as we did for the previous migration from php5 to 7.
While for the first one, we do have already to install the specific version of php.

You talked about running the php helper instead of simply restoring the config.
The issue https://github.com/YunoHost/issues/issues/1536 pushes us into the same direction. If we go that way (and I'll probably work on that soon) we will rather prefer to run the helper during the restore script.
For the config being lost, we can either restore after the helper or just rely on the config-panel option about overwriting config files.

kay0u:
Yeah, it sounds good the way you you depict the situation. Plus the fact that we should stop using the php-fpm template and use [`--usage --footprint`](https://github.com/YunoHost/yunohost/pull/881/files) instead

PR #881, opened 2020-02-09
New helpers for PHP and extra apt repo

maniackcrudelis:
## The problem

Some really useful new helpers for php are available and quite stable in https://github.com/YunoHost-Apps/Experimental_helpers

## Solution

- Add the new helper `ynh_install_php` to allow to install custom php version from sury **properly** without breaking everything !
- Add the new helper `ynh_install_extra_app_dependencies`. It allows to install a package from an extra repository **properly** !
Some helpers, used by this helper are needed by `ynh_install_php`.
- Add the new helper `ynh_get_scalable_phpfpm` to allow to create a php-fpm config set for the device without a template.
- Add the new helper `ynh_check_ram` to get the amount of ram of the device. Needed for `ynh_get_scalable_phpfpm`.

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps


data/helpers.d/apt
```diff
@@ -233,8 +248,11 @@ ynh_install_app_dependencies () {
             if ! grep -nrq "sury" /etc/apt/sources.list*
             then
                 # Re-add sury
-                echo "deb https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/sury.list
-                wget -O /etc/apt/trusted.gpg.d/sury.gpg https://packages.sury.org/php/apt.gpg    
+                ynh_install_extra_repo --repo="https://packages.sury.org/php/ $(lsb_release -sc) main" --key="https://packages.sury.org/php/apt.gpg" --name=extra_php_version
+
+                # Pin this sury repository to prevent sury of doing shit
+                ynh_pin_repo --package="*" --pin="origin \"packages.sury.org\"" 200 --name=extra_php_version
+                ynh_pin_repo --package="php7.0*" --pin="origin \"packages.sury.org\"" 600 --name=extra_php_version --append
```
kay0u: According to https://github.com/YunoHost/yunohost/pull/880
```suggestion
                ynh_pin_repo --package="php$YNH_DEFAULT_PHP_VERSION*" --pin="origin \"packages.sury.org\"" 600 --name=extra_php_version --append
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+	# Add an extra repository for those packages
+	ynh_install_extra_repo --repo="https://packages.sury.org/php/ $(lsb_release -sc) main" --key="https://packages.sury.org/php/apt.gpg" --priority=995 --name=extra_php_version
+
+	# Install requested dependencies from this extra repository.
+	# Install php-fpm first, otherwise php will install apache as a dependency.
+	ynh_add_app_dependencies --package="php${phpversion}-fpm"
+	ynh_add_app_dependencies --package="php$phpversion php${phpversion}-common $package"
+
+	# Set php7.0 back as the default version for php-cli.
+	update-alternatives --set php /usr/bin/php7.0
```
kay0u:
```suggestion
	update-alternatives --set php /usr/bin/php$YNH_DEFAULT_PHP_VERSION
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+	local phpversion
+	local package
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+	package=${package:-}
+
+	# Store phpversion into the config of this app
+	ynh_app_setting_set $app phpversion $phpversion
+
+	if [ "$phpversion" == "7.0" ]
```
kay0u:
```suggestion
	if [ "$phpversion" == "$YNH_DEFAULT_PHP_VERSION" ]
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+	then
+		ynh_die "Do not use ynh_install_php to install php7.0"
```
kay0u:
```suggestion
		ynh_die "Do not use ynh_install_php to install php$YNH_DEFAULT_PHP_VERSION"
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+	# Install php-fpm first, otherwise php will install apache as a dependency.
+	ynh_add_app_dependencies --package="php${phpversion}-fpm"
+	ynh_add_app_dependencies --package="php$phpversion php${phpversion}-common $package"
+
+	# Set php7.0 back as the default version for php-cli.
+	update-alternatives --set php /usr/bin/php7.0
+
+	# Pin this extra repository after packages are installed to prevent sury of doing shit
+	ynh_pin_repo --package="*" --pin="origin \"packages.sury.org\"" 200 --name=extra_php_version
+	ynh_pin_repo --package="php7.0*" --pin="origin \"packages.sury.org\"" 600 --name=extra_php_version --append
```
kay0u:
```suggestion
	ynh_pin_repo --package="php$YNH_DEFAULT_PHP_VERSION*" --pin="origin \"packages.sury.org\"" 600 --name=extra_php_version --append
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+}
+
+# Remove the specific version of php used by the app.
+#
+# usage: ynh_install_php
+ynh_remove_php () {
+	# Get the version of php used by this app
+	local phpversion=$(ynh_app_setting_get $app phpversion)
+
+	if [ "$phpversion" == "7.0" ] || [ -z "$phpversion" ]
```
kay0u:
```suggestion
	if [ "$phpversion" == "$YNH_DEFAULT_PHP_VERSION" ] || [ -z "$phpversion" ]
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+	then
+		if [ "$phpversion" == "7.0" ]
```
kay0u:
```suggestion
		if [ "$phpversion" == "$YNH_DEFAULT_PHP_VERSION" ]
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+		then
+			ynh_print_err "Do not use ynh_remove_php to install php7.0"
```
kay0u:
```suggestion
			ynh_print_err "Do not use ynh_remove_php to install php$YNH_DEFAULT_PHP_VERSION"
```

data/helpers.d/php
```diff
@@ -2,18 +2,47 @@
...
 
+	# The default behaviour is to use the template.
+	use_template="${use_template:-1}"
+	usage="${usage:-}"
+	footprint="${footprint:-}"
+	if [ -n "$usage" ] || [ -n "$footprint" ]; then
+		use_template=0
+	fi
+
 	# Configure PHP-FPM 7.0 by default
```
kay0u:
```suggestion
	# Configure PHP-FPM $YNH_DEFAULT_PHP_VERSION by default
```

data/helpers.d/php
```diff
@@ -2,18 +2,47 @@
...
 	phpversion="${phpversion:-7.0}"
```
kay0u:
```suggestion
	phpversion="${phpversion:-$YNH_DEFAULT_PHP_VERSION}"
```

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+
+	# Add an extra repository for those packages
+	ynh_install_extra_repo --repo="https://packages.sury.org/php/ $(lsb_release -sc) main" --key="https://packages.sury.org/php/apt.gpg" --priority=995 --name=extra_php_version
+
+	# Install requested dependencies from this extra repository.
+	# Install php-fpm first, otherwise php will install apache as a dependency.
+	ynh_add_app_dependencies --package="php${phpversion}-fpm"
+	ynh_add_app_dependencies --package="php$phpversion php${phpversion}-common $package"
+
+	# Set php7.0 back as the default version for php-cli.
```
kay0u:
```suggestion
	# Set php$YNH_DEFAULT_PHP_VERSION back as the default version for php-cli.
```

maniackcrudelis:
Rather prefer to wait for one or the other of those PR to be merged and then modify the other.
Anyway, there's going to be conflicts between them.

kay0u:
Is it ready to merge?

I think it would be great to merge this PR then [Use a dedicated php service for each app](https://github.com/YunoHost/yunohost/pull/915) and finally [Pave the way to migration to php7.3 and following](https://github.com/YunoHost/yunohost/pull/880)

What do you think?

maniackcrudelis:
That's ok for me, no change this PR.

data/helpers.d/php
```diff
@@ -2,18 +2,47 @@
...
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
 
+	# The default behaviour is to use the template.
+	use_template="${use_template:-1}"
+	usage="${usage:-}"
+	footprint="${footprint:-}"
+	if [ -n "$usage" ] || [ -n "$footprint" ]; then
+		use_template=0
+	fi
```
alexAubin, line 44: Hmokay, but then is it really so useful to define this use_template argument ? Can't we just have it as an internal variable that is false only if `usage` or `footprint` is defined ..?

data/helpers.d/apt
```diff
@@ -218,6 +219,20 @@ ynh_install_app_dependencies () {
...
+    then
+        # Replace version specifications by relationships syntax
+        # https://www.debian.org/doc/debian-policy/ch-relationships.html
+        # Sed clarification
+        # [^(\<=\>] ignore if it begins by ( or < = >. To not apply twice.
+        # [\<=\>] matches < = or >
+        # \+ matches one or more occurence of the previous characters, for >= or >>.
+        # [^,]\+ matches all characters except ','
+        # Ex: 'package>=1.0' will be replaced by 'package (>= 1.0)'
+        dependencies="$(echo "$dependencies" | sed 's/\([^(\<=\>]\)\([\<=\>]\+\)\([^,]\+\)/\1 (\2 \3)/g')"
```
alexAubin, line 233: myeyes

data/helpers.d/apt
```diff
@@ -266,3 +316,223 @@ ynh_remove_app_dependencies () {
...
+
+#=================================================
+
+# Install packages from an extra repository properly.
+#
+# usage: ynh_install_extra_app_dependencies --repo="repo" --package="dep1 dep2" [--key=key_url] [--name=name]
+# | arg: -r, --repo - Complete url of the extra repository.
+# | arg: -p, --package - The packages to install from this extra repository
+# | arg: -k, --key - url to get the public key.
+# | arg: -n, --name - Name for the files for this repo, $app as default value.
```
alexAubin, line 328: For my own culture: can this helper be used also during upgrade ?

E.g. I install mongodb from an extra apt repo ... Then later I want to upgrade the app and its dependency (in the upgrade script). I can and should call this helper again such that it should upgrade the mongodb and that should work, right ?

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+	sed --in-place "/$YNH_APP_INSTANCE_NAME:$phpversion/d" "/etc/php/ynh_app_version"
+
+	# If no other app uses this version of php, remove it.
+	if ! grep --quiet "$phpversion" "/etc/php/ynh_app_version"
+	then
+		# Purge php dependences for this version.
+		ynh_package_autopurge "php$phpversion php${phpversion}-fpm php${phpversion}-common"
+		# Remove the service from the admin panel
+		yunohost service remove php${phpversion}-fpm
+	fi
```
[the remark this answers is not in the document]
maniackcrudelis: You're right.
There was a warning for php7.0, but nothing for php7.3.

In both the install and remove part, we need a condition to exclude the default version.

data/helpers.d/php
```diff
@@ -65,3 +149,235 @@ ynh_remove_fpm_config () {
...
+# Set as 'medium', the process manager will be at dynamic. If the service is idle, a number of children
+#   equal to pm.min_spare_servers will stay alive. So the service can be quick to answer to any request.
+#   The number of children can grow if needed. The footprint can stay low if the service is idle, but
+#   not null. The impact on the proc is a little bit less than 'ondemand' as there's always a few
+#   children already available.
+# Set as 'high', the process manager will be set at 'static'. There will be always as many children as
+#   'pm.max_children', the footprint is important (but will be set as maximum a quarter of the maximum
+#   RAM) but the impact on the proc is lower. The service will be quick to answer as there's always many
+#   children ready to answer.
+ynh_get_scalable_phpfpm () {
```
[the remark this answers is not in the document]
maniackcrudelis: Huh, the doc should be somewhere it can be read, for sure.
And yes, the helper is designed to be internal, only used by the php helper.

Usage and footprint are to be set by the packager into the script, according to the need of the app.
But can also be modified into the config-panel by the admin. Cause usage can be different of what expected by the packager.

data/helpers.d/php
maniackcrudelis: Same concept than that previous PR https://github.com/YunoHost/yunohost/pull/364/commits/bba393c45c5ee45a942f6c7be697e7a2415e14b1 on fail2ban.

Which was actually in the other way, a template option was added.
We indeed don't **need** that variable, but it think it's nice to have it to make the code clear.

The argument on its own is not mandatory, its default value is 1.

data/helpers.d/apt
maniackcrudelis: Agree, it burns...
That's why it's explained just before ;)

data/helpers.d/apt
maniackcrudelis: I think so yes, this helper will add your specific repo, upgrade the package, then remove the repo.

data/helpers.d/php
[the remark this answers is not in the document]
maniackcrudelis: Added to my todo list about those PR.
However, rather prefer to have #880 merged with that PR to fix that problem. Since 880 will add that default php version we need.

maniackcrudelis:
For your information, I'm working locally on this PR, before testing it.
I already passed `ynh_install_php` and `ynh_remove_php` to 'internal', and integrated those helpers to `ynh_add_fpm_config` and `ynh_remove_fpm_config`, so those 2 are the only "available" helpers for packagers.

That way, `ynh_add_fpm_config`, if used with --phpversion, will installed an extra php version, if not already the default one.
And `ynh_remove_fpm_config` relay to `ynh_app_setting_get $app phpversion` to remove any extra php version that is not the default one.

Note that doesn't impact the apps already using the other helpers. But we will discourage their usage.

maniackcrudelis:
A first batch of fixes.
What's explained on my previous comment and the corrections from my tests with Drupal, Moodle and Nextcloud.
For now, I only tried from new installs with the new helpers. (But with full package check tests)

Tomorrow I'll try with an install with the old helpers, and then upgrade and backup from the new ones.
Still with package check.

Those 3 apps (that are initially just 3 apps using php7.3 and level 7) have completely different implementations of the helpers.
The good news is that it worked perfectly with the new helpers. Without having to change anything on the apps. (Except Moodle because it sources the experimental helpers **after** the officials, but that's all)

data/helpers.d/php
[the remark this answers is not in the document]
maniackcrudelis: Should be fixed by https://github.com/YunoHost/yunohost/pull/881/commits/017b0e929c7f6a07f7828013316da7fcc3fe80f5
If the helper is called to remove the default php version, it won't do it.
So, for an app that have installed php7.3 on stretch, after a migration to buster, if the app calls directly ynh_remove_php, the helper will refuse to remove php7.3 as it became the default version for buster.

kay0u:
> (Except Moodle because it sources the experimental helpers **after** the officials, but that's all)

Ouuups, should be fixed by: https://github.com/YunoHost-Apps/moodle_ynh/pull/17

maniackcrudelis:
As planned, I tried the 3 same apps with an install with their own helpers.
And then all the other tests with the new helpers, so to have kind of a real migration for already installed apps.

It worked out of the box. No modification done on this current PR.

I'm now trying a "proper" migration of nextcloud package. So, no more `ynh_install_php` or `ynh_remove_php`, only `ynh_add_fpm_config`

For now, I've noticed 2 main "issues"
The order of `ynh_add_fpm_config` and `ynh_install_app_dependencies` is important, if I don't use `extra_pkg_dependencies`. Not really an "issue", just a point to look for.
The second is more of a real "issue", Nextcloud  has a lot of non existing packages as dependencies... A lot of php7.x-stuff that does not exist aside of php-stuff.
And globally, `ynh_install_app_dependencies` does not work... I try to figure that out...

maniackcrudelis:
That time when everything goes nice and smooth.
You feel comfortable and confident, you know all will be ok with that code.
...
And you slightly lift an innocent lid
...
No luck dude, it was an `apt-pandora` box !!! You're doomed...

maniackcrudelis:
Eventually, we need the pin at 995 when installing dependencies related to sury.
Which means that we need to keep `$extra_pkg_dependencies` but to use it with `ynh_add_fpm_config` instead `ynh_install_php`.

By the way, I enhanced the logging of apt, what a hell to debug...

And finally, this PR is fully tested !
So, it works with an app using its own helper, or at least it doesn't interfere with it.
For such apps, the migration is transparent.

And to migrate the code of an app to those helpers is quite easy, mainly to give up the other helpers and add the `$extra_pkg_dependencies` to `ynh_add_fpm_config`

Just to be sure though, tomorrow I will run again the tests with the last commits.
And I'm waiting for the end of a current test to push a migration PR for nextcloud.

data/helpers.d/php
```diff
@@ -33,23 +106,82 @@ ynh_add_fpm_config () {
...
-	cp ../conf/php-fpm.conf "$finalphpconf"
-	ynh_replace_string --match_string="__NAMETOCHANGE__" --replace_string="$app" --target_file="$finalphpconf"
-	ynh_replace_string --match_string="__FINALPATH__" --replace_string="$final_path" --target_file="$finalphpconf"
-	ynh_replace_string --match_string="__USER__" --replace_string="$app" --target_file="$finalphpconf"
-	ynh_replace_string --match_string="__PHPVERSION__" --replace_string="$phpversion" --target_file="$finalphpconf"
+
+	if [ $use_template -eq 1 ]
+	then
+                # Usage 1, use the template in ../conf/php-fpm.conf
+                cp ../conf/php-fpm.conf "$finalphpconf"
```
JimboJoe, line 113: Seeing your use of this helper in restore in https://github.com/YunoHost-Apps/nextcloud_ynh/pull/292, I've tried it, but with an app that uses a template (so with a `php-fpm.conf`): it fails here because in the restoration context, th file is located in `../settings/conf`.
I'm adding this comment here because I'm not sure how to address that, and that there aren't any other side effects...
So to avoid having to set a "context" (install/upgrade or restore), shall we just test for `../conf/php-fpm.conf` and fall back to `../settings/conf/php-fpm.conf` otherwise...?

data/helpers.d/apt
JimboJoe: I confirm that's not the case. I had to specify `"grafana (>=7.0.0)"` to get it to be upgraded from 6.7.3. I think we should come up with something to fix that... but I'm really not comfortable with that piece of code about dependencies... :scream:

data/helpers.d/php
yalh76: I had the same issue during restore on after implementing as done on nextcloud_ynh :
- https://github.com/YunoHost-Apps/drupal_ynh/pull/24/commits/57a2239c2c3c12f54ef55a685f5a90a2bcd152c7
- https://github.com/YunoHost-Apps/kimai2_ynh/pull/24/commits/93968e7326d7aa1d6361437ed061a5b4a0ce9829

So as a workaround, for now I switched to not using a template:
`ynh_add_fpm_config --usage=low --footprint=low --package="$extra_php_dependencies"`

JimboJoe: I proposed a PR to fix that: https://github.com/YunoHost/yunohost/pull/1001

PR #882, opened 2020-02-19
Reorder smtpd_sender_restrictions

yalh76:
To let send without login when coming from localhost

## The problem

When trying to send mail from apps on Yunohost, for now the app need to first authenticate, which means: have to create a dedicated user for the app

## Solution

Reorder smtpd_sender_restrictions to allow to send mail without login from mynetworks

## PR Status

Ready

## How to test

For example, install mobilizon_ynh and try to register a user. Without this patch, you will find in `/var/log/mail.info` something like `NOQUEUE: reject: RCPT from localhost[127.0.0.1]: 553 5.7.1 <xxx@xxxx.xxx>: Sender address rejected: not logged in;`

With the patch, all goes well, mail sent and received

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/postfix/main.cf
+++ b/data/templates/postfix/main.cf
@@ -124,8 +124,8 @@ smtpd_helo_restrictions =
 
 # Requirements for the sender address
 smtpd_sender_restrictions =
-    reject_sender_login_mismatch,
     permit_mynetworks,
+    reject_sender_login_mismatch,
     permit_sasl_authenticated,
     reject_non_fqdn_sender,
     reject_unknown_sender_domain,


rds13:
👍

yalh76:
Maybe to be tested if it doesn't allow a user using rainloop or roundcube to send mails without authentication...

yalh76:
@maniackcrudelis and @alexAubin: maybe you have an opinion about that small change and possible consequences...

yalh76:
@kayou can it be merged at least in 3.8 ?

alexAubin:
Had a quick look to try to understand the implications ...

From what I see it was added a while ago https://github.com/YunoHost/yunohost/pull/330/files and we probably didn't thought much about which position to set this

[This stack overflow answer](https://serverfault.com/questions/640756/postfix-login-mismatch-sender-address-rejected/640763#640763) also mentions `reject_authenticated_sender_login_mismatch` which might what we're looking for ...

Idk, naively I'd suggest keeping the current order and using `reject_authenticated_sender_login_mismatch` but that's more an intuition than me really understanding the implications ...

yalh76:
Ok then, I will test that on https://github.com/YunoHost-Apps/wekan_ynh/issues/60 and let you know if `reject_authenticated_sender_login_mismatch` is enough to let wekan send mails without creating a dedicated "mail" user :p

yalh76:
Will be replaced by https://github.com/YunoHost/yunohost/pull/977

PR #884, opened 2020-02-29
Persist cookies between multiple ynh_local_curl calls for the same app

JimboJoe:
## The problem

Some app installations require the use of multiple API calls through HTTP, keeping cookies along the way. Current `ynh_local_curl` implementation doesn't enable that behavior.

## Solution

Persist cookies in an app-dedicated `/tmp/ynh_$app_cookie.txt` file, in order to keep cookies for an app, and segregate the different apps installation.

## PR Status

Implemented in Piwigo package [here](https://github.com/YunoHost-Apps/piwigo_ynh/pull/45).

## How to test

Test manually or via the Piwigo example (the token retrieval wouldn't work without that change).

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -198,6 +198,7 @@ ynh_setup_source () {
 }
 
 # Curl abstraction to help with POST requests to local pages (such as installation forms)
+# For multiple calls, cookies are persisted between each call for the same app
 #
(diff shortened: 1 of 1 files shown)


Psycojoker:
Since you've approved, I'll merge it.

PR #885, opened 2020-03-07
Fix getops

kay0u:
## The problem

Infinite loop due to getops (if you want to dig it: https://paste.yunohost.org/acinuquleh)

## Solution

Parameter Substitution magic (force the substitution at the beginning or at the end)

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -80,10 +80,10 @@ ynh_handle_getopts_args () {
 				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]}-/--${args_array[$option_flag]}\\TOBEREMOVED\\-}"
 				# And replace long option (value of the option_flag) by the short option, the option_flag itself
 				# (e.g. for [u]=user, --user will be -u)
-				# Replace long option with = (which is at the beginning of the argument)
-				arguments[arg]="${arguments[arg]/#--${args_array[$option_flag]}/-${option_flag} }"
-				# And long option without = (which is at the end of the arguement)
-				arguments[arg]="${arguments[arg]/%--${args_array[$option_flag]%=}/-${option_flag}}"
+				# Replace long option with = (match the beginning of the argument)
+				arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]}/-${option_flag} /")"
+				# And long option without = (match the whole line)
+				arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]%=}$/-${option_flag} /")"
 			done
 		done
 
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -80,10 +80,10 @@ ynh_handle_getopts_args () {
 				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]}-/--${args_array[$option_flag]}\\TOBEREMOVED\\-}"
 				# And replace long option (value of the option_flag) by the short option, the option_flag itself
 				# (e.g. for [u]=user, --user will be -u)
-				# Replace long option with =
-				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]}/-${option_flag} }"
-				# And long option without =
-				arguments[arg]="${arguments[arg]//--${args_array[$option_flag]%=}/-${option_flag}}"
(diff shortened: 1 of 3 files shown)


maniackcrudelis:
I don't understand what was the issue and how do you fix it

kay0u:
On stretch we have the same bug but it's not triggered because of the order of the loop.

https://github.com/YunoHost/yunohost/blob/a4f03b87a091a419aad72728e70a243930235a74/data/helpers.d/getopts#L61
doesn't loop in the same order on stretch and buster. (I don't understand why, probably because we have bash 5.0* on buster and 4.4* on stretch).

on Stretch:
> k (manifest_key)
m (manifest)

on Buster:
> m (manifest)
k (manifest_key)

maniackcrudelis:
Found a way to strengthen the substitution with a sed.
I replaced the 2 substitutions we had by only one:
```
arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]}/-${option_flag} /")"
```
And it's even easier to read. Try it on your own and tell me if it works also for you?

kay0u:
https://paste.yunohost.org/ligivaxogo.js

Doesn't work for me.

Example here with: `ynh_handle_getopts_args --permission main --add visitors`
Line 38:
> arguments[arg]=--permission
echo --permission | sed 's/^--permission=/-p /'

There is no "=" at the end, so the sed command doesn't work as expected.

maniackcrudelis:
Indeed...

So keep 2 different substitutions, but like that:
```
arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]}/-${option_flag} /")"
arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]%=}$/-${option_flag} /")"
```

kay0u:
I still get an error:

https://paste.yunohost.org/vuvexakoxa.bash

> -d parameter requires an argument.

With an optional argument. But I like the idea.

maniackcrudelis:
> + ynh_mysql_connect_as --user=root --password=toto --database=
[...]
> + args_array=([u]=user= [p]=password= [d]=database=)

if you indeed added --database= with no arguments, that's an error on your side.

kay0u:
https://github.com/YunoHost/yunohost/blob/270bc8f39cc376b69487848bdedff8b27d079e4c/data/helpers.d/mysql#L68-L69

maniackcrudelis:
But anyway, if the line end like that
`ynh_mysql_connect_as --user=root --password=toto --database=`
It can't work.
Optional or not is not handle by this helper. `--database=` with no argument is clearly an error. If the argument is optional, then if should not be into the command line.

kay0u:
It's probably a mistake in the helper, the fact is that with my solution this bug is not triggered.
But I can still fix the helper itself.

maniackcrudelis:
Answering 
> -d parameter requires an argument.

is the correct behavior. The option --database= requires an argument.
But I guess the code wasn't triggering that bug until then.

kay0u:
Just for science, to get the same behavior, we must surround `${arguments[arg]}` with double quote.
Which would give:
```
arguments[arg]="$(echo "${arguments[arg]}" | sed "s/^--${args_array[$option_flag]}/-${option_flag} /")"
arguments[arg]="$(echo "${arguments[arg]}" | sed "s/^--${args_array[$option_flag]%=}$/-${option_flag} /")"
```

Well I'll t

maniackcrudelis:
> Just for science, to get the same behavior, we must surround ${arguments[arg]} with double quote.

Absolutely right !

kay0u:
Should we take this patch back into stretch branches?

maniackcrudelis:
I don't think so, it works that way. There's no reason to change anything.

kay0u:
Okay, so if you agree, it's ready to merge

kay0u:
@alexAubin Uuuh, you can merge it now if you want, but if needed I can recreate the PR with a proper base :-)

(It will fix the buster CI)

alexAubin:
Ugh yea sorry for the rebase mess, I rebased this branch, thanks for the black magic debugging and fix

PR #886, opened 2020-03-09
Fix getops again

kay0u:
## The problem

`echo` may interpret arguments during substitution :

https://paste.yunohost.org/inalofovev.bash

In this example, `--db_name=bar` is replaced by `-n bar` here

https://github.com/YunoHost/yunohost/blob/cd0b0c8892cdd5c25b2a2458806dd4a20dbd088d/data/helpers.d/getopts#L84

and one the next line: 
https://github.com/YunoHost/yunohost/blob/cd0b0c8892cdd5c25b2a2458806dd4a20dbd088d/data/helpers.d/getopts#L86

`echo` see the -n argument.

## Solution

double quote `${arguments[arg]}`

Con:
Empty arguments is no longer detected as an issue (for example: `ynh_mysql_connect_as --user=root --password=toto --database=`, more info [here](https://github.com/YunoHost/yunohost/pull/885#issuecomment-596228521) on the previous discussion)

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -81,9 +81,9 @@ ynh_handle_getopts_args () {
 				# And replace long option (value of the option_flag) by the short option, the option_flag itself
 				# (e.g. for [u]=user, --user will be -u)
 				# Replace long option with = (match the beginning of the argument)
-				arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]}/-${option_flag} /")"
+				arguments[arg]="$(echo "${arguments[arg]}" | sed "s/^--${args_array[$option_flag]}/-${option_flag} /")"
 				# And long option without = (match the whole line)
-				arguments[arg]="$(echo ${arguments[arg]} | sed "s/^--${args_array[$option_flag]%=}$/-${option_flag} /")"
+				arguments[arg]="$(echo "${arguments[arg]}" | sed "s/^--${args_array[$option_flag]%=}$/-${option_flag} /")"
 			done
 		done
 


kay0u:
@maniackcrudelis do you have a better fix to avoid the problem I showed about empty arguments?

maniackcrudelis:
> Empty arguments is no longer detected as an issue (for example: ynh_mysql_connect_as --user=root --password=toto --database=, more info here on the previous discussion)

I hardly see how this is related, but anyway, that's another problem.
Double quotes are what I was suggesting on xmpp right now.

kay0u:
> > Empty arguments is no longer detected as an issue (for example: ynh_mysql_connect_as --user=root --password=toto --database=, more info here on the previous discussion)
> 
> I hardly see how this is related, but anyway, that's another problem.
> Double quotes are what I was suggesting on xmpp right now.

Without double quote:
`parse_arg '-u root' '-p toto' -d`

With double quote:
`parse_arg '-u root' '-p toto' '-d '`

maniackcrudelis:
The second result is what we expect.
The detection of the lack of argument should be detected later. This code has as only purpose to shrink the line to the small options.

kay0u:
@alexAubin You can merge it when you want.

I'm going to leave this bug for now because it's been like this for a while.

alexAubin:
> into YunoHost:buster-unstable from kay0u:fix-getops

yo dawg you aint in the core dev yet ?

added you bruh

kay0u:
I'm just not confident enough yet :x

Thanks

PR #887, opened 2020-03-10
[helpers.d/postgresql] support for buster version 11

arthurlutz:
## The problem

detecting old version of postgres

## Solution

add detection for buster version

## How to test

For example install ynh_funkwhale see https://github.com/YunoHost-Apps/funkwhale_ynh/issues/98

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/postgresql
+++ b/data/helpers.d/postgresql
@@ -269,14 +269,17 @@ ynh_psql_test_if_first_run() {
 		local psql_root_password="$(ynh_string_random)"
 		echo "$psql_root_password" >$PSQL_ROOT_PWD_FILE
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Uh but I already made a similar fix [here](https://github.com/YunoHost/yunohost/commit/1ce8cebaae8142d71d9f8d0fce0112c8b4f62a50), dunno if it's really meaningful to add this to the migration stuff that will only be ran from stretch ?

arthurlutz:
I found your fix on the buster-unstable branch after, sorry for the noise. It might be nice to document (or point me to some content) about the difference between the work you're doing on buster-unstable and migration-buster.

PR #888, opened 2020-03-12
Allow dots in application name (fix #1525)

southerntofu:
## The problem

Described in [#1525](https://github.com/YunoHost/issues/issues/1525)

## Solution

Add a literal dot in the regex to validate application names.

## PR Status

Tested by @Gofannon. Seems to fix the issue for them.

Doctests were not run due to my failure at setting up the dev environment. I'll come back to you when i succeed

## How to test

Installing and removing an app with a dot in the name should not fail. This can be tried with the `diagrams.net` application.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -67,7 +67,7 @@ re_github_repo = re.compile(
 )
 
 re_app_instance_name = re.compile(
-    r'^(?P<appid>[\w-]+?)(__(?P<appinstancenb>[1-9][0-9]*))?$'
+    r'^(?P<appid>[\w\.-]+?)(__(?P<appinstancenb>[1-9][0-9]*))?$'
 )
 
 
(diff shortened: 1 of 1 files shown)


Josue-T:
Hello, as we have decided to use the dot to separate the permission name and the application in own permissions system I don't think that it's really a good idea to accept this.

maniackcrudelis:
Sounds to me that the problem isn't that dots should be allowed, but on the contrary that dots shouldn't be allowed during the installation.
https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines.md#yep-11

Gofannon:
> Sounds to me that the problem isn't that dots should be allowed, but on the contrary that dots shouldn't be allowed during the installation.
> https://github.com/YunoHost/doc/blob/master/packaging_apps_guidelines.md#yep-11

Agreed. It should be forbidden in the first place.
I didn't run `package_check` to verify the compliance against YEP rules. I installed it straight on YunoHost server.

kay0u:
As explained, we will not merge this PR, so I am closing it. Feel free to reopen it.

Thanks anyway for your contribution! <3

PR #889, opened 2020-03-14
[enh] Tell apt to explain what's wrong when there are unmet dependencies

alexAubin:
## The problem

When a dependency cannot be installed, we currently get something like : 

```
Info: DEBUG - The following packages will be REMOVED:
Info: WARNING - E: Packages need to be removed but remove is disabled.
Info: DEBUG -   appname-ynh-deps
Info: WARNING - Unable to install dependencies
```

which usually corresponds to a dependency of appname-ynh-deps that cannot be installed ... but doesn't tell us what / why ...

## Solution

Instead, we can use `apt-get check` to tell us what's the issue with some output like : 

```
WARNING - E: Packages need to be removed but remove is disabled.
DEBUG - + apt-get check
DEBUG - Reading package lists...
DEBUG - Building dependency tree...
DEBUG - Reading state information...
DEBUG - You might want to run 'apt --fix-broken install' to correct these.
DEBUG - The following packages have unmet dependencies:
DEBUG -  pixelfed-ynh-deps : Depends: libvpx4 but it is not installable
DEBUG -                      Depends: libmagickwand-6.q16-3 but it is not installable or
DEBUG -                               libmagickwand-6.q16-6 but it is not installed
DEBUG -                      Depends: pngquant but it is not installed
DEBUG -                      Depends: jpegoptim but it is not installed
DEBUG -                      Depends: gifsicle but it is not installed
DEBUG -                      Depends: php7.3-bcmath but it is not installed
DEBUG -                      Depends: php7.3-pgsql but it is not installed
DEBUG -                      Depends: php7.3-zip but it is not installed
WARNING - E: Unmet dependencies. Try 'apt --fix-broken install' with no packages (or specify a solution).
DEBUG - + ynh_die '--message=Unable to install dependencies'
WARNING - Unable to install dependencies
```

where we can pinpoint the issue to : `pixelfed-ynh-deps : Depends: libvpx4 but it is not installable` (the other ones are only "not installed" but implied to be installable)

## PR Status

Made a quick test on my side and works as expected

## How to test

Try to install an app that has unmet dependencies (possibly edit the install script and add a dummy non-existent dependency)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/apt
+++ b/data/helpers.d/apt
@@ -186,6 +186,9 @@ ynh_package_install_from_equivs () {
     (cd "$TMPDIR"
     equivs-build ./control 1> /dev/null
     dpkg --force-depends -i "./${pkgname}_${pkgversion}_all.deb" 2>&1)
+    # If install fails we use "apt-get check" to try to debug and diagnose possible unmet dependencies
+    # Note the use of { } which allows to group commands without starting a subshell (otherwise the ynh_die wouldn't exit the current shell). 
+    # Be careful with the syntax : the semicolon + space at the end is important!
     ynh_package_install -f || { apt-get check 2>&1; ynh_die --message="Unable to install dependencies"; }
     [[ -n "$TMPDIR" ]] && rm -rf $TMPDIR	# Remove the temp dir.
 
--- a/data/helpers.d/apt
+++ b/data/helpers.d/apt
@@ -186,7 +186,7 @@ ynh_package_install_from_equivs () {
     (cd "$TMPDIR"
     equivs-build ./control 1> /dev/null
     dpkg --force-depends -i "./${pkgname}_${pkgversion}_all.deb" 2>&1)
-    ynh_package_install -f || ynh_die --message="Unable to install dependencies"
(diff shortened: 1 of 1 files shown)


data/helpers.d/apt
```diff
@@ -186,7 +186,7 @@ ynh_package_install_from_equivs () {
     (cd "$TMPDIR"
     equivs-build ./control 1> /dev/null
     dpkg --force-depends -i "./${pkgname}_${pkgversion}_all.deb" 2>&1)
-    ynh_package_install -f || ynh_die --message="Unable to install dependencies"
+    ynh_package_install -f || { apt-get check 2>&1; ynh_die --message="Unable to install dependencies"; }
```
maniackcrudelis: Maybe a small comment here could be useful to explain the usage of {} as a group of commands without starting a sub shell, that has to end with a ;

Otherwise, a wonderful idea !

PR #894, opened 2020-03-22
Allow public apps with no sso tile

kay0u:
## The problem

https://forum.yunohost.org/t/yunohost-3-7-spooky-testing-call-for-feedback/9385/49?u=kayou
https://forum.yunohost.org/t/yunohost-3-7-spooky-testing-call-for-feedback/9385/50?u=kayou

## Solution

...

## PR Status

Not fully tested, any though about it?

## How to test

...

## Validation

- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


kay0u:
I've tried every case I can think of, it works for me...

We were supposed to release version 3.7 this week, so I guess it's delayed, again?

Josue-T:
> I thought about it a bit and it _seems_ okay, but I really can't say for sure that there aren't any piece of code (maybe in `app_ssowatconf` or related) doing some assumptions about this ... Also gotta be careful about what happens in `app_install` (and in particular the `_migrate_legacy_permission`) and in the migration procedure.

I'll try this week to update all of my apps with group-permission support. So it will be a good occasion to see if we still have some bug or not.

PR #901, opened 2020-03-28
Unit tests / lint / cleaning for translation files

alexAubin:
## The problem

Inconsistencies in format of locale string break YunoHost ... Several of them likely introduced because of the use of autotranslation (or negligence idk)

## Solution

Add a linter meant to be run in travis ?

Still wip ... Not sure how to inteface this with travis exactly

## PR Status

WIP

## How to test

Run `python scriptname.py`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


tests/remove_stale_string.py
```diff
@@ -0,0 +1,19 @@
...
+
+reference = json.loads(open(locale_folder + "en.json").read())
+
+for locale_file in locale_files:
+
+    print(locale_file)
+    this_locale = json.loads(open(locale_folder + locale_file).read(), object_pairs_hook=OrderedDict)
+    this_locale_fixed = {k:v for k, v in this_locale.items() if k in reference}
+
+    json.dump(this_locale_fixed, open(locale_folder + locale_file, "w"), indent=4, ensure_ascii=False)
```
Psycojoker:
```suggestion
    json.dump(this_locale_fixed, open(locale_folder + locale_file, "w"), indent=4, ensure_ascii=False, sort_keys=True)
```

I'm surprised you didn't needed that option but normally you should to avoid breaking order?

alexAubin: We can, but the locale files are not ordered in the first place so that would create huge diff that reorders everything :s (Ultimately someday we could try to configure weblate to automatically reorder things when it creates commits I guess)

Psycojoker: Oh ok, I though they were '^'

I'm really surprised that the order is unchanged then '-'

alexAubin: Yup, I had to load the json using an OrderedDict

alexAubin:
(I removed untranslated strings from ar.json after discussing with ButterflyOfFire on Mastodon, agreed that this was not expected)

alexAubin:
Willing to add those to travis.yml but that's my first time dealing with travis ...

I believe we disabled it in github at some point because nobody was really paying attention to the python linter results ? :/ 

(Also I'll tried to explain / organize a bit more the different script, maybe turn them into actual pytest scripts rather than custom stuff...)

tests/remove_stale_string.py
```diff
@@ -0,0 +1,19 @@
+import re
+import json
+import glob
+from collections import OrderedDict
+
```
alexAubin: Btw I don't know where to put this small script as it's not really a test but more some sort of automatic cleaning tool ...

alexAubin:
(In fact I still need to turn check_locale_format_consistency.py into tests for tox as well)

kay0u:
You can merge it if you have done

PR #902, opened 2020-03-28
Remove redirected_urls after postinstall

kay0u:
## The problem

...

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tools.py
+++ b/src/yunohost/tools.py
@@ -303,9 +303,6 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
     else:
         ssowat_conf = read_json('/etc/ssowat/conf.json.persistent')
 
-    if 'redirected_urls' not in ssowat_conf:
-        ssowat_conf['redirected_urls'] = {}
-
     write_to_json('/etc/ssowat/conf.json.persistent', ssowat_conf)
     os.system('chmod 644 /etc/ssowat/conf.json.persistent')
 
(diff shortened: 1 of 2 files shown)


src/yunohost/tools.py
```diff
@@ -306,8 +306,6 @@ def tools_postinstall(operation_logger, domain, password, ignore_dyndns=False,
     if 'redirected_urls' not in ssowat_conf:
         ssowat_conf['redirected_urls'] = {}
```
alexAubin: Imho we can also remove this `if`

(I guess keeping the other stuff that read/write the conf.json.persistent doesn't hurt so that at least the file exists already, so people willing to tweak it won't have to create it manually)

src/yunohost/user.py
```diff
@@ -192,22 +192,6 @@ def user_create(operation_logger, username, firstname, lastname, mail, password,
         'loginShell': '/bin/false'
     }
 
-    # If it is the first user, add some aliases
-    if not ldap.search(base='ou=users,dc=yunohost,dc=org', filter='uid=*'):
-        attr_dict['mail'] = [attr_dict['mail']] + aliases
```
alexAubin: We need this tho

kay0u:
Thanks for your remarks

PR #905, opened 2020-03-30
add ynh_permission_has_user

kay0u:
## The problem

https://github.com/YunoHost/example_ynh/pull/111#discussion_r400354745

## Solution

...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -303,8 +303,6 @@ user:
                     arguments:
                         permission:
                             help: Name of the permission to fetch info about
-                            extra:
-                                pattern: *pattern_username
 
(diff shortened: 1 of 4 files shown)


maniackcrudelis:
Meanwhile, we can use that other helper https://github.com/YunoHost-Apps/Experimental_helpers/blob/master/ynh_permission_has_user/ynh_permission_has_user

It works the same way, so as soon as this helper is released, it will replaced seamlessly the experimental one.

kay0u:
Thanks for the fix, I just saw that (tested python part, but not bash >.<)

PR #909, opened 2020-04-01
Add conflict rule against apache2 and bind9

alexAubin:
## The problem

c.f. discussion on Yunohost apps, some app requires php-uploadprogress with has a "recommend libapache2-whatever which in turns install apache.

But we've seen many situations leading to apache or bind9 being installed which breaks critical services like nginx and dnsmasq...

## Solution

Bringing this once again because we really need to do something about it on the long term.

It's been discussed that some advanced user may want to install apache2 on the side, but fuck it, if they need it they can still force it using some `dpkg --force-conflicts` somehow.

## PR Status

Yolocommited

## How to test

Try to install apache or bind with this rule enabled..

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -43,6 +43,7 @@ Conflicts: iptables-persistent
  , yunohost-config-dovecot, yunohost-config-slapd
  , yunohost-config-nginx, yunohost-config-amavis
  , yunohost-config-mysql, yunohost-predepends
+ , apache2, bind9
 Replaces: moulinette-yunohost, yunohost-config
  , yunohost-config-others, yunohost-config-postfix
  , yunohost-config-dovecot, yunohost-config-slapd


Josue-T:
Well, I'm one of the user which use bind9 with dnsmasq on the same machine... just hope that too many user willl be impacted by this.

Josue-T:
Hello,

For some other who have some issue about this I tried to bypass the error with `dpkg --force-conflicts` but it don't really solve the issue because after apt is happy with this. The only way that I found to fix this issue with a stable state of apt is to rename the bind package with a new name. I wrote a small script for this.

```bash
#!/bin/bash

set -eu

pkg=bind9
tmp=$(mktemp -d)

chown _apt $tmp
chmod 700 $tmp
cd $tmp

apt-get download $pkg
dpkg-deb -x $pkg*.deb package
dpkg-deb --control $pkg*.deb package/DEBIAN
sed -i "s|Package: $pkg|Package: $pkg-no-conflict|g" package/DEBIAN/control
dpkg -b package ${pkg}_no_conflict.deb
apt-get install ./${pkg}_no_conflict.deb
```

Before to run this script you will need to remove temporary the bind9 (or apache) package.

PR #912, opened 2020-04-02
No need for mysql root password

alexAubin:
## The problem

Another change on top of the removal of the backup/restore of mysql password.

This one is a bit more touchy though ... so to be discussed.

The problem is that managing that mysql password in /etc/yunohost/mysql is a bit annoying (gotta call a damn dpkg-reconfigure to change or define it wtf) and it's actually a security risk to use it in command line as the password will show up to other unix users in `ps -ef` ... Same thing can be said for psql btw ...

As I realized today, we don't need to do this anymore. Mariadb ships a auth_socket/unix_socket plugin that allows to authenticate just from the fact that you're already root, which makes more sense than some funky password.

## Solution

Adapt helpers to call `mysql` without password, and auto-migrate to enable unix_socket as auth plugin for root in the regen-conf (it's in fact only needed for people that come from the wheezy/jessie era)

## PR Status

Tested the migration part but didn't carefully test the helpers...

## How to test

Zblerg :| 

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/hooks/conf_regen/34-mysql
+++ b/data/hooks/conf_regen/34-mysql
@@ -19,6 +19,7 @@ do_post_regen() {
       # dpkg-reconfigure will initialize mysql (if it ain't already)
       # It enabled auth_socket for root, so no need to define any root password...
       # c.f. : cat /var/lib/dpkg/info/mariadb-server-10.3.postinst | grep install_db -C3
+      MYSQL_PKG="$(dpkg --list | sed -ne 's/^ii  \(mariadb-server-[[:digit:].]\+\) .*$/\1/p')"
       dpkg-reconfigure -freadline -u "$MYSQL_PKG" 2>&1
 
       systemctl -q is-active mariadb.service \
--- a/data/hooks/conf_regen/34-mysql
+++ b/data/hooks/conf_regen/34-mysql
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 set -e
-MYSQL_PKG="$(dpkg --list | sed -ne 's/^ii  \(mariadb-server-[[:digit:].]\+\) .*$/\1/p')"
 . /usr/share/yunohost/helpers
 
 do_pre_regen() {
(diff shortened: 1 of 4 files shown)


data/hooks/conf_regen/34-mysql
```diff
@@ -1,58 +1,41 @@
...
+          password="$(cat /etc/yunohost/mysql)"
+          # Enable plugin unix_socket for root on localhost
+          mysql -u root -p"$password" <<< "GRANT ALL PRIVILEGES ON *.* TO 'root'@'localhost' IDENTIFIED WITH unix_socket WITH GRANT OPTION;"
+      fi
+
+      # If now we're able to login without password, drop the mysql password
+      if echo "" | mysql
+      then
+          rm /etc/yunohost/mysql
+      fi
```
Josue-T, line 29: Maybe we can manage more this situation. Because in this case at each regen-conf, we will try to update the privilege in mysql and don't really make sens as if the first time it fail the other time it will probably also fail...

Josue-T:
Can we also do this for postgresql ?

data/hooks/conf_regen/34-mysql
alexAubin: Hmyea I guess at least we can display a warning ...

I don't know much about what else to do though ... If enabling the auth_socket plugin doesn't work, that means that either the mysql didn't work, or that there's some issue with the structure or permission of the mysql db ... In both cases, that means there are bigger issues on the system and we can't fix it automagically. Yet, we do need this if we expect to be able to run mysql without password in other parts of yunohost like the mysql helpers ...

alexAubin:
>Can we also do this for postgresql ?

Yes I think we can do this, just gotta add the right stuff in pg_hba.conf I think ... I'll try to have a look

data/hooks/conf_regen/34-mysql
Josue-T: OK, Yes,

We can just add a warning.

kay0u:
https://github.com/YunoHost/test_apps/pull/8 should fix tests

data/hooks/restore/11-conf_ynh_mysql
```diff
@@ -1,42 +1,5 @@
...
-}
-
-# store new root password
-echo "$new_pwd" | tee /etc/yunohost/mysql
-chmod 400 /etc/yunohost/mysql
-
-# reload the grant tables
-mysqladmin -s -u root -p"$new_pwd" reload
+# This is a dummy empty file as a workaround for
+# https://github.com/YunoHost/issues/issues/1553 until it is fixed
```
kay0u, line 5: https://github.com/YunoHost/yunohost/pull/927 should fix that?

alexAubin: Yes indeed ... that's from the other PR ... I don't know why I kept it in the end, we should remove it from the idontknowwhicone-unstable branch as well

zamentur:
I mark it in "work needed":

- [ ] Some apps depends of /etc/yunohost/mysql : https://github.com/search?q=org%3AYunoHost-Apps+%22%2Fetc%2Fyunohost%2Fmysql%22&type=code
- [ ] Conflicts to solve

alexAubin:
Somewhat related if we want to do mysql operation without knowing the root password ... https://github.com/YunoHost/issues/issues/1731#issuecomment-772152848

alexAubin:
Merging to move forward

PR #913, opened 2020-04-03
Update nginx SSL conf with new Mozilla recommendation

alexAubin:
## The problem

Nginx ciphers and other security settings not up to date with Mozilla's recommendation (in particular w.r.t. using TLSv1 and 1.1)

## Solution

c.f. https://ssl-config.mozilla.org/#server=nginx&version=1.10.3&config=intermediate&openssl=1.1.1l&ocsp=false&guideline=5.4

Though we can't have any "modern" configuration with 1.10.3... (lack supports for TLS1.3 I believe) Need buster to have nginx 1.14...

## PR Status

Yolocommited, hopefully not breaking everything

## How to test

Regen nginx conf, clear browser cache and test that it works.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/security.conf.inc
+++ b/data/templates/nginx/security.conf.inc
@@ -1,24 +1,22 @@
-{% if compatibility == "modern" %}
-# Ciphers with modern compatibility
-# https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=nginx-1.6.2&openssl=1.0.1t&hsts=yes&profile=modern
-# The following configuration use modern ciphers, but remove compatibility with some old clients (android < 5.0, Internet Explorer < 10, ...)
-ssl_protocols TLSv1.2;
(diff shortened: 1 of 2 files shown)


frju365:
I may ask the question : What would be the risk in fact to have more recent version of nginx, which doesn't come from debian repos but nginx directly ? Because in fact, nginx repo is normally enough stable, and could be a good thing to have latest features regarding security , DDOS mitigation...

https://nginx.org/en/linux_packages.html#Debian

alexAubin:
I would say there's a risk in terms of creding catastrophic conflicts in some random situations ... We've seen this with sury which ended up removing php7.0-common because php7.3-curl conflicted with it ... (I don't remember exactly) And in the end we should just finish Buster instead of investing energy in "having the new stuff without Buster" ... :/

Josue-T:
Sury also break pip here : https://github.com/YunoHost-Apps/pgadmin_ynh/issues/3

alexAubin:
Ogod ... now that you mention it, i generated these configs with openssl 1.1.1l instead of 1.1.0l

Might be why the core CI is broken with this PR ?

alexAubin:
(In fact, the ssl 1.1.0l requirement doesn't change anything except for Dovecot)

PR #914, opened 2020-04-03
Update mailstack ciphers

alexAubin:
## The problem

Configurations are not up to date with Mozilla's recommendation ...

## Solution

Apply recommendation

For dovecot that sounds pretty legit and okay ... For postfix, I'm less confident about what I'm doing ... in particular because there's no recommendation for the `smtp_` settings, which I believe define what happens for the interaction with other servers

## PR Status

Yolocommited :s 

## How to test

Uh try to configure an email client and submit some mails to deliver to external servers

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/dovecot/dovecot.conf
+++ b/data/templates/dovecot/dovecot.conf
@@ -12,10 +12,25 @@ protocols = imap sieve {% if pop3_enabled == "True" %}pop3{% endif %}
 
 mail_plugins = $mail_plugins quota
 
-ssl = yes
+###############################################################################
+
+# generated 2020-04-03, Mozilla Guideline v5.4, Dovecot 2.2.27, OpenSSL 1.1.1l, intermediate configuration
+# https://ssl-config.mozilla.org/#server=dovecot&version=2.2.27&config=intermediate&openssl=1.1.1l&guideline=5.4
+
+ssl = required
+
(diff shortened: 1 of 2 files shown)


data/templates/dovecot/dovecot.conf
```diff
@@ -12,10 +12,25 @@ protocols = imap sieve {% if pop3_enabled == "True" %}pop3{% endif %}
...
+ssl = required
+
 ssl_cert = </etc/yunohost/certs/{{ main_domain }}/crt.pem
 ssl_key = </etc/yunohost/certs/{{ main_domain }}/key.pem
-ssl_protocols = !SSLv3
+
+ssl_dh_parameters_length = 2048
+
+# intermediate configuration
+ssl_protocols = TLSv1.2 TLSv1.3
```
Josue-T, line 28: Are you sure that TLSv1.3 is already supported on buster ?

alexAubin: I was suspicious too but that's indeed what Mozilla recommends with Dovecot 2.2.27 (= on Stretch) and tweaking the conf with this, Dovecot still works (or at least the service starts...)

PR #915, opened 2020-04-03
Use a dedicated php service for each app

maniackcrudelis:
## The problem

Discussion about using a dedicated php service for each app https://github.com/YunoHost/issues/issues/1536

## Solution

It works perfectly

## PR Status

We should probably reload the main service during an upgrade though.

## How to test

Use this config file as an extra helper to source into the scripts.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


maniackcrudelis:
Indeed, and I was expecting that when starting. And then I forgot...

But it appears that I have:
- A master process (master master)
- 2 pool for www
- Another master process, for jirafeau
- And 3 pool for jirafeau

If I stop the service php for jirafeau, I get a 502 on Jirafeau and the process aren't there anymore.
If I do reload the global php service, I get back my pools for jirafeau. (But not its master process, logic)
And I can't restart anymore my jirafeau pool, because another instance listen to it...

alexAubin:
(Naively I see that the global conf has `include=/etc/php/7.0/fpm/pool.d/*.conf` so we could imagine having the file in the same location ... but not ending with .conf ... Dunno if that's a good idea tho)

maniackcrudelis:
Same problem as explain above.

alexAubin:
(Nvm about my previous idea, that's just moving the problem somewhere else, many apps have hard-coded references to php7.0-fpm service..., like to reload it ...)

maniackcrudelis:
You're right they have...

data/helpers.d/php
```diff
@@ -45,7 +45,27 @@ ynh_add_fpm_config () {
...
+	# Create a config for a dedicated php-fpm service for the app
+	echo "
+[Unit]
+Description=PHP $phpversion FastCGI Process Manager for $app
+After=network.target
+
+[Service] 
+Type=notify
+PIDFile=/run/php/php${phpversion}-fpm-$app.pid
+ExecStart=/usr/sbin/php-fpm${phpversion} --nodaemonize --fpm-config $finalphpconf --pid /run/php/php${phpversion}-fpm-$app.pid
```
kay0u, line 58: On my php-fpm service if I also have these lines:

```
ExecStartPost=-/usr/lib/php/php-fpm-socket-helper install /run/php/php-fpm.sock /etc/php/7.0/fpm/pool.d/www.conf 70
ExecStopPost=-/usr/lib/php/php-fpm-socket-helper remove /run/php/php-fpm.sock /etc/php/7.0/fpm/pool.d/www.conf 70
```

But... I have no idea what is it for

kay0u:
Do we have to change this line [too](https://github.com/YunoHost/example_ynh/blob/15bf6f665e108a6086122158ae7d042660b66626/conf/nginx.conf#L21), or at least the nginx helper ?

data/helpers.d/php
maniackcrudelis: On stretch ?

maniackcrudelis:
We don't change the place of the socket, no need to change nginx conf.

maniackcrudelis:
Thinking about it, I would really not go for a dirty patch that change the path during the install or whatever.
It would be inconsistent with the scripts themselves.

If possible I rather prefer to have a proper way to do it. Considering, we would also have to manage the backup and restore script.

What about a settings into settings.yml, set by the php helper to says "Ok, that app has moved to a dedicated php service." It would be done during an install or an upgrade.
Then the php, backup and restore helper could rely to that settings to know what to do.

It would be an easy way to migrate all apps properly.
By the way, it could throw a warning about the need to migrate properly the app if the backup or restore still use the old path.

alexAubin:
Alright but again that means we have to take care of legacy apps for the next X years, both technically and in terms of user support ... who are going to be confused because there's a php-fpm service AND a nextcloud service (for example), and when they restart php-fpm it restarts some apps but not nextcloud because it's on its own, etc ...

Can you elaborate on what issue may arise with hacking the script on the fly ? Granted that's not pretty, but so far for stuff like php5 / php7.0 / php7.3 it works kinda well.

maniackcrudelis:
The solution I suggest (the last one) would end up upgrading apps as soon as there's an install or an upgrade, since in such situation we use the helper.
So, not too much legacy apps after that.
Meanwhile, as the backup script would still use the old path, we could print a warning to ask to migrate.
And after some time package linter could consider that using that path is a problem, as we do often.

I don't think it would raise any issue, just that to migrate magically from the core on the fly would not be consistent with what the package says.
Sometimes, it just harder to debug when that kind of things happen because it's not consistent with the scripts themselves.

Just this morning a packager was telling that it should absolutely move from php5 to 7 otherwise it wouldn't work (very old package...).

But I admit that finally, there's some migration code in both solution to do the job.

data/helpers.d/php
kay0u: Yes

maniackcrudelis: Where was this file ? I didn't have one !

kay0u: I don't have it on buster btw.

For example (still stretch):
```
root@ynh-dev-stretch:/ynh-dev# cat /etc/systemd/system/multi-user.target.wants/php7.3-fpm.service 
[Unit]
Description=The PHP 7.3 FastCGI Process Manager
Documentation=man:php-fpm7.3(8)
After=network.target

[Service]
Type=notify
ExecStart=/usr/sbin/php-fpm7.3 --nodaemonize --fpm-config /etc/php/7.3/fpm/php-fpm.conf
ExecStartPost=-/usr/lib/php/php-fpm-socket-helper install /run/php/php-fpm.sock /etc/php/7.3/fpm/pool.d/www.conf 73
ExecStopPost=-/usr/lib/php/php-fpm-socket-helper remove /run/php/php-fpm.sock /etc/php/7.3/fpm/pool.d/www.conf 73
ExecReload=/bin/kill -USR2 $MAINPID

[Install]
WantedBy=multi-user.target
```

kay0u: Would it be "sury" that adds these lines? (Do you install/upgrade php via "sury" on your side?).

maniackcrudelis:
Ok, I worked on that.

Adding a migration part into ynh_backup, ynh_restore_file and ynh_systemd_action. Aside of php of course.

I use jirafeau.
- Install from apps list
I use a local jirafeau package on which I only add a source to my helpers. No others modifications.
- Upgrade to my package. Php migrate properly to a dedicated service.
- Backup. A warning tell me that I should migrate my app, but it works perfectly.
- Restore. Again a warning to tell me to migrate, all ok.
But only apparently, my restore script does not restore the service... It doesn't even know there is a service for a dedicated php.
Actually, the file went to the standard pool. Maybe a mistake on my side, but any way, without the service it won't work.
And with the file there, my remove fails.

As much as I like this idea of having a dedicated service, I don't think trying to automatically migrate is a so good idea.
I think we should just add that as an option, and let's packager migrate themselves.

Later we could says that do not migrate will reduce the level.

data/helpers.d/php
maniackcrudelis: Damn, I missed that directory multi-user.target.wants !
But I have the same script as before
```
PIDFile=/run/php/php7.0-fpm.pid
ExecStart=/usr/sbin/php-fpm7.0 --nodaemonize --fpm-config /etc/php/7.0/fpm/php-fpm.conf
ExecReload=/bin/kill -USR2 $MAINPID
```

Nope, so sury on my side

kay0u: I don't have sury on my buster server either. Anyway, I still don't know the purpose of these lines

maniackcrudelis: php-fpm-socket-helper does not exist on my setup, I guess that close the investigation...

kay0u: Mine: https://paste.yunohost.org/inuqiduril.bash

maniackcrudelis: Looks like a default socket to use and update before start and after a stop.
I think we can ignore this in our case.

data/helpers.d/php
```diff
@@ -45,7 +64,27 @@ ynh_add_fpm_config () {
...
+ExecReload=/bin/kill -USR2 \$MAINPID
+
+[Install]
+WantedBy=multi-user.target
+" > ../conf/$fpm_service
+
+	# Create this dedicated php-fpm service
+	ynh_add_systemd_config --service=$fpm_service --template=$fpm_service
+
```
kay0u, line 86:
```suggestion
	yunohost service add $fpm_service --log_type "systemd"
```

data/helpers.d/php
```diff
@@ -61,7 +100,10 @@ ynh_remove_fpm_config () {
 		fpm_config_dir="/etc/php/7.0/fpm"
 		fpm_service="php7.0-fpm"
 	fi
+
+	# Remove the dedicated service php-fpm service
+	ynh_remove_systemd_config --service=$fpm_service
+
```
kay0u, line 106:
```suggestion
	yunohost service remove $fpm_service
```

data/helpers.d/php
maniackcrudelis: Yes indeed.
But...
There's actually an error log in /var/log/php7.0-fpm.log and another part of the log in syslog.
Would be interesting to have everything in the same place.

maniackcrudelis: `php_admin_value[error_log]` and `php_admin_flag[log_errors]` are supposed to be the keys to use in the conf file.
But maybe only for the errors.

Meanwhile, I got the stop and start log into syslog and some info into /var/log/php-fpm.log...

Interesting fact, php-fpm.log is the default value for [php-fpm.conf](https://github.com/php/php-src/blob/master/sapi/fpm/php-fpm.conf.in#L24)
Looks like a recurrent problem, .d conf does not really overwrite the value, despite the doc saying it does...

To use a `/etc/php/7.0/dedicated-fpm/php-fpm.conf` does not change anything...

I do not see any proper solution...

maniackcrudelis: Finally succeed to have all logs into a specific file for each service.
I tried with success with 3 apps, each running on their dedicated service.

Here again, to manage the backup/restore, it would be a lot easier to run the helper again.

maniackcrudelis:
This PR is almost ready. Should we merge it in #881 ? Or wait for 881 to be merged to resolve conflict ?
Or just wait to be sure we want that ?

kay0u:
I think I'll merge #881 and #926, if everybody is agree indeed. Once this is done, we can resolve the conflicts and have a clear vision of this PR

data/helpers.d/php
```diff
@@ -2,32 +2,67 @@
...
 	phpversion="${phpversion:-7.0}"
 
-	local fpm_config_dir="/etc/php/$phpversion/fpm"
-	local fpm_service="php${phpversion}-fpm"
+	# Do not use a dedicated service by default
+	dedicated_service=${dedicated_service:-0}
+
+	if [ $dedicated_service -eq 1 ]
+	then
+		local fpm_service="php${phpversion}-fpm-$app"
```
alexAubin, line 27: I think we should really consider having the name of the dedicated fpm service just being `$app` ... Otherwise it's gonna be such a pain to remember how these are called when trying to help users. When we package nodejs apps (or whichever other techno than php) we usually just call the service `$app` because there's no point in calling it `nodejs10.3-$app`. 

It'd be so much simpler to just be able to run for example `systemctl status nextcloud` and `journalctl -u nextcloud` (and also seeing just 'nextcloud' in the webadmin service list)

```suggestion
		local fpm_service="$app"
```

maniackcrudelis: Mmh, at least something like $app-php
We're talking here about php, not the app itself.

alexAubin: Depends what you mean by the app itself ... Again with nodejs apps and others, it's common practice to call the service $app and not $app-nodejs or $app-$tech :confused:

maniackcrudelis: Feel like node apps are more dependent to node, and thus to remove the node part in such apps is not a problem for me.
Php apps are dependent from nginx, php and fpm. So for me, considering that it's only the php-fpm part that we're talking about, and not all the app, for me would be better to have a service name that clearly said that it's only php.
Also, there's no such thing here as restarting the app. As you could do with etherpad for example.
If you change the config file for etherpad, you have to restart etherpad, and so node with it.
If you do so with wordpress, you don't have to restart anything, you will have to only if you change the php-fpm config for wordpress.

maniackcrudelis:
```suggestion
		local fpm_service="${app}-phpfpm"
```
What about that ? Easy to find the service for an app, and quite clear about what it is.

data/helpers.d/php
```diff
@@ -205,13 +278,22 @@ ynh_remove_fpm_config () {
...
+		# Remove the global php-fpm conf
+		ynh_secure_remove --file="$fpm_config_dir/php-fpm-$app.conf"
+		# Remove the service from the list of services known by Yunohost
+		yunohost service remove $fpm_service
+	elif ynh_package_is_installed --package="php${phpversion}-fpm"; then
 		ynh_systemd_action --service_name=$fpm_service --action=reload
 	fi
 
+	ynh_secure_remove --file="$fpm_config_dir/pool.d/$app.conf"
+	ynh_secure_remove --file="$fpm_config_dir/conf.d/20-$app.ini" 2>&1
```
kay0u, line 295:
```suggestion
	ynh_exec_warn_less ynh_secure_remove --file="$fpm_config_dir/conf.d/20-$app.ini"
```

data/helpers.d/php
kay0u: I vote for the last one!

alexAubin: Hmf yea let's go for this  ¯\\\_\(ツ)\_/¯

maniackcrudelis: :stuck_out_tongue_closed_eyes: :trophy:

PR #916, opened 2020-04-04
[enh] Setup all XMPP components for each "parent" domains

pitchum:
## The problem

The problem is described in issue [#1516](https://github.com/YunoHost/issues/issues/1516).

## Solution

Move components config from the main metronome.cfg.lua file to each vhost config file.

## PR Status

I've tested by myself and I think it's ready to be merged.

## How to test

**Not on production server**

With [ynh-dev](https://github.com/YunoHost/ynh-dev) and use-git, checkout my branch.
```
./ynh-dev use-git yunohost
cd /ynh-dev/yunohost
git remote add pitchum https://github.com/pitchum/yunohost
git checkout pitchum/fix-1516-separate-xmpp-vhosts -b fix-1516-separate-xmpp-vhosts
```
Add at least one secondary domain with LE certificate.
```
yunohost domain add somedomain.net
yunohost domain cert-install somedomain.net
```
Then create a user on this domain (option *-m* is important here):
```
yunohost user create romeo -f Romeo -l Capulet -p xxxxx -m romeo@somedomain.net
```

Then connect to this new XMPP account using Gajim or Dino or whichever good XMPP client you want and try to create chatrooms on muc.somedomain.net, upload files, etc.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
As already discussed, we currently have no notion of subdomains in YunoHost or being able to enable/disable mail or xmpp explicitly for some domain, though that will come in the future ... Still, that's troublesome in the sense that we gonna generate this config for every domain, and many people have instances with 10+ domains because they host some apps on specific domains and have no need for xmpp features ... This is already an issue I have with the current DNS diagnosis system ...

Anyway, wondering if you would be okay if we enable this only for "parent domains" (or the other way around, we exclude subdomains) ?

What I mean by parent / subdomain in this context is : 
- foo.com is apriori a parent domain ...
- bar.foo.com is considered a subdomain (but only if foo.com is registered on yunohost)
- (consequently, foo.nohost.me is considered as a parent domain except if you added nohost.me which is unlikely)

The test is pretty easy to write as : 
```
for domain in all_domains:
    parent_domain = domain.split(".", 1)[1]
    is_subdomain = parent_domain in all_domains
```

and we could imagine having a `yunohost domain list --exclude-subdomains` to only have "root domains" (or whatever you call these)

pitchum:
> and we could imagine having a `yunohost domain list --exclude-subdomains` to only have "root domains" (or whatever you call these)

Challenge accepted! :)

src/yunohost/certificate.py
```diff
@@ -639,15 +639,13 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
...
-            csr.add_extensions([crypto.X509Extension("subjectAltName", False, "DNS:" + subdomain)])
-        except YunohostError:
-            logger.warning(m18n.n('certmanager_warning_subdomain_dns_record', subdomain=subdomain, domain=domain))
+    # Include xmpp-upload subdomain in subject alternate names
+    subdomain="xmpp-upload." + domain
+    try:
+        _dns_ip_match_public_ip(get_public_ip(), subdomain)
+        csr.add_extensions([crypto.X509Extension("subjectAltName", False, "DNS:" + subdomain)])
+    except YunohostError:
+        logger.warning(m18n.n('certmanager_warning_subdomain_dns_record', subdomain=subdomain, domain=domain))
```
alexAubin, line 648:
```suggestion
    from yunohost.domain import domain_list
    # For "parent" domains, include xmpp-upload subdomain in subject alternate names
    if domain in domain_list(exclude_subdomains=True)["domains"]:
        subdomain="xmpp-upload." + domain
        try:
            _dns_ip_match_public_ip(get_public_ip(), subdomain)
            csr.add_extensions([crypto.X509Extension("subjectAltName", False, "DNS:" + subdomain)])
        except YunohostError:
            logger.warning(m18n.n('certmanager_warning_subdomain_dns_record', subdomain=subdomain, domain=domain))
```

:warning: not tested

pitchum:
Thanks for your vigilance @alexAubin, I was so happy to have implemented the change for metronome's config that I forgot to take care of the certificate generation :)
I've committed your suggested change and tested it. It's OK.

src/yunohost/domain.py
```diff
@@ -41,24 +41,29 @@
...
 
-def domain_list():
+def domain_list(exclude_subdomains=False):
     """
     List domains
 
     Keyword argument:
         filter -- LDAP filter used to search
         offset -- Starting number for domain fetching
         limit -- Maximum number of domain fetched
```
kay0u, line 51: Unrelated, but these arguments have been removed here: https://github.com/YunoHost/yunohost/commit/56513bbdb121aa654cff9bbb28241addbf0fb7ab YunoHost 2.7.1, we should consider removing them from the comments :-)

PR #917, opened 2020-04-05
Get rid of domain-specific acme-challenge snippet, use a single snippet included in every conf

alexAubin:
## The problem

c.f. https://forum.yunohost.org/t/audit-nginx-avec-gixy/10797

The current acme-challenge location should in fact end with `/` according to the current code ... but it wasn't the case in the past. And for some reason the code didn't overwrite the file if it was already there (and it's not called during renew either)... Anyway 

## Solution

The code dates back from my very early days in Yunohost where I wasn't aware of the regenconf system. Instead, we should just include a common snippet and include it in all nginx confs.

Though when doing so, we have to take care of legacy (sigh) and make sure the nginx conf is indeed really updated before getting rid of the old file.

## PR Status

Yolocommited, not tested

## How to test

Setup a domain with letsencrypt (or maybe just manually add the old 000-acmechallenge.conf, no need to actually fetch the cert) and run the new regen-conf, check that the old 000-acmechallenge.conf does get removed only if the domain conf is correctly updated.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/acme-challenge.conf.inc
+++ b/data/templates/nginx/plain/acme-challenge.conf.inc
@@ -1,0 +1,5 @@
+location ^~ '/.well-known/acme-challenge/'
+{
+        default_type "text/plain";
+        alias /tmp/acme-challenge-public/;
+}
--- a/data/hooks/conf_regen/15-nginx
+++ b/data/hooks/conf_regen/15-nginx
@@ -110,6 +110,21 @@ do_post_regen() {
     mkdir -p "/etc/nginx/conf.d/${domain}.d"
   done
 
+  # Get rid of legacy lets encrypt snippets
+  for domain in $domain_list; do
+      # If the legacy letsencrypt / acme-challenge domain-specific snippet is still there
+      if [ -e /etc/nginx/conf.d/${domain}.d/000-acmechallenge.conf ]
+      then
+          # And if we're effectively including the new domain-independant snippet now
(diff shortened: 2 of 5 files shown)


data/templates/nginx/plain/acme-challenge.conf.inc
```diff
@@ -0,0 +1,5 @@
+location ^~ '/.well-known/acme-challenge/'
```
kay0u, line 1: When is this file copied to the right place during the installation process?

alexAubin: https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/conf_regen/15-nginx#L24
called by https://github.com/YunoHost/yunohost/blob/unstable/data/hooks/conf_regen/15-nginx#L11
itself called by https://github.com/YunoHost/yunohost/blob/unstable/debian/postinst#L12

(though maybe the tests disagree :D )

alexAubin: (Uh I linked that from unstable instead of stretch-unstable for some reason ... but shouldnt have changed since)

alexAubin: The postinstall includes a regen-conf for nginx though so it should all be consistent if you at least re-run the postinstall ?

kay0u: Yeah, I just saw that! All right then!

PR #920, opened 2020-04-05
Semantic improvements for PHP helpers PR (in particular RAM helper)

alexAubin:
Here's me being again obsessed with semantics ¯\\\_\(ツ)\_/¯

In particular proposing to:
- split `ynh_check_ram` into `ynh_get_ram` (to get RAM infos) and `ynh_require_ram` (which uses `ynh_get_ram` to check that a specific RAM info is larger than a requirement)
- make it explicit wether we're interested in the total amount of RAM, or the free amount of RAM ... both not to be mistaken with the "available" amount of RAM which to me is a third different info (c.f. `free -h`)

### DIFF
--- a/data/helpers.d/hardware
+++ b/data/helpers.d/hardware
@@ -10,16 +10,16 @@
 ynh_get_ram () {
 	# Declare an array to define the options of this helper.
 	declare -Ar args_array=( [f]=free [t]=total [s]=ignore_swap [o]=only_swap )
-    local free
-    local total
+	local free
+	local total
 	local ignore_swap
 	local only_swap
 	# Manage arguments with getopts
 	ynh_handle_getopts_args "$@"
 	ignore_swap=${ignore_swap:-0}
 	only_swap=${only_swap:-0}
-    free=${free:-0}
-    total=${total:-0}
+	free=${free:-0}
+	total=${total:-0}
 
 	local total_ram=$(vmstat --stats --unit M | grep "total memory" | awk '{print $1}')
 	local total_swap=$(vmstat --stats --unit M | grep "total swap" | awk '{print $1}')
@@ -43,9 +43,9 @@ ynh_get_ram () {
 			# Use only the amount of free swap
 			ram=$free_swap
 		fi
-    elif [ $total -eq 1 ]
-    then
-	    local ram=$total_ram_swap
+	elif [ $total -eq 1 ]
+	then
+		local ram=$total_ram_swap
 		if [ $ignore_swap -eq 1 ]
 		then
 			# Use only the amount of free ram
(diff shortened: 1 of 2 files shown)


data/helpers.d/hardware
```diff
@@ -53,20 +55,46 @@ ynh_check_ram () {
 			# Use only the amount of free swap
 			ram=$total_swap
 		fi
+    else
+        echo "Uhoh, you should choose --free or --total when using ynh_get_ram" >&2
```
maniackcrudelis, line 59: Something more formal maybe ?
In any case, a ynh_print_something would be better.

data/helpers.d/hardware
```diff
@@ -53,20 +55,46 @@ ynh_check_ram () {
...
+}
 
-	# If no RAM is required, return the amount of available ram.
-	else
-		echo $ram
-	fi
+# Return 0 or 1 depending if the system has a given amount of RAM+swap free or total
+#
+# usage: ynh_require_ram [--amount=RAM required in Mb] [--free|--total] [--ignore_swap|--only_swap]
+# | arg: -a, --amount - The amount to require, in Mb
```
maniackcrudelis, line 69: Talking about semantics, I prefer `required` than `amount`

data/helpers.d/hardware
```diff
@@ -53,20 +55,46 @@ ynh_check_ram () {
...
+    local free
+    local total
+	local ignore_swap
+	local only_swap
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+	amount=${amount:-0}
+    # Dunno if that's the right way to do, but that's some black magic to be able to
+    # forward the bool args to ynh_get_ram easily?
+    free=${free:+--free}
```
maniackcrudelis, line 87: It works indeed. Same as :- except that it replace the content only if the variable **is** set.

data/helpers.d/hardware
```diff
@@ -53,20 +55,46 @@ ynh_check_ram () {
...
+    # Declare an array to define the options of this helper.
+	declare -Ar args_array=( [a]=amount= [f]=free [t]=total [s]=ignore_swap [o]=only_swap )
+    local amount
+    local free
+    local total
+	local ignore_swap
+	local only_swap
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+	amount=${amount:-0}
```
maniackcrudelis, line 84: I don't think that argument should be optional, otherwise, what's the point ?

maniackcrudelis:
obsessed sounds like the right word...

But finally, what's the point of this ?

maniackcrudelis:
@alexAubin if you do agree with my corrections, let's merge.

PR #922, opened 2020-04-07
YunoHost "as a python lib" ? Clean usr/bin/yunohost and yunohost-api ... Move moulinette initialization and other stuff to src/yunohost/__init__.py 

alexAubin:
## The problem

Hmf dunno why I'm working on this as it's clearly not a priority.

Just discussed with a user today that wanted to manage 300+ user using a python script ... Ended up willing to write a script that would simply do something like 

```
from yunohost.user import create_user

for user in some_list:
    create_user(...)
```

Ended up looking at the mess in usr/bin/yunohost which hasn't been cleaned in a while ... Many options do not make sense anymore and there's a dozen of variables created to define logging levels, all with the-same-value-but-slightly-different-in-some-case ...

## Solution

Simplify this mess ... 

Regarding the possibility to write scripts, for now it looks something like : 
```python
import sys
sys.path.insert(0, "/usr/lib/moulinette/")
import yunohost
yunohost.init() # possibly set debug=True if needed, but when writing script supposedly you ain't debugging yunohost itself, and full log is still available in yunohost-cli.log and yunohost log list...

from yunohost.user import user_create

user_create(username="foo", ...)
```

## PR Status

Ideally, propagate the new `__init__` thing to unit tests as well ... 

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/bin/yunohost
+++ b/bin/yunohost
@@ -4,9 +4,9 @@
 import os
 import sys
 import argparse
+import glob
 
 import moulinette
-from moulinette.actionsmap import ActionsMap
 from moulinette.interfaces.cli import colorize, get_locale
 
 # Directory and file to be used by logging
@@ -123,6 +123,7 @@ def _retrieve_namespaces():
     extensions = [n for n in ActionsMap.get_namespaces() if n.startswith('ynh_')]
     return ['yunohost'] + extensions
 
+
 # Stupid PATH management because sometimes (e.g. some cron job) PATH is only /usr/bin:/bin ...
 default_path = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 if os.environ["PATH"] != default_path:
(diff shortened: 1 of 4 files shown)


kay0u:
ostinstall was broken >.<

Psycojoker:
If tests passes you can merge for me.

kay0u:
> Propagate changes from moulinette/simplify-interface-init

Reminders: if you want to use a specific branch of moulinette or SSOwat during the CI, use the same name of the the current YunoHost branch (here you could have named moulinette branch `clean-usr-bin-yunohost`)... Or we can wait until moulinette/simplify-interface-init to get merged as well :-)

alexAubin:
>Reminders: if you want to use a specific branch of moulinette or SSOwat during the CI, use the same name of the the current YunoHost branch (here you could have named moulinette branch clean-usr-bin-yunohost)... Or we can wait until moulinette/simplify-interface-init to get merged as well :-)

Indeed :+1: 

I duplicated the other branch on moulinette to have the same name than this branch and tests now pass :+1: 

I suggest to merge this in 4.0 to not interfere with 3.8.x

alexAubin:
PR depends on https://github.com/YunoHost/moulinette/pull/245

PR #923, opened 2020-04-09
[enh] Misc improvements for diagnosis

alexAubin:
## The problem

Several things to polish in diagnosis's design

####  :wrench:  General design
- [x] Add a 'yunohost diagnosis get' to get one specific raw info
- [x] Simplify the way we write/generate test items
- [x] Display full traceback when a diagnosis miserably fails
- [X] Improve wording of many messages ?
- [X] Propagate important changes in strings to other strings (thanks to Kayou :cupid: )

#### :satellite:  Internet connectivity or IPv6-related
- [x] Add details about global and local IPs
- [x] Make sure there's no AAAA record if no IPv6
- [x] Add a check about hairpinning though dunno where exactly to put it é_è
- [x] http reachability should be done in both IPv4 and IPv6
- [x] Port exposure should be done in both IPv4 and IPv6

#### :bookmark: DNS records
- [x] Only report warning for XMPP DNS records for now
- [x] Do not report an error for DNS TXT records that differs because of trailing `;`
- [x] More flexible check for SPF record (ignore ip4, ip6 because 'a mx' is enough)

#### :globe_with_meridians: Web 
- [x] Diagnose multiple domains at once in web diagnosis

#### :mailbox_with_mail: Mail 

- [x] c.f. ljf's PR to be merged

#### :bar_chart: Services, regen-conf, system resources
- [x] Sometime testing service configuration fails because supposedly "nginx" command does not exist, suspecting a PATH issue when ran from cron job ...
- [x] Tweak disk usage report (sometimes partition are like total 50 Mo and yunohost triggers an error despite the fact that there's still >50% space available on partition)
- [x] Manually modified system configuration : add a tip about regen-conf command to see diff and force regen
- [x] Find a way to handle stupid postgresql ... when we check the status we should check postgresql@9.6 (dunno the exact name). It's not an official service tho so ...

#### :video_game: Webadmin

- [x] Get rid of stupid call to diagnosis just to get the version number, let's just have a simple "yunohost tools versions" for this
- [x] Improve formatting / display of DNS records errors , and other HTML tags to improve readability
- [X] Refresh / rerun diagnosis when opening diagnosis page
- [X] (c.f. previous item) Review cache duration value compared to cost of running the test
- [ ] Make it more obvious that categories panel that are collapsed can be expanded when clicking on them 

##### :soon: Postponed to another PR / another iteration someday™

- Maybe (or maybe not) we should trigger a rediagnosis of dnsrecords, webexposure when adding a domain ...
- Use the results of the diagnosis dnsrecords + web exposure to know if a domain is ready for ACME... (possibly we can also display the current state on domain page)
- Maybe (or maybe not) have a "SSL certificate" category in diagnosis ?
- Maybe we should rename / rework the 'security' category which currently only consist in checks about Meltdown ... Or at least make it explicit what check is made.... Otherwise people might be too confident that "my server is secure" because the diagnosis say so ... Long term plan is to implement [this](https://github.com/YunoHost/issues/issues/1438) among other things, but clearly out of scope for 3.8

## PR Status

- [x] Merge ljf's PR
- [x] Ideally test the whole thing on a fresh VPS with IPv6, using a nohost.me domain and a "custom" domain ... Validate that all results make sense before/after configuring things properly

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/hooks/diagnosis/10-ip.py
```diff
@@ -121,13 +131,12 @@ def ping(protocol, target):
...
 
-    def resolvconf_is_symlink(self):
-        return os.path.realpath("/etc/resolv.conf") == "/run/resolvconf/resolv.conf"
-
-    def resolvconf_points_to_localhost(self):
-        file_ = "/etc/resolv.conf"
-        resolvers = [r.split(" ")[1] for r in read_file(file_).split("\n") if r.startswith("nameserver")]
-        return resolvers == ["127.0.0.1"]
+    def good_resolvconf(self):
+        content = read_file(file_).strip().split("\n")
```
zamentur, line 135: Traceback (most recent call last):
  File "/usr/lib/moulinette/yunohost/diagnosis.py", line 168, in diagnosis_run
    code, report = hook_exec(path, args={"force": force}, env=None)
  File "/usr/lib/moulinette/yunohost/hook.py", line 335, in hook_exec
    returncode, returndata = _hook_exec_python(path, args, env, loggers)
  File "/usr/lib/moulinette/yunohost/hook.py", line 453, in _hook_exec_python
    ret = module.main(args, env, loggers)
  File "/usr/share/yunohost/hooks/diagnosis/10-ip.py", line 159, in main
    return IPDiagnoser(args, env, loggers).diagnose()
  File "/usr/lib/moulinette/yunohost/diagnosis.py", line 387, in diagnose
    items = list(self.run())
  File "/usr/share/yunohost/hooks/diagnosis/10-ip.py", line 44, in run
    good_resolvconf = self.good_resolvconf()
  File "/usr/share/yunohost/hooks/diagnosis/10-ip.py", line 135, in good_resolvconf
    content = read_file(file_).strip().split("\n")
NameError: global name 'file_' is not defined

alexAubin: Damn yes sorry, forgot to actually fix this ... Fixed naow!

data/hooks/diagnosis/21-web.py
```diff
@@ -43,19 +53,39 @@ def run(self):
...
+        # hairpinning issues, the server may be correctly exposed on the
+        # outside, but from the outside, it will be as if the port forwarding
+        # was not configured... Hence, calling for example
+        # "curl --head the.global.ip" will simply timeout...
+        if at_least_one_domain_ok:
+            ipv4 = Diagnoser.get_cached_report_item("ip", {"test": "ipv4"})
+            global_ipv4 = ipv4.get("data", {}).get("global", {})
+            if global_ipv4:
+                try:
+                    requests.head("http://" + ipv4, timeout=5)
```
zamentur, line 78: I am Groot
trsnalation: ipv4 is not a string here it's a dict, so we should use global_ipv4 instead of ipv4 ?

data/hooks/diagnosis/50-systemresources.py
```diff
@@ -12,22 +12,24 @@ class SystemResourcesDiagnoser(Diagnoser):
 
     def run(self):
 
+        MB = 1024**2
+        GB = 1024**2
```
kay0u, line 16:
```suggestion
        GB = 1024**3
```

alexAubin:
So uh that was a lot of commit ... I think I'm pretty much done for this iteration and ljf was almost done too with his PR ... Hopefully we can merge and test everything tomorrow on a real VPS (it's difficult to test it in a LXC not actually exposed to the internetz)

data/hooks/diagnosis/21-web.py
```diff
@@ -4,58 +4,153 @@
...
+                results[ipversion] = r["http"]
             except Exception as e:
                 raise YunohostError("diagnosis_http_could_not_diagnose", error=e)
 
-            if r["status"] == "ok":
+        for domain in domains:
+
+            # If both IPv4 and IPv6 (if applicable) are good
+            if all(results[ipversion][domain]["status"] == "ok" for ipversion in ipversions):
+                if 4 in ipversions:
```
kay0u, line 105: Why only with ipv4?

data/hooks/diagnosis/21-web.py
```diff
@@ -4,58 +4,153 @@
...
+                    self.do_hairpinning_test = True
                 yield dict(meta={"domain": domain},
                            status="SUCCESS",
-                           summary=("diagnosis_http_ok", {"domain": domain}))
-            else:
-                detail = r["code"].replace("error_http_check", "diagnosis_http") if "code" in r else "diagnosis_http_unknown_error"
+                           summary="diagnosis_http_ok")
+            # If both IPv4 and IPv6 (if applicable) are failed
+            elif all(results[ipversion][domain]["status"] != "ok" for ipversion in ipversions):
+                detail = results[4 if 4 in ipversions else 6][domain]["status"]
```
kay0u, line 112: Why not both ipversions here?

data/hooks/conf_regen/43-dnsmasq
```diff
@@ -50,6 +50,21 @@ do_pre_regen() {
 do_post_regen() {
   regen_conf_files=$1
 
+  # Fuck it, those domain/search entries from dhclient are usually annoying
+  # lying shit from the ISP trying to MiTM
+  if grep -q -E "^ *(domain|search)" /run/resolvconf/resolv.conf
```
kay0u, line 55: Whoukay... I didn't understand what I read :-/

src/yunohost/domain.py
```diff
@@ -448,21 +448,16 @@ def _build_dns_conf(domain, ttl=3600):
...
-    if ipv4:
-        spf_record += ' ip4:{ip4}'.format(ip4=ipv4)
-    if ipv6:
-        spf_record += ' ip6:{ip6}'.format(ip6=ipv6)
-    spf_record += ' -all"'
-
     mail = [
         ["@", ttl, "MX", "10 %s." % domain],
-        ["@", ttl, "TXT", spf_record],
+        ["@", ttl, "TXT", "v=spf1 a mx -all"],
```
kay0u, line 460: Why this change?

data/hooks/diagnosis/21-web.py
alexAubin: Hm I dunno it's mostly because later I do the hairpinning test using the global IPv4 ... I guess we could do it using the IPv6 too but hmpf :|

data/hooks/diagnosis/21-web.py
alexAubin: Because I was thinking that it's probably the same error for both ip versions and we don't want to display the message twice ... But yea that's a bit weird :s

data/hooks/conf_regen/43-dnsmasq
alexAubin: Meh yea I was probably a bit brutal here after being annoyed so many times with this ... The thing is that /etc/resolv.conf can contain `search` lines, for example `search foo`. Such that each time you query a domain, it will also test if `youdomain.foo` exists ... And typically this is set via the DHCPclient by shitty ISPs (for existence at home I have `search numericable`...) to lie about domains that are non-existent and I regularly end up with certificate warning when trying to reach a website because numericable is answering ... (And sometime even when contacting real existing domains because of idk temporary DNS resolution issue). It also happens on LXC/LXD which injects `search lxd` (not sure why, maybe to more easily contact other LXC...)

And it also messed up the check of the content of this file in the diagnosis ... We could also just ignore the corresponding lines but hmpf I just find this completely stupid and annoying so the following lines enforce that there's no `search` thing in the file)

src/yunohost/domain.py
alexAubin: ljf explained to me that in fact, having `ip4:11.22.33.44` or `ip6:11:22:33..`  is in fact redudant with the `a` and `mx` in this record. `a` and `mx` meaning is to declare the IP being A, AAAA and MX records as legitimate to send emails. So it's redundant in the context of our recommendation to re-add `ip4` etc ...

 (And as a side note it also makes it possibly more difficult for people with dynamic IP if the record depends on the IP, they have to update another record instead of just the A / AAAArecord)

You could however, in more complex setup, have other machines that you want to allow to send email (typically if you have a Gandi / OVH / ... whatever mailbox on the same domain) and in such case you might end up wiling to add the IP of those extra machines with `ip4:` and `ip6:`. Which is why later in diagnosis I ignore the parts about `ip4` and `ip6`, also to not raise a warning/error for people that configured with the old recommendation.

PR #926, opened 2020-04-12
Add default php in backup

kay0u:
## The problem

`YNH_DEFAULT_PHP_VERSION` is not defined

## Solution

Add `YNH_DEFAULT_PHP_VERSION` env in backup.

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -2,7 +2,7 @@
 
 # Declare the actual php version to use.
 # A packager willing to use another version of php can override the variable into its _common.sh.
-YNH_DEFAULT_PHP_VERSION=${YNH_DEFAULT_PHP_VERSION:-7.0}
+YNH_DEFAULT_PHP_VERSION=7.0
 YNH_PHP_VERSION=${YNH_PHP_VERSION:-YNH_DEFAULT_PHP_VERSION}
 
 # Create a dedicated php-fpm config
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -2,7 +2,7 @@
 
 # Declare the actual php version to use.
 # A packager willing to use another version of php can override the variable into its _common.sh.
-if [ -n "$YNH_DEFAULT_PHP_VERSION" ]; then
+if [ -n "${YNH_DEFAULT_PHP_VERSION:-}" ]; then
 	YNH_PHP_VERSION=${YNH_PHP_VERSION:-YNH_DEFAULT_PHP_VERSION}
 fi
 
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -2,7 +2,7 @@
 
 # Declare the actual php version to use.
 # A packager willing to use another version of php can override the variable into its _common.sh.
-YNH_PHP_VERSION=${YNH_PHP_VERSION:-$YNH_DEFAULT_PHP_VERSION}
+YNH_PHP_VERSION=${YNH_PHP_VERSION:-${YNH_DEFAULT_PHP_VERSION:-7.0}}
 
 # Create a dedicated php-fpm config
 #
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -1,8 +1,8 @@
 #!/bin/bash
 
+YNH_DEFAULT_PHP_VERSION=7.0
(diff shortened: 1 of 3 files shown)


data/helpers.d/php
```diff
@@ -2,7 +2,7 @@
 
 # Declare the actual php version to use.
 # A packager willing to use another version of php can override the variable into its _common.sh.
-YNH_PHP_VERSION=${YNH_PHP_VERSION:-$YNH_DEFAULT_PHP_VERSION}
+YNH_PHP_VERSION=${YNH_PHP_VERSION:-${YNH_DEFAULT_PHP_VERSION:-7.0}}
```
kay0u, line 5: Another solution:

```suggestion
if [ -n "$YNH_DEFAULT_PHP_VERSION" ]; then
	YNH_PHP_VERSION=${YNH_PHP_VERSION:-YNH_DEFAULT_PHP_VERSION}
fi
```

alexAubin, line 5: Uh but is this change still needed despite the fix in backup.py ?

kay0u: Yeah --'

https://github.com/YunoHost/yunohost/blob/b0cd37aecad25bacd74c765101473d8ca8150d7d/data/hooks/backup/05-conf_ldap#L7

This line fails the tests, because there is no `YNH_DEFAULT_PHP_VERSION` defined.
We could move `set -eu` right after sourcing helpers, but I think it would be better if no errors are thrown when we sourcing all our helpers.

alexAubin: Uh indeed ... Or we could inject the var in hook_exec() directly ... dunno if that's a better design or not ... nvm let's just keep your code :+1:

kay0u:
I don't know, I think my suggestion is better than what I commit...

data/helpers.d/php
maniackcrudelis: If you tried it (that syntax is sometimes touchy), I prefer your first proposition, on a single line.
Otherwise, I'd go for 2 lines, one for each variable to keep it clear and simple.

kay0u: In this case, if we redefine YNH_DEFAULT_PHP_VERSION with a new value, why add it to the core? I mean, let's take it out of app.py, and put it here?

maniackcrudelis: You think it would be easier to have only the definition of the variable here ?
I don't know what's going into the core about it, so I couldn't say.

In my case, I use a definition in this file for my test, I wasn't running on unstable. It worked.
But, then, to have another line with the definition of the variable would be better for clarity.

kay0u: I don't like the idea of having several location for the same thing... Another solution would store it into a file? `/etc/yunohost/default_php_version`

maniackcrudelis: Do we need to have the default version somewhere else than here ?

kay0u: Currently, it is not used in the core, so I don't know.... I'd say, no

maniackcrudelis: I feel like the only place we need to know that would be the apps scripts.
So for me, it's fine to have that info only in that file.

data/helpers.d/php
```diff
@@ -2,7 +2,8 @@
 
 # Declare the actual php version to use.
 # A packager willing to use another version of php can override the variable into its _common.sh.
-YNH_PHP_VERSION=${YNH_PHP_VERSION:-$YNH_DEFAULT_PHP_VERSION}
+YNH_DEFAULT_PHP_VERSION=${YNH_DEFAULT_PHP_VERSION:-7.0}
```
maniackcrudelis, line 5:
```suggestion
YNH_DEFAULT_PHP_VERSION=7.0
```
Obviously not defined since the purpose of this line is to define it.
Also, to force the definition is a good way to be sure no apps script can change that value in its _common.

kay0u: Since I'm going to remove `YNH_DEFAULT_PHP_VERSION` from the core, it seems to me that it's legit.

kay0u:
Thanks for fixing my mess!

PR #927, opened 2020-04-12
[fix] Restoration of custom hooks / missing restore hooks

kay0u:
## The problem

...

## Solution

Add `manager` argument in `BackupMethod.create`, rework `set_system_targets`

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/backup.py
+++ b/src/yunohost/backup.py
@@ -752,7 +752,7 @@ class BackupManager():
 
         for method in self.methods:
             logger.debug(m18n.n('backup_applying_method_' + method.method_name))
-            method.mount_and_backup(self)
+            method.mount_and_backup()
             logger.debug(m18n.n('backup_method_' + method.method_name + '_finished'))
 
     def _compute_backup_size(self):
@@ -851,7 +851,7 @@ class RestoreManager():
         self.info = backup_info(name, with_details=True)
         self.archive_path = self.info['path']
         self.name = name
-        self.method = BackupMethod.create(method)
+        self.method = BackupMethod.create(method, self)
         self.targets = BackupRestoreTargetsManager()
 
     #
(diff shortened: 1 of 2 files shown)


src/yunohost/backup.py
```diff
@@ -969,24 +972,25 @@ def unknown_error(part):
...
             # Otherwise, add it from the archive to the system
             # FIXME: Refactor hook_add and use it instead
-            custom_restore_hook_folder = os.path.join(CUSTOM_HOOK_FOLDER, 'restore')
-            filesystem.mkdir(custom_restore_hook_folder, 755, True)
             for hook_path in hook_paths:
                 logger.debug("Adding restoration script '%s' to the system "
                              "from the backup archive '%s'", hook_path,
                              self.archive_path)
-            shutil.copy(hook_path, custom_restore_hook_folder)
+                self.method.copy(hook_path, custom_restore_hook_folder)
```
alexAubin, line 993: I'm trying this on a dev instance ... 
- I created an archive with a custom hook (or rather : I removed the default restore hook for conf_cron)
- But then when I get to this line, I have 
   - `hook_path = "/hooks/restore/32-conf_cron"`
   - `custom_restore_hook_folder = /etc/yunohost/hooks.d/restore`
- which ends up with the "real" file created/restored being : `/etc/yunohost/hooks.d/restore/hooks/restore/32-conf_cron` ... instead of `/etc/yunohost/hooks.d/restore/32-conf_cron` 
- annnnnd I don't know how to properly fix this :| Maybe there's some trick to do somewhere to tweak the destination path somehow ...

But apart from this, that fix sounds epic o.O

kay0u: Thank you for the review, https://github.com/YunoHost/yunohost/pull/927/commits/cb5ca42df01ba61067d6d77001b596597bd05f3c should fix this issue :-)

kay0u:
I targeted `no-need-for-mysql-root-password` branch, but this fix is relevant for `stretch-unstable` I guess?

alexAubin:
Yes indeed :+1:

kay0u:
There is no ci on this branch :-( (because `no-need-for-mysql-root-password` is not up to date with `stretch-unstable`). How can I change the targeted branch, without adding change from `no-need-for-mysql-root-password`?

alexAubin:
@kay0u : not sure exactly, I would say rebase the branch or something like re-creating a fresh branch an cherry-picking commits then force-push on the existing branch ?

kay0u:
Done :-)

alexAubin:
Yolomerging to move on

PR #929, opened 2020-04-14
Allow to overwrite pm.max_children

maniackcrudelis:
## The problem

For some use case, an admin may want to force the limit for the number of max_children for php.
For example, an instance of nextcloud alone on a server with a big load.
The only way for now to do that is to overwrite the config file manually.

## Solution

Allow a config key `php_forced_max_children` into the settings.yml to overcome the limit.
This settings has as purpose to be included into a config_panel, along with other php paramaters.
I'll do it later, probably with nextcloud first.

## PR Status

Ready to be reviewed

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -402,7 +402,7 @@ ynh_get_scalable_phpfpm () {
         php_max_children=$max_proc
     fi
 
-    # Get an potential forced value for php_max_children
+    # Get a potential forced value for php_max_children
     local php_forced_max_children=$(ynh_app_setting_get --app=$app --key=php_forced_max_children)
     if [ -n "$php_forced_max_children" ]; then
         php_max_children=$php_forced_max_children
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -402,6 +402,12 @@ ynh_get_scalable_phpfpm () {
         php_max_children=$max_proc
     fi
 
+    # Get an potential forced value for php_max_children
+    local php_forced_max_children=$(ynh_app_setting_get --app=$app --key=php_forced_max_children)
+    if [ -n "$php_forced_max_children" ]; then
+        php_max_children=$php_forced_max_children
+    fi
+
     if [ "$php_pm" = "dynamic" ]
     then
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
Yep, can be merged.
I already started to implement it into config_panel on my side

data/helpers.d/php
```diff
@@ -402,6 +402,12 @@ ynh_get_scalable_phpfpm () {
         php_max_children=$max_proc
     fi
 
+    # Get an potential forced value for php_max_children
+    local php_forced_max_children=$(ynh_app_setting_get --app=$app --key=php_forced_max_children)
+    if [ -n "$php_forced_max_children" ]; then
+        php_max_children=$php_forced_max_children
```
kay0u, line 408: Btw, shouldn't we check the value (could be done in config panel) to avoid negative values. We could do:

```suggestion
        php_max_children=$(at_least_one $php_forced_max_children)
```

maniackcrudelis: This function is used because we have division before, which could end up with unexpected values.
But $php_forced_max_children is from a value into the config panel, with a big warning saying that you shouldn't probably even change the value.
But If an user willingly turn that value to a negative one... Well, it will fail... And that will be a good lesson to have been so stupid...

kay0u: Well, it's a way of seeing things :smile:

maniackcrudelis: By the way, in parallel, I've added an action to reset the php-fpm config.
So even if an user fuck up his config, it can still reset to a default one (same for nginx and systemd. And also a last one... But I keep the suspense on that last one :wink:)

kay0u: hoo can't wait! :open_mouth:

data/helpers.d/php
```diff
@@ -402,6 +402,12 @@ ynh_get_scalable_phpfpm () {
         php_max_children=$max_proc
     fi
 
+    # Get an potential forced value for php_max_children
```
JimboJoe, line 405:
```suggestion
    # Get a potential forced value for php_max_children
```

PR #934, opened 2020-04-15
Abort if up to date manifest

yalh76:
a simplification of the comments to let understand it without more explanations

I think switching between gt and lt in the comments add confusion...


Josue-T:
Thanks

PR #935, opened 2020-04-16
Remove legacy settings

Josue-T:
## The problem

Too many code to manage for old settings protected_uris, unprotected_uris, skipped_uris.

## Solution

Migrate all settings, to some permissions

## PR Status

Just tested qui unit test and it pass. It fail on the here probably because there still need some migrations.

## How to test

Checkout the branch. Try to install some app legacy app and see that some permission are made for the legacy settings.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


Aeris1One:
Will this PR remove compatibility with apps that does not migrated yet to new permission system ? (Those which always uses protected_uris, unprotected_uris, skipped_uris)

Josue-T:
> Will this PR remove compatibility with apps that does not migrated yet to new permission system ? (Those which always uses protected_uris, unprotected_uris, skipped_uris)

No we try to keep the compatibility for most of cases.

Josue-T:
Hello,

Made some tests with this following apps:
- wordpress
- nextcloud
- gitea
- etherpad_mypads
- leed

I found some regression and some bug which are now fixed. So I think it should be OK to go forward with this.

locales/en.json
```diff
@@ -440,6 +440,7 @@
     "migration_0018_failed_to_migrate_iptables_rules": "Failed to migrate legacy iptables rules to nftables: {error}",
     "migration_0018_failed_to_reset_legacy_rules": "Failed to reset legacy iptables rules: {error}",
     "migration_0019_add_new_attributes_in_ldap": "Add new attributes for permissions in LDAP database",
+    "migration_0019_migrate_old_app_settings": "Migrate old apps settings 'skipped_uris', 'unprotected_uris', 'protected_uris' in permissions system.",
     "migration_0019_backup_before_migration": "Creating a backup of LDAP database and apps settings prior to the actual migration.",
     "migration_0019_can_not_backup_before_migration": "The backup of the system could not be completed before the migration failed. Error: {error:s}",
     "migration_0019_migration_failed_trying_to_rollback": "Could not migrate… trying to roll back the system.",
```
ericgaspar, line 446:
```suggestion
    "migration_0019_migration_failed_trying_to_rollback": "Unable to migrate... trying to roll back the system.",
```

ericgaspar (changes requested):

alexAubin:
Copypasta of [this comment](https://github.com/YunoHost/yunohost/pull/861#issuecomment-693529756) : 

Follow-up of yesterday's discussion about non-trivial legacy permissions. Here's a dump of what exists in the wild (only apps declared as working) : 

https://drop.devloprog.org/r/kep83fHuY_#kPT86CnRL7Zf7eqqyvNZflfT4DHRgO2Q/GmKC71do7k=

2020-09-16-184735_1366x768_scrot

alexAubin:
Soooo following the previous small study, and as a solution to the issue of "what do with non-trivial legacy permissions" I'm proposing this commit https://github.com/YunoHost/yunohost/pull/935/commits/2b4e14cca7b27ce21a88d18de290671015149f67 that hard-code a small description of what the uri/regexes correspond to ...

I'm far from being 100% sure about the exactitude of the info though ... (in fact im ~sure about only ~50ish% or these) but at least we can discuss the principle and iterate on the names if you have some more insight

alexAubin:
Sooooo uh idk, proposing to merge this PR into permission_protection and keep the discussion going there ? It looks to me that this PR is the right way to go and we just need to clarify some implementation details about it, which can be done in the other PR (and will allow easier testing and iteration)

alexAubin:
Let's-a-go then o/

PR #939, opened 2020-04-18
Make nodejs helpers easier to use

maniackcrudelis:
## The problem

Someone make me realize recently that those helpers aren't easy to understand and use.
A quick look to the apps that actually use the helper (A lot more that I have imagined !) shows that indeed, that's a mess, approximately as many different way to implement the helper than apps that use it...
And that right as well that when it's about node, nothing is easy !

## Solution

Add aliases for node and npm and new variables to ease the usage of the helpers.
Add explanation about how to really use the variables and aliases.

## PR Status

Ready.

## How to test

Try to use it on your own node app, that's way easier indeed !

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps


tituspijean:
Tested on the PR linked above, and approved.
I had to do `ynh_exec_as $app $ynh_node_load_PATH $ynh_npm` instead of what's written in the comments.

maniackcrudelis:
Does that mean that after you removed PATH=$PATH into your ynh_exec_as, it didn't fix your problem, and you had to add $ynh_node_load_PATH ?

tituspijean:
No. See https://github.com/YunoHost-Apps/ztncui_ynh/pull/5/commits/fefa49295a0acaa71f02fd6d405b7640a239c6f0: `PATH=$PATH` in `ynh_exec_as` and your helper were never tested together.

maniackcrudelis:
To use alias was only because it felt more natural to use an alias like a real binary than a variable to actually replace an actual binary.

If I remember well, npm wanted to used node, which wasn't into PATH.

PR #946, opened 2020-04-21
[fix] Be more robust against manually modified domain conf

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1568 and other situations reported on the forum where a manually-modified conf file (either by an app or because of weird other subtle bug that are still to be identified) results in catastrophic situations : the conf file is not removed, but certs are, resulting in nginx not being able to load the configuration because it can't find the certs.

## Solution

During domain_remove(), we should enforce this conf file to be removed. This is kind of a adhoc fix as it doesnt fix the root cause of the issue but still better than instances exploding.

I added some tests to cover this (@kay0u : dunno if anything specific is needed on the CI if we add a new test file) 

Also increased the robustness of the nginx regen-conf for cases where the conf is broken or nginx fails to restart/reload

## PR Status

Tested and working on my side

## How to test

Uuuuuh run the tests :D 

Or manually : add a domain, change someting stupid in the conf file of the domain, try to remove the domain.
Before : an attempt to reload nginx will be triggered but will miserably crash because the conf didnt get removed
Now : the conf is removed and nginx can reload succesfully

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/tests/test_regenconf.py
+++ b/src/yunohost/tests/test_regenconf.py
@@ -1,0 +1,80 @@
+import glob
+import os
+import pytest
+import shutil
+import requests
+
+from conftest import message, raiseYunohostError
+
+from moulinette import m18n
+from moulinette.utils.filesystem import mkdir
+
+from yunohost.domain import _get_maindomain, domain_add, domain_remove, domain_list
+from yunohost.utils.error import YunohostError
+from yunohost.regenconf import manually_modified_files, _get_conf_hashes, _force_clear_hashes
+
+TEST_DOMAIN = "secondarydomain.test"
+TEST_DOMAIN_NGINX_CONFIG = "/etc/nginx/conf.d/secondarydomain.test.conf"
+
+def setup_function(function):
+
+    _force_clear_hashes([TEST_DOMAIN_NGINX_CONFIG])
+    clean()
+
+def teardown_function(function):
+
+    clean()
+    _force_clear_hashes([TEST_DOMAIN_NGINX_CONFIG])
+
+def clean():
+
+    assert os.system("pgrep slapd >/dev/null") == 0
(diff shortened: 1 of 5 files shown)


src/yunohost/tests/test_regenconf.py
```diff
@@ -0,0 +1,80 @@
+import glob
```
kay0u, line 1: MOAR TESTS \o/

Please could you add:

```yml
test-regenconf:
  extends: .tests
  stage: tests
  script:
    - cd src/yunohost
    - py.test tests/test_regenconf.py
```

At the end of this file?

kay0u: I can explain a little if you want:

```yml
test-regenconf:
```
Is the name, you can see the names of all the jobs here: https://gitlab.com/yunohost/yunohost/pipelines/137564286

```yml
  extends: .tests
```
Here we use the parameters from the `.tests` section. https://github.com/YunoHost/yunohost/blob/b392efdf85f3f9528772ff76dd40ad3239613bef/.gitlab-ci.yml#L5-L26

```yml
  stage: tests
```
Uuh, the stage (we have two stage for now, Postinstall and Tests) https://gitlab.com/yunohost/yunohost/pipelines/137564286

```yml
  script:
    - cd src/yunohost
    - py.test tests/test_regenconf.py
```

And finally the bash script to run during the test of this section.

alexAubin: alrightythen

Done in https://github.com/YunoHost/yunohost/pull/946/commits/817e45108631f1979e1443584713ea36a7d26643 , we'll see if CI is happy ;)

PR #948, opened 2020-04-21
Fix getopts with empty parameters

maniackcrudelis:
## The problem

Getopts "bugs" if you do something like `my_helper --param1="value1" `
Notice the trailing space at the end.
This will be interpreted as a another value for the first parameter. So you will have `param1="value1;"`
Not a big problem, but it can be, and actually it may fail apt with a "php7.0;-fpm"

## Solution

To not add a trailing space is an obvious solution, but not enough in case of optional parameter added by a script.
So, ignore empty parameters.

## PR Status

Tested and working as expected.
Should be merge as soon as possible as it breaks the current php helpers with nextcloud.

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 : 

@YunoHost/apps

### DIFF
--- a/data/helpers.d/getopts
+++ b/data/helpers.d/getopts
@@ -147,26 +147,30 @@ ynh_handle_getopts_args () {
 									break
 								fi
 							else
-								# Else, add this value to this option
-								# Each value will be separated by ';'
-								if [ -n "${!option_var}" ]
-								then
-									# If there's already another value for this option, add a ; before adding the new value
-									eval ${option_var}+="\;"
-								fi
+                                # Ignore empty parameters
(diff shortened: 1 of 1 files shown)


alexAubin:
(Dunno if that's needed in stable or testing as well but we can make a hotfix release if needed)

maniackcrudelis:
Well, that's an issue with the helper on its own, even on stable.

But it actually breaks nextcloud on testing with the last php helper.
And I expect the same with other apps in same conditions.

However, as we haven't seen that bug before, I guess it's not a big deal to just follow the common process and go to testing with it.

PR #951, opened 2020-04-22
[wip] YunoHost "as a python lib" ? Move moulinette initialization and other stuff to src/yunohost/__init__.py

alexAubin:
## The problem

Continuation of https://github.com/YunoHost/yunohost/pull/922 to : 
- try to clean the mess happening in bin/yunohost and bin/yunohost-api
- have a way for people to easily write scripts that call yunohost functions (well i don't even know if that's a good idea in terms of avoiding people to break their own server...)

## Solution

Factorize some stuff into src/yunohost/__init__.py ...hopefully improving the readability of bin/yunohost and bin/yunohost-api ... (well that huge wall of json to configure logging is still a pain but can't easily be simplified...)

Regarding the possibility to write scripts, for now it looks something like : 

```python
import sys
sys.path.insert(0, "/usr/lib/moulinette/")
import yunohost
yunohost.init_logging() # possibly set debug=True if needed, but when writing script supposedly you ain't debugging yunohost itself, and full log is still available in yunohost-cli.log and yunohost log list...
yunohost.init_i18n()

from yunohost.user import user_create

user_create(username="foo", ...)

```

## PR Status

WIP but sortof working

## How to test

Run the yunohost command with various options, same for yunohost-api ... Also test the case where /etc/yunohost/installed doesnt exist yet...

And test to manually import and run yunohost functions...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/__init__.py
+++ b/src/yunohost/__init__.py
@@ -6,8 +6,9 @@ import sys
 import glob
 
 import moulinette
+from moulinette import m18n
 from moulinette.utils.log import configure_logging
-
+from moulinette.interfaces.cli import colorize, get_locale
 
 def is_installed():
     return os.path.isfile('/etc/yunohost/installed')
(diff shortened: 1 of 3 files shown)


Psycojoker:
LGTM if tests passes :)

src/yunohost/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+        host=host,
+        port=port,
+        routes={('GET', '/installed'): is_installed_api},
+        use_websocket=True
+    )
+    sys.exit(ret)
+
+
+def extensions():
+    # This is probably not used anywhere, but the actionsmap and code can be
```
Psycojoker, line 56: It's mentionned in YEP 4.7 https://yunohost.org/#/packaging_apps_guidelines so we need to trash this YEP if we remove this (probably) unused feature

alexAubin:
Let's merge this into the other PR to try to reduce the number of PRs opened ...

PR #952, opened 2020-04-23
[fix] Stupid "KeyError: label" weird error due to improper use of logger.exception in app.py

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1549

The root cause of this issue is that we use logger.exception instead of logger.error in app.py (in particular during app installs). Logger.exception is meant to display the stacktrace of some unknown / unexpected exception ... But here it's some sort of "expected" exception where we have a ~clear idea of what we're catching and what to do. So we meant to use logger.error instead (it was me that refactored this code a few months ago and didn't know the difference)

The reason why it displayed that "KeyError" is, in my understanding, because it tries to fetch the latest exception. But since there's none when the app fails (because we only check the return code of the system call) it just use some completely unrelated exception. In fact, when there's no exception about the 'key error' thing, it still prints None ...

## Solution

Use logger.error instead of logger.exception.

## PR Status

Tested and working for me

## How to test

Try to install a broken app like zeronet, or any app install that you know will fail. Check the `tail` of the log generated.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -512,7 +512,7 @@ def app_upgrade(app=[], url=None, file=None):
             upgrade_failed = True if upgrade_retcode != 0 else False
             if upgrade_failed:
                 error = m18n.n('app_upgrade_script_failed')
-                logger.exception(m18n.n("app_upgrade_failed", app=app_instance_name, error=error))
+                logger.error(m18n.n("app_upgrade_failed", app=app_instance_name, error=error))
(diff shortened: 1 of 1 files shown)


Psycojoker:
We've discussed it on the chat and that's a minor fix that totally makes sens.

I had similar problem at $dayjob.

PR #953, opened 2020-04-23
[fix] Make sure to return '/' and not empty string for stuff on domain root

alexAubin:
## The problem

This is a small regression introduced in 3.7 when refactoring app_ssowatconf and app_map : when an app is installed on the root of a domain, `app_map` should return something like 

`domain.tld/: foo`

and not : 

`domain.tld: foo`

(notice the trailing /)

The lack of `/` break some checks like here https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/app.py#L1109 which in turns trigger https://github.com/YunoHost/issues/issues/1572 (it should not be possible to make default app an app that's already on the root ...)

(Note that the makedefault system is still heavily buggy but can't be fixed until we properly implement domain settings, which is also needed to enable/disable mail/xmpp per-domain, and we also have to think about how to handle nasty backup/restore or change-url cases...)

## Solution

Force the path var to be at least `/` if it's empty

## PR Status

Tested `app_map` but not sure about all the implications this change can have .. Let's see what the tests say.

## How to test

Install an app on root, check what `app_map` returns with/without this fix

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -239,7 +239,7 @@ def app_map(app=None, raw=False, user=None):
                 perm_domain, perm_path = perm_url.split("/", 1)
                 perm_path = "/" + perm_path.rstrip("/")
 
-            perm_path = perm_path if perm_path != "" else "/"
+            perm_path = perm_path if perm_path.strip() != "" else "/"
 
             return perm_domain, perm_path
 
@@ -1269,7 +1269,7 @@ def app_ssowatconf():
                 perm_domain, perm_path = perm_url.split("/", 1)
                 perm_path = "/" + perm_path.rstrip("/")
 
-            perm_path = perm_path if perm_path != "" else "/"
+            perm_path = perm_path if perm_path.strip() != "" else "/"
 
             return perm_domain + perm_path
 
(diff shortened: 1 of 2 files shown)


src/yunohost/app.py
```diff
@@ -239,6 +239,8 @@ def _sanitized_absolute_url(perm_url):
                 perm_domain, perm_path = perm_url.split("/", 1)
                 perm_path = "/" + perm_path.rstrip("/")
 
+            perm_path = perm_path if perm_path != "" else "/"
```
Psycojoker, line 242:
```suggestion
            perm_path = perm_path if perm_path.strip() != "" else "/"
```

Not sure it's needed but if that can prevent a bug...

src/yunohost/app.py
```diff
@@ -1267,6 +1269,8 @@ def _sanitized_absolute_url(perm_url):
...
```
Psycojoker, line 1272:
```suggestion
            perm_path = perm_path if perm_path.strip() != "" else "/"
```

Same patch.

PR #955, opened 2020-04-23
Keep track of 'parent' operation in operation loggers

alexAubin:
## The problem

So hmyeah clearly not the priority but I got tired of seeing so many different operations when running `yunohost log list` ... Following the integration of group/permissions we now have many operations which contains "sub" operations.

For example : adding a user will create an operation logger, but it will also create a "child" operation to create a group and update group all_users.
Similarly, "app_install" will create permissions and update it. (not even mentionning if the install fails and app is autoremoved, which also delete permissions etc..)

With this, we quickly end up with so many operations that it's less straightforward to find the log of the original "parent" operation you are interested in ...

Would be nice to be able to improve the CLI .. not sure how, maybe only list "root" operations by default in "yunohost log list" (i don't really know yet what to aim for exactly...) but at least be able to tweak the log list view in the webadmin to highlight "root" operations ... or display "child" operations differently ...

## Solution

Add a notion of "parent" for operations. Not really used for anything yet.

In the most simple case : 
- Operation A is started. There's no other active operation logger, so it's the root.
- Operation B is started. It finds that instance A is still active and therefore defines its parent as being A.
- Operation B is closed
- Operation A is closed
- Operation C is started. There's no other active operation logger, so it's the root.
- Operation C is closed

However there are more complex cases where we need black magic:
- Operation A is started
- A script or hook is ran (typically an app script)
- The script or hook contains `yunohost foo bar` (for example, a domain creation, or a regenconf)
- Operation B is started (so in a different process than A)
- Operation B identifies that it's not in the "root" process. It looks at which .log files are being used by parent processes until it fines the log of operation A being actively written by some parent process. Therefore it defines its parent as being A.
- Operation B is closed
- Operation A is closed

Writing this, I realize that it doesn't cover the self-upgrade system that completely operates in an unrelated process child and this won't work, but I guess we can live with it.

## PR Status

Tested and working on my side

## How to test

For example, create a user. Use "yunohost log list" to list recent operations. Look into the yaml file and you should see that stuff related to permission/group have a "parent" key set.

For the more advanced case, I used a dummy app containing a "yunohost domain add foo.com". Checked that log for the domain creation indeed gets related to the app install operation even though they run in different processes.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -28,6 +28,7 @@ import os
 import re
 import yaml
 import collections
+import glob
 
 from datetime import datetime
 from logging import FileHandler, getLogger, Formatter
@@ -334,6 +335,8 @@ class OperationLogger(object):
     This class record logs and metadata like context or start time/end time.
     """
 
+    _instances = []
+
     def __init__(self, operation, related_to=None, **kwargs):
         # TODO add a way to not save password on app installation
         self.operation = operation
@@ -344,6 +347,8 @@ class OperationLogger(object):
         self.logger = None
         self._name = None
         self.data_to_redact = []
+        self.parent = self.parent_logger()
+        self._instances.append(self)
 
         for filename in ["/etc/yunohost/mysql", "/etc/yunohost/psql"]:
             if os.path.exists(filename):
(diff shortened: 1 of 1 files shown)


src/yunohost/log.py
```diff
@@ -354,6 +359,50 @@ def __init__(self, operation, related_to=None, **kwargs):
...
+        #
+        # The strategy is :
+        # 1. list 20 most recent log files
+        # 2. iterate over the PID of parent processes
+        # 3. see if parent process has some log file open (being actively
+        # written in)
+        # 4. if among those file, there's an operation log file, we use the id
+        # of the most recent file
+
+        recent_operation_logs = sorted(glob.iglob("/var/log/yunohost/categories/operation/*.log"), key=os.path.getctime, reverse=True)[:20]
```
alexAubin, line 387:
```suggestion
        recent_operation_logs = sorted(glob.iglob(OPERATIONS_PATH + "*.log"), key=os.path.getctime, reverse=True)[:20]
```

src/yunohost/log.py
```diff
@@ -354,6 +359,50 @@ def __init__(self, operation, related_to=None, **kwargs):
...
+
+        import psutil
```
Psycojoker, line 389: Should be at the top level of the file '^'

src/yunohost/log.py
```diff
@@ -475,8 +525,10 @@ def close(self, error=None):
         self.ended_at = datetime.utcnow()
         self._error = error
         self._success = error is None
+
         if self.logger is not None:
             self.logger.removeHandler(self.file_handler)
+            self.file_handler.close()
```
Psycojoker, line 531: Is it like an unrelated fix? Or to prevent from appearing as a parent while things are finished?

alexAubin: It's a related fix. Without this, the current code doesnt close the file even after `removeHandler()` and the strategy of looking at `proc.open_files()` doesnt work (well, it works but could catch the wrong file)

src/yunohost/log.py
alexAubin: Uh yeah these days I became obsessed with lazyloading everything but it seems pretty fast to load on RPi so I'll put it on top x)

alexAubin:
Tests are legitimately not happy because no lock is set during tests and the strategy is defeated ... I need to think about what to do with this because that would also affect cases where yunohost is used as a lib. We should either enforce acquiring the lock, or find another strategy...

PR #959, opened 2020-04-26
Add a sleep 2 before systemctl

yalh76:
## The problem
https://github.com/YunoHost/issues/issues/1573

## Solution
Add a sleep 2 before systemctl 

## PR Status
Ready and tested

## How to test
Do recurring backup of an app

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/systemd
+++ b/data/helpers.d/systemd
@@ -121,6 +121,8 @@ ynh_systemd_action() {
         action="reload-or-restart"
     fi
 
+    # Workaround for https://github.com/YunoHost/issues/issues/1573
+    sleep 2
     systemctl $action $service_name \
         || ( journalctl --no-pager --lines=$length -u $service_name >&2 \
         ; test -e "$log_path" && echo "--" >&2 && tail --lines=$length "$log_path" >&2 \


alexAubin:
Hmmmmm shall we increase it to something like 5 to cover slower hardware ?

maniackcrudelis:
Huh, we have a lot of ynh_systemd_action in app scripts and helpers. A sleep 5 will soon become a sleep 20 or 30

alexAubin:
Ah yeah, didn't think about that, nevermind xD

maniackcrudelis:
Also, it's really a very specific problem that we didn't really spot...
ynh_systemd_action start to follow the log before doing any action, but somehow, sometime it missed the first line !

yalh76:
Maybe for now a `sleep 2` could be a good start ?

alexAubin:
Do you folks want this as a hotfix for 3.7.x as well ?

maniackcrudelis:
Really not a hot fix.
And actually wondering if really worth it at all

PR #960, opened 2020-04-26
Try to show smarter / more useful logs by filtering irrelevant lines like set +x etc

alexAubin:
## The problem

This is again me procrastinating instead of fixing bugs for 3.8 lol

Anyway, just stumbled on some debug logs that was showing a lot of "set +x" or "Helper used in legacy mode" or "local foo" which ain't so helpful (c.f. also the screenshot in https://github.com/YunoHost/yunohost-admin/pull/294 ) . It's not unusual to see half the lines of the 20ish last lines of some debug log filled with these

## Solution

 So in the same vein as [this PR](https://github.com/YunoHost/yunohost-admin/pull/294), this is an attempt to increase the likelyhood that users would be able to identify the issue themselves by removing some lines which we know aren't gonna be relevant. This is also meant to help more advanced users / app packagers because this hack is also applied to debug info show during app script failures.

The goal would be to apply this filter when viewing a log in the webadmin. Though it would *not* be applied if the user clicks "show more lines" of share the logs on yunopaste. (Though depending on how 100% confident that these lines ain't useful we could also enable it all the time but idk ...)

## PR Status

Tested and working on my side

## How to test

Install an app that you know will fail, check `yunohost log display` with/without -f

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -35,7 +35,7 @@ ynh_exit_properly () {
         ynh_clean_setup	# Call the function to do specific cleaning for the app.
     fi
 
-    ynh_die	# Exit with error status
+    exit 1  # Exit with error status
 }
 
 # Exits if an error occurs during the execution of the script.
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -35,7 +35,10 @@ ynh_exit_properly () {
         ynh_clean_setup	# Call the function to do specific cleaning for the app.
     fi
 
-    exit 1  # Exit with error status
+    # Exit with error status
+    # We don't call ynh_die basically to avoid unecessary 10-ish
+    # debug lines about parsing args and stuff just to exit 1..
+    exit 1
 }
 
 # Exits if an error occurs during the execution of the script.
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -916,6 +916,7 @@ def dump_app_log_extract_for_debugging(operation_logger):
 
     filters = [
         r"set [+-]x$",
+        r"set [+-]o xtrace$",
         r"local \w+$",
         r"local legacy_args=.*$",
         r".*Helper used in legacy mode.*",
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -920,7 +920,7 @@ def dump_app_log_extract_for_debugging(operation_logger):
         r"local legacy_args=.*$",
         r".*Helper used in legacy mode.*",
         r"args_array=.*$",
-        r"declare -Ar args_array$",
+        r"local -A args_array$",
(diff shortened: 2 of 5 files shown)


alexAubin:
Here's a before/after comparison
Capture du 2020-04-26 20-57-28

src/yunohost/app.py
```diff
@@ -911,6 +911,19 @@ def dump_app_log_extract_for_debugging(operation_logger):
     with open(operation_logger.log_path, "r") as f:
         lines = f.readlines()
 
+    filters = [
+        r"set [+-]x$",
+        r"local \w+$",
+        r"local legacy_args=.*$",
+        r".*Helper used in legacy mode.*",
+        r"args_array=.*$",
+        r"declare -Ar args_array$",
```
maniackcrudelis, line 920: `declare -Ar` is about to be replaced by `local -A`

alexAubin:
Added a small commit that replaces `ynh_die` by just `exit 1` in `ynh_exit_properly` 

..reason is because that last `ynh_die` leads to : 

```
2020-04-28 10:50:47,189: WARNING - 
2020-04-28 10:50:47,190: DEBUG - + ynh_die
2020-04-28 10:50:47,190: DEBUG - + local legacy_args=mc
2020-04-28 10:50:47,190: DEBUG - + args_array=([m]=message= [c]=ret_code=)
2020-04-28 10:50:47,190: DEBUG - + local -A args_array
2020-04-28 10:50:47,190: DEBUG - + local message
2020-04-28 10:50:47,190: DEBUG - + local ret_code
2020-04-28 10:50:47,191: DEBUG - + ynh_handle_getopts_args
2020-04-28 10:50:47,191: DEBUG - + set +o xtrace
2020-04-28 10:50:47,191: DEBUG - + ret_code=1
2020-04-28 10:50:47,191: DEBUG - + echo ''
2020-04-28 10:50:47,191: DEBUG - + exit 1
```

(note the empty warning message why make it sounds like it comes from something buggy ... but nope, just an `echo "" >&2`)

So basically a huge pile of stuff just to `exit 1` ...

data/helpers.d/utils
```diff
@@ -35,7 +35,7 @@ ynh_exit_properly () {
         ynh_clean_setup	# Call the function to do specific cleaning for the app.
     fi
 
-    ynh_die	# Exit with error status
+    exit 1  # Exit with error status
```
maniackcrudelis, line 38: :grin: ...

A comment :grin: ?

Honestly, I'm quite sure sooner or later we're going to revert to ynh_die if we don't have a comment to explain it was actually made on purpose.

PR #961, opened 2020-04-27
[enh] Real CSP headers for the webadmin

alexAubin:
## The problem

Angry red CSP errors in JS console on the webadmin (when accessed via a domain) because of websocket

## Solution

~~Add websocket to default-src~~

~~Dunno the implication but considering we're in Report-Only anyway, that just removes the error ...~~

Well, I ended up digging into implementing an actual CSP ... Though I might have forgotten something ... c.f. https://github.com/YunoHost/yunohost/pull/961#issuecomment-620903303

## PR Status

Tested and working on my side

## How to test

Go to the webadmin and open JS console

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/nginx/plain/yunohost_admin.conf.inc
+++ b/data/templates/nginx/plain/yunohost_admin.conf.inc
@@ -6,6 +6,9 @@ location /yunohost/admin/ {
     default_type text/html;
     index index.html;
 
+    more_set_headers "Content-Security-Policy: upgrade-insecure-requests; default-src 'self'; connect-src 'self' https://raw.githubusercontent.com https://paste.yunohost.org wss://$host; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval'; object-src 'none';";
+    more_set_headers "Content-Security-Policy-Report-Only:";
+
     # Short cache on handlebars templates
     location ~* \.(?:ms)$ {
       expires 5m;
--- a/data/templates/nginx/security.conf.inc
+++ b/data/templates/nginx/security.conf.inc
@@ -22,7 +22,7 @@ ssl_prefer_server_ciphers off;
 # https://wiki.mozilla.org/Security/Guidelines/Web_Security
 # https://observatory.mozilla.org/
 more_set_headers "Content-Security-Policy : upgrade-insecure-requests";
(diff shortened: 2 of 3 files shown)


zamentur:
IMHO this kind of CSP settings could be configured apps by apps.
On https://infosec.mozilla.org/guidelines/web_security#content-security-policy i read:

> Care needs to be taken with data: URIs, as these are unsafe inside script-src and object-src (or inherited from default-src).

If we need this for the yunohost-admin, we could put a meta tag instead:

`<meta http-equiv="Content-Security-Policy" content="default-src https: data: wss:">`

zamentur:
merge in 3 days

alexAubin:
Discussion made me want to dig a bit more ...

Considering the webadmin is a very specific piece of software that we know, we can reduce the scope of the CSP to exactly what we need ... Though of course that's a bit more risky than just in report-only mode.

So here's what I ended up with : 

- `upgrade-insecure-requests; ` : Probably unnecessary because we enforce https by other means but whatever that doesn't hurt ...
- `default-src 'self'; ` : This is the fallback for every other rule that ain't explicitly defined. We only allow 'self' (I believe it should be equivalent to `https//$host` ?)
- `connect-src 'self' https://raw.githubusercontent.com https://paste.yunohost.org wss://$host;` : This is the stuff we gonna interact with ... : 
    - the API at https://domain.tld/yunohost/api (that's covered by `self`, idk if we could reduce it to only a few specific paths)
    - github raw content is needed because of the custom app install mechanism ... Though we could add something in yunohost's api such that it's the backend that feches the manifest (c.f. also the will to make the mechanism independant from github)
    - paste is needed because of one remaining page (I believe 'services' ?) that paste log using ajax ... in the mid-term that should be done on the backend as well like we do for regular log sharing (done with `yunohost log display --share`) that also allow anonymization now
- `style-src 'self' 'unsafe-inline'; - we could get rid of this if we remove any inline `style="..."` in our html (I'm a big culprit)
- `script-src 'self' 'unsafe-eval';` - there's a piece of sammy js that needs eval at some point ~.~
- `object-src 'none';` - we don't want no java/flash

PR #962, opened 2020-04-28
[enh] c.f. issue 1365, asking a mail address during user creation is confusing ux, people try enter their existing mail addresses

ekhae:
## The problem
Users tend to create mails with other domains than yunohost domain, that was generating an error.
...

## Solution
Only request a user name for email, that is assignated with a yunohost domain.
...

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -112,7 +112,7 @@ user:
                     help: Domain for the email address and xmpp account
                     extra:
                         pattern: &pattern_domain
-                            - !!str ^([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+([^\W\d_]{2,})$
+                            - !!str ^([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+((xn--)?[^\W\d_]{2,})$
                             - "pattern_domain"
                 -q:
                     full: --mailbox-quota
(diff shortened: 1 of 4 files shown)


Aeris1One:
Principle agreement

alexAubin:
I forced-push the branch with a cherry-pick because this was initially developed on top of buster-unstable and the history was a bit messy

Still needs to make a proper review of this, there will probably be a few tests to fix

zamentur:
I have done the fix for chinese domain here, as i see it, and to avoid to resolve conflict later.

alexAubin:
(Actually, there are probably some fixes to be done in the unit tests for this ... and this is also pretty likely to break some CI and other automatic deployment scripts)

data/actionsmap/yunohost.yml
```diff
@@ -97,15 +97,6 @@ user:
                         pattern: &pattern_lastname
                             - !!str ^([^\W\d_]{2,30}[ ,.'-]{0,3})+$
                             - "pattern_lastname"
-                -m:
-                    full: --mail
```
alexAubin, line 101: Too many apps and other misc script using that option in the wild, dropping it will make many things explode, so instead I'll implement a legacy behavior/deprecation...

PR #968, opened 2020-05-02
Simplify / optimize reading version of yunohost packages...

alexAubin:
## The problem

On my RPi2, running `yunohost --version` takes a goddamn crazy four seconds, two of them due to the code loading `apt.Cache()` which takes two seconds just by itself.

Also the code of `get_installed_version` is crazy complicated for what it achieves ...

## Solution

Simplify and speed up everything by reading first line of changelog stored in `/usr/share/doc/` ... that's a trick that's only reliable for yunohost packages (other may have debian.changelog.gz instead) but we only use these functions for this so ...

Tested on the RPi, the operation of reading this for the 4 packages takes 0.2s (instead of 2s)

## PR Status

Tested and working on my side

## How to test

Run `yunohost --version`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/utils/packages.py
+++ b/src/yunohost/utils/packages.py
@@ -63,7 +63,7 @@ def meets_version_specifier(pkg_name, specifier):
     # context
     assert pkg_name in YUNOHOST_PACKAGES
     pkg_version = get_ynh_package_version(pkg_name)["version"]
-    pkg_version = pkg_version.split("~")[0]
+    pkg_version = re.split(r'\~|\+|\-', pkg_version)[0]
     pkg_version = version.parse(pkg_version)
 
     # Extract operator and version specifier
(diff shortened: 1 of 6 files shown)


src/yunohost/utils/packages.py
```diff
@@ -21,15 +21,12 @@
 import re
 import os
 import logging
-from collections import OrderedDict
 
-import apt
-from apt_pkg import version_compare
```
alexAubin, line 27: aaaaannnnnd right after pressing the button I realize `version_compare` is still used in the damn Specifier / SpecifierSet which is also completly overkilled (iirc we only use it to check the version requirement in app manifest ... and it's always something like `"< x.y.z"` and we could just do this using any reasonable lightweight lib or whatever ... like the one Josue uses in his PR ...

alexAubin:
I haz le green tests :eyes:

PR #972, opened 2020-05-03
Add a timeout to wget in helpers

alexAubin:
## The problem

c.f. https://forum.yunohost.org/t/yunohost-command-not-ending/11177/7

Upgrading an app resulted in a never-ending command blocking all other yunohost commands

## Solution

`wget` was hanging indefinitely .. not sure exactly why but it's pretty clear that the issue is a lack of timeout. `wget` only has a read-timeout of 900 seconds enabled (meaning : if nothing received during 900 sec once downloaded started, command times out) but no timeout on dns queries and opening TCP connection except what system libs implement. (c.f. `man wget`)

We can add `--timeout 900` to set 900 sec timeout to all these operations.

N.B. : this doesn't mean that the overall operation can't take more than 900 secs, just that if absolutely nothing happens during 900 secs, the command exits.

## PR Status

Yolocommited, not tested ...

## How to test

Well I don't know how to mock a wget hanging forever ... But do we really need a test :s 

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/apt
+++ b/data/helpers.d/apt
@@ -436,7 +436,8 @@ ynh_install_extra_repo () {
     if [ -n "$key" ]
     then
         mkdir --parents "/etc/apt/trusted.gpg.d"
-        wget --quiet "$key" --output-document=- | gpg --dearmor | $wget_append /etc/apt/trusted.gpg.d/$name.gpg > /dev/null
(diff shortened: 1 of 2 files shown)


Psycojoker:
Yolo mergin'

PR #977, opened 2020-05-05
[helper] App users ldap

Josue-T:
## The problem

- App can't authenticate in LDAP for extended search request. Some app need this. By example mastodon
- App can't cleanly send and receive email.

## Solution

- Create a user in LDAP in a specific place

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


zamentur:
Aleks says me yesterday that it may be possible to change dovecot config to allow system user to send mail and may be to have some inbox.

Are we sure we want to create ldap users for this ?

alexAubin:
(c.f. https://github.com/YunoHost/yunohost/pull/815 )

Gredin67:
up @zamentur @alexAubin see https://github.com/YunoHost-Apps/synapse_ynh/issues/362

Gredin67:
@tituspijean maybe?

Josue-T:
> Superseded by #815 :|

Well not completly because, by example synapse need a LDAP user to authenticate to ldap to use ldap filters. So no it will not completly fix the issue.

alexAubin:
I don't understand why synapse needs a LDAP user to bind with, basically any anonymous users has read access, and there are dozens of other apps that do implement LDAP searches without being authenticated ... has it been doubled checked that synapse does really need a LDAP user+password for this, and why ? x_x

Josue-T:
It look like a know issue here: https://github.com/matrix-org/matrix-synapse-ldap3/issues/169#issuecomment-1364747945

But I would say I didn't tested on the last release if the issue still be here.

PR #979, opened 2020-05-06
[enh] service.py cleanup  + add tests for services

alexAubin:
## The problem

I was about to implement `service_update` without realizing that apps already updated service by just calling `service_add` again which just overwrite currently existing services and that's fine

Nevertheless found a few things to clean

## Solution

Clean stuff, depecate log_type, and ended up writing a few tests :| 

## PR Status

Tested and working on my side

## How to test

Run the tests and zblerg

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/src/yunohost/service.py
+++ b/src/yunohost/service.py
@@ -61,27 +61,27 @@ def service_add(name, description=None, log=None, log_type="file", test_status=N
     """
     services = _get_services()
 
-    services[name] = {}
+    services[name] = service = {}
 
     if log is not None:
         if not isinstance(log, list):
             log = [log]
 
-        services[name]['log'] = log
+        service['log'] = log
 
         if not isinstance(log_type, list):
             log_type = [log_type]
 
         if len(log_type) < len(log):
-            log_type.extend([log_type[-1]] * (len(log) - len(log_type))) # extend list to have the same size as log
+            log_type.extend([log_type[-1]] * (len(log) - len(log_type)))  # extend list to have the same size as log
 
         if len(log_type) == len(log):
-            services[name]['log_type'] = log_type
+            service['log_type'] = log_type
         else:
             raise YunohostError('service_add_failed', service=name)
 
     if description:
-        services[name]['description'] = description
+        service['description'] = description
     else:
         # Try to get the description from systemd service
         out = subprocess.check_output("systemctl show %s | grep '^Description='" % name, shell=True).strip()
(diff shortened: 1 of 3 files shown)


src/yunohost/service.py
```diff
@@ -296,10 +296,10 @@ def service_status(names=[]):
         # the hack was to add fake services...
         # we need to extract regenconf from service at some point, also because
         # some app would really like to use it
-        if services[name].get("status", "") is None:
+        if infos.get("status", "") is None:
```
kay0u, line 299: I'm confused:

```suggestion
        if infos.get("status") is None:
```

src/yunohost/service.py
```diff
@@ -314,8 +314,8 @@ def service_status(names=[]):
 
         else:
             translation_key = "service_description_%s" % name
-            if "description" in services[name] is not None:
-                description = services[name].get("description")
+            if "description" in infos is not None:
+                description = infos.get("description")
             else:
                 description = m18n.n(translation_key)
```
kay0u, line 320:
```suggestion
            description = infos.get("description")
            if not description:
                description = m18n.n(translation_key)
```

src/yunohost/service.py
```diff
@@ -415,41 +415,33 @@ def service_log(name, number=50):
...
     result = {}
 
     # First we always add the logs from journalctl / systemd
     result["journalctl"] = _get_journalctl_logs(name, number).splitlines()
 
-    for index, log_path in enumerate(log_list):
-        log_type = log_type_list[index]
+    for log_path in log_list:
+        # log is a file, read it
+        if not os.path.isdir(log_path):
```
kay0u, line 431:
```suggestion
        if os.path.isfile(log_path):
```

src/yunohost/service.py
alexAubin: Eeeeeh there's a not-obvious subtlety there ... Most services don't have this "status" key ... but some do have `"status" = None` and these are the one we want to catch with that if. Maybe we could have `if "status" in infos and infos["status"] is None` ... but that doesn't make it more explicit. We could also improve the comment..

src/yunohost/service.py
alexAubin: There's also a subtlety here which should clearly be improved as well .. but basically you want to cover the case where a service has `--log /some/dir` and the dir doesn't exists yet ...

With the current code it does work, but if check `isfile`, it will say "nope", will then proceed to the rest of the code an try to `os.listdir("/some/dir")` which will throw an expection if folder doesn't exists :s

kay0u: I don't know, maybe this code is clearer then ?

```python
if os.path.isfile(log_path):
    result[log_path] = _tail(log_path, number)
else if not os.path.isdir(log_path):
    continue
```

src/yunohost/service.py
kay0u: I understand, sorry for the noise.

src/yunohost/service.py
```diff
@@ -415,41 +414,33 @@ def service_log(name, number=50):
...
+            result[log_path] = _tail(log_path, number) if os.path.exists(log_path) else []
+            continue
```
alexAubin, line 432:
```suggestion
        if os.path.isfile(log_path):
            result[log_path] = _tail(log_path, number)
            continue
        elif not os.path.isdir(log_path):
            result[log_path] = []
            continue
```

src/yunohost/service.py
alexAubin: Alrighty :stuck_out_tongue_winking_eye: Commited in https://github.com/YunoHost/yunohost/commit/f25e07fd829cd179393762d5c23a6f9f2670ed1c

PR #980, opened 2020-05-06
Check for fatal code errors

alexAubin:
## The problem

Realized we had again code with undefined var...

## Solution

Add a check with tox that shall fail, but only catches really fatal stuff like undefined vars / redefining vars etc ... Ignore stuff like "import not at the top" or "unused var" ... should make wider tests in the future, but c.f. previous work from Kayou, we need a lot of cleaning to make tox/flake8 really happy.

## PR Status

Tested on my side

## How to test

Run the test o/

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,10 +9,16 @@ skip_install=True
 deps =
   pytest >= 4.6.3, < 5.0
   pyyaml >= 5.1.2, < 6.0
+  flake8 >= 3.7.9, < 3.8
 commands =
     pytest {posargs}
 
 [testenv:lint]
 skip_install=True
 commands = flake8 src doc data tests
+deps = flake8
+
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -9,10 +9,16 @@ skip_install=True
...
+  flake8 >= 3.7.9, < 3.8
 commands =
     pytest {posargs}
 
 [testenv:lint]
 skip_install=True
 commands = flake8 src doc data tests
 deps = flake8
+
+[testenv:invalidcode]
```
alexAubin, line 21: idk how to call this tho, feel free to suggest a different name if you have better ideas

PR #981, opened 2020-05-06
Default 'ask' questions for common app manifest args

alexAubin:
## The problem

This is me procrastinating again because of stupid issues spotted on the forum ...

c.f. https://forum.yunohost.org/t/installation-a-la-racine-conflicte-avec-apps-installees/11222

- In the past, gitlab had a message in manifest about path saying only '/' was accepted
- somehow this message got translated on weblate
- since then, message got updated
- translation did not
- user was confused, thought app could only be installed on /

## Solution

... it's complete madness that each package reimplement these messages. Jibec a few years ago already showed me a spreasheet showing how on the translator side, it was just about translating the same thing over and over with just the app name changing ... For all the common ones like domain, path, password, is_public and admin, we can have a default message from the core. This will also make translation process for these easier...

If packagers want to add more specific info to these, they can through the `help` key ...

## PR Status

Tested and working on my side

PR incoming for the webadmin

The only thing not covered now is custom app install from the webadmin ... But this should be adressed also for other reason (it shouldn't be the JS client trying to absolutely fetching the manifest.json from raw.githubuserstuff)

## How to test

Run `yunohost app catalog --full` or `yunohost app info [some_installed_app] --full` and check the manifest strings

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -2071,12 +2073,63 @@ def _get_manifest_of_app(path):
...
+
+        # We only support questions for install so far, and for other
+        if script_name != "install":
+            continue
+
+        for arg in arg_list:
+
+            # Do not override 'ask' field if provided by app ?... Or shall we ?
+            #if "ask" in arg:
+            #    continue
```
alexAubin, line 2122: Note that this is where we decide if this applied or not by default ... I was going to only do this for app that do not define the "ask" key but hmpf

alexAubin:
Yolomergin :| 

https://www.youtube.com/watch?v=yOMj7WttkOA

PR #982, opened 2020-05-06
During app installs, set default answer for user-type args to main user

alexAubin:
## The problem

We have a default answer for domains, but it's a bit annoying to type the username all the time for admin args when we could guess what's the main user

## Solution

Guess the main user by checking if the user has an alias to root@maindomain.tld

## PR Status

Tested on my side

## How to test

Try to install an app that require a user, e.g. wordpress

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/app.py
```diff
@@ -2439,9 +2439,16 @@ def _parse_args_in_yunohost_format(args, action_args):
...
                 elif arg_type == 'user':
                     msignals.display(m18n.n('users_available'))
-                    for user in user_list()['users'].keys():
+                    users = user_list()['users']
+                    for user in users.keys():
                         msignals.display("- {}".format(user))
 
+                    root_mail = "root@%s" % _get_maindomain()
+                    for user in users.keys():
+                        if root_mail in user_info(user)["mail-aliases"]:
```
kay0u, line 2448: is it possible to have several users with this aliase?

alexAubin: I don't think so ?

PR #985, opened 2020-05-08
Rely on diagnosis for letsencrypt elligibility

alexAubin:
## The problem

Meh, I was about to update the messages related to certificates to point to the diagnosis to help debugging, but hmpf that was just yet again reminding me that the results of what the current cert checks are inconsistent with what the diagnosis do >.>

## Solution

Ended up tweaking the logic because anyway there's only one way to test this which is on the battlefield

## PR Status

Partially tested ... needs more tests on a real setup with a properly exposed domain. Will do soonish

## How to test

Try to run `yunohost domain cert-status`

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/certificate.py
```diff
@@ -596,10 +596,9 @@ def _prepare_certificate_signing_request(domain, key_file, output_folder):
     # For "parent" domains, include xmpp-upload subdomain in subject alternate names
     if domain in domain_list(exclude_subdomains=True)["domains"]:
         subdomain = "xmpp-upload." + domain
-        try:
-            _dns_ip_match_public_ip(get_public_ip(), subdomain)
+        if dig(subdomain, "A", resolvers="force_external") == ("ok", [get_public_ip()]):
```
Josue-T, line 599: And what about IPv6 ?

alexAubin: Hmyea indeed, it may be not so hard to fetch the result from the diagnosis system as well, will have a look

alexAubin:
Tested on a semi-real server and working \o/ Yolomergin

PR #987, opened 2020-05-11
[enh] New config-panel mechanism

zamentur:
# The problem

VPN Client need to upload some files in config panel. 
https://github.com/YunoHost/issues/issues/1589
The config panel current system is to complicated (too much redundant code to write).

# Solution
config

## CLI / API design
This CLI / API has been rewrited in order to find a way to get a future uniform CLI/API for Global, Domain, App, User configuration/settings.

The old design of `yunohost app setting APP KEY [-v VALUE] [-d]` is not ok cause set actions are define with a GET HTPP verb...

Currently, it's like `yunohost settings` aka global settings but on steroid. 
 
#### Get a description of available forms with current values
```
$ yunohost app config get vpnclient
```
#### Get a specific value
```
$ yunohost app config get vpnclient manual.vpn.server_port
```
#### Export values in a file as json
```
$ yunohost app config get vpnclient manual -m export --output-as json > /root/verysimple.cube
```
#### Why define --mode instead of a --fields option ?
I think it's boring for user to specify each fields wanted. I think we could offers some kind of pre defined views called mode here.

#### Change settings inside a specific form
In interactive way
```
$ yunohost app config set vpnclient manual
```
Note: Current values are prefilled, so you can just type enter to keep like it was.

Prefilled with some answers
```
$ yunohost app config set vpnclient manual --args "server_port=443&server_proto=tcp"
```
Or
```
$ yunohost app config set vpnclient manual --args-file /root/verysimple.cube
```

#### Change a specific setting
```
$ yunohost app config set vpnclient manual.vpn.server_port -v 443
```

## Config panel format
You can see an example of toml here: https://github.com/YunoHost/test_apps/blob/master/config_app_ynh/config_panel.toml

### Default config script
With simple use case, packagers doesn't need anymore to write a config script, the toml is enough. If the config script doesn't exist, it will be added with this default code:
```
#!/bin/bash
source /usr/share/yunohost/helpers
ynh_abort_if_errors
final_path=$(ynh_app_setting_get $app final_path)
ynh_app_config_run $1
```

### Option short key have to be unique in all the config_panel.toml
So you can't have 
```
[manual.vpn.server_ip]
[advanced.dns.server_ip]
```
Indeed the real variable name is server_ip and here you have a conflict.
This, simplifies some part of the code, and avoid to have to use long name and do ./_ conversion.
In future, it could be decide to manipulate with FORM_ID like "manual_server_ip" or "advanced_server_ip"

### No retrocompatibility
All previous config panel should be rewritten into 1.0 version.
The main breaking change is environment variables are shorten and lower case.
However, write a config panel in this new version is easier than 0.1 format.

#### `email`, `url`, `date`, `time`, `color`, `range`
This types are quite trivial, it refers to HTML5 dedicated input. Some pre-validation are defined for them.

#### `text`
This type allows to have a multiline input.
In cli mode, it runs your default editor to allow you to edit the multiline text.

#### `tags`
You could now select tags like here https://bootstrap-vue.org/docs/components/form-tags
You can defined an allowed list of choices with choices argument
The pattern attribute applies on all element

#### `file`
An input file question. It allows only single file (no multifile or folder).

The file mechanism inside moulinette is not used, cause these file questions are defined in toml not in action map :/
Instead we use a crazy base64 method. So this file input is not designed to support big file upload.

#### `markdown` and `alert`
Markdown and alert component allow to display some static or dynamic info to user.

#### `bind` argument mechanism
This new argument allows to define where read and write the value bind to the question.
##### Default behaviour
If you have not defined a specific getter/setter (see bellow), and without source argument it will read and save the value in app settings yaml file.

##### Read / write in a var of a configuration file
If you want to read and save the value into a variable (called like the option name) of a file (json, yaml, ini, php, py ...) you can do:
```
bind = ":__FINALPATH__/conf.json"
```
If you want to read and save the value into an other variable (email in the example) of a file (json, yaml, ini, php, py ...) you can do:
```
bind = "email:__FINALPATH__/conf.json"
```
Note: This mechanism is language agnostic, however it's monoline: you can't save multiline text or file in a variable with this method. If you need to save multiline content in a configuration variable, you should do it via a specific getter/setter.

If you want to read and save the value into an other variable (email in the example) after a regex in a file (json, yaml, ini, php, py ...) you can do:
```
bind = "^\[user\]>email:__FINALPATH__/conf.json"
```
##### Read / write an entire file
If you have a question of type `file`or `text` you could want to save the content into a specific path on the system.
```
bind = "__FINALPATH__/logo.png"
```
##### Simple getter
Here is an example of custom getter you can put in a config script
```
get__current_date() {
echo "$(date)"
}
```

##### Getter which rewrite an option
A more advanced example, where we rewrite the option argument
scripts/config
```bash
get__status() {
    if [ -f "/sys/class/net/tun0/operstate" ] && [ "$(cat /sys/class/net/tun0/operstate)" == "up" ]
    then
    cat << EOF
style: success
ask:
  en: Your VPN is running :)
EOF
    else
    cat << EOF
style: danger
ask:
  en: Your VPN is down
EOF
    fi
}
```
config_panel.toml
```
        [main.cube.status]
        ask = "Custom getter alert"
        type = "alert"
        style = "info"
        bind = "null" # no behaviour on
```

##### Setter
```
set__admin_password() {
    if [ -z "${admin_password}" ]
    then
        python -c "import crypt; print(crypt.crypt(\"${admin_password}\", \"\$6\$saltsalt\$\"))" > $final_path/password
        ynh_print_info "The hash of your password has been registered in $final_path/password"
    fi
}
```
I suggest to display info for user about change that has been made to help them to understand a bit what's going.

### Validation and internal values
Validation can be made with regex through `pattern` argument 
```
        pattern.regexp = '^.+@.+$'
        pattern.error = 'An email is required for this field'
```

You can also restrict several types with a choices list.
```
        choices.option1 = "Plop1"
        choices.option2 = "Plop2"
        choices.option3 = "Plop3"
```

Some other type specific argument exist like 
| type | validation arguments |
| -----  | --------------------------- |
| `number`, `range` | `min`, `max`, `step` |
| `file` | `accept` |
| `tags` | `limit` |
| `boolean` | `yes` `no` |

Finally, if you need specific or multi variable validation, you can use custom validators function:
```
validate__admin_password() {
    if [[ "${#admin_password}" -lt 8 ]]; then echo 'Too short password'; fi
}
```
Note: this validators is a bad example, cause password type questions are already checked...

### `visible`
A visible property allow to display section or question based on a condition
```
[manual.auth.password]
visible = "login && ! private_key"
```
Note: visible are not managed in CLI

### `redact` confidential info
Password are redact automatically in operation log, but you can add redact args for a specific option if needed.

#### Restart a service at the end
You can use the `services` key to specify which service need to be reloaded or restarted
```
services = [ 'nginx', '__APP__' ]
```
This argument could be on panel, section, or question.

#### Overwrite config panel mechanism
All main configuration helpers are overwritable, example:
```
ynh_app_config_apply() {
    
    # Stop vpn client
    touch /tmp/.ynh-vpnclient-stopped
    systemctl stop ynh-vpnclient

    _ynh_app_config_apply

    # Start vpn client
    systemctl start ynh-vpnclient
    rm -f /tmp/.ynh-vpnclient-stopped

}
```

## Philosophy
**I think we should discuss in next meeting how to avoid bad use of all of these things to avoid to lose users with too big/complicated configuration panels**

## PR Status

- [ ] Write english locale
- [x] Test app install
- [x] Boolean : what if we want to save 'yes' into the config file ?
- [x] Is config_panel the right terms ? may be helpers should be called like `ynh_app_forms_get` ? 
- [ ] panels in the code could be called forms instead...
- [x] Unit test
- [x] Don't generate each time the file checksum for upgrade
- [ ] list all reserved keywords and trigger an error if an option use one of this keyword
- [ ] Documentation for packagers

## How to test

You need :
- https://github.com/YunoHost/moulinette/pull/290
- https://github.com/YunoHost/yunohost-admin/pull/366
- https://github.com/YunoHost/test_apps/tree/master/config_app_ynh

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


data/helpers.d/apt
```diff
@@ -189,17 +193,37 @@ ynh_package_install_from_equivs () {
     LC_ALL=C equivs-build ./control 1> /dev/null
     dpkg --force-depends --install "./${pkgname}_${pkgversion}_all.deb" 2>&1)
 
-    ynh_package_install --fix-broken || \
+    # Let's try to see if install will work using dry-run. It it fails,
+    # it could be because the pinning of sury is blocking some package install
+    # c.f. for example: https://github.com/YunoHost/issues/issues/1563#issuecomment-623406509
+    # ... In that case, we use an ugly hack were we'll use a tweaked
+    # preferences.d directory with looser contrains for sury...
+    if ! ynh_package_install --fix-broken --dry-run >/dev/null 2>&1 && [ -e /etc/apt/preferences.d/extra_php_version ]
```
kay0u, line 201: Why did you add back this dirty patch?

src/yunohost/app.py
```diff
@@ -1601,6 +1604,23 @@ def app_config_apply(operation_logger, app, args):
...
+                    # Upload files from API
+                    # A file arg contains a string with "FILENAME:BASE64_CONTENT"
+                    if option["type"] == "file" and msettings.get('interface') == 'api':
+                        if upload_dir is None:
+                            upload_dir = tempfile.mkdtemp(prefix='tmp_configpanel_')
+                        filename, args[generated_name] = args[generated_name].split(':')
+                        logger.debug("Save uploaded file %s from API into %s", filename, upload_dir)
+                        file_path = os.join(upload_dir, filename)
+                        try:
+                            with open(file_path, 'wb') as f:
```
randomstuff, line 1621: Note that as `filename` is chosen by the user, this gives arbitrary file write (i.e. `os.path.join("/foo", "/etc/passwd") == "/etc/passwd"`) through this API.

data/helpers.d/apt
zamentur: i don't

src/yunohost/app.py
zamentur: fixed

Psycojoker:
We uses bottle as the web backend in the moulinette, it might be worth it looking at the builtin features for file upload like http://bottlepy.org/docs/dev/api.html?highlight=fileupload#bottle.FileUpload and see if we can integrate that directly into yunohost code '-' (and that's secured reviewed code for bottle ... normally)

src/yunohost/app.py
randomstuff: Does it actually fix the issue?

~~~
>>> os.path.normpath("/tmp/" + "../../etc/passwd")
'/etc/passwd'
~~~

zamentur:
> We uses bottle as the web backend in the moulinette, it might be worth it looking at the builtin features for file upload like http://bottlepy.org/docs/dev/api.html?highlight=fileupload#bottle.FileUpload and see if we can integrate that directly into yunohost code '-' (and that's secured reviewed code for bottle ... normally)

I already use Bottle for file argument in moulinette. However, here it's a bit more difficult cause file question are contained in toml not in actionmap, that's why i use this base64 approach.

zamentur:
- [ ] Validate pattern
- [ ] API file args not working
- [ ] VisibleIf in cli
- [ ] Unit test
- [ ] Check for file managed by the app to avoid the file to be considered as manually modified

zamentur:
Note: if you want you can start to test in CLI mode

src/yunohost/app.py
```diff
@@ -2737,46 +2855,79 @@ def parse_question(self, question, user_answers):
...
 
-        if question.value is None:
-            text_for_user_input_in_cli = self._format_text_for_user_input_in_cli(
-                question
-            )
-
-            try:
-                question.value = Moulinette.prompt(
-                    text_for_user_input_in_cli, self.hide_user_input_in_prompt
+<<<<<<< HEAD
```
kay0u, line 2858: :smile:

src/yunohost/app.py
```diff
@@ -2936,43 +3109,176 @@ class NumberArgumentParser(YunoHostArgumentFormatParser):
...
 
     def parse_question(self, question, user_answers):
-        question = super(NumberArgumentParser, self).parse_question(
+        question_parsed = super().parse_question(
             question, user_answers
         )
+        question_parsed.min = question.get('min', None)
+        question_parsed.max = question.get('max', None)
+        if question_parsed.default is None:
+            question_parsed.default = 0
```
alexAubin, line 3118: Eeeeh what if the question is optional ? (Dunno if that matters or not)

src/yunohost/app.py
```diff
@@ -2737,46 +2856,76 @@ def parse_question(self, question, user_answers):
...
-        # we don't have an answer, check optional and default_value
-        if question.value is None or question.value == "":
-            if not question.optional and question.default is None:
-                raise YunohostValidationError(
-                    "app_argument_required", name=question.name
-                )
-            else:
+
+            # Apply default value
+            if question.value in [None, ""] and question.default is not None:
```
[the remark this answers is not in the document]
zamentur: With boolean or integer value False or 0 means something. Don't know if at this step it could be of that types...
Maybe we should check [] to for Tags args

src/yunohost/app.py
```diff
@@ -1764,169 +1760,226 @@ def app_action_run(operation_logger, app, action, args=None):
...
+    from yunohost.hook import hook_exec
+
+    # Add default config script if needed
+    config_script = os.path.join(APPS_SETTING_PATH, app, "scripts", "config")
+    if not os.path.exists(config_script):
+        logger.debug("Adding a default config script")
+        default_script = """#!/bin/bash
+source /usr/share/yunohost/helpers
+ynh_abort_if_errors
+final_path=$(ynh_app_setting_get $app final_path)
```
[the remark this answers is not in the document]
zamentur: final_path is needed to replace __FINAL_PATH__ in `service` and `source` properties.

alexAubin:
Note that the are some important python errors found by the CI:

```
src/yunohost/app.py:39:1: F401 'moulinette.interfaces.cli.colorize' imported but unused
src/yunohost/app.py:57:1: F401 'yunohost.utils.config' imported but unused
src/yunohost/app.py:1763:5: F811 redefinition of unused 'config' from line 57
src/yunohost/app.py:1773:5: F811 redefinition of unused 'config' from line 57
src/yunohost/utils/i18n.py:21:1: F401 'moulinette.Moulinette' imported but unused
src/yunohost/utils/config.py:24:1: F401 'toml' imported but unused
src/yunohost/utils/config.py:41:1: F811 redefinition of unused '_get_services' from line 40
src/yunohost/utils/config.py:147:27: F821 undefined name 'errors'
src/yunohost/utils/config.py:696:21: F821 undefined name 'shutil'
```

alexAubin:
Tests are passing

12TIvbgMTrGhhu

data/helpers.d/utils
```diff
@@ -473,6 +473,197 @@ ynh_replace_vars () {
...
+    local legacy_args=fka
+    local -A args_array=( [f]=file= [k]=key= [a]=after=)
+    local file
+    local key
+    local after
+    # Manage arguments with getopts
+    ynh_handle_getopts_args "$@"
+    after="${after:-}"
+
+    [[ -f $file ]] || ynh_die "File $file does not exists"
```
kay0u, line 520:
```suggestion
    [[ -f $file ]] || ynh_die --message="File $file does not exists"
```

data/helpers.d/utils
```diff
@@ -473,6 +473,197 @@ ynh_replace_vars () {
...
+    local -A args_array=( [f]=file= [k]=key= [v]=value= [a]=after=)
+    local file
+    local key
+    local value
+    local after
+    # Manage arguments with getopts
+    ynh_handle_getopts_args "$@"
+    after="${after:-}"
+
+    [[ -f $file ]] || ynh_die "File $file does not exists"
```
kay0u, line 595:
```suggestion
    [[ -f $file ]] || ynh_die --message="File $file does not exists"
```

src/yunohost/app.py
```diff
@@ -52,8 +52,13 @@
     mkdir,
 )
 
-from yunohost.service import service_status, _run_service_command
 from yunohost.utils import packages
+from yunohost.utils.config import (
+    ConfigPanel,
+    parse_args_in_yunohost_format,
+    Question,
```
kay0u, line 59: I don't understand why you moved `parse_args_in_yunohost_format` and `Question` into the config.py file since it's related to app install, upgrade, etc...

At least the app.py file is shorter now

src/yunohost/app.py
```diff
@@ -640,36 +633,13 @@ def app_upgrade(app=[], url=None, file=None, force=False, no_safety_backup=False
         # Execute the app upgrade script
         upgrade_failed = True
         try:
```
kay0u, line 635: there is a `try`, `finally`, but no except, is it ok to do that?

src/yunohost/app.py
```diff
@@ -640,36 +633,13 @@ def app_upgrade(app=[], url=None, file=None, force=False, no_safety_backup=False
...
-
-            error = m18n.n("unexpected_error", error="\n" + traceback.format_exc())
-            logger.error(
-                m18n.n("app_upgrade_failed", app=app_instance_name, error=error)
+            upgrade_failed, failure_message_with_debug_instructions = hook_exec_with_script_debug_if_failure(
+                extracted_app_folder + "/scripts/upgrade",
+                env=env_dict,
+                operation_logger=operation_logger,
+                error_message_if_script_failed=m18n.n("app_upgrade_script_failed"),
+                error_message_if_failed=lambda e: m18n.n("app_upgrade_failed", app=app_instance_name, error=e)
```
kay0u, line 641: I don't understand this line, especially the e variable gave to `error` arg O.o

kay0u (commented):
WIP review,

`--message=` getops is not really necessary, it's just to keep it unified.

src/yunohost/tests/test_app_config.py
```diff
@@ -0,0 +1,198 @@
...
+
+def test_app_config_get(config_app):
+
+    assert isinstance(app_config_get(config_app), dict)
+    assert isinstance(app_config_get(config_app, full=True), dict)
+    assert isinstance(app_config_get(config_app, export=True), dict)
+    assert isinstance(app_config_get(config_app, "main"), dict)
+    assert isinstance(app_config_get(config_app, "main.components"), dict)
+    # Is it expected that we return None if no value defined yet ?
+    # c.f. the whole discussion about "should we have defaults" etc.
```
kay0u, line 109: uuuh, where can i read this discussion?

src/yunohost/hook.py
```diff
@@ -447,10 +447,10 @@ def _hook_exec_bash(path, args, chdir, env, user, return_format, loggers):
             raw_content = f.read()
         returncontent = {}
 
-        if return_format == "json":
+        if return_format == "yaml":
```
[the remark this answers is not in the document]
zamentur: Normally, json is included in yaml, so if a script return a json, it could be read with read_yaml.

However as it's not the same parser, that's true that some small differiencies could exists.

For now i don't even know if some hook returns json ?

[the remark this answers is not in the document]
zamentur: I made this change cause it's clearly easier to write yaml in bash than json.

src/yunohost/app.py
zamentur: Because now Question and ConfigPanel are used to create the commands `yunohost domain config get|set` too.

And because the app.py is too looooooong and it's boring to navigate between 4000+ lines.

src/yunohost/app.py
zamentur: try finally without except is valid in python. It allows to execute the code in the finally even if an exception is raised

src/yunohost/tests/test_app_config.py
zamentur: Sorry it was an IRL discussion about the fact that default attribute should not be used in classic App Config Panel (but in Domain Config Panel it could be used). Indeed, if you bind to the settings, the settings should be registered. If you bind to a var in a file, this line need to exists in the file, so this var is read too (=> no default needed in this case).

At the end of the discussion, we think we should return an error if a value is not provided in the case of App Config Panel.

src/yunohost/app.py
alexAubin: Eeeeh yeah sorry it's a bit odd, but it's later used in the function hook_exec_with_debug_stuff (don't remember the exact name) ... and the error message includes a sub-error message depending on the nature of the error (app script crashed, manual Ctrl+C, or exception in yunohost code) ... though most of the time it's the app script that crashed but meh ... I just tried to reproduce the current code behavior while at the same time reducing the epic lengthy toast of boring error management code

PR #988, opened 2020-05-11
[fix] postfix per-domain destination concurrency

taziden:
## The problem

Postfix has this very peculiar behavior where the behavior of some config items changes depending on the value. 
Here, if `smtp_destination_concurrency_limit` is set to 1, then according to http://www.postfix.org/postconf.5.html#default_destination_concurrency_limit it doesn't mean "1 concurrent mail per domain, but per recipient address".

## Solution

So, if set to 1, it means we can send any volume of e-mails concurrently (with a 5s delay) if all recipient addresses are different.
In order to avoid this, we should increase the value to restore the expected behavior (concurrency per domain, not per recipient).

## PR Status

...

## How to test

Send datalove <3

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/templates/postfix/main.cf
+++ b/data/templates/postfix/main.cf
@@ -170,7 +170,7 @@ smtpd_milters = inet:localhost:11332
 milter_default_action = accept
 
 # Avoid to send simultaneously too many emails
-smtp_destination_concurrency_limit = 1
+smtp_destination_concurrency_limit = 2
 default_destination_rate_delay = 5s
 
 # Avoid email adress scanning


alexAubin:
Hmmmm idk what to think about this ... I think we asked ourselves that question when implementing it the first time (not sure)

Thing is, you could be hosting a mailing list with many (50+) adresses on gmail.com but then each time you would send an email to the mailing list, it would take 50*5 = 250 seconds for all the mails to be sent. Maybe that's legit but I would be tempted to reduce the duration to 1sec ..?

taziden:
> Hmmmm idk what to think about this ... I think we asked ourselves that question when implementing it the first time (not sure)
> 
> Thing is, you could be hosting a mailing list with many (50+) adresses on gmail.com but then each time you would send an email to the mailing list, it would take 50*5 = 250 seconds for all the mails to be sent. Maybe that's legit but I would be tempted to reduce the duration to 1sec ..?

One of the first thing to do when you administer a mail server with some volume, is put throttling in place for popular destinations that will consider you as a spammer  if you keep the default postfix settings.
You'll find plenty of docs on the webz about that, like this one https://wiki.deimos.fr/Postfix:_limit_outgoing_mail_throttling.html

You definitely want to be on the safe side and not have your mail server to have a bad reputation with Google/Yahoo/Microsoft and so on if you administer, say a mailing list with many addresses on gmail. Yes, delivery time will increase but at least it will be delivered.

Here, by setting smtp_destination_concurrency_limit to 1, what happens is that you send way too much e-mails to Google in little time and Google will flag you as a potential spammer and defer your e-mails. It happened to me :-) 

Here is a config I use on a server which send 18k mails each week (a website newsletter), among which a big part, maybe half is for gmail addresses. When we had set smtp_destination_concurrency_limit to 1 by mistake, thinking it was the concurrency per domain, we've been blocked.
Today, here are the settings we use for Gmail : 

```
smtp_initial_destination_concurrency = 2
smtp_destination_concurrency_limit = 5
smtp_destination_recipient_limit = 5
smtp_destination_rate_delay = 1s
```
Each second we deliver 5 e-mails to Gmail.

For Yahoo, La Poste, Hotmail and others who are not that permissive, we increased the rate_delay to 5s and set the concurrency limit to 2.

With such a setup, we're not blocked by any of the big providers.

taziden:
> Thing is, you could be hosting a mailing list with many (50+) adresses on gmail.com but then each time you would send an email to the mailing list, it would take 50*5 = 250 seconds for all the mails to be sent. Maybe that's legit but I would be tempted to reduce the duration to 1sec ..?

In the current setup, with such a mailing list, the server would get blacklisted (temporarily) quickly.
With my proposal, if you have 50 e-mails to deliver to Gmail, it would take 125s in total, not 250s.
Each batch, you send 2 e-mails, with 5 seconds between each batch. So, 50/2*5=125 if I'm not mistaken.
You could decrease the rate delay or increase the concurrency limit in order to shorten this delay but ~2mins seem reasonable to me in this scenario. Especially if you're sure not to be taken as a spammer for that.

alexAubin:
So you're proposing we add

```
smtp_destination_concurrency_limit = 5
smtp_destination_recipient_limit = 5
```

as well ? What you explain seems sensible and based on real-life experience rather than random guess so that's fine for me :+1:

taziden:
I'd keep the concurrency limit low (thus 2) so as to avoid blacklisting by the annoying ones, such as Orange, La Poste, Free and Hotmail/Microsoft.
I'd say the PR is OK as it is. No need to change anything else for the moment. If people complain that the total delivery time is a problem, we'll see then.

Good for you?

PR #991, opened 2020-05-16
[enh] CI V2

kay0u:
## Solution

One way to improve the CI is to build and install debs before running tests instead of using the symbolic link as we do for ynh-dev.

- [x] Building debs in the beginning
- [x] Run install/postinstall and an update to the current version with these debs
- [x] Performing tests with the installed debs

Needs https://github.com/YunoHost/yunohost-ci/pull/1

The new version of yunohost-ci preinstalls all YNH dependencies (no need to reinstall them at each test)

## PR Status

Ready

## How to test

Result [here](https://gitlab.com/kay0u/yunohost/pipelines/148010171)
Test report [here](https://gitlab.com/kay0u/yunohost/pipelines/148010171/test_report)

[pipeline status](https://gitlab.com/kay0u/yunohost/-/commits/ci-build-&-install-deb) [coverage report](https://gitlab.com/kay0u/yunohost/-/commits/ci-build-&-install-deb)

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


kay0u:
We could improve the CI with this:

https://docs.gitlab.com/ee/ci/yaml/README.html#artifact-downloads-with-needs

And this (to be able to test migration to buster for example):
https://docs.gitlab.com/ee/ci/yaml/#artifact-downloads-between-pipelines-in-the-same-project

.gitlab-ci.yml
```diff
@@ -1,8 +1,76 @@
...
+    - cd $DEB_TO_BUILD
+    - debuild -us -uc
+
+build-metronome:
+  extends: .build-stage
+  variables:
+    DEB_TO_BUILD: "metronome"
+  script:
+    - cd $DEB_TO_BUILD
+    - dpkg-buildpackage -rfakeroot -uc -b -d
```
alexAubin, line 59: Hmmm I don't know if that's really worth it to build metronome considering the time it takes ? (It's huuuuuuge compared to how often the CI runs) Also if we go this way we should also add python-miniupnpc builds for Buster CI which also is huge. Just having them preinstalled like other dependencies is good enough imho

kay0u: Metronome builds as fast as the other packages on my computer. We'll see how long it takes on the CI.

I don't know, maybe it's completely overkill to build deb on each build, to build deb at every build, but at least it removes the problems when adding/removing dependencies on YNH.

kay0u: On the other hand, I can preinstall all the necessary dependencies for the CI :
https://github.com/YunoHost/yunohost/pull/991/files#diff-96edf7a6f008de9e928d04e1ae5e12a5R19
https://github.com/YunoHost/yunohost/pull/991/files#diff-96edf7a6f008de9e928d04e1ae5e12a5R113

kay0u: I removed the construction of the metronome deb, we can easily add it again, as well as python-miniupnpc

PR #992, opened 2020-05-16
Better error message when psql is not there for database_exists

autra:
## The problem

If app packager checks too soon for db existence (before postgres is installed) the error message is not very clear.

## Solution

Just add a litte error message if psql cannot be found. The install crashes the same, but the packager now knows why, hopefully.

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/postgresql
+++ b/data/helpers.d/postgresql
@@ -187,7 +187,13 @@ ynh_psql_database_exists() {
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
 
-    if ! sudo --login --user=postgres PGUSER="postgres" PGPASSWORD="$(cat $PSQL_ROOT_PWD_FILE)" psql -tAc "SELECT datname FROM pg_database WHERE datname='$database';" | grep --quiet "$database"
+    # if psql is not there, we cannot check the db
+    # though it could exists.
+    if ! command -v psql
+    then
+      ynh_print_err -m "PostgreSQL is not installed, impossible to check for db existence."
+      return 1
+    elif ! sudo --login --user=postgres PGUSER="postgres" PGPASSWORD="$(cat $PSQL_ROOT_PWD_FILE)" psql -tAc "SELECT datname FROM pg_database WHERE datname='$database';" | grep --quiet "$database"
     then
         return 1
     else


data/helpers.d/postgresql
```diff
@@ -187,7 +187,12 @@ ynh_psql_database_exists() {
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
 
-    if ! sudo --login --user=postgres PGUSER="postgres" PGPASSWORD="$(cat $PSQL_ROOT_PWD_FILE)" psql -tAc "SELECT datname FROM pg_database WHERE datname='$database';" | grep --quiet "$database"
+    # if postgres user doesn't exist, we consider the db does not as well
```
autra, line 190: aaaand forgot to edit the comment.

data/helpers.d/postgresql
```diff
@@ -187,7 +187,12 @@ ynh_psql_database_exists() {
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
 
-    if ! sudo --login --user=postgres PGUSER="postgres" PGPASSWORD="$(cat $PSQL_ROOT_PWD_FILE)" psql -tAc "SELECT datname FROM pg_database WHERE datname='$database';" | grep --quiet "$database"
+    # if psql is not there, we cannot check the db
+    # though it could exists.
+    if ! command -v psql
+    then
+      ynh_die --message="PostgreSQL is not installed, impossible to check for db existence"
```
alexAubin, line 194: Uuuuh I'm wondering if this helper ynh_psql_database_exists is called directly or indirectly during remove script, that could exit the remove script prematurely..?

autra: yes you're right ("If it can happen..."). OMG I don't know if there is a way out of this then.

alexAubin: I'm guessing you had a real-life context that led you to implement this, soooo uuuh what was it ? :P 

Naively just adding a warning could be good enough ?

autra: Well it's exactly what I wrote actually : I've decided to put all the checks (does the user/home dir/db already exists?) at the beginning of my script. Then of course it failed because I didn't have psql yet. I just thought that it should log correctly in this case, because I felt like it took me too much time to check what was happening.

But you're right : I should keep the old behaviour but just add a warning. I'm not yet sure how bash behaves in this case though... (what does it do? Returns 0? Returns nothing?)

alexAubin: Hmmmm logging a warning is just a message and is idependant of what the function returns. If you want it to return 1, then just write `return 1`

yalh76: In addition `ynh_psql_database_exists` seems to only be use in `ynh_psql_remove_db` that is only used in the removal script... My opinion is: no need to check if postgresl is installed or not

alexAubin:
Related : https://github.com/YunoHost/yunohost/pull/993

data/helpers.d/postgresql
autra: > In addition `ynh_psql_database_exists` seems to only be use in `ynh_psql_remove_db` that is only used in the removal script... My opinion is: no need to check if postgresl is installed or not

If I read correctly the code of `ynh_psql_create_db`, I think we should alway call this function before attempting to create a db, only for the fact that `ynh_psql_create_db` modifies owner and rights quite violently. I actually do this check in `diaspora_ynh`. (actually maybe another fix should be to make ynh_psql_create_db call this function directly? Idk)

autra:
I've updated with a (hopefully) correct return. Should match the old behaviour.

zamentur (approved):
LGTM (untested)

PR #999, opened 2020-05-23
Use system hostname by default for avahi

Gia90:
## The problem

The yunohost machine is registered as `yunohost.local` even when a different system hostname is set.
Running `avahi-resolve-address <yunohost machine IP>` always returns `yunohost.local`.

## Solution

Removing the `host-name` line in `/etc/avahi/avahi-daemon.conf` and restarting the `avahi-daemon` service solved the issue.
As described in the [avahi-demon manual](https://linux.die.net/man/5/avahi-daemon.conf):
> host-name= Set the host name avahi-daemon tries to register on the LAN. **If omited defaults to the system host name as set with the sethostname() system call**.

### DIFF
--- a/data/templates/avahi-daemon/avahi-daemon.conf
+++ b/data/templates/avahi-daemon/avahi-daemon.conf
@@ -19,7 +19,7 @@
 # file!
 
 [server]
-host-name=yunohost
+#host-name=yunohost
 domain-name=local
 #browse-domains=0pointer.de, zeroconf.org
 use-ipv4=yes


alexAubin:
Hmmmm but can you clarif what's the avahi hostname after this PR then ? Is it `$(hostname -s).local` ?

Note that having `yunohost.local` is a "feature" because some people access their server though yunohost.local ... or at least before postinstall ... because it's much easier to remember than having to find out what's the local IP. So I don't know what behavior we want here exactly

Gia90:
The avahi hostname will become `$(hostname).local`.

I was quite confused after reading these lines in the [post install page](https://yunohost.org/#/postinstall):
> If you don't own a domain name and don't want a .nohost.me, .noho.st or .ynh.fr, you can use a local domain. The idea is to configure your router to redirect a local domain name to your server. **For example you could create the yunohost.local domain redirecting to your server in your router**, and now every device on the network will be redirected to your server when accessing yunohost.local. More information on how to setup a local domain can be found here.

The confusion comes from the fact that hostnames ending with `.local` should only be resolved through mDNS (avahi) and not by a DNS (like the one in the router) as stated in the [mDNS RFC](https://tools.ietf.org/html/rfc6762):
> Any DNS query for a name ending with ".local." MUST be sent to the mDNS IPv4 link-local multicast address 224.0.0.251 (or its IPv6 equivalent FF02::FB).

Using `yunohost.local` as hostname for both avahi and the router DNS isn't a good idea since it may generate conflicts in the hostname resolution.

alexAubin:
Indeed, clearly the doc is confusing and the whole story about .local should be clarified as noted by these issues: 

https://github.com/YunoHost/issues/issues/1447
https://github.com/YunoHost/issues/issues/1464
https://github.com/YunoHost/issues/issues/1418

but overall we don't have much expertize about this and some details are fuzzy.

When you mention "router", do you mean router as a hardware (internet box) or are you thinking about dnsmasq ? (anyway both sounds wrong)

I'm not sure what what the author of the doc snippet you pasted had in mind ... I think it was mostly about adding "yunohost.local" as a yunohost domain such that a user could then install app on this domain, without the need of anything DNS record or /etc/hosts tweaking (this is in fact a recurrent question). But one implication currently is that some bits of config are added to dnsmasq config but as you explain that indeed seems weird/conflicting with avahi. However we could add an exception for the .local domain in dnsmasq's regenconf. But that's beyond the scope of this PR.

Anyway, I think overall I agree with your PR an it makes sense :+1: . Trying to summarize:
- before postinstall, we can make it such that pre-installed images (like ARM or x86 ISO) have a default hostname like "yunohost" and therefore we can advise users to connect to the machine with yunohost.local (to avoid having to find out the local IP)
- after postinstall, machine can be accessed "$hostname.local" which is more meaningful  than yunohost.local and avoid conflicts in more advanced setups with multiple yunohost instances on the same network (can even be useful for dev anyway).
- we "just" need to document/explain/advertise this properly to users.

Gia90:
When I said "router", I was referring to the "internet box", to use the same terminology used in the post install page.

I really like your idea, it will simplify a lot [this procedure](https://yunohost.org/#/ssh) and make yunohost more user friendly :)

This PR will need some additional work obviously and if merged in this state may even cause issues, since I just realized that there are other config files which rely on `yunohost.local`:
https://github.com/YunoHost/yunohost/search?q=%22yunohost.local%22&unscoped_q=%22yunohost.local%22

alexAubin:
I'm discover this small trick : https://serverfault.com/a/986437

That allows to easily define multiple .local alias for the machine

This way we can keep `yunohost.local` *and* `<hostname>`.local which imho is what we want because we want to be able to tell people that they can reach their yunohost using `yunohost.local` *and* that it keeps working even after the postinstall (otherwise that might be weird for people ... they just postinstalled the server and suddently the URL they were on doesn't work anymore)

zamentur:
I think this new PR https://github.com/YunoHost/yunohost/pull/1112 is more advanced on this topic, so i close this PR.

Thanks for your contribution :)

PR #1004, opened 2020-05-25
Test for args parsing 2

Psycojoker:
Then end of: https://github.com/YunoHost/yunohost/pull/989

idk if anyone really wants to review, it's boring as hell and was super boring to write.


alexAubin:
Yolomergin because it's "just" tests

PR #1006, opened 2020-05-26
Fix sury v99999

kay0u:
## The problem

Sury, again (the current solution seems to work, but it's not really conventional)

## Solution

Apply Pin recommended by @taziden 
I don't know if it's better this way, tell me

## PR Status

Tested with:
- Install nextcloud
- Install Weblate ([buster](https://github.com/YunoHost-Apps/weblate_ynh/tree/buster) branch)
- Install Snipeit ([fix-buster](https://github.com/YunoHost-Apps/snipeit_ynh/tree/fix-buster) branch)
- Install Synapse

No problems so far

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/apt
+++ b/data/helpers.d/apt
@@ -193,37 +193,17 @@ ynh_package_install_from_equivs () {
     LC_ALL=C equivs-build ./control 1> /dev/null
     dpkg --force-depends --install "./${pkgname}_${pkgversion}_all.deb" 2>&1)
 
-    # Let's try to see if install will work using dry-run. It it fails,
-    # it could be because the pinning of sury is blocking some package install
-    # c.f. for example: https://github.com/YunoHost/issues/issues/1563#issuecomment-623406509
-    # ... In that case, we use an ugly hack were we'll use a tweaked
-    # preferences.d directory with looser contrains for sury...
-    if ! ynh_package_install --fix-broken --dry-run >/dev/null 2>&1 && [ -e /etc/apt/preferences.d/extra_php_version ]
-    then
-        cp -r /etc/apt/preferences.d/ /etc/apt/preferences.d.tmp/
-        sed 's/^Pin-Priority: .*/Pin-Priority: 600/g' -i /etc/apt/preferences.d.tmp/extra_php_version
-        local apt_tweaks='--option Dir::Etc::preferencesparts=preferences.d.tmp'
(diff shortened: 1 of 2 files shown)


Josue-T:
Look like good.

Maybe the question would if we install synapse which need the libxml and if we also install an other php app which need php7.4-xml.

kay0u:
I guess it's related to this [issue](https://github.com/YunoHost-Apps/synapse_ynh/issues/193)
let's play with php7.4-xml:
```
root@domain:/ynh-dev# apt install php7.4-xml
php7.4-xml is already the newest version (7.4.6-1+0~20200514.17+debian9~1.gbp0f509e).
0 upgraded, 0 newly installed, 0 to remove and 5 not upgraded.

root@domain:/ynh-dev# apt-cache policy libxml2-dev
libxml2-dev:
  Installed: 2.9.9+dfsg-1+0~20200226.5+debian9~1.gbp3b6674
  Candidate: 2.9.9+dfsg-1+0~20200226.5+debian9~1.gbp3b6674
  Version table:
 *** 2.9.9+dfsg-1+0~20200226.5+debian9~1.gbp3b6674 995
        995 https://packages.sury.org/php stretch/main amd64 Packages
        100 /var/lib/dpkg/status
     2.9.4+dfsg1-2.2+deb9u2 500
        500 http://deb.debian.org/debian stretch/main amd64 Packages
        500 http://security.debian.org/debian-security stretch/updates/main amd64 Packages

root@domain:/ynh-dev# yunohost app install synapse 
```
and the logs of the install: (spoiler, everything is installed)
https://paste.yunohost.org/raw/viguzesahe

Josue-T:
Thanks for the feedback

kay0u:
the question is what to do with servers already screwed up by Sury?

alexAubin:
Doesn't this change already take care of the smooth migration of existing sury repo ? (Well, they'll have to run an app script that tries to install php stuff for it to actually be applied ... but it sounds precisely like the situation we're trying to fix ... though granted that it would be cleaner to have a "real" migration so that there's no sudden weird conf change just by attempting to install an app but I don't know if that's worth the effort of implementing a migration...)

alexAubin:
Alright folks I'm yolomerging because somebody again had an issue with sury today and I'm tired of that shit é_è

PR #1007, opened 2020-05-26
Add pre-defined DHE group and set up Nginx to use it

SohKa:
## The problem

By default, OpenSSL uses 1024-bit non-custom Diffie-Hellman primes (reported in [this issue](
https://github.com/YunoHost/issues/issues/93)). In consequence it weakens Nginx's security (reported in [this issue](https://github.com/YunoHost/issues/issues/1611)).

## Solution

As explained [here](https://github.com/YunoHost/issues/issues/93#issuecomment-633510357), Mozilla now recommends to use pre-defined DHE groups. By using them, it would fix both issues given above (in the "problem" section).

## PR Status

Tested on my server by following @alexAubin's instructions given [here](https://github.com/YunoHost/issues/issues/93#issuecomment-634075780). :smiley:

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


alexAubin:
ping @Josue-T : that might also be relevant for apps that need this like Synapse ?

Josue-T:
> ping @Josue-T : that might also be relevant for apps that need this like Synapse ?

Yes :+1: coturn, the turnserver of synapse need this. Not specifically synapse (probably because synapse only use elliptic curve).

SohKa:
I hesitated to ship the 3072-bit DHE group instead of the 2048-bit group. A lot of people advise to directly use the 3072-bit one (because the 2048-bit one will progressively phase out), but in the end I decided to stick with Mozilla's recommendation (2048-bit). Has anyone a strong opinion about this ?

alexAubin:
Hmmm idk it looks like it's not a big deal to update this later so I'm a huge partisan of sticking as much as we can to the Mozilla's recommendation so that we can avoid asking ourselves existential questions about security ¯\\\_(ツ)_/¯

SohKa:
Thanks for your improvements and the merge @alexAubin ! :)

PR #1010, opened 2020-05-28
[metronome config] Allow several other mime types to be uploaded

larchange:
https://github.com/YunoHost/issues/issues/1608

## The problem

metronome has a limited set of allowed mime types to be uploaded. It exclude some common files extensions such as pdf, doc, odt, and other.

Also conversations (android client) uses pgp extension when sending attached files with PGP encryption.

All the above will failed to be sent with the message: "Invalid filename or unallowed type"

## Solution

Change metronome's configuration file to include all these extensions
I added all the mime types on https://developer.mozilla.org/fr/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

and manually add pgp and apk

## PR Status

Tested on my yunohost instance
...

## How to test

- With any client try to send a pdf (or any other type not in metronome default list (/usr/lib/metronome/modules/mod_http_upload.lua))
- apply the change 
- try again

=======

- With conversations enable PGP encryption and try to send a picture
- apply the change
- Try again

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


larchange:
@pitchum this pull request is easy to test.
I don't think there is any drawbacks to allow all these files extensions. Except if @maranda  has something to say ? why not include this in metronome directly ?

pitchum:
After some tests I'm not very happy with this PR. Mainly because of the existence of that option *http_file_allowed_mime_types* in metronome.

1. other servers that I've tested (ejabberd and prosody) do not enforce such restrictions by default
2. the parameter actually only checks the file extension, not the mime type (it'd better be named *http_file_allowed_file_extensions* to avoid misinterpretation)
3. it still allows sharing a malicious file by changing the extension
4. it prevents sharing a legit file if the extension is not listed, people will have to learn how to rename their files before sharing (or they're just going to give up XMPP)
5. I don't want to maintain a list of allowed extensions over time, such a list will never be exhaustive enough

My opinion is that we should get rid of that restriction. As far as I can see it's not possible to disable this behaviour in metronome's configuration.

@maranda what are your thoughts about this?

maranda:
@pitchum the problem is that YunoHost insists on using the internal HTTP upload module instead of the external one, if you do recall I had already explained my concerns to you on that regard but they were refused.

However Metronome needs to have MIME list because it needs to at least check that the content you upload matches the one you requested and also restricts what gets uploaded so that your Metronome instance doesn't become a can of vermins.

I don't know what Ejabberd does but for Prosody I'm rather sure it just picks up the list from a file through the HTTP stack the problem is that it still doesn't do any, even partial, MIME assertion.
Which brings up the main problem of using the internal module, that proper HTTP servers don't have the size limitations on files that Metronome does have, could have proper static file processors, and otherwise perform much better.

So I repeat that the internal module needs to be replaced with the external one to offload the file servicing to NGINX that's the best solution. Also the internal module will soon be deprecated.

larchange:
@pitchum do you remember why you don't want the external one ?

If not I can work on the change, if I understand correctly it's just some nginx configuration and it should be transparent for the user.

@maranda so what you are telling us is that a list of mime types is necessary in order to have a clean metronome. Could it fallback to a default mime type (at least configurable) as yunohost doesn't want to maintain a list of what the users can or cannot send. My own opinion is to allow to send any files.

maranda:
@larchange see https://github.com/maranda/metronome/issues/523

maranda:
By the way the current default MIME types shipped, are the safest most commonly exchanged media formats. In my experience not much else gets sent.

larchange:
I agree that the list that I added in my commit is a bit big ! your commit (https://github.com/maranda/metronome/commit/2ad644bac0edd49e283eac516678dfad260c280e) is perfect. We don't need to maintain the full list in Yunohost.

I think we can add in Yunohost pgp, pdf and doc (and all office), odt (and all libreoffice) and maybe apk, tar.gz, zip and leave the rest up until a user complain.

For the internal vs external this PR is orthogonal, it will still be needed with external.

maranda:
@larchange no it won't because there's no MIMEs list mantained since that duty is transferred to the upstream HTTP service.

larchange:
OK understand. @pitchum what do you think ?

pitchum:
I'd be happy to test a PR introducing the external module :)

larchange:
@pitchum. OK give me a little bit of time i will submit the PR :)

zamentur:
@larchange What's the status of this PR

larchange:
@zamentur I'm not currently working on it.

data/templates/metronome/domain.tpl.cfg.lua
```diff
@@ -70,6 +70,84 @@ Component "xmpp-upload.{{ domain }}" "http_upload"
...
+      ["csh"] = "application/x-csh",
+      ["css"] = "text/css",
+      ["csv"] = "text/csv",
+      ["doc"] = "application/msword",
+      ["docx"] = "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+      ["eot"] = "application/vnd.ms-fontobject",
+      ["epub"] = "application/epub+zip",
+      ["gif"] = "image/gif",
+      ["gz"] = "application/x-compressed",
+      ["html"] = "text/html",
```
randomstuff, line 95: Wouldn't allowing HTML, XHTML and SVG open the server to [XSS from users](https://xmpp.org/extensions/xep-0363.html#server)?

larchange:
I opened a new PR  #1165 which solve this one

larchange:
this PR can be closed, it's not a really good solution, and it's superseded by  #1165

Issue #77, opened 2015-07-29
Dnsmasq IPv6


src386:
Hi,
Dnsmasq is configured with A, MX, TXT and SRV records but no AAAA.
It would be a nice feature to add.
Regards.

PR #1012, opened 2020-05-31
Update n to 6.5.1

ericgaspar:
## The problem
- N not up to date
- May not fix this issue [#77](https://github.com/YunoHost/package_check/issues/77)

## Solution
- Bring N up to date with upstream

## PR Status

...

## How to test

...

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -126,7 +126,7 @@ ynh_install_nodejs () {
     test -x /usr/bin/npm && mv /usr/bin/npm /usr/bin/npm_n
 
     # If n is not previously setup, install it
-    if ! test $(n --version > /dev/null 2>&1)
+    if ! $n_install_dir/bin/n --version > /dev/null 2>&1
     then
         ynh_install_n
     elif dpkg --compare-versions "$($n_install_dir/bin/n --version)" lt $n_version
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -129,7 +129,7 @@ ynh_install_nodejs () {
     if ! test $(n --version > /dev/null 2>&1)
     then
         ynh_install_n
-    elif dpkg --compare-version "$($n_install_dir/bin/n --version | cut -d" " -f2)" lt $n_version
+    elif dpkg --compare-versions "$($n_install_dir/bin/n --version)" lt $n_version
     then
         ynh_install_n
     fi
--- a/data/helpers.d/nodejs
+++ b/data/helpers.d/nodejs
@@ -125,7 +125,7 @@ ynh_install_nodejs () {
     test -x /usr/bin/npm && mv /usr/bin/npm /usr/bin/npm_n
 
     # If n is not previously setup, install it
-    if ! type $(n --version > /dev/null 2>&1)
+    if ! test $(n --version > /dev/null 2>&1)
     then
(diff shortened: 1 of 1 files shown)


data/helpers.d/nodejs
```diff
@@ -125,7 +125,10 @@ ynh_install_nodejs () {
     test -x /usr/bin/npm && mv /usr/bin/npm /usr/bin/npm_n
 
     # If n is not previously setup, install it
-    if ! test $(n --version > /dev/null 2>&1)
+    if ! type $(n --version > /dev/null 2>&1)
```
alexAubin, line 128: Not sure to understand why change `test` to `type` ? Note that if `n` is not installed, I'm guessing `$(n --version)` will return empty string, but `type` (with no arg) returns 0 and `test` (with no arg) returns 1

So it looks to me like the behavior changes in a unexpected way but maybe I'm missing something

ericgaspar: damn! this is a typo

data/helpers.d/nodejs
```diff
@@ -126,6 +126,9 @@ ynh_install_nodejs () {
 
     # If n is not previously setup, install it
     if ! test $(n --version > /dev/null 2>&1)
+    then
+        ynh_install_n
+    elif dpkg --compare-version "$($n_install_dir/bin/n --version | cut -d" " -f2)" lt "6.5.1"
```
kay0u, line 131: To avoid any problems, I would create a new `n_version` variable at the top of this file and use it where needed.

data/helpers.d/nodejs
```diff
@@ -126,6 +127,9 @@ ynh_install_nodejs () {
 
     # If n is not previously setup, install it
     if ! test $(n --version > /dev/null 2>&1)
+    then
+        ynh_install_n
+    elif dpkg --compare-version "$($n_install_dir/bin/n --version | cut -d" " -f2)" lt $n_version
```
kay0u, line 132: It's `--compare-versions` ;-)
And for me, there's no need to cut. Do you remember why you added it?

```suggestion
    elif dpkg --compare-versions "$($n_install_dir/bin/n --version)" lt $n_version
```

ericgaspar: aaah! one mistake per line 😱
I got it [here](https://github.com/YunoHost-Apps/Experimental_helpers/pull/52/files) (but the mistakes are mine...).
Thanks anyway for reviewing my crappy code.

data/helpers.d/nodejs
```diff
@@ -126,6 +127,9 @@ ynh_install_nodejs () {
 
     # If n is not previously setup, install it
     if ! test $(n --version > /dev/null 2>&1)
```
kay0u, line 129:
```suggestion
    if ! $n_install_dir/bin/n --version > /dev/null 2>&1
```

After some tests, i found that the current `test` doesn't work. n is reinstalled each time.
Can someone else test this helper please?

ericgaspar: @kay0u I have tested your code with this configurations:
- install `n` when `n` is not installed
- install `n` when `n`< 6.5.1
- do nothing when `n` already installed

It seems to work...

PR #1013, opened 2020-06-01
WIP: Refactoring yunohost format arguments parsing

Psycojoker:
## The problem

The current argument parsing code is rigged with quite some problems:

- everything is mixed together
- everything is parsed then everything is asked at once, this is very confusing and hard to extend
- all different kinds or arguments are parsed in the same place mixing all the things
- when "ask" is not present behavior is super weird
- code is hard to understand
- parsing individual arguments is closely tide to walking the data structure
- etc....

## Solution

After having written test for everything (oh god this was so freaking boring), refactor everything.

The approach is to factorize everything is a main generic parsing argument class then inherit from it for all arguments.

This will allow to extends this argument system and had new kind of arguments very easily.

Also everything is easier to read now and everything is nicely put in its small place.

And since we have tests we can refactor it again is needed in the future.

## PR Status

Tests pass and this is a very isolated part and now well tested part of the system so this shouldn't change anything at all for the users.

The only BREAKING change is that it will raises now if one app dev adds a default value to a password type of argument because I never want that.

## How to test

Launch tests or install an application.

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


locales/en.json
```diff
@@ -13,6 +13,7 @@
     "app_already_up_to_date": "{app:s} is already up-to-date",
     "app_argument_choice_invalid": "Use one of these choices '{choices:s}' for the argument '{name:s}'",
     "app_argument_invalid": "Pick a valid value for the argument '{name:s}': {error:s}",
```
kay0u, line 15: Please, remove this keys:

- `app_argument_invalid`
- `domains_available`
- `pattern_password_app`
- `users_available`

Psycojoker: I'm going to use them actually to make better error messages, just haven't found the energy yet.

Psycojoker: Normally done

Psycojoker:
Added some missing stuff for password and custom errors messages.

Should be way closer to being mergeable.

src/yunohost/app.py
```diff
@@ -2525,128 +2736,14 @@ def _parse_args_in_yunohost_format(user_answers, argument_questions):
...
-                    msignals.display(m18n.n('users_available'))
-                    users = user_list()['users']
-                    for user in users.keys():
-                        msignals.display("- {}".format(user))
-
-                    root_mail = "root@%s" % _get_maindomain()
-                    for user in users.keys():
-                        if root_mail in user_info(user)["mail-aliases"]:
-                            question_default = user
-                            break
```
alexAubin, line 2572: That stuff about default answer for user-type question disappeared during conflict/merge, looking at the new structure it should be pretty straightforward to reintegrate but busy right now with other reviews

alexAubin (changes requested):

src/yunohost/app.py
```diff
@@ -2514,6 +2514,217 @@ def _parse_args_for_action(action, args={}):
...
+
+
+class UserArgumentParser(YunoHostArgumentFormatParser):
+    argument_type = "user"
+
+    def parse_question(self, question, user_answers):
+        from yunohost.user import user_list
+
+        question = super(UserArgumentParser, self).parse_question(question, user_answers)
+        question.choices = user_list()["users"]
```
kay0u, line 2685:
```suggestion
        question.choices = user_list()["users"]
        if question.default is None:
            root_mail = "root@%s" % _get_maindomain()
            for user in question.choices.keys():
                if root_mail in user_info(user)["mail-aliases"]:
                    question.default = user
                    break
```

alexAubin: Yolo let's see, thanks for the fix !

src/yunohost/app.py
Psycojoker: Ah indeed, quite a lot of other PR has modified this code, sorry for not moving fast enough :< 

(don't hesitate to push on it ™)

alexAubin:
Tests are passing ! Yolomergin !

PR #1014, opened 2020-06-03
Fix regen-conf mechanism for SSH

alexAubin:
## The problem

Another tricky update for the regenconf ...

Realized today while helping a user that we have this funky behavior that if you run `yunohost tools regen-conf` it won't regen the conf of SSH. Digging further with ljf, we realize that there are many scenario with problematic behaviors.

Scenario 1 : a user setup a fresh instance, and during the install script, choose to *not* let yunohost manage the ssh conf
Problem: unless this admin eventually runs `yunohost tools regen-conf ssh --force`, there will be absolutely no info / reminder that the recommended ssh conf should be used. There won't even be a hash in regenconf.yml

Scenario 2 : we update the ssh configuration
Problem: when admins upgrades their server, the automatic `regen-conf` during upgrade won't update the ssh conf ... and the regen-conf won't even tell that there are pending changes to be applied or anything if you don't run `yunohost tools regen-conf` (even with --force) unless you explicitly specify --force

Scenario 3 : the admin manually edits the ssh conf
Problem: as for scenario 2, apart from the diagnosis screen saying the file got manually modified, running `regen-conf` won't tell anything about ssh.

## Solution

- Fix postinstall such that, if the admin chose to keep the original ssh conf, it will be seen as manually modified by the regenconf (instead of "nothing")
- For servers where there's no hash for the ssh conf in regenconf.yml, automatically add the pending conf hash.
- Running the regen-conf should always also regen the conf of ssh ...
- But in case the ssh conf got manually modified and the admin runs `yunohost tools regen-conf --force`, we don't overwrite the ssh conf (you need to specify ssh explicitly) ... I think that's mainly for stuff like, if you changed the ssh port, it could easily have catastrophic consequences if you run the `regen-conf --force` witthout thinking too much about it ...

## PR Status

Done, tested using provided tests

## How to test

Run the tests o/ ?

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


src/yunohost/regenconf.py
```diff
@@ -185,6 +186,20 @@ def _pre_call(name, priority, path, args):
...
         failed_regen = {}
 
+        # Here we are doing some weird legacy shit
+        # The thing is, on some very old or specific setup, the sshd_config file
+        # was absolutely not managed by the regenconf ...
+        # But we now want to make sure that this file is managed.
+        # However, we don't want to overwrite a specific custom sshd_config
+        # which may make the admin unhappy ...
+        # So : if the hash for this file does not exists, we set the hash as the
+        # hash of the pending configuration ...
```
kay0u, line 196:
```suggestion
        # hash of the current configuration ...
```

alexAubin: Because if we put the current hash, the conf is going to be replaced by the new configuration ... but that's not what we want, we want the conf to be flagged as manually modified

kay0u: I didn't understand the point of this hash, my bad

PR #1018, opened 2020-06-15
Simplify apt/php dependencies helpers

alexAubin:
## The problem

Following https://forum.yunohost.org/t/pdoexception-queryexception-could-not-find-driver/11848/18 I went berserk about the apt/php dependency helpers that are overcomplicated and error prone.

The explanation behind the story is that the app was using : 
```
ynh_install_php  # + some options to install some packages
[...]
ynh_install_app_dependencies "some other packages"
```

But `ynh_install_app_dependencies` behavior is such that it forgets about any dependency previously installed, which later led to the removal of the php dependency of the app (in this case, php-pgsql)

Apparently the "right" way is to use the `--package` option in `ynh_add_fpm_config` to install specific php dependency but imho that's neither obvious, semantic nor good sense to install some apt dependency at the beginning of the install, and some other at some other time - but only if they are not about the default php version ...

More generally, the current way to install apt dependency is hella confusing... There's 
- `ynh_install_app_dependency`
- `ynh_add_app_dependency`
- `ynh_install_extra_app_dependency`
- `ynh_install_php`
- `ynh_add_fpm_config --package`

(not to mention the associated removal helpers and add_repo and pin etc...)

This is creating overly complex stuff and bugs, both in apps (c.f. previous example + all the apps that need to split "regular" dependency and php dependency) and in the helpers themselves (spotted during this refactoring), and is hell to debug because there are dozens of call to apt and other things. 

While it's understandable because of the history of the whole thing and the lack clear strategy for sury, now that we have a better understanding and clarified what to do with php, we should simplify all of this before creating additional technical debt...

## Solution

If you need to install apt dependencies, you should call `ynh_install_app_dependencies`. Period. And you should be able to call it multiple time without it forgetting about previously installed deps. If sury needs to be installed, then it can be automatically determined because of the presence of a php7.X-foo in the list that does not corresponds to the vanilla php version.

This PR is therefore aiming to get rid of `ynh_add_app_dependency`, `ynh_install_php`, and the `--package` option of `ynh_add_fpm_config`. Since we can't do this in one go because some app depends on it, for now they are kept while app's code should be adapted ... then we can remove them in a follow-up PR.

## PR Status

Only partially tested on pixelfed for now but should be more tested to validate that there's no regression on other apps ...

## How to test

Zblerg try to install apps that depends on previously mentionned helpers

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/data/helpers.d/apt
+++ b/data/helpers.d/apt
@@ -360,7 +360,6 @@ ynh_add_sury() {
 #
 # usage: ynh_add_app_dependencies --package=phpversion [--replace]
 # | arg: -p, --package=     - Packages to add as dependencies for the app.
-# | arg: -r, --replace      - Replace dependencies instead of adding to existing ones.
 #
 # Requires YunoHost version 3.8.1 or higher.
 ynh_add_app_dependencies () {
@@ -368,24 +367,10 @@ ynh_add_app_dependencies () {
     local legacy_args=pr
     local -A args_array=( [p]=package= [r]=replace)
     local package
-    local replace
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
-    replace=${replace:-0}
 
-    local current_dependencies=""
-    if [ $replace -eq 0 ]
-    then
-        local dep_app=${app//_/-}	# Replace all '_' by '-'
-        if ynh_package_is_installed --package="${dep_app}-ynh-deps"
-        then
-            current_dependencies="$(dpkg-query --show --showformat='${Depends}' ${dep_app}-ynh-deps) "
-        fi
-
-        current_dependencies=${current_dependencies// | /|}
-    fi
-
-    ynh_install_app_dependencies "${current_dependencies}${package}"
+    ynh_install_app_dependencies "${package}"
 }
 
 # Remove fake package and its dependencies
@@ -450,7 +435,7 @@ ynh_install_extra_app_dependencies () {
     ynh_install_extra_repo --repo="$repo" $key --priority=995 --name=$name
 
     # Install requested dependencies from this extra repository.
-    ynh_add_app_dependencies --package="$package"
+    ynh_install_app_dependencies --package="$package"
 
     # Remove this extra repository after packages are installed
     ynh_remove_extra_repo --name=$app
--- a/data/helpers.d/apt
+++ b/data/helpers.d/apt
@@ -209,6 +209,8 @@ ynh_package_install_from_equivs () {
     ynh_package_is_installed "$pkgname"
 }
 
+YNH_INSTALL_APP_DEPENDENCIES_REPLACE="true"
+
 # Define and install dependencies with a equivs control file
 #
 # This helper can/should only be called once per app
@@ -247,6 +249,24 @@ ynh_install_app_dependencies () {
         # [^,]\+ matches all characters except ','
         # Ex: 'package>=1.0' will be replaced by 'package (>= 1.0)'
         dependencies="$(echo "$dependencies" | sed 's/\([^(\<=\>]\)\([\<=\>]\+\)\([^,]\+\)/\1 (\2 \3)/g')"
+    fi
+
+    # The first time we run ynh_install_app_dependencies, we will replace the
+    # entire control file (This is in particular meant to cover the case of
+    # upgrade script where ynh_install_app_dependencies is called with this
+    # expected effect) Otherwise, any subsequent call will add dependencies
+    # to those already present in the equivs control file.
+    if [[ $YNH_INSTALL_APP_DEPENDENCIES_REPLACE == "true" ]]
+    then
+        YNH_INSTALL_APP_DEPENDENCIES_REPLACE="false"
+    else
+        local current_dependencies=""
+        if ynh_package_is_installed --package="${dep_app}-ynh-deps"
+        then
+            current_dependencies="$(dpkg-query --show --showformat='${Depends}' ${dep_app}-ynh-deps) "
+        fi
+        current_dependencies=${current_dependencies// | /|}
+        dependencies="$current_dependencies $dependencies"
     fi
 
     #
@@ -290,6 +310,9 @@ EOF
     rm /tmp/${dep_app}-ynh-deps.control
     ynh_app_setting_set --app=$app --key=apt_dependencies --value="$dependencies"
 }
+
+
+
 
 # Add dependencies to install with ynh_install_app_dependencies
 #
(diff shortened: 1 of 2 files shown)


data/helpers.d/apt
```diff
@@ -332,7 +382,26 @@ ynh_add_app_dependencies () {
 # Requires YunoHost version 2.6.4 or higher.
 ynh_remove_app_dependencies () {
     local dep_app=${app//_/-}	# Replace all '_' by '-'
+
+    local current_dependencies=""
```
kay0u, line 386: This variable is not used? Could you remove the `if ynh_package_is_installed --package="${dep_app}-ynh-deps"`?

data/helpers.d/apt
```diff
@@ -288,39 +320,57 @@ EOF
...
+
     ynh_app_setting_set --app=$app --key=apt_dependencies --value="$dependencies"
+
+    if [[ -n "$specific_php_version" ]]
+    then
+        # Set the default php version back as the default version for php-cli.
+        update-alternatives --set php /usr/bin/php$YNH_DEFAULT_PHP_VERSION
+
+        # Store phpversion into the config of this app
+        ynh_app_setting_set $app phpversion $specific_php_version
```
kay0u, line 332:
```suggestion
        ynh_app_setting_set --app=$app --key=phpversion --value=$specific_php_version
```

data/helpers.d/php
```diff
@@ -335,43 +335,13 @@ ynh_install_php () {
...
-    # Pin this extra repository after packages are installed to prevent sury of doing shit
-    for package_to_not_upgrade in "php" "php-fpm" "php-mysql" "php-xml" "php-zip" "php-mbstring" "php-ldap" "php-gd" "php-curl" "php-bz2" "php-json" "php-sqlite3" "php-intl" "openssl" "libssl1.1" "libssl-dev"
-    do
-        ynh_pin_repo --package="$package_to_not_upgrade" --pin="origin \"packages.sury.org\"" --priority="-1" --name=extra_php_version --append
-    done
-
-    # Advertise service in admin panel
-    yunohost service add php${phpversion}-fpm --log "/var/log/php${phpversion}-fpm.log"
+    ynh_install_app_dependencies "$package"
+    ynh_app_setting_set $app phpversion $phpversion
```
kay0u, line 344:
```suggestion
    ynh_app_setting_set --app=$app --key=phpversion --value=$specific_php_version
```

But this key should already be added in `ynh_install_app_dependencies`

fflorent:
> This PR is therefore aiming to get rid of ynh_add_app_dependency, ynh_install_php, and the --package option of ynh_add_fpm_config. Since we can't do this in one go because some app depends on it, for now they are kept while app's code should be adapted ... then we can remove them in a follow-up PR.

If I testifying can help, I couldn't restore an application (Garradin) because of this confusion.

The app installs debian dependencies using `ynh_add_fpm_config`, and use `ynh_remove_fpm_config` in the remove script without removing the package list (still, the issue is understandable as the remove script should be symmetric with the install script, so we can expect `ynh_remove_fpm_config` to remove them).

Thus, this is what happened to me:
 - I wanted to upgrade Garradin;
 - There were some issues with the upgrade due to package dependencies (which is another problem);
 - Yunohost removed Garradin, but not completely as the debian package named `garradin-ynh-deps` (containing deb dependencies for Garradin) wasn't removed;
 - Yunohost attempted to restore the backup, but it failed because there were conflicts due to `garradin-ynh-deps`;
 - Result: Yunohost was unable to restore the backup after an upgrade failure.

I would warmly welcome this change after a code review, any status of this PR?

data/helpers.d/apt
```diff
@@ -282,39 +320,57 @@ EOF
...
+    then
+        # Set the default php version back as the default version for php-cli.
+        update-alternatives --set php /usr/bin/php$YNH_DEFAULT_PHP_VERSION
+
+        # Store phpversion into the config of this app
+        ynh_app_setting_set --app=$app --key=phpversion --value=$specific_php_version
+
+        # Integrate new php-fpm service in yunohost
+        yunohost service add php${specific_php_version}-fpm --log "/var/log/php${phpversion}-fpm.log"
+    fi
```
alexAubin, line 336:
```suggestion
    else
         ynh_app_setting_set --app=$app --key=phpversion --value=$YNH_DEFAULT_PHP_VERSION
    fi
```

We probably want to set the phpversion setting even if using Debian's default version ?

data/helpers.d/apt
```diff
@@ -281,39 +320,61 @@ EOF
...
+ynh_add_sury() {
+
+    # Add an extra repository for those packages
+    ynh_install_extra_repo --repo="https://packages.sury.org/php/ $(ynh_get_debian_release) main" --key="https://packages.sury.org/php/apt.gpg" --name=extra_php_version --priority=600
+
+    # Pin this extra repository after packages are installed to prevent sury from doing shit
+    for package_to_not_upgrade in "php" "php-fpm" "php-mysql" "php-xml" "php-zip" "php-mbstring" "php-ldap" "php-gd" "php-curl" "php-bz2" "php-json" "php-sqlite3" "php-intl" "openssl" "libssl1.1" "libssl-dev"
+    do
+        ynh_pin_repo --package="$package_to_not_upgrade" --pin="origin \"packages.sury.org\"" --priority="-1" --name=extra_php_version --append
+    done
```
alexAubin, line 357: Ah maybe we can remove this because it's now managed in the core iirc ?

alexAubin:
>Also I wonder if the control file could not be reset by the python script right before invoking the upgrade script. Thus `YNH_INSTALL_APP_DEPENDENCIES_REPLACE` wouldn't be useful, right?

Hmyeah we could do that, but I'm worried that this leave a lot of possibilities for things to wrong, in the sense that typically an upgrade script will look like : 

```
1. fetch settings
2. perform safety backup
3. start upgrading the app ... and somewhere around there (re)do the ynh_install_app_dependencies
```

but then if the safety backup crashes (or anything else before ynh_install_app_dependencies - one could argue that the safety backup should be managed in the core oneday™ but that's not the point ;P), apt/dpkg will end up in a state where there are dangling dependencies that may get auto-removed ... Instead it's more robust / less issue-prone to replace or "actualize" dependencies in sort of an atomic way.

(Ooooooooor the whole apt dependency stuff could be managed in a declarative way in the json, but that doesnt cover cases where dependency may dynamically depends on conditions like debian flavor or app options, and then it gets complicated :P)

PR #1020, opened 2020-06-15
Backup to uncompressed tar by default

alexAubin:
## The problem

When people have large amount of data to backup, we can safely assume that these data are media like picture or video which are most probably already compressed ... Therefore, using compressed .tar.gz archives compared to uncompressed .tar is expected to be a huge waste of CPU and time while at the same time increasing the risk of corrupted archive.

I ran a small test on a VPS : 

```
# Install a ~dummy app
yunohost app install hextris

# Add a 1ish GB file from /dev/urandom to simulate large amount of already-compressed data
dd if=/dev/urandom of=/var/www/hextris/pwet bs=10M count=100
```

Backing-up the app to a .tar.gz archive took about 75 sec

Compare to 10 sec when backing up to a .tar archive. 

The size of both output was basically the same (random data are already 'maximally compressed')

## Solution

Backup to .tar file (instead of .tar.gz) by default ... Though we gotta keep compatibility with previous archives of course @_@

## PR Status

Only tested on simple case so far ... will see what the automatic tests say

## How to test

...


alexAubin:
To be discussed later™, as reported by Kayou during today's meeting, possibly : 
- have a way to switch to compressed tar ? (possibly using the backup method mechanism...)
- maybe have a global setting to change the default backup method "permanently"

src/yunohost/settings.py
```diff
@@ -71,6 +71,7 @@ def is_boolean(value):
         "choices": ["intermediate", "modern"]}),
     ("pop3.enabled", {"type": "bool", "default": False}),
     ("smtp.allow_ipv6", {"type": "bool", "default": True}),
+    ("backup.compress_tar_archives", {"type": "bool", "default": False}),
```
zamentur, line 74: Missing description ?

zamentur (approved):
LGTM

PR #1021, opened 2020-06-17
create admin home folder

kay0u:
## The problem

After post install the folder `/home/admin` doesn't exist.

## Solution

Create it as we do for a regular user: https://github.com/YunoHost/yunohost/blob/6c2163589fd684002bd9f2724b500ffd2dd30d19/src/yunohost/user.py#L209-L215

## PR Status

...

## How to test

...


alexAubin:
Yolomerging, we'll see if the tests are happy (for now they are failing because ran in stretch image if i understand correctly)

PR #860, opened 2019-12-06
[enh] Sort alphabetically the domain list

cyxae:
## The problem

Domains are randomly displayed when running "yunohost domain list"
 yunohost/issues#1355

## Solution

Sort domains alphabetically, by top domains

## PR Status

...

## How to test

Add many domains and subdomains
Run "yunohost domain list"

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #1022, opened 2020-06-21
Add default domain to API response

Salamandar:
## The problem

The web admin interface needs to know the default domain to select it in the drop down lists. (app install).

https://github.com/YunoHost/issues/issues/1633

...

## Solution

Add a `default` value that contains the index of the default domain in the list.

...

## PR Status
Needs another PR on yunohost-admin repo.
...

## How to test

Just try to install a new app, the domain selection list should be alphabetically sorted and the main domain should be selected.
...


src/yunohost/domain.py
```diff
@@ -61,7 +61,11 @@ def domain_list(exclude_subdomains=False):
                 continue
         result_list.append(domain)
 
-    return {'domains': result_list}
+    result_list.sort()
```
Psycojoker, line 64: You can use `sorted(result_list)` bellow to save a line if you want (but that won't be a in-place sort)

Salamandar: I think I prefer as it is, because result_list is used twice below (and we don't want sort() to run twice, do we ? ;) )

zamentur:
I am not sure the cli behaviour is what we want:
```
root@ynh:/ynh-dev# yunohost domain list
default: 0
domains: ynh.local
```
In cli we might put default: ynh.local or just nothing... What's you point of view ?

src/yunohost/domain.py
```diff
@@ -61,7 +61,11 @@ def domain_list(exclude_subdomains=False):
...
+    return {
+        'domains': result_list,
+        'default': result_list.index(_get_maindomain())
```
zamentur, line 67: I think this index research should be done in web admin to improve cli display AND avoid postinstall failure (it search the index of 'yunohost.org' which is not in domains list).
```suggestion
        'default': _get_maindomain()
```

zamentur (changes requested):

zamentur:
Linked to https://github.com/YunoHost/yunohost-admin/pull/306

zamentur:
We should merge this way to sort domain i think: https://github.com/YunoHost/yunohost/pull/860/files

src/yunohost/domain.py
alexAubin:
```suggestion
```

Domain sorting is to be addressed by https://github.com/YunoHost/yunohost/pull/860

src/yunohost/domain.py
```diff
@@ -61,7 +61,10 @@ def domain_list(exclude_subdomains=False):
                 continue
         result_list.append(domain)
 
-    return {'domains': result_list}
+    return {
+        'domains': result_list,
+        'default': _get_maindomain()
```
alexAubin, line 66:
```suggestion
        'main': _get_maindomain()
```

The semantic "main" should be kept (though iirc we kind of use "default domain" in places in the webadmin views but in cli we have for example `yunohost tools main-domain` etc... )

PR #1023, opened 2020-06-21
Remove files still present after domain removal

Salamandar:
## The problem
Some files (DKIM keys, history of ssl keys, nginx config) are still present after domain removal.
This aims to fix this.

https://github.com/YunoHost/issues/issues/1629

The main issue with this fix is that the creation of those file are in `hooks.d`, but hooks don't provide a removal process, so it has to be in python… I would have prefered creation and deletion of those files to be in the same source files.


zamentur (approved):
LGTM, tested

src/yunohost/domain.py
```diff
@@ -221,6 +221,29 @@ def domain_remove(operation_logger, domain, force=False):
...
+    # Remove metronome
+    os.system('rm -rf /var/xmpp-upload/{}'.format(domain))
+    domain_encoded = quote(domain, safe='').lower().replace('.', '%2e').replace('-','%2d').replace('_', '%5f')
+    os.system('rm -rf /var/lib/metronome/muc%2e' + domain_encoded)
+    os.system('rm -rf /var/lib/metronome/pubsub%2e' + domain_encoded)
+    os.system('rm -rf /var/lib/metronome/vjud%2e' + domain_encoded)
+    os.system('rm -rf /var/lib/metronome/xmpp%2dupload%2e' + domain_encoded)
+    os.system('rm -rf /var/lib/metronome/{}'.format(domain_encoded))
+
     regen_conf(names=['nginx', 'metronome', 'dnsmasq', 'postfix'])
```
alexAubin, line 247: Idk what to think about this ... I understand the motivation and necessity to do something about this, but as you explain in the original post, the initial creation is done by the regen_conf hooks. The deletion procedure which you are thinking about is actually a trick of creating an empty file (you'll see some `touch $pending_dir/etc/foo/bar` ..)

In its current state, I suspect this code will lead to a bunch of file flagged as "manually modified/removed" by the regenconf - though tests should be ran to confirm that hypothesis...

Anyway, it'd really be preferable to have this managed in a consistent way inside the regenconf hooks - though clearly it's not easy to implement because some tricks have to be found to identify files that correspond to old-domain-that-you-don't-know-anymore-existed ... ideally would have to implement a new mechanism in the regen conf for this >_>

alexAubin (changes requested):

src/yunohost/domain.py
zamentur: None of these files or folders are currently managed by regenconf.yml (i made the test).
But i agree with the idea it should be distribute inside some action in regenconf hooks or something like that.

alexAubin: Yeah they might not appear in regenconf.yml, but still some stuff like dkim keys or XMPP-related folders are created by the regenconf hooks anyway (dunno what to think about this)

for example : https://github.com/YunoHost/yunohost/blob/dev/data/hooks/conf_regen/31-rspamd#L29

Neustradamus:
Any news on it?

PR #1029, opened 2020-07-12
[enh] Force HTTPOnly and Secure for all cookies

zamentur:
## The problem

HTTP allow to specify some secure options about cookies to avoid some attacks.

## Solution

Here i suggest to add HTTPOnly and secure for each cookies of all apps.

I am not sure of this approach, we need to think it with several examples of apps

## PR Status

To discuss

## How to test

...

### DIFF
--- a/data/templates/nginx/security.conf.inc
+++ b/data/templates/nginx/security.conf.inc
@@ -30,6 +30,9 @@ more_set_headers "X-Download-Options : noopen";
 more_set_headers "X-Permitted-Cross-Domain-Policies : none";
 more_set_headers "X-Frame-Options : SAMEORIGIN";
 
+# Force HTTPOnly and Secure for all cookies
+proxy_cookie_path ~$ "; HTTPOnly; Secure;";
+
 # Disable gzip to protect against BREACH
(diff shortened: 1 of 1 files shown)


Psycojoker:
Can you provide the related document you've used for deciding of that PR plz?

zamentur:
I wrote this following advice of a person i meet and who knows a lot of things about xss injection.

"HTTPOnly" avoid the use of cookies on client side.
About HTTPOnly, it might broke some apps using cookies client side, but without, it could be possible to exploit injection to read some auth cookies and gain a complete access on an app.

"Secure" avoid a cookie to be sent if we are in an http session. 
Normally, HSTS protect all browsers which have connected through https before. And we put a redirection several cases (all ?)
I think secure could be dangerous for yunohost with a reverse proxy in front and with an hacked nginx conf to send all in http. But we don't know exactly how secure are this kind of setup, as the https is managed by the reverse proxy.

alexAubin:
(Turns out https://infosec.mozilla.org/guidelines/web_security does recommend at least putting 'Secure' on *all* cookies - not sure about about HTTPonly)

alexAubin:
To be addressed in https://github.com/YunoHost/yunohost/pull/1290

PR #1034, opened 2020-08-05
[zsh] Addition of a completion script automatically generated

buzuck:
## The problem
There is no auto-completion suggestions for ZSH.

## Solution
Add auto-completion to ZSH, by generating that completion file from
`data/actionmap/yunohost.yml`.

This completion script makes use of the caching capabilities of ZSH to provide elements of completion that can take some time to generate, such as the list of the applications.

## PR Status

- Generation of the completion script done in a development environment
- Tested on my YunoHost production instance, v3.x
- This is the first time I tried creating a ZSH completion script. If anyone has some experience in this, any comment is welcome! :slightly_smiling_face: 

There is also a few comments and todos in the headers of `yunohost.yml` and `zsh_completion.py`.

_I tried to keep the code commented. I hope it is not overcomplicated, and that is quite readable..._ :sweat_smile: 

## How to test

The completion file is created with:
```
python3 zsh_completion.py [--debug <OUTPUT_FILE>]
```

If no argument is given, the script tries to write the completion script into `/usr/local/share/zsh/site-functions/_yunohost`. The targeted directory has to be present in the variable `$fpath` for ZSH to find the completion script.

Install zsh with apt, enter the zsh shell and then run `compinstall` with the default values to enable completion.

## Validation

* [ ] Principle agreement 0/2 :
* [ ] Quick review 0/1 :
* [ ] Simple test 0/1 :
* [ ] Deep review 0/1 :
  * [ ] Items added to `yunohost.yml`
  * [ ] Overall architectural code choices for `zsh_completion.py`
  * [ ] Code style, naming, ...
  * [ ] Validate the caching mechanism of ZSH
  * [ ] _Optional: Review the ZSH completion script_


data/actionsmap/zsh_completion.py
```diff
@@ -0,0 +1,665 @@
...
+# Link about this globbing system:
+# http://zsh.sourceforge.net/Doc/Release/Expansion.html#Filename-Generation
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# - Optimization:
+#   - caching mecanism: invalidate the cache afer some commands? Hard, the
+#   cache is local to user
+#   - implement a zstyle switch, to change the cache validity period?
+
+ZSH_COMPLETION_FILE = '/usr/local/share/zsh/site-functions/_yunohost'
```
alexAubin, line 37: Hey there !

Small issue to fix for the debian packaging : that file won't be packaged. Instead, if we stick to something similar what's done for the bash completion : the output should go to some folder inside the yunohost/ tree
([for example bash completion goes into bash-completion.d/yunohost](https://github.com/YunoHost/yunohost/blob/dev/data/actionsmap/yunohost_completion.py#L16)) then gotta add [a rule inside debian/install to tell it the source/destination during the actual install of the .deb](https://github.com/YunoHost/yunohost/blob/dev/debian/install#L3)

buzuck: Thanks! I didn't see this Debian packaging aspect. I just pushed something in this regard.

buzuck:
While going through the Debian's packaging file, I saw Jinja2 is required 
've also seen that jinja is required by YunoHost. I refrained using it, as I wanted to have as little dependencies as possible. Jinja could definitely help generating this ZSH completion file, but I have no idea if this lib is available when the package is being build :confused:

alexAubin:
Ah yes indeed

What matters here is wether or not the lib is in Build-Depends : 

https://github.com/YunoHost/yunohost/blob/dev/debian/control#L5

And you can see that it is ;) We already use it at build time ... not sure to remember what for, maybe some doc generation

buzuck:
Great! I'll be rewriting my Python code then :smiley: 
----
And perchance, If anyone around has a knowledge of the ZSH completion framework, I would be interested to get a review on this topic, as this is my very first ZSH completion plugin :sweat_smile:

alexAubin:
Closing because the PR has been stuck for 2 or 3 years ... though not 100% clear to me ... Feel free to complain to re-open if there's intent to finish it though ...

PR #1035, opened 2020-08-06
enable custom relayhost for postfix

khimaros:
## The problem

Fixes https://github.com/YunoHost/issues/issues/1649

## Solution

New settings for relayhost, sasl user, and sasl password with accompanying template values.

## PR Status

WIP. Tested manual setup on Google Cloud f1-micro instance on **stretch-stable**. my primary concern is if `settings` are an appropriate way to store secrets in yunohost. fine for api keys, but probably not so for passwords.

## How to test

set up your smtp relay, eg. with sendgrid. create api keys/sasl credentials.

```
# yunohost settings set smtp.relayhost '[smtp.sendgrid.net]:587'
# yunohost settings set smtp.relayhost_user 'apikey'
# yunohost settings set smtp.relayhost_pass '<YOUR_API_KEY>'

# yunohost service regen-conf postfix
# systemctl restart postfix
```
and then send an email from your host.


alexAubin:
Eh that looks like duplicate work of https://github.com/YunoHost/yunohost/pull/773 ? :confused:

khimaros:
Nice catch, @alexAubin. This does appear to be duplicate work and an almost identical approach. Actually, the similarity is encouraging as a new contributor as it validates my mental model!

I searched for an issue matching this description and couldn't find one so I filed https://github.com/YunoHost/issues/issues/1649 -- in hindsight, I should have also searched PRs before starting this work.

I'm happy to pickup this work if the original contributor is unresponsive, or if you'd prefer to target buster-unstable (as this PR does) instead of stretch-unstable. According to @Psycojoker, "I don't think we'll backport new features to stretch unstable"

alexAubin:
Hey there,

sorry for the delay / slow interaction. That's indeed pretty close to the other PR :stuck_out_tongue_winking_eye:. Though unfortunately I would tend to favor the old one (#773) because it's already more release-ready (this one would have a few technical details to polish, e.g. the regen-conf script shouldn't edit directly files in /etc/ but rather in $pending_conf but granted that it's not easy to understand from a first-contributor point of view because the regen conf mechanism is tricky... nevertheless overall things looked pretty good)

Nevertheless comparing the two, there are a few interesting things to discuss : 

1.

```
{% if smtp_relayhost %}
smtp_tls_security_level = encrypt
{% else %}
smtp_tls_security_level = may
{% endif %}
```

Could you elaborate on why you tweak this param depending on the relayhost ? Is that because you're only effectively interacting with the relay and you know it can encrypt, so better enforce encryption ?

2. in `smtp_tls_exclude_ciphers` you changed `aNULL` to `NULL` but no idea what's that ?

3. You added `smtp_sasl_tls_security_options = noanonymous` in addition to `smtp_sasl_security_options`, do you know what/why ? (I should probably just RTFM for that one)

khimaros:
> Hey there,
> 
> sorry for the delay / slow interaction. That's indeed pretty close to the other PR stuck_out_tongue_winking_eye. Though unfortunately I would tend to favor the old one (#773) because it's already more release-ready (this one would have a few technical details to polish, e.g. the regen-conf script shouldn't edit directly files in /etc/ but rather in $pending_conf but granted that it's not easy to understand from a first-contributor point of view because the regen conf mechanism is tricky... nevertheless overall things looked pretty good)
> 

It seems that activity on that PR has dropped off.

Would you like me to incorporate this feedback into my PR and move forward with merge?

> Nevertheless comparing the two, there are a few interesting things to discuss :
> 
> 
> ```
> {% if smtp_relayhost %}
> smtp_tls_security_level = encrypt
> {% else %}
> smtp_tls_security_level = may
> {% endif %}
> ```
> 
> Could you elaborate on why you tweak this param depending on the relayhost ? Is that because you're only effectively interacting with the relay and you know it can encrypt, so better enforce encryption ?
> 

That's the idea. According to [docs](http://www.postfix.org/postconf.5.html#smtp_tls_security_level): "[encrypt] is not an appropriate default for systems delivering mail to the Internet."

>     1. in `smtp_tls_exclude_ciphers` you changed `aNULL` to `NULL` but no idea what's that ?

I misinterpreted this as a typo and should have confirmed. It looks like `aNULL` is the correct value here, however, it might be worth considering `smtp_tls_mandatory_ciphers = medium` or `high` here in the relayhost case.

>     2. You added `smtp_sasl_tls_security_options = noanonymous` in addition to `smtp_sasl_security_options`, do you know what/why ? (I should probably just RTFM for that one)

I cargo-cult-ed this from a few examples I found for using an SMTP relay for mail routing. It looks like [default](http://www.postfix.org/postconf.5.html#smtp_sasl_security_options) is `noplaintext, noanonymous`, so this can be removed.

alexAubin:
Soooo I merged https://github.com/YunoHost/yunohost/pull/773 to move forward with this after @ljf integrated the small change regarding `smtp_tls_security_level = encrypt`

Sorry for the whole confusion and delay about reimplementing something that was already done a PR (the PR was asleep mostly because we were focusing on Buster since a few months but now trying to catch up with stuff). Even though that was a perfectly legit and good first contribution :|. Nevertheless we can include you in the contributor for that release since your work kind of counts as a at least a commit / review :stuck_out_tongue_winking_eye: (and hopefully trying to be symbolicly thankful)

PR #1037, opened 2020-08-17
Update acme_tiny to 4.1.0

kay0u:
## The problem

acme_tiny is not up to date

## Solution

Update acme_tiny to the [latest version](https://github.com/diafygi/acme-tiny/releases)

## PR Status

Not tested, but we should keep this lib up to date, right?

## How to test

...


alexAubin:
Eeeeh yolomergin ?

PR #1041, opened 2020-08-28
Add apt conf regen hook

alexAubin:
## The problem

There are some weird inconsistencies related to the sury pinning policy and php package installs ... that's complicated to explain and i'm half drunk but basically it makes everything simpler and consistent to just manage the sury pinning policy from a regen-conf hook (even if sury isn't enabled ... it doesn't hurt to set the pinning policy anyway)

That way we can make sure that in the future, if we have to update the pinning policy, there's no weird stuff required about "having to re-run an install or upgrade of an app that uses sury for the pinning policy to get upgraded". It just gets upgraded because it's handled by the regenconf hook no matter what.

## Solution

Implement a regenconf hook for apt, more specifically to handle the apt pinning policy

## PR Status

Done, tested and working on my side

## How to test

Run the regen-conf, check the content of /etc/apt/preferences.d/extra_php_version


kay0u (approved):
I like the idea! Cheers 🍻

PR #1050, opened 2020-09-10
[fix] Reduce right given to ynh users with ssh

zamentur:
## The problem

...

## Solution

Forbid standard user to use several ssh feature. Only admin and root are allowed to do that.
Remove 'sftpusers' unknown user

## PR Status

Ready

## How to test

...

### DIFF
--- a/data/templates/ssh/sshd_config
+++ b/data/templates/ssh/sshd_config
@@ -70,20 +70,11 @@ Subsystem sftp internal-sftp
 # Disallow unallowed users to use SSH as a VPN-like or a X11 server
 AllowTcpForwarding no
 AllowStreamLocalForwarding no
-GatewayPorts no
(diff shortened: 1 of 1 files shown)


data/templates/ssh/sshd_config
```diff
@@ -66,12 +66,26 @@ AcceptEnv LANG LC_*
...
+# Disallow unallowed users to use SSH as a VPN-like or a X11 server
+AllowTcpForwarding no
+AllowStreamLocalForwarding no
+GatewayPorts no
+X11Forwarding no
+PermitTTY no
+PermitUserRC no
+
+# Chroot standard users
+ChrootDirectory /home/%u
```
zamentur, line 79: May be this could broke some user habits...

PR #1054, opened 2020-09-19
Fix ynh_app_upstream_version : restore ability to read manifest

autra:
## The problem

This function was supposed to be able to read from the manifest (see usage and documentation), but actually was only parsing an env var, thus breaking ynh_check_app_version_changed

## Solution

Restoring the capacity to read a manifest

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -417,7 +417,7 @@ ynh_read_manifest () {
     jq ".$manifest_key" "$manifest" --raw-output
 }
 
-# Read the upstream version from the manifest
+# Read the upstream version from the manifest, or from the env variable $YNH_APP_MANIFEST_VERSION if not given
 #
 # usage: ynh_app_upstream_version [--manifest="manifest.json"]
 # | arg: -m, --manifest=    - Path of the manifest to read
@@ -437,7 +437,13 @@ ynh_app_upstream_version () {
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
 
-    version_key=$YNH_APP_MANIFEST_VERSION
+    if [[ "$manifest" != "" ]] && [[ -e "$manifest" ]];
+    then
+      version_key=$(ynh_read_manifest --manifest="$manifest" --manifest_key="version")
+    else
+      version_key=$YNH_APP_MANIFEST_VERSION
+    fi
+
     echo "${version_key/~ynh*/}"
 }
 


alexAubin (approved):
Super quick review, LGTM

Anybody from @YunoHost/apps has some additional insight about this ?

Josue-T:
Hello,

Can you explain more why it broke `ynh_check_app_version_changed`. Don't really understand why...

autra:
Hi @Josue-T, sure! It is called [here with the -m flag](https://github.com/YunoHost/yunohost/blob/dev/data/helpers.d/utils#L493), so the calling function expects it to read the manifest, and [here](https://github.com/YunoHost/yunohost/blob/dev/data/helpers.d/utils#L495) where it's actually expecting to read the env variable.

This is more a quickfix than anything here. I guess <span class="pl-en">ynh_check_app_version_changed</span> at least should be deprecated as we have now this info in an env variable. But maybe some scripts relies on ynh_app_upstream_version, so I chose to fix it instead.

Josue-T:
Well OK but I don't understand what it change as normally these two things should have exactly the same result.

autra:
> Well OK but I don't understand what it change as normally these two things should have exactly the same result.

I don't understand what you don't understand 😆 : what are those "things that should have the same results" that you are referring to?

In short, we need to compare 2 upstream versions of an app package : the current one, and the one that will get installed. `YNH_APP_MANIFEST_VERSION` contains the *future* app version. 

Maybe I'm gonna comment the code of [the relevant part of `ynh_check_app_version_changed`](https://github.com/YunoHost/yunohost/blob/bca4cd0c51ad6452ca66af7afac815f56b7651e9/data/helpers.d/utils#L492) to be more explicit:

```python
    # this gets the *current* app version
    local current_version=$(ynh_read_manifest --manifest="/etc/yunohost/apps/$YNH_APP_INSTANCE_NAME/manifest.json" --manifest_key="version" || echo 1.0)
    # /!\ BUG HERE => this *wants* to get the upstream part of the current app version, but fails without my patch
    local current_upstream_version="$(ynh_app_upstream_version --manifest="/etc/yunohost/apps/$YNH_APP_INSTANCE_NAME/manifest.json")"
    # this get the *future* app version, and btw defaulting to 1.0 is *not* a good idea imho
    # and WILL bite us very severely when we don't expect to
    local update_version=$(ynh_read_manifest --manifest="../manifest.json" --manifest_key="version" || echo 1.0)
    # this gets the *future* upstream app version, because it is in the YNH_APP_MANIFEST_VERSION env var set by app.py#app_upgrade before calling this script
    local update_upstream_version="$(ynh_app_upstream_version)"
```

In other word, we call `ynh_app_upstream_version` twice, but only the second time works correctly.

Josue-T:
Oh I got the issue now. Thanks.

Josue-T (approved):

PR #1058, opened 2020-10-03
Fix redis hook (bash does not allow empty function)

Chostakovitch:
## The problem

The redis hook script contains an empty bash function. Bash does not support empty function.

## Solution

Add a dummy instruction inside the function (equivalent to true)

## PR Status

Done

## How to test

`yunohost tools regen-conf redis`


alexAubin:
Yolomerging !

PR #1060, opened 2020-10-08
We don't need app-type argument for manifest questions ?

alexAubin:
## The problem

Weeell we have 200 lines of code to test it but ... this feature is used by literally no app ever ... (at least in the app catalog). Proposing to remove it.

To give a bit of context : this was identified by axolotle who's working on a bit refactoring of the webadmin, and found that is was painful to handle this arg type (because then you gotta make an html select filled with the available apps, and just meh.)

## Solution

Buuuurn the unused features :|

## PR Status

Done / yolocommitted

## How to test

Uh run all the other tests I guess ?


Psycojoker:
iirc that was for when we had this agenda backend that needed to choose between 2 agenda server (and that was confusing everyone)

I would really had prefer that we removed that before I had to write all those super painful tests /o\

alexAubin:
Yeah I feel a bit bad just deleting 200 lines of code that looks like super hard work :|

Psycojoker:
Worst case we can restore it no? It was a lot of (super mega) boring copy/pasta and tuning

PR #1062, opened 2020-10-09
Refactor local CA management for more consistency and simpler postinstall

alexAubin:
# The problem

Postinstall is a bit lengthy and complex and the local CA management is a huge mess with bits everywhere. 

## Solution

This is an attempt to simplify / uniformize most aspects of the local CA management, and also one of the first pieces towargs heavy simplification of the postinstall process as a whole

## PR Status

Kind of tested but I'm battling with slapd which refuses to let me re-run the postinstall -_- ...

## How to test

Run the postinstall and also test changing the main domain, see if that's correctly propagated on the local CA


zamentur:
I set work needed as CI checks failed

PR #1064, opened 2020-10-11
Move ldap initialization before the postinstall

alexAubin:
## The problem

Another poke for the "postinstall is too complex and lengthy". There's no necessity to run the LDAP initialization during the postinstall, and also it's complicated to properly reset and rerun the postinstall when it failed for some reason... Also initializing the LDAP database before postinstall opens up the ability to recommend people installing on arm boards to directly connect using the "admin" user instead of the confusing "root the first time but then admin user"

## Solution

Init the ldap database during the .deb install

## PR Status

Done, tested on my side

## How to test

Theoretically it's just about running "dpkg-reconfigure yunohost" (after making sure that /etc/yunohost/installed doesnt exists yet) and running the postinstall.

But dpkg-reconfigure yunohost calls the installed .deb's postinst script ... but this PR changes it and it's not trivial to propagate it so uuuuh I just manually ran the various "bash /usr/share/.../... init" stuff that you can find in the changelog to simulate the "dpkg-reconfigure yunohost"


src/yunohost/__init__.py
```diff
@@ -53,6 +53,7 @@ def check_command_is_valid_before_postinstall(args):
 
     allowed_if_not_postinstalled = ['tools postinstall',
                                     'tools versions',
+                                    'tools shell',
```
zamentur, line 56: Is it usable from the API ?
The risk is we could forget this, the day we secure the access to `tools postinstall` from the API...

alexAubin: No it's not, the only route exposed to unauthenticated users is the postinstall : https://github.com/YunoHost/yunohost/blob/dev/data/actionsmap/yunohost.yml#L1383-L1384

(then the postinstall is ran and /etc/yunohost/installed is created so it can't be re-run anymore)

zamentur: Ok, so, it's solved

PR #1066, opened 2020-10-13
Do not advertise upgrades for bad-quality apps

alexAubin:
## The problem

In some cases, Yunohost may tell the user that some upgrade is available, which in fact introduces regressions to the app. For example libreto was level 7 but recently dropped to level 2, then breaking backup etc...

## Solution

Only advertise upgrades for good-quality apps

## PR Status

Yolocommited but should be working

## How to test

Eeeh idk try to install some bad quality apps and check if the "upgradable" key in "yunohost app info <theapp>" is "bad_quality"

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -182,19 +182,29 @@ def _app_upgradable(app_infos):
     from packaging import version
 
     # Determine upgradability
-    # In case there is neither update_time nor install_time, we assume the app can/has to be upgraded
 
-    # Firstly use the version to know if an upgrade is available
-    app_is_in_catalog = bool(app_infos.get("from_catalog"))
+    app_in_catalog = app_infos.get("from_catalog")
     installed_version = version.parse(app_infos.get("version", "0~ynh0"))
     version_in_catalog = version.parse(app_infos.get("from_catalog", {}).get("manifest", {}).get("version", "0~ynh0"))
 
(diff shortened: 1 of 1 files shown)


kay0u:
Uuh, it sounds good like that, but in the case of libreto, it wouldn't prevent people from upgrading it, because the CI may takes some time to do its job, and update the levels

alexAubin:
Weeeell yup indeed there would be a lap of time of a few days during which the app would still be available for upgrades ... but at least once it's flagged level 2 in the app catalog the upgrade would not be advertised ... I don't know what's the exact frequency of upgrades of app levels in the app catalog but it seems sort of reasonable ... 

or at least my goal was more about covering the case of "1 year ago you installed a level7 app which is now level 3 (or notworking) ... and you don't check app upgrades every so often - but some day you decide to do a bunch of upgrade and yunohost should definitely *not* advertise an upgrade to a low-quality / not working state.

So indeed this PR doesn't solve 100% of the issue if the app just got upgraded but not validated by the CI. The "real" accurate way would be to have a CI/release workflow such that "new commits are not pushed to the app catalog until it got tested by the CI"... But that's way more tricky to implement (though of course would be super nice to have) and doesn't conflict with this quick/easy fix (actually that's just complementary)

alexAubin:
Sooo uuuh is it okay to merge this ?

kay0u (approved):

Psycojoker:
Shouldn't we merge this?

PR #1067, opened 2020-10-14
[fix] intermediate_certificate is now included in signed certificate

Psycojoker:
Fix https://github.com/YunoHost/issues/issues/1644

## The problem

Describe in issue.

## Solution

Don't add the intermediate certificate, it's now included in the signed certificate.

## PR Status

Work on prod for me ™

## How to test

- choose a signed domain
- look in /etc/yunohost/certs/domain.tld/crt.pem
- observe that the 2 last entries are duplicated
- apply the patch
- yunohost domain cert-renew domain.tld --force
- look at the file again (**USING ABSOLUTE PATH**, the symlink to the folder is **changed** so if you have cd in the folder you'll be looking at the old file), last 2 entries are not duplicated anymore

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -539,12 +539,6 @@ def _fetch_and_enable_new_certificate(domain, staging=False, no_checks=False):
 
         raise YunohostError('certmanager_cert_signing_failed')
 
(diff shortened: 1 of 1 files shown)


alexAubin (approved):
Quick review, LGTM :+1:

PR #1068, opened 2020-10-17
Refuse the installation of an app with a point in its name

cyxae:
## The problem

As described in https://github.com/YunoHost/issues/issues/1525

## Solution

Refuse app names with point in app_install

## PR Status

To be tested

## How to test

Run yunohost app install [app name with a point]
For example run yunohost app install foo.bar
And it should refuse the installation

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -705,7 +705,7 @@ def app_install(operation_logger, app, label=None, args=None, no_remove_on_failu
 
     raw_app_list = _load_apps_catalog()["apps"]
 
-    if app in raw_app_list or ('@' in app) or ('http://' in app) or ('https://' in app):
+    if app in raw_app_list or ('@' in app) or ('http://' in app) or ('https://' in app) or ('.' in app):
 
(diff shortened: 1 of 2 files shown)


src/yunohost/app.py
```diff
@@ -714,11 +714,15 @@ def confirm_install(confirm):
         # extract "bar" and test if we know this app
         elif ('http://' in app) or ('https://' in app):
             app_name_to_test = app.strip("/").split("/")[-1].replace("_ynh", "")
+        elif '.' in app:
```
zamentur, line 717: What if app contains a path like ./myapp_ynh ? or even ../myapp_ynh ?

I think this checks should be done on the `app_id` obtained from the manifest, see around line 751...

zamentur (changes requested):
Thanks for your contribution.
I think we need to support relative path in app_location. The "." detection should be done on manifest app id. 

I suggest to check with a restrictive regex instead of just detect if there is a point. I think we don't need to support accent or smiley utf-8 char.

PR #1069, opened 2020-10-31
Diagnosis: report usage of backports repository in apt's sources.list

alexAubin:
## The problem

Saw again somebody with backports installed, creating stupid issues

## Solution

Discourage people from having backports enabled using a check in the diagnosis

(Could also go further and report packages installed from backports (typically there's a ~bpo in the version name) but that raises the question of how to downgrade them and if that's a good idea..)

## PR Status

Done, tested and working on my side

## How to test

Add a dummy line with `deb foo-backports` in /etc/apt/sources.list.d/whatever.list then run the diagnosis


zamentur (approved):
LGTM (tested and reviewed)

PR #1070, opened 2020-11-02
Define a new "yunohost app search" command

cyxae:
## The problem

According to the issue https://github.com/YunoHost/issues/issues/1360 there is no such search function as one would expect before installing anything

## Solution

Create a simple function to search in the YunoHost app catalog, in the app name or description

## PR Status

Ready to test

## How to test

It defines a new command which you can use as :
yunohost app search [string]
for example :
yunohost app search caldav
yunohost app search agendav


zamentur (approved):
LGTM tested

PR #1071, opened 2020-11-02
Add an option to disable the 'YunoHost' panel overlay in web apps

cyxae:
## The problem

According to the issue https://github.com/YunoHost/issues/issues/1669 there are problems in some apps with the presence of the little Yunohost square overlay in web apps 

## Solution

Add an option to remove this little overlay in nginx configuration

## PR Status

My implementation of boolean test in bash is quite dirty, since we have to retrieve a python boolean and then use it in bash. I ended up doing a string comparison with "false". There may be a cleaner way to do it.

I think there’s also an issue when the file "yunohost_panel.conf.inc" is completely empty, so the solution I found is to left a single "#" in it as it is just a comment.

## How to test

Just change the setting :
`yunohost settings set 'panel_overlay.enabled' --value false`
and then the nginx configuration is updated automatically. You may need to reopen a new tab (clear cache) to test that the overlay has appeared/disappeared


cyxae:
Since the command line says `yunohost service regen-conf` is deprecated, I also replaced every occurrence of it in "settings.py" by `yunohost tools regen-conf`

data/hooks/conf_regen/15-nginx
```diff
@@ -47,6 +47,12 @@ do_pre_regen() {
 
   # install / update plain conf files
   cp plain/* "$nginx_conf_dir"
+  # remove the panel overlay if this is specified in settings
+  panel_overlay=$(yunohost settings get 'panel_overlay.enabled')
+  if [ $panel_overlay = false ] || [ $panel_overlay = False ]
+  then
+    echo "#" > "${nginx_conf_dir}/yunohost_panel.conf.inc"
```
zamentur, line 54: There is an issue if i set it to False and next to True, the overlay is still hidden :/

cyxae: I think I had this problem but if you clear the cache of the browser after setting it to True, it may work

zamentur: It was a mistake from me, i read you code and don't understand it correctly. So it's solved

src/yunohost/settings.py
```diff
@@ -74,6 +74,7 @@ def is_boolean(value):
     ("smtp.relay.user", {"type": "string", "default": ""}),
     ("smtp.relay.password", {"type": "string", "default": ""}),
     ("backup.compress_tar_archives", {"type": "bool", "default": False}),
+    ("panel_overlay.enabled", {"type": "bool", "default": True}),
```
zamentur, line 77: I guess we could have more option around ssowat (like themas...)

I suggest to call that : ssowat.panel_overlay.enabled or something like that

cyxae: Indeed that would make sense

cyxae: Done !

zamentur (approved):
LGTM, untested

ChvckN0rri5:
> 
``` 
yunohost settings set 'panel_overlay.enabled' --value false
```

Please update with:
```
yunohost settings set 'ssowat.panel_overlay.enabled' --value false
```

image

PR #1073, opened 2020-11-08
Upgrade n version to upstream

ericgaspar:
## The problem
- New version upstream
...

## Solution
- Upgrade to version 6.7.0
...

## PR Status

...

## How to test

...


kay0u (approved):
Have you tested it with some apps? (as it's not a major update, it shouldn't break anything but I've already lost all my faith with that)

yalh76 (approved):

PR #714, opened 2019-04-17
[fix] ynh_psql_remove_db fail

zamentur:
## The problem

The psql remove database helper fail with several message

## Solution

Remove small error in db exist test

## PR Status

Ready

## How to test

...

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

PR #1075, opened 2020-11-17
[enh] in order to be able to change user password in CLI without writing it in clear (issue #714)

cricriiiiii:
## The problem

To update a user password in CLI, the only way is to write the new password in clear using "yunohost user update username -p new_password". This puts the new password in clear in some logs. This is issue https://github.com/YunoHost/issues/issues/714.

## Solution

This PR allows to call the "yunohost user update" command with option -p but without an argument, in which case the admin is prompted for the new password.

## PR Status

Ready for tests.

## How to test

...

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -165,8 +165,11 @@ user:
                     full: --change-password
                     help: New password to set
                     metavar: PASSWORD
+                    nargs: "?"
+                    const: 0
                     extra:
                         pattern: *pattern_password
+                        comment: good_practices_about_user_password
                 --add-mailforward:
                     help: Mailforward addresses to add
                     nargs: "*"
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -169,7 +169,6 @@ user:
                     const: 0
                     extra:
                         pattern: *pattern_password
-                        comment: good_practices_about_user_password
                 --add-mailforward:
                     help: Mailforward addresses to add
                     nargs: "*"
--- a/src/yunohost/user.py
+++ b/src/yunohost/user.py
@@ -324,7 +324,9 @@ def user_update(operation_logger, username, firstname=None, lastname=None, mail=
     if lastname and firstname:
         new_attr_dict['cn'] = new_attr_dict['displayName'] = [firstname + ' ' + lastname]
 
-    if change_password:
+    if change_password is not None:
+        if not change_password:
+            change_password = msignals.prompt(m18n.n("ask_password"), True, True)
         # Ensure sufficiently complex password
         assert_password_is_strong_enough("user", change_password)
 
--- a/src/yunohost/user.py
+++ b/src/yunohost/user.py
@@ -324,8 +324,12 @@ def user_update(operation_logger, username, firstname=None, lastname=None, mail=
     if lastname and firstname:
         new_attr_dict['cn'] = new_attr_dict['displayName'] = [firstname + ' ' + lastname]
 
+    # change_password is None if user_update is not called to change the password
     if change_password is not None:
-        if not change_password:
+        # when in the cli interface if the option to change the password is called
+        # without a specified value, change_password will be set to the const 0.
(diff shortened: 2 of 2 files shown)


cricriiiiii:
This introduces a new behavior which puzzles me a little bit. When running the command without any option "yunohost user update username", before returning a success (for not updating anything) it prints the comment which is intended for the password modification. In the actionmap it is placed under 
`
...
-p:
    ...
    extra:
        comment: good_practices_about_user_password
` 
Is this supposed to happen or is it a problem with the parser or something else ?

src/yunohost/user.py
```diff
@@ -324,7 +324,9 @@ def user_update(operation_logger, username, firstname=None, lastname=None, mail=
     if lastname and firstname:
         new_attr_dict['cn'] = new_attr_dict['displayName'] = [firstname + ' ' + lastname]
 
-    if change_password:
+    if change_password is not None:
+        if not change_password:
```
zamentur, line 328: The moulinette framework generate an HTTP API and a CLI with the same code.

You should add a test to check you are in the CLI context.

```suggestion
        if not change_password:
```
        if not change_password and msettings.get('interface') == 'cli':
```

Comment (cricriiiiii) [reply src/yunohost/user.py]: Oh right, I didn't realize that was the case, I added this. thanks!

File: data/actionsmap/yunohost.yml [hunk 2]
```diff
@@ -165,8 +165,11 @@ user:
                     full: --change-password
                     help: New password to set
                     metavar: PASSWORD
+                    nargs: "?"
+                    const: 0
                     extra:
                         pattern: *pattern_password
+                        comment: good_practices_about_user_password
```
zamentur, line 172: This line created a bug: the advices are displayed even if no password is changed...
I don't know how to avoid this bug without removing this line :/
```
root@ynh:/ynh-dev# yunohost user update ljfa -f LOL
You are now about to define a new user password. The password should be at least 8 characters long—though it is good practice to use a longer password (i.e. a passphrase) and/or to a variation of characters (uppercase, lowercase, digits and special characters).
```

zamentur (changes requested):
Thanks for this contribution.
I think we should find a solution for the bug about password advices display all the time before to merge.

data/actionsmap/yunohost.yml
alexAubin: (I believe it was already an issue priori to this PR so not really a regression)

alexAubin: (I think the proper fix involve obscure tweaking of the moulinette ...)

zamentur: What do we do ? we remove the advice or we keep it on each update command event if there are no option change-password ?

alexAubin: To me this is to be addressed in another PR ... in the meantime it's not really a huge deal ... this issue has been here for a while now and we can live with it ...

alexAubin (approved):
Not tested but LGTM

cricriiiiii:
Thanks for the reviews and comments!
Concerning the bug, it seems that moulinette is not really structured to have different "extra" parameters for different options of the same action. But actually, the "action" "update" in the category "user" should probably be a "subcategory" and all its different options should be different "actions" of this "subcategory". Then the bug would be easily fixed without tweaking moulinette I think.
What do you think?

data/actionsmap/yunohost.yml
zamentur: ok so i set the ready to merge flag

zamentur:
About your suggestion about subcategories, i don't know if it's a good idea i plan the discussion during the next meeting (this tuesday).

data/actionsmap/yunohost.yml
alexAubin: "Une fois n'est pas coutume", zamentur was in fact right, we should probably remove that line

zamentur:
So we discuss during the meeting to avoid to display the advice message, so finally we decided to remove this advice.

cricriiiiii:
Great! Many thanks!

PR #1076, opened 2020-11-21
Automatically translate legacy rules in /etc/ssowat/conf.json.persistent

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1671

## Solution

Yunohost should translate legacy rules in /etc/ssowat/conf.json.persistent into the new format

While actually looking into implementing the initial idea from last meeting, though (which was about "loading only conf.json and not .persistent anymore", I was concerned with the following : 
- this requires changing the code of ssowat (which is doable but meh)
- if you want to just change the theme or logging level, it's annoying for the user to have to run "yunohost app ssowatconf" - many people will forget about this

So for now I decided to *only* translate the legacy rules (unprotected_urls, etc...) into the new permission structure *inside* /etc/ssowat/conf.json.persistent, and to keep everything else working the way it's working now. 

## PR Status

Tested and working on my side

## How to test

Add something like `"unprotected_uris": ["foo/bar"]` or `"unprotected_regex": ["foo/bar"]` to `/etc/ssowat/conf.json.persistent`. Run `yunohost app ssowatconf`, and the content of the json file should change

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1420,6 +1420,9 @@ def app_ssowatconf():
     with open('/etc/ssowat/conf.json', 'w+') as f:
         json.dump(conf_dict, f, sort_keys=True, indent=4)
 
+    from utils.legacy import translate_legacy_rules_in_ssowant_conf_json_persistent
+    translate_legacy_rules_in_ssowant_conf_json_persistent()
+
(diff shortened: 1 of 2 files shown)


kay0u (approved):
:+1:

Josue-T (approved):

PR #1082, opened 2020-11-25
Add number arg parser

kay0u:
## The problem

No number parser, used [here](https://github.com/YunoHost-Apps/wordpress_ynh/blob/e24fe50fccd27b403d5d51c299f72e3284dc1427/config_panel.toml#L54) for example

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -2593,6 +2593,29 @@ class UserArgumentParser(YunoHostArgumentFormatParser):
                             error=m18n.n('user_unknown', user=question.value))
 
 
+class NumberArgumentParser(YunoHostArgumentFormatParser):
+    argument_type = "number"
+    default_value = ""
+
+    def parse_question(self, question, user_answers):
+        question = super(NumberArgumentParser, self).parse_question(question, user_answers)
+
+        if question.default is None:
+            question.default = 0
+
+        return question
+
+    def _post_parse_value(self, question):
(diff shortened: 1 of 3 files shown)


src/yunohost/app.py
```diff
@@ -2593,6 +2593,29 @@ def _raise_invalid_answer(self, question):
...
+            question.default = 0
+
+        return question
+
+    def _post_parse_value(self, question):
+        if isinstance(question.value, int):
+            return super(NumberArgumentParser, self)._post_parse_value(question)
+
+        try:
+            return int(question.value)
```
Psycojoker, line 2613: you can use `str.isdigit()` instead here if you want

src/yunohost/app.py
```diff
@@ -2593,6 +2593,29 @@ def _raise_invalid_answer(self, question):
...
+        except ValueError:
+            raise YunohostError('app_argument_invalid', name=question.name,
+                            error=m18n.n('invalid_number'))
```
kay0u, line 2616:
```suggestion
        if isinstance(question.value, str) and question.value.isdigit():
            return int(question.value)

        raise YunohostError('app_argument_invalid', name=question.name,
                            error=m18n.n('invalid_number'))
```

src/yunohost/app.py
kay0u: Thank you!

PR #1080, opened 2020-11-25
Create a PR on new tags (for new YNH version)

kay0u:
## Solution

Related to:
https://github.com/YunoHost/doc/pull/1448
https://gitlab.com/yunohost/yunohost/-/jobs/872132228
https://gitlab.com/yunohost/yunohost/-/blob/9af3873a1f83a1c242d0b534e0a8111bd3d80b88/.gitlab/ci/doc.gitlab-ci.yml

I use a token which has only access to `public_repo` on yunohost-bot, registered on [this page](https://gitlab.com/yunohost/yunohost/-/settings/ci_cd) (accessible only by admin members)

## PR Status

...

## How to test

...

PR #1085, opened 2020-11-26
add branch name in the helper template

kay0u:
##  Solution

Add the current branch during doc generation. Should we use the commit hash instead to avoid "desync"?

We can see the result of the job here: https://yunohost.gitlab.io/-/yunohost/-/jobs/874635832/artifacts/doc/helpers.html

## PR Status

...

## How to test

...


alexAubin:
> Should we use the commit hash instead to avoid "desync"?

Hmmmyes maybe ?

I guess it depends on the intention : do we want to keep the doc in sync with the very-latest stuff in dev (in that case keeping the branch name dev is okay) or if we only push the doc change when releasing a stable ? (I haven't 100% followed the other branch for autopush from gitlab builds)

kay0u:
I think we want to keep the stable version of the doc, therefore I did this https://github.com/YunoHost/yunohost/pull/1080 which should create a PR on each new YunoHost version.

alexAubin:
So yup I would say we wanna point to the commit ideally :+1:

doc/generate_helper_doc.py
```diff
@@ -3,10 +3,19 @@
 import os
 import glob
 import datetime
+import subprocess
 
+def get_current_git_branch():
+    p = subprocess.Popen("git branch -a --contains | grep remote | cut -d'/' -f3 ", shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
kay0u, line 9:
```suggestion
    p = subprocess.Popen("git rev-parse --verify HEAD", shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```

PR #1090, opened 2020-12-01
[enh] Adding composer helper

yalh76:
## Solution

Adding composer experimental helper

## PR Status

Ready and tested on several apps

### DIFF
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -560,3 +560,63 @@ ynh_get_scalable_phpfpm () {
         fi
     fi
 }
+
+readonly YNH_DEFAULT_COMPOSER_VERSION=1.10.17
+# Declare the actual composer version to use.
+# A packager willing to use another version of composer can override the variable into its _common.sh.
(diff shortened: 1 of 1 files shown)


kay0u (approved):

data/helpers.d/php
```diff
@@ -560,3 +560,63 @@ ynh_get_scalable_phpfpm () {
         fi
     fi
 }
+
+readonly YNH_DEFAULT_COMPOSER_VERSION=1.10.17
```
zamentur, line 564: What's the reasons not to have composer 2.0.8 as default version (the current stable version) ?
Probably a naive question i know :)

tituspijean: When we updated the helpers there were multiple common Composer packages that were yet to be compatible with v2. That was breaking our apps... so we sticked to v1.

I will try to check tomorrow if they are updated. There was one from Wikimedia for example.

yalh76: Because mostly app still use composer 1 ^^

PR #1096, opened 2020-12-05
Less madness for hook exec

alexAubin:
## The problem

Currently app scripts are run in such a way that this kind of thing appears when you run `ps -ef` on the machine as a random user : 

```
sh -c 'YNH_APP_DOMAIN=domain.tld YNH_APP_SECRET=superpassword ./install appname domain.tld superpassword'
```

## Solution

- Do not provide "regular" arguments to app scripts anymore, everything should be fetch using env variables. Unfortunately a handful of apps are still fetching args using `domain=$1`-style but it should only be like a handful of them, low-quality apps
- "Hide" the env variables by provinding them to the `env` arguments (of subprocess.Popen which is being call_async_output) instead of providing them explicitly in the bash command that's being run

## PR Status

Tested and working on my side on a simple app, but should be tested on the CI-core and CI-apps to have a better view of possible regressions

## How to test

Uuuuh run any app operation like `yunohost app install`, check ps -ef


zamentur:
I think post_user_create hooks are also affected. And i think nextcloud use $1...

alexAubin:
My god indeed the password is passed as argument >_>

I think we could just remove all the args to post_user_create except for the username and email because according to this magic grep, I find no app using $3, 4, 5 ...

```
 grep -nr '=\$' $(find ./ -name "post_*") | grep "user_create"
```

Also there's no affected app hook (post_app_install etc) affected by the currently proposed change

zamentur:
I added user_update hook with env dict in same commit.

Psycojoker:
LGTM following a quick review

zamentur:
I set the work needed tag as gitlab ci test failed

alexAubin:
Tests are passing on https://gitlab.com/yunohost/yunohost/-/pipelines/236864198 :tada:

PR #1098, opened 2020-12-08
[fix] Backup/restore dkim keys

zamentur:
## The problem

Dkim keys are not backuped

## Solution

Backup and restore it

## PR Status

Untested (and written in 2 minutes before a meeting...)

## How to test

...


alexAubin:
Yolomerging i guess since tests are passing

PR #1105, opened 2020-12-11
add all_users in protected_urls

kay0u:
## The problem

After migration, nobody can access to a `protected_uris` in ssowat `conf.json.persistent`

## Solution

Add all users in this permission.

Should we add it to `skipped_urls` and `unprotected_urls`?

## PR Status

...

## How to test

...


kay0u:
pipeline https://gitlab.com/yunohost/yunohost/-/pipelines/228616791

PR #1107, opened 2020-12-17
handle change php version

kay0u:
## The problem

We don't handle the change of php version of an app, and if a custom version was installed, it's never uninstalled

## Solution

Remove the old phpfpm conf (and php-fpm package if not needed)

## PR Status

Done

## How to test

Tested with nextcloud (install an old version which used php7.2, and upgrade it with this new helper)


alexAubin (approved):
LGTM

yalh76 (approved):
Quick review LGTM..

PR #1108, opened 2020-12-18
do not reload php7.3 too fast

kay0u:
## The problem

When we run `systemctl reload-or-restart php7.3-fpm` too fast, the service may end up being stopped.

(Try: `systemctl reload-or-restart php7.3-fpm && systemctl reload-or-restart php7.3-fpm && systemctl reload-or-restart php7.3-fpm && systemctl reload-or-restart php7.3-fpm` then `systemctl status php7.3-fpm`)

## Solution

Avoid 2 reload for nothing (Should we reload right before the `ynh_die`?)

## PR Status

...

## How to test

...

### DIFF
--- a/data/helpers.d/php
+++ b/data/helpers.d/php
@@ -272,9 +272,9 @@ WantedBy=multi-user.target
         # Validate that the new php conf doesn't break php-fpm entirely
         if ! php-fpm${phpversion} --test 2>/dev/null
         then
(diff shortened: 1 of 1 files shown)


alexAubin (approved):
Arf anéfé j'avais pas réalisé que j'avais laissé un reload en dessous :sweat_smile:

PR #1109, opened 2020-12-20
firewall: force source port for UPnP.

leo-lb:
miniupnpc uses a random source port by default, the issue is that the
firewall rule to allow destination port 1900 incoming is unused because
the UPnP server will use the random source port as destination port in
the reply which iptables will block. Forcing the source port to be 1900
will ensure the UPnP server also uses that as destination port in the
reply and pass the firewall.

python-miniupnpc 2.0 or later is required for this change to have any
effect, it is otherwise silently ignored. A debian package upgrade is in
the works for official Yunohost repos.

Closes https://github.com/YunoHost/issues/issues/1463

### DIFF
--- a/src/yunohost/firewall.py
+++ b/src/yunohost/firewall.py
@@ -336,7 +336,7 @@ def firewall_upnp(action='status', no_refresh=False):
 
     # Refresh port mapping using UPnP
     if not no_refresh:
-        upnpc = miniupnpc.UPnP()
+        upnpc = miniupnpc.UPnP(localport=1)
         upnpc.discoverdelay = 3000
 
         # Discover UPnP device(s)


leo-lb:
Debian Buster package patch (it only provided python3 module)

Patch on top of https://salsa.debian.org/miniupnp-team/miniupnpc/-/commit/c815db181694c3e4af8b5e8e33b6147d7bf5d8d1 (or newer if it works), it is current Debian Buster's `$ apt source miniupnpc`

```
From a404fb4aa3e5ed0f0d5a8da17c06288ed212716b Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?L=C3=A9o=20Le=20Bouter?= <lle-bout@zaclys.net>
Date: Mon, 21 Dec 2020 00:33:47 +0100
Subject: [PATCH] Add back Python2 package

---
 debian/control                  | 21 +++++++++++++++++++++
 debian/python-miniupnpc.install |  2 ++
 debian/rules                    |  2 +-
 3 files changed, 24 insertions(+), 1 deletion(-)
 create mode 100644 debian/python-miniupnpc.install

diff --git a/debian/control b/debian/control
index daf5ed2..8845715 100644
--- a/debian/control
+++ b/debian/control
@@ -7,6 +7,8 @@ Uploaders:
 Build-Depends:
  debhelper (>= 11~),
  dh-python,
+ python-dev,
+ python-setuptools,
  python3-dev,
  python3-setuptools,
 Standards-Version: 4.1.4
@@ -96,3 +98,22 @@ Description: UPnP IGD client lightweight library Python 3 bindings
  other IGD daemon and is ideal for using on low memory device for this reason.
  .
  This package contains the Python 3 bindings.
+
+Package: python-miniupnpc
+Architecture: any
+Section: python
+Depends: ${misc:Depends}, ${shlibs:Depends}, ${python:Depends}
+Description: UPnP IGD client lightweight library Python bindings
+ The UPnP protocol is supported by most home adsl/cable routers and Microsoft
+ Windows 2K/XP. The aim of the MiniUPnP project is to bring a free software
+ solution to support the "Internet Gateway Device" part of the protocol. The
+ MediaServer/MediaRenderer UPnP protocol is also becoming very popular.
+ .
+ Miniupnpc aims at the simplest library possible, with the smallest footprint
+ and no dependencies to other libraries such as XML parsers or HTTP
+ implementations. All the code is pure ANSI C. Compiled on a x86 PC, the
+ miniupnp client library have less than 15KB code size. For instance, the upnpc
+ sample program is around 20KB. The miniupnp daemon is much smaller than any
+ other IGD daemon and is ideal for using on low memory device for this reason.
+ .
+ This package contains the Python bindings.
\ No newline at end of file
diff --git a/debian/python-miniupnpc.install b/debian/python-miniupnpc.install
new file mode 100644
index 0000000..773342a
--- /dev/null
+++ b/debian/python-miniupnpc.install
@@ -0,0 +1,2 @@
+usr/lib/python2*/dist-packages/miniupnpc-*.egg-info
+usr/lib/python2*/dist-packages/miniupnpc.so
diff --git a/debian/rules b/debian/rules
index d06d1c1..6f47fe7 100755
--- a/debian/rules
+++ b/debian/rules
@@ -10,7 +10,7 @@ export PYBUILD_SYSTEM=distutils
 
 
 %:
-	dh $@ --with python3 --buildsystem=makefile
+	dh $@ --with python3 --with python2 --buildsystem=makefile
 
 override_dh_auto_build:
 	dh_auto_build
-- 
2.29.2

```

alexAubin:
Followup of discussions in private with @leo-lb  : The previous patch has been applied on our deb build chain (on top of debian/2.1.1 miniupnpc) and released in testing so maybe we can have some bits of feedback quick

>Closes YunoHost/issues#1463

Our code is still kinda catastrophic and should be refactored though, but granted that may indeed improve the situation :sweat_smile:

alexAubin:
(To be merged in next 4.1 iteration if people agree)

kay0u (approved):
:+1:

kay0u:
In fact, opening a port using upnp works, but not closing it (we already discuss that in the chat)

leo-lb:
> In fact, opening a port using upnp works, but not closing it (we already discuss that in the chat)

And I'm going to fix that shortly keep committing to it and delaying but shouldnt do that for too long.

alexAubin:
Sooo uh not sure to fully understand what's the status for this ? Is it doable to fix the issue regarding port closing ?

PR #1112, opened 2020-12-28
Adding mDNS for .local domains

tituspijean:
## The problem

`yunohost.local` domain broadcasting through mDNS is useful but a bit unflexible.

## Solution

- ~~Add a base `avahi-alias` service to publish as many domains as required~~
- ~~Add checks in `domain.py` to add these services whenever the user adds a domain ending by `.local`~~
- Use [Python-Zeroconf lib](https://github.com/jstasiak/python-zeroconf) to deploy a system service that broadcasts any .local domain over the network

## PR Status

- [x] Feature can be tested
- [x] Feature is integrated in diagnosis system
- [x] Seemingly no need for migrations
- [x] Feature is integrated in the regen_conf
- [x] Code complete and ready for review

## How to test

- Try using a `.local` domain during the postinstall
- Try adding a new `.local` domain
- Try adding `.local` subdomains: diagnosis should warn you it will not work. (BTW if someone can have a look to the Python library and figure out if/how CNAME records can be sent out...)


Psycojoker:
I don't know if you've already planned that but this will likely need a data migration for existing domains (I'm not expecting that people really uses those but who knows)

zamentur:
- [ ] Integrate this change inside regenconf (to support "restore use case" and "migrations use case")
- [ ] Test the install/remove change url on this kind of domain
- [ ] Test "User portal" on this kind of domain
- [ ] Change DNS diagnstic to avoid alert due to missing DNS records

tituspijean:
Unfortunately I have been unable to make .local subdomains discoverable (like `test.domain.local`)... I will investigate a bit more into it this weekend but I feel I will have to close this draft PR.

tituspijean:
For reference, this [project](https://github.com/jstasiak/python-zeroconf) is a tad more complete. It would enable publishing more complex mDNS entries, and we could call it directly from within Python scripts.

tituspijean:
So, I have restarted the work on the matter. Instead of spawning multiple services that rely on `avahi-publish`, I rely on `python-zeroconf` mentioned above.

- [x] Retrieval of all `.local` domains
- [x] Make broadcast work (run `pip3 install zeroconf rich` then `python3 mdns.py`)
- [ ] Create settings to select onto which interfaces broadcast should happen
- [ ] Create a service that runs this script

Based of zamentur's list of chores above, I guess that:

> Integrate this change inside regenconf (to support "restore use case" and "migrations use case")

Regenconf would be a simple "systemctl restart". Migration would not be needed if we make the system setting empty by default.

> Test the install/remove change url on this kind of domain

Should not be a problem with the new approach.

tituspijean:
Status: service works, but the unregistration part ends with
`
RuntimeError: Task <Task pending coro=<AsyncZeroconf.async_close() running at /usr/local/lib/python3.7/dist-packages/zeroconf/asyncio.py:261> cb=[gather.<locals>._done_callback() at /usr/lib/python3.7/asyncio/tasks.py:664]> got Future <Task pending coro=<AsyncEngine._async_cache_cleanup() running at /usr/local/lib/python3.7/dist-packages/zeroconf/_core.py:148> wait_for=<Future cancelled>> attached to a different loop
`

I do not know enough about async IO to understand why.

alexAubin (approved):
- [ ] Check how to handle the enabling of the new service on legacy systems
- [ ] Check systemd dependency to network stack and what's default target
- [x] Build python3-zeroconf on our infra

alexAubin:
Yolo ~

PR #1114, opened 2020-12-30
[mod](user_create) only ask for one letter for first/last name

Psycojoker:
## The problem

[Credic O](https://en.wikipedia.org/wiki/C%C3%A9dric_O) can't use YunoHost.

## Solution

Accept his last name and same for the first name.

## PR Status

Test

## How to test

> yunohost user create testeu

Answer "a" and "b" to first and last name.


kay0u (approved):

JimboJoe (approved):
:laughing:

PR #1118, opened 2020-12-31
Normalize conf template handling for nginx, php-fpm, systemd and fail2ban using ynh_add_config

alexAubin:
## The problem

Conf handling is a mess, we have tons of ynh_replace_string all over the place, and some ugly --othervar options to replace "other vars", lack of consistency, etc

## Solution

Use the new `ynh_add_config` helper to consistently handle template hydration (and backup/save of the checksum ugh)

## PR Status

Yolocommited

## How to test

Gotta install some apps idk

### DIFF
--- a/data/helpers.d/nginx
+++ b/data/helpers.d/nginx
@@ -2,60 +2,25 @@
 
 # Create a dedicated nginx config
 #
-# usage: ynh_add_nginx_config "list of others variables to replace"
-#
-# | arg: list - (Optional) list of others variables to replace separated by spaces. For example : 'path_2 port_2 ...'
+# usage: ynh_add_nginx_config
 #
 # This will use a template in ../conf/nginx.conf
-#   __PATH__      by  $path_url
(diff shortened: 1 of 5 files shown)


data/helpers.d/nginx
```diff
@@ -64,8 +29,6 @@ ynh_add_nginx_config () {
         ynh_replace_string --match_string="^#root_path_only" --replace_string="" --target_file="$finalnginxconf"
     fi
 
-    ynh_store_file_checksum --file="$finalnginxconf"
```
alexAubin, line 67: Hmpf re-reading this, we can't just handle the checksum inside ynh_add_config because we keep editing the damn config afterwards because of thos #root_path_only etc ... I guess we should handle these in ynh_add_config as well ...

kay0u: Or maybe handle `root_path_only` before calling this script?

yalh76 (approved):
++
114 lines removed, great ^^

PR #1119, opened 2020-12-31
Simpler and more consistent logging initialization

Psycojoker:
## The problem

We have zillion of loggers in moulinette and we only listen to one in yunohost, that's totally overkill.

For example we are missing this kind of error right now:

image

## Solution

Move every advanced logger to moulinette.core.

Change some loggers to avoid noise.

This is combined with this PR https://github.com/YunoHost/moulinette/pull/263

## PR Status

Seems to be working, changes are small.


zamentur:
I set work needed cause of the Aleks comment on the other PR:
> I'm a bit confused because in Yunohost we do have many different loggers basically for each file (for example here https://github.com/YunoHost/yunohost/blob/dev/src/yunohost/service.py#L43 ) and they all happen to be logged consistently o.O ?

alexAubin:
(It's not really blocking though, it'd just be a bonus to understand what's going on)

alexAubin:
So I was pretty confused and realized that if (parts of) moulinette wasn't logged, it's probably because we had `"handlers": []` for the "moulinette" logger ... So imho it makes more sense and is simpler to fix this directly rather than tweak all logger name in every file ...

In the process, I also refactored / unified the logging config between cli and api because there's in fact no reason to have 2 different dicts, they were pretty much the same.

alexAubin:
I'm still not sure what to do about the "unable to retrieve string for default locale" error .. I know we have a few places where this error is triggered as part of ~nominal stuff, and so far it wasn't displayed so it didnt confuse the user ... but if this is displayed when there's in fact no problem, it's definitely going to be confusing ...

PR #1120, opened 2021-01-01
[enh] display domain_path of app on app list

Psycojoker:
## The problem

When listing the applications with `yunohost app list` we don't show by default were the application is installed. People end up having to learn about the existence of `yunohost app map` just for that and it's a bit stupid (and it doesn't display the same information).

image

## Solution

Add, when domain and path is available for an app, the `domain_path` key like this:

image

## PR Status

Tested and working.

## How to test

Just do a `yunohost app list`

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -163,6 +163,9 @@ def app_info(app, full=False):
         'version': local_manifest.get('version', '-'),
     }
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Yolomerging because sounds like a small legit enh

PR #1125, opened 2021-01-01
Add '--remove-apps' (and '--force') options to "yunohost domain remove"

Psycojoker:
## The problem

Sometime when removing a domain some applications are still installed on it, in this case we cancel and ask the user to remove or move it by hand. This is tedious and we can simply automate that.

## Solution

Add the option '--remove-apps' which will list the apps to remove then prompt the user for confirmation.

Also add the option '-y/--yes' to bypass the prompt.

Example session:

image

image

## PR Status

Working.

## How to test

- add a domain
- install apps on it ("yunohost app install my_webapp")
- remove it with "yunohost domain remove"
- remove it with "--remove-apps"
- remove it with "--remove-apps --yes"


data/actionsmap/yunohost.yml
```diff
@@ -448,6 +448,14 @@ domain:
                     help: Domain to delete
                     extra:
                         pattern: *pattern_domain
+                -r:
+                    full: --remove-apps
+                    help: Remove apps installed on the domain
+                    action: store_true
+                -y:
+                    full: --yes
```
alexAubin, line 456: Only minor comment I have is that maybe we should name this --force to be consistent with other commands ? This seems to be our de-facto standard for the "don't prompt any warning and just do it"

alexAubin (approved):

src/yunohost/domain.py
```diff
@@ -206,10 +208,21 @@ def domain_remove(operation_logger, domain, force=False):
         settings = _get_app_settings(app)
         label = app_info(app)["name"]
         if settings.get("domain") == domain:
-            apps_on_that_domain.append("    - %s \"%s\" on https://%s%s" % (app, label, domain, settings["path"]) if "path" in settings else app)
+            apps_on_that_domain.append((app, "    - %s \"%s\" on https://%s%s" % (app, label, domain, settings["path"]) if "path" in settings else app))
 
     if apps_on_that_domain:
-        raise YunohostError('domain_uninstall_app_first', apps="\n".join(apps_on_that_domain))
+        if remove_apps:
+            if not yes:
```
alexAubin, line 215: Ah we probably also want to check that `msettings.interfacestuff == "cli"` here also

data/actionsmap/yunohost.yml
Psycojoker: image

Turns out we already have this option, it's just not exposed on the CLI API.

I'm going to expose it and use it then.

Psycojoker: While ... I'm not sure it's the best idea? Force will do the removal process even if the domain is unknown, which is weird. Apparently it's an option that you've added? I guess you'll have a better opinion then '^'

src/yunohost/domain.py
Psycojoker: Done

Psycojoker:
Done, thx for the review :) 
I'm just not 100% sure about the "-f" thingy

data/actionsmap/yunohost.yml
alexAubin: Uuuuh yeah indeed, didn't realize that ... But I guess it's good enough, thanks for the change :+1:

PR #1129, opened 2021-01-03
Add multimedia helpers and hooks

alexAubin:
## The problem

Multimedia helpers are overly complex, downloading stuff from an external repo (why?) just to have a bunch of helper/hooks ...

Also today an user reported this https://forum.yunohost.org/t/cant-install-app-transmission-nextcloud/13826 where everything exploded because of unecessary sudos all over the place ...

## Solution

Add the multimedia helpers and hooks in the core ...

This supersedes https://github.com/YunoHost/yunohost/pull/525

Note that I did not port `ynh_multimedia_movefolder` because it doesn't seem to be used in any app

## PR Status

Yolo commited, not tested

## How to test

Try to install an app that rely on the multimedia helpers. In fact not so many of them do : airsonic, (ampache), calibreweb, couchpotato, minidlna, navidrome, nextcloud, transmission

### DIFF
--- a/data/helpers.d/multimedia
+++ b/data/helpers.d/multimedia
@@ -1,0 +1,87 @@
+readonly MEDIA_GROUP=multimedia
+readonly MEDIA_DIRECTORY=/home/yunohost.multimedia
+
+# Initialize the multimedia directory system
+#
+# usage: ynh_multimedia_build_main_dir
+ynh_multimedia_build_main_dir() {
+
+    ## Création du groupe multimedia
+    groupadd -f $MEDIA_GROUP
+
+    ## Création des dossiers génériques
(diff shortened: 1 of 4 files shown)


kay0u:
Should we handle backup of this kind of folders in the core?

alexAubin:
Good question ... I guess it's a special case also related to the "big data" stuff ... It's not clear to me what app put in these exactly compared to /home/yunohost.$app (or yunohost.app/$app) ... Also it's typically gonna weight a lot more than usual backup soooo zblerg idk, it's the whole story of "we should probably have different kind of backups" and polish the experience according to this

alexAubin:
Tested today, merging

PR #1132, opened 2021-01-10
Add the -F flag to prevent grep from using the user password as a regex pattern

Saxodwarf:
## The problem
Issue #[1534](https://github.com/YunoHost/issues/issues/1435)
## Solution
Add the -F flag to the grep command.

## PR Status

...

## How to test

...


alexAubin (approved):
sounds legit to me

(Is it expected that the PR was created as a draft though ?)

Saxodwarf:
Not really, I just wasn't sure about what to write in the PR's description, so I wanted your opinion first.

alexAubin:
This is a legit small fix so let's yolomerge it, thanks for you contribution :+1:

Saxodwarf:
You're welcome, thanks for all the good work on Yunohost !

PR #1145, opened 2021-01-21
Report low total space for rootfs

alexAubin:
## The problem

c.f. [this stupid issue](https://github.com/YunoHost/issues/issues/1697) and stupid VPS providers with 8GB root partitions

## Solution

Complain about it in the diagnosis and at the beginning of postinstall

## PR Status

Tested and working on my side

## How to test

Run the diagnosis or postinstall on a system with low total space (or tweak values in the code)


data/hooks/diagnosis/50-systemresources.py
```diff
@@ -103,6 +103,26 @@ def run(self):
...
+        #
+
+        main_disk_partitions = [d for d in disk_partitions if d.mountpoint in ['/', '/var']]
+        main_space = sum([psutil.disk_usage(d.mountpoint).total for d in main_disk_partitions])
+        if main_space < 10 * GB:
+            yield dict(meta={"test": "rootfstotalspace"},
+                       data={"space": human_size(main_space)},
+                       status="ERROR",
+                       summary="diagnosis_rootfstotalspace_critical")
+        if main_space < 14 * GB:
```
kay0u, line 120: In messages you talk about 16GB but here you put 14, is it intended?

alexAubin: hmyeah idk that's pretty much arbitrary number, the difference is more like a margin to not have a warning if you have like 15.5 GB ... 16 GB recommended, 15 GB maybe ok, but if you have 14 GB this is definitely an issue

kay0u: Seems legit!

kay0u (approved):

PR #1146, opened 2021-01-21
Misc command rename and alias

alexAubin:
## The problem

- [x] bored of having "diagnosis show" but "log display"
- [x] having "GET /log/display" is not restful, should be "GET /log/<logname>" 
- [x] bored of telling people to run "log display foobar --share" when we could have "log share foobar"
- [x] bore of running "diagnosis run" but "migrations migrate" (which calls method run)

## Solution

Rename, aliases etc

## PR Status

Tested and working on my side
Gotta adapt some stuff in webadmin too ...

## How to test

Run the commands idk

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -1519,9 +1519,11 @@ tools:
                         action: store_true
 
               ### tools_migrations_migrate()
-              migrate:
+              run:
                   action_help: Run migrations
-                  api: POST /migrations/migrate
+                  api: POST /migrations/run
+                  deprecated_alias:
+                     - migrate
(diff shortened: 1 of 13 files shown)


data/actionsmap/yunohost.yml
```diff
@@ -1519,9 +1519,11 @@ tools:
                         action: store_true
 
               ### tools_migrations_migrate()
```
kay0u, line 1521:
```suggestion
              ### tools_migrations_run()
```

kay0u: please also rename the function

kay0u (approved):

PR #1147, opened 2021-01-21
Abort postinstall if /etc/yunohost/apps ain't empty

alexAubin:
## The problem

Once more today, saw this issue where /etc/yunohost/installed mysteriously disappeared following some failed stuff ... (as well as yunohost-admin getting uninstalled ... why the hell ?!). Then user reinstalled yunohost-admin, and the webadmin was like "hey let's run the postinstall" which then miserably crashed

## Solution

Test at the beginning of the postinstall if /etc/yunohost/apps exists, and abort if there's at least one app folder inside ... That should be a decent indicator that the system was postinstalled in the past and something funky happened

## PR Status

Yolocommited

## How to test

Idk try to run postinstall with non-empty /etc/yunohost/apps

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -6,10 +6,18 @@ do_configure() {
   rm -rf /var/cache/moulinette/*
 
   if [ ! -f /etc/yunohost/installed ]; then
-      bash /usr/share/yunohost/hooks/conf_regen/01-yunohost init
-      bash /usr/share/yunohost/hooks/conf_regen/02-ssl init
-      bash /usr/share/yunohost/hooks/conf_regen/06-slapd init
-      bash /usr/share/yunohost/hooks/conf_regen/15-nginx init
+
+      # If apps/ is not empty, we're probably already installed in the past and
+      # something funky happened ...
+      if [ -d /etc/yunohost/apps/ ] && ls /etc/yunohost/apps/* 2>/dev/null
+      then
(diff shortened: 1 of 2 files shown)


kay0u (approved):

debian/postinst
```diff
@@ -6,10 +6,18 @@ do_configure() {
...
 
   if [ ! -f /etc/yunohost/installed ]; then
-      bash /usr/share/yunohost/hooks/conf_regen/01-yunohost init
-      bash /usr/share/yunohost/hooks/conf_regen/02-ssl init
-      bash /usr/share/yunohost/hooks/conf_regen/06-slapd init
-      bash /usr/share/yunohost/hooks/conf_regen/15-nginx init
+
+      # If apps/ is not empty, we're probably already installed in the past and
+      # something funky happened ...
+      if [ -d /etc/yunohost/apps/ ] && ls /etc/yunohost/apps/* 2>/dev/null
```
kay0u, line 12:
```suggestion
      if [ -d /etc/yunohost/apps/ ] && ls /etc/yunohost/apps/* >/dev/null 2>&1
```

kay0u (approved):
:+1:

kay0u:
Do you want a hotfix for 4.1 or do we merge it to 4.2?

alexAubin:
Hmnah 4.2 is fine ... turns out the issue the original issue was related to the user uninstalling yunohost because of apache2 >_> When a user decided they would just fuck up the system, not sure what we can do ...

PR #1148, opened 2021-01-21
Prevent the installation of apache2 ...

alexAubin:
## The problem

Apparently people don't give a shit about the goddamn apt trying to tell you that installing apache2 will remove yunohost and you should not do this. And they type the "Yes, do as I say!" or whatever without wondering what the fuck. And then everything explode like apt warned.

## Solution

Fucking pin the goddamn apache2 package. Same for bind9 because who knows what users are capable of.

## PR Status

Tested on my side

## How to test

Run the regen conf for `apt`, then try to install apache2 / apache2-bin / bind9


kay0u (approved):
You could add also:

If you remove these lines, you accept the general sales conditions of YunoHost™ and you agree to pay 30€/hour for any additional support

:heart:

PR #1149, opened 2021-01-22
Fix dyndns madness

alexAubin:
## The problem

- Dyndns update happens way too often, every two fucking minutes is fucking madness
- If any ~long yunohost operation is ongoing, [dyndns updates command will queue up every 2 fucking minutes creating a huge mess](https://forum.yunohost.org/t/there-is-always-something-running/14003/5) and slowness all over the place
- Removing a dyndns doesnt remove the dyndns keys, then you can re-add a dyndns domain
- [What the actual fuck why was all this not already fixed 5 years ago](https://randomoverload.org/wp-content/uploads/2016/08/a9ech72150388.jpg)

## Solution

- Handle the dyndns cron from the regenconf
- Set the cron job to every 10 minutes
- Make sure that it doesn't queue up a dyndns command if there's already a yunohost command running (c.f. lock)
- Remove dyndns keys when removing a domain

## PR Status

Not really tested, gotta do a full test of all 

## How to test

Add a domain, subscribe it, check the "at" command do work, check the cron job work, check domain deletion removes the key


src/yunohost/dyndns.py
```diff
@@ -118,7 +119,8 @@ def dyndns_subscribe(operation_logger, subscribe_host="dyndns.yunohost.org", dom
         subscribe_host -- Dynette HTTP API to subscribe to
 
     """
-    if len(glob.glob('/etc/yunohost/dyndns/*.key')) != 0 or os.path.exists('/etc/cron.d/yunohost-dyndns'):
+
+    if _guess_current_dyndns_domain(dyn_host) != (None, None):
```
kay0u, line 123:
```suggestion
    if _guess_current_dyndns_domain(subscribe_host) != (None, None):
```

src/yunohost/dyndns.py
```diff
@@ -168,9 +170,17 @@ def dyndns_subscribe(operation_logger, subscribe_host="dyndns.yunohost.org", dom
...
-    logger.success(m18n.n('dyndns_registered'))
+    # Yunohost regen conf will add the dyndns cron job if a private key exists
+    # in /etc/yunohost/dyndns
+    regen_conf("yunohost")
 
-    dyndns_installcron()
+    # Add some dyndns update in 2 and 4 minutes from now such that user should
+    # not have to wait 10ish minutes for the conf to propagate
+    cmd = "at -M now + {t} >/dev/null 2>&1 <<< \"/bin/bash -c 'yunohost dyndns update'\""
+    subprocess.check_call(cmd.format(t="2 min"), shell=True)
```
kay0u, line 180: I don't really understand why you do this, and why you run it two times?

Can we instead just run `yunohost dyndns update` (or directly call the function `dyndns_update` to avoid any lock problem) in regen conf?

alexAubin: Yeah sorry i should have explained that in the comment

Thing is, after registering the domain on dynette, for some yet unkown reason, trying "yunohost dyndns update" right away (in the next 1~2ish minutes) won't work because it doesn't recognize the key ... that might have to do with the dynette cron (on server side) but not sure ...

So that's why i set a job a +2 and +4 minutes (because now the client cron only runs every 10 minutes)

Not sure what you mean about the lock ?

kay0u: Ok, I understand now, seems legit then.

> Not sure what you mean about the lock ?

I was talking about running yunohost command, but I was not sure if we can run yunohost command in the regenconf because of moulinette lock etc.

kay0u (approved):
Not fully tested, but LGTM

alexgarel:
Wow I'm happy to see this PR as I come to fill an issue about it. On my raspberrypi, the old strategy makes the computer goes out of memory (because of accumulation of dyndns update) while upgrading, killing the installation process in the middle :'( I recovered but that was a bit though. So thankes for this PR !

PR #1150, opened 2021-01-22
fix cache ynh-admin vuejs

kay0u:
## The problem

We don't have `.ms` file anymone

## Solution

Add cache to files `.js|css|png|jpg|jpeg|gif|ico|json|woff|woff2|ttf|eot`

## PR Status

...

## How to test

...


Axolotle:
Hey ! Not sure we need this, dist files should be hashed so maybe we can set an indefinite cache. Gonna check this.

Axolotle:
Everything is hashed except for what's in `app/public`, infinite cache ?

kay0u:
> Infinite cache ?

What do you mean by that?

Should I remove all cache?

Axolotle:
Hum sorry i don't know what i'm doing/saying with things related to nginx ; )

What i can say is that:
Files at the root should not be cached at all (maybe i should move the favicon into assets so it can be hashed).
So basically it is only the `index.html` that contains links like `<link href=/yunohost/admin/js/views/user/info.a048ca3c.js rel=prefetch>`.

For the rest, everything is hashed and linked into the index so those should be cached forever or with an expiration to `1y` or something like that.

Also, should i add gzipping in the build pipeline ?

kay0u:
After having discussed it, the cache seems to be managed internally so it can normally be deleted

alexAubin (approved):

Axolotle (approved):

PR #1151, opened 2021-01-22
Add ynh_exec_as to official

ericgaspar:
- Add `ynh_exec_as` to the core


kay0u (commented):
As I said in matrix, I think this helper should rather be in the file user, but it's not a strong opinion. Any thought about it @YunoHost/apps ?

ericgaspar:
I agree. I have changed `ynh_exec_as` to user section

alexAubin:
Having this in "user" sounds good ! Let's merge !

PR #1154, opened 2021-01-26
fix upnp closing port

kay0u:
## The problem

We don't close upnp port

## Solution

To fix this issue I can either refactor firewall and make it easy to read or add more mess because I'm too lazy for the previous solution. I let you discover which solution I choose.

Create a new list with of `UDP_TO_CLOSE` and `TCP_TO_CLOSE` port.

## PR Status

...

## How to test

...


alexAubin (approved):
Super quick review, I guess that's good enough :+1:

PR #1159, opened 2021-01-31
[enh] Support SMTPS Relay

grenagit:
## The problem

The configuration of an SMTP relay supporting only SMTPS and not STARTTLS is not functional.

Issue [1725](https://github.com/YunoHost/issues/issues/1725)

## Solution

Add `smtp_tls_security_level = encrypt` and `smtp_tls_wrappermode = yes` when `smtp.relay.port` is set to `465`.

## PR Status

...

## How to test

...


alexAubin (approved):
Sounds good to me ? Thanks for the contribution !

PR #1162, opened 2021-02-02
fix linter, remove allow_failure for corresponding tests

kay0u:
## Solution

Fix all errors in the linter
I'm not sure about what I did in the file [src/yunohost/domain.py](https://github.com/YunoHost/yunohost/blob/bfd7257a8c886081d5a54880215e2f839772b34b/src/yunohost/domain.py#L607-L624)
Do we really want to remove `allow_failure: true`?

## PR Status

...

## How to test

...

### DIFF
--- a/.gitlab/ci/lint.gitlab-ci.yml
+++ b/.gitlab/ci/lint.gitlab-ci.yml
@@ -7,7 +7,6 @@ lint37:
   stage: lint
   image: "before-install"
   needs: []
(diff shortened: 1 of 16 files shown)


alexAubin (approved):
LGTM but I would keep `allow_failure: true`, otherwise we gonna have constant job failing because of linter stuff :/ But black should take care of most of it

PR #1165, opened 2021-02-10
[xmpp] setup http_upload_external

larchange:
## The problem

http upload is currently managed by metronome, but it's a deprecated part of metronome which has a limited mime types handling (For example pdf or doc file are not in the default handled mime types).

http upload is not the responsibility of metronome and need to be outsource.

## Solution

use the external http upload way describe on metronome documentation.

## PR Status

PENDING check/test/approval

## How to test

- Try to send a pdf file, 
- confirm the transfer failure
- apply the patch
- regenerate configurations for metronome and nginx
- try to send again a pdf and other kinds of files
- see the success .. or not !

### DIFF
--- a/data/hooks/conf_regen/12-metronome
+++ b/data/hooks/conf_regen/12-metronome
@@ -5,6 +5,12 @@ set -e
 do_pre_regen() {
   pending_dir=$1
 
+  # generate secrets for xmpp upload
+  [[ -f /etc/yunohost/http_file_secret ]] \
+     || tr -dc A-Za-z0-9 </dev/urandom | head -c 64 > /etc/yunohost/http_file_secret
+  [[ -f /etc/yunohost/http_file_delete_secret ]] \
+     || tr -dc A-Za-z0-9 </dev/urandom | head -c 64 > /etc/yunohost/http_file_delete_secret
+
   cd /usr/share/yunohost/templates/metronome
 
   # create directories for pending conf
@@ -14,6 +20,8 @@ do_pre_regen() {
 
   # retrieve variables
   main_domain=$(cat /etc/yunohost/current_host)
+  http_file_secret=$(cat /etc/yunohost/http_file_secret)
+  http_file_delete_secret=$(cat /etc/yunohost/http_file_delete_secret)
 
   # install main conf file
   cat metronome.cfg.lua \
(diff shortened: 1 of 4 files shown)


larchange:
@maranda finally we may have external http upload
@pitchum here the PR for you to test :) Let me know what piece of code if any are not at the correct locations.

larchange:
maybe we can add 
add_header Content-Security-Policy "default-src 'none'; frame-ancestors 'none';";

as advised in https://xmpp.org/extensions/xep-0363.html#server pointed by @randomstuff

data/hooks/conf_regen/12-metronome
```diff
@@ -14,6 +20,8 @@ do_pre_regen() {
 
   # retrieve variables
   main_domain=$(cat /etc/yunohost/current_host)
+  http_file_secret=$(cat /etc/yunohost/http_file_secret)
+  http_file_delete_secret=$(cat /etc/yunohost/http_file_delete_secret)
```
zamentur, line 24: We probably need to rename those files

data/hooks/conf_regen/12-metronome
```diff
@@ -24,9 +32,20 @@ do_pre_regen() {
...
     cat domain.tpl.cfg.lua \
       | sed "s/{{ domain }}/${domain}/g" \
+      | sed "s/{{ HTTP_FILE_SECRET }}/${http_file_secret}/g" \
+      | sed "s/{{ HTTP_FILE_DELETE_SECRET }}/${http_file_delete_secret}/g" \
       > "${metronome_conf_dir}/${domain}.cfg.lua"
   done
 
+  # create the upload endpoint
+  [[ -d "/var/www/metronome/xmmp-upload" ]] \
+    || mkdir -p /var/www/metronome/upload/
```
zamentur, line 42: Seems a bug "xmpp-upload" vs "upload"

zamentur:
We should backup and restore the xmpp-upload dir and the secret you added.

data/hooks/conf_regen/12-metronome
larchange: sure, what name do you propose ?

data/hooks/conf_regen/12-metronome
larchange: aie ! yes it's a bug... will change all these soon

larchange:
> We should backup and restore the xmpp-upload dir and the secret you added.

Do we really want to backup the xmpp-upload folder ? if we spawn a new instances somewhere else, metronome will not keep track of the files, and they will stay forever in the folder ...

data/hooks/conf_regen/12-metronome
larchange: fixed

data/templates/metronome/share.tpl.php
```diff
@@ -0,0 +1,189 @@
+<?php
```
zamentur, line 1: Do you know if equivalent script exist in python (we would avoid to add php in dependence of yunohost just for xmpp upload feature

larchange: This should not be complicated to rewrite it in python. I just took the official metronome example script. I can do it when I have time

larchange: Could we iterate instead of having the perfect PR ? Because this PR is quite useful.

zamentur:
I suggest to use this perl implementation instead https://github.com/weiss/ngx_http_upload . No new dependencies, no daemon to run.

There is also a flask implementation: https://github.com/horazont/xmpp-http-upload

In any case, Metronome should be a yunohost app, cause a lot of server don't use it.

alexAubin:
Closing because it's been 2 years and the plan is more to go towards metronome-as-an-app rather than handling all its complexity from YunoHost core ...

PR #1168, opened 2021-02-16
Enhance the jinja template for bash helpers

Salamandar:
This merge request implement multiple changes in the helper doc template :

* use `</br>` instead of `<space><space>` for line break (spaces at the end of line may be removed by text editors)
* Remove a lot of empty lines created by the Jinja templates (use of `{%- … %}` instead of simple `{% … %}`)

* the biggest change I think : ` {{ h.details.replace('\n', '</br>').replace('_', '\_') }} ` is replaced by ` {{ h.details }} `

My explanation for the latest change : I think the generated markdown should be humanly readable (if not, what's the interest of markdown ?), and those replacements are not needed if the source files are correctly formatted as markdown.


alexAubin:
Uuuuh have you tested the rendering though (specifically talking about the last change) ? I guess now that we're on Grav the rendering is a bit different so things may indeed change ... but I'm pretty sure we at least need the `.replace('_', '\_')` ... that's because helper names are typically `ynh_foo_bar` but the markdown parser may interpret the `_` as emphasis ...

Salamandar:
> that's because helper names are typically ynh_foo_bar but the markdown parser may interpret the _ as emphasis ...

Yeah, well, why not write it with quotes instead ? (in the shell file i mean)
```
`ynh_foo_bar`
```

The issue with this `replace()` is that we can't use backticks (code blocks) : underscores are replaced in the code blocks too… And that's problematic.

Salamandar:
For the moment, you can't get a correct code block. You can only get one of those : 
* no code block
Capture d’écran du 2021-02-16 14-41-09

* code block with weirdly escaped underscores :
  
Capture d’écran du 2021-02-16 14-41-20

alexAubin:
Ah yes, I guess I'm starting to understand what you mean :sweat_smile: 

Well yup let's go for this then, it just sound a bit fastidious to review/adapt all the helper doc but it's good if somebody is motivated to do it :yum:

Salamandar:
Well I'm already doing it. Not for everything, of course. But little by little everything will be OK.

alexAubin (approved):

kay0u (approved):

alexAubin:
(So uuuuuh it's not clear to me, shall we already merge this or are we waiting for some fixes on helpers ?)

Salamandar:
@alexAubin I guess this could be merged, since the doc is "broken" already. I'll work on more fixes progressively.

PR #1169, opened 2021-02-22
(TBD) Don't raise an exception if legacy permission migration fails

alexAubin:
## The problem

Follow-up of several stories where the legacy permission migration failed because either:
- weird regexes where `.` was replaced by `-` (or viceversa) for reasons
- the domain doesn't exists anymore because people used `change_url` and the script did not propagate the change to the regex...

Then the entire migration (or backup restore) explodes and the user has to manually fix the setting file ...

## Solution

Do not miserably fail if the reason for failure is domain unknown

## PR Status

Yolocommited, not tested

## How to test

Zblerg

### DIFF
--- a/src/yunohost/permission.py
+++ b/src/yunohost/permission.py
@@ -879,7 +879,7 @@ def _validate_and_sanitize_permission_url(url, app_base_path, app):
         path = "/" + path
 
         if domain.replace("%", "").replace("\\", "") not in domains:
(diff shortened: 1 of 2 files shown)


alexAubin:
I wrote this but don't even know if we do really want this behavior ... One issue is that if we the permission ain't properly migrated here, it won't be at all, and later during the app upgrade, the app may not even realize that it should migrate to the new system because the legacy permission doesn't even exists anymore...

PR #1170, opened 2021-02-22
[metronome] activate module pubsub

larchange:
## The problem

Pub sub XEP060 allow several features see https://xmpp.org/extensions/xep-0060.html
Some problems happens when not activated:
- Avatars are not synchronized when contacts are not connected
- OMEMO need PUBSUB XEP060 and XEP0163 to correctly exchange keys https://xmpp.org/extensions/xep-0384.html#intro-overview 

## Solution

Activate the module

## PR Status

Config done, PENDING approval

## How to test

publish a new avatar, see if it synchronize with other device
connect a new user with a new device, disconnect it talk to him/her with OMEMO activated. message should be correctly encrypted even for the disconnected devices.


zamentur (approved):
LGTM (untested)

PR #1179, opened 2021-03-03
[fix] Add backup for multimedia files

zamentur:
## The problem

Multimedia directory is not backup and i receive request from borg_ynh users about that

## Solution

Add a hook for saving multimedia directory

About legacy management, i don't know exactly what to do between:

- Nothing the multimedia dir is included by default after upgrade to the new yunohost version (could be problematic with daily backup on some instances)
- Create a migration (with an explanation and ability to skip) that deactivate the multimedia backup by creating a .nobackup inside this dir 
- Don't include by default this hook on yunohost backup create action

## PR Status

Ready

## How to test

...

### DIFF
--- a/data/hooks/backup/18-data_multimedia
+++ b/data/hooks/backup/18-data_multimedia
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Exit hook on subcommand error or unset variable
+set -eu
+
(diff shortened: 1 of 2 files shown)


alexAubin:
To be merged in 4.2-and-half

PR #1181, opened 2021-03-08
Feature: Add postfix backup in yunohost hooks

trogeat:
## The problem

Today postfix configuration is not included in backups.
We need backups so i patched Yunohost code ans hope it will fit community needs

## Solution

Updated mail backup hook to add aliases and postfix configuration.

## PR Status

Review and approval needed.

## How to test

make a backup/restore

### DIFF
--- a/data/hooks/backup/22-conf_mail
+++ b/data/hooks/backup/22-conf_mail
@@ -2,8 +2,9 @@
 
 source /usr/share/yunohost/helpers
 ynh_abort_if_errors
-YNH_CWD="${YNH_BACKUP_DIR%/}/conf/dkim"
+YNH_CWD="${YNH_BACKUP_DIR%/}/conf/mail"
 mkdir -p "$YNH_CWD"
 cd "$YNH_CWD"
 
 ynh_backup --src_path="/etc/dkim"
+ynh_backup --src_path="/etc/postfix"
--- a/data/hooks/restore/22-conf_mail
+++ b/data/hooks/restore/22-conf_mail
@@ -1,9 +1,16 @@
 #!/bin/bash
 
-backup_dir="$1/conf/dkim"
+backup_dir="$1/conf/mail"
 
 cp -a $backup_dir/etc/dkim/. /etc/dkim
 
 chown -R root:root /etc/dkim
 chown _rspamd:root /etc/dkim
 chown _rspamd:root /etc/dkim/*.mail.key
+
+cp -a $backup_dir/etc/postfix/. /etc/postfix
+chown -R root:root /etc/postfix
+
+cp -a $backup_dir/aliases /etc/aliases
+chown -R root:root /etc/aliases
+postalias /etc/aliases


alexAubin:
Can you elaborate on why you need / want to backup this

The postix conf is automatically regenerated by the regen-conf mecanism of Yunohost, so there's no real point in backing it up / restoring it

Historically we indeed did stuff like backup/restore the nginx conf, but the same things apply : it's regenerated by the regenconf

At best, this is harmless because the regenconf will just overwrite stuff ... but in some edge case scenarios, this may in fact break the freshly restored system (e.g. if restoring on a newer version of yunohost) because having an old-format conf on a new system may trigger unexpected stuff ...

trogeat:
In our case we added aliases to create kind of mailing list +virtual file and we wanted to be able to backup/restore these files.

trogeat:
Can i submit something in a different way that you would accept ? :-)

alexAubin:
Zblerg I dunno, I'm currently pushing in the "opposite" direction of what's proposed here, in https://github.com/YunoHost/yunohost/pull/1184 and https://github.com/YunoHost/yunohost/pull/1185 - such that as explained before, we don't backup/restore the nginx conf anymore because it's supposed to be automatically regenerated.

But then yeah, what about manually modified files and custom use cases ?

In #1185 I add a backup/restore hook named conf_manually_modified_files which should cover part of your case. Though I'm guessing that you also may have some alias db or whatever on the side that are not "registered" in Yunohost (Yunohost is only aware of the file it actively controls). I see several directions to try to cover this: 
- find a way to make yunohost's regen conf "aware" of some files even if it doesn't actively modify them though i have no idea exactly how that should work
- on your very own instance, add a custom backup/restore hook that would cover your specific need about the postfix conf ... it's not really *that* hard, it's just boring as hell and you need to know where to put the stuff.
- which brings the third solution: have a small mecanism in yunohost that allows to easily declare a bunch of additional files to backup. Let's say for example `/etc/yunohost/extra_files_to_backup_restore.txt` with one line = one path/glob to backup/restore. Could be managed by `conf_manually_modified_files` which we could rename `conf_manual_customizations` or idk.

zamentur:
Note, if your choice is on the custom hook for your own instance only, you can find an hook example here: https://yunohost.org/en/backup/include_exclude_files .

A 4th solution, could be to create a PR to add some mailing list feature in yunohost based on ldap groups. I already did some test to get the "Reply to the list" button in mail client.

Is it possible for you to document what you have changed in configuration to get your feature ? It could help to create a future PR.

alexAubin:
Closing cf today's discussion:
- https://github.com/YunoHost/yunohost/pull/1185 partially addressed the issue
- https://github.com/YunoHost/yunohost/pull/1539 as well
- you may also want to look at https://yunohost.org/en/packaging_apps_hooks#backup and corresponding forum thread "Share your hooks" for examples

PR #1183, opened 2021-03-09
Rework the authenticator system, split the LDAP stuff into auth part and utils part

alexAubin:
## The problem

- It's weird that all the LDAP auth / utils happens in Moulinette despite being quite Yunohost specific (I mean each time we want to tweak it we have to do a PR on both yunohost and moulinette)
- The LDAP auth versus utils part are in fact too separate things yet handled by the same piece of code ... but in one case we only login with the admin identity just as an authentication mecanism, and then the rest of the code re-logs as root to have all the interface about search() etc... Anyway to me it just makes more sense to have them in too separate files
- c.f work on user password reset / subscribe, ideas around having a proper API for user stuff rather than the current mess in SSOwat, and the research on friend-2-friend / federation : for all of these we'll want specific authentication layer that are not based on the LDAP admin

## Solution

Move the LDAP auth and utils into Yunohost, in two separate files

## PR Status

Tested and working on my dev LXC

## How to test

Run commands in CLI and test the webadmin

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -33,18 +33,9 @@
 #     Global parameters     #
 #############################
 _global:
-    configuration:
-        authenticate:
-            - api
-        authenticator:
-            default:
-                vendor: ldap
-                help: admin_password
-                parameters:
-                    uri: ldap://localhost:389
-                    base_dn: dc=yunohost,dc=org
-                    user_rdn: cn=admin,dc=yunohost,dc=org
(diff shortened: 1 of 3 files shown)


alexAubin:
(Btw one related topic to this is to possibly oneday™ move to a simpler auth mecanism for the webadmin that wouldnt require slapd being up - we just need some place to store the admin password and keep it in sync. That would make the webadmin login more resilient against slapd being broken)

PR #1185, opened 2021-03-10
Don't backup stuff that ain't relevant for backup

alexAubin:
## The problem

At best it makes no sense to backup the SSH conf, the ssowat conf, the certificates, the XMPP conf or the nginx conf. At worst, it'll create inconsistencies or make the restore explode, as experienced a couple times by users, because of funky edge cases.

## Solution

Don't backup/restore these.

Add a backup/restore hook for manually modified files

Add a backup/restore hook for xmpp data, which weren't properly backuped.

## PR Status

Yolocommited, needs testing

## How to test

Idk make backup / restore of stuff


zamentur:
I don't find where certificates are regenerated ? 
If we don't backup it we need to ask let's encrypt for certs.This could means downtime with TLS warning due to dns cache with LE... In more user should think to ask for certificate at the good time.

PR #1186, opened 2021-03-11
Add option to skip including big data in the backup

kuba-orlik:
(See https://github.com/YunoHost/issues/issues/1751)

## The problem

Some backup setups cause ynh cli to copy way more data than needed to create a full snapshot.

## Solution

Add option to exclude directories marked as bid (is_big) from the backup archive

## PR Status

Tested on my instance

## How to test

Run `SKIP_BIG_DATA=1 yunohost backup create --method copy -o somedir`

### DIFF
--- a/data/helpers.d/backup
+++ b/data/helpers.d/backup
@@ -60,24 +60,31 @@ CAN_BIND=${CAN_BIND:-1}
 #
 # Requires YunoHost version 2.4.0 or higher.
 # Requires YunoHost version 3.5.0 or higher for the argument --not_mandatory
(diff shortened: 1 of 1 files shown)


alexAubin (approved):
So follow-up of the discussion on the original issue : I'm not sure to understand what's the difference with what `BACKUP_CORE_ONLY` does already ?

(Granted that the semantic is a mess and it's kind of a "hidden" option)

kuba-orlik:
...you're right. Thanks for pointing it out

PR #1187, opened 2021-03-12
Improve error semantic

alexAubin:
## The problem

Follow-up of the rework of the webadmin in VueJS : there are some pretty obvious "bad UX" stuff in the way errors are handled ... which is in fact due to the lack of proper semantic for exceptions. Basically all MoulinetteError / YunohostError right now trigger a 400 error with no distinction between 
- "validation / form error" (for example trying to create a user that already exists, therefore interface should stay on the form and display the error there)
- "an app script miserably failed" (which should redirect to the corresponding log so hopefully people will share it)

## Solution

See also https://github.com/YunoHost/moulinette/pull/271

- Introduce some `MoulinetteValidationError` and `YunohostValidationError` that corresponds to code 400 / Bad request.
- Classical `MoulinetteError` and `YunohostError` correspond to code 500 / Internal server error.
- Have also a `MoulinetteAuthenticationError` corresponding to 401 / Unauthorized though that http code was already pretty much well handled through other mecanism
- When an exception occurs in a `@is_unit_operation`, injects a reference to the log that can later be caught by the webadmin to redirect to the appropriate log page. This is a generalization of https://github.com/YunoHost/yunohost/pull/1077 which was only applied on app install failures

## PR Status

- [ ] Also cover cases where the OperationLogger is created manually instead of using `@is_unit_operation`
- [ ] Propagate the changes to the error handling in the webadmin

## How to test

Uuuuh try to trigger various errors from the webadmin

### DIFF
--- a/tests/test_i18n_keys.py
+++ b/tests/test_i18n_keys.py
@@ -29,8 +29,8 @@ def find_expected_string_keys():
     #    # i18n: foo
     p1 = re.compile(r"m18n\.n\(\n*\s*[\"\'](\w+)[\"\']")
     p2 = re.compile(r"YunohostError\(\n*\s*[\'\"](\w+)[\'\"]")
-    p2 = re.compile(r"YunohostValidationError\(\n*\s*[\'\"](\w+)[\'\"]")
-    p3 = re.compile(r"# i18n: [\'\"]?(\w+)[\'\"]?")
+    p3 = re.compile(r"YunohostValidationError\(\n*\s*[\'\"](\w+)[\'\"]")
+    p4 = re.compile(r"# i18n: [\'\"]?(\w+)[\'\"]?")
 
     python_files = glob.glob("src/yunohost/*.py")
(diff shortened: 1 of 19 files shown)


tests/test_i18n_keys.py
```diff
@@ -25,9 +25,11 @@ def find_expected_string_keys():
     # Try to find :
     #    m18n.n(   "foo"
     #    YunohostError("foo"
+    #    YunohostValidationError("foo"
     #    # i18n: foo
     p1 = re.compile(r"m18n\.n\(\n*\s*[\"\'](\w+)[\"\']")
     p2 = re.compile(r"YunohostError\(\n*\s*[\'\"](\w+)[\'\"]")
+    p2 = re.compile(r"YunohostValidationError\(\n*\s*[\'\"](\w+)[\'\"]")
```
kay0u, line 32: You should probably create a new variable, and use it in the for loop below instead of overwrite p2.

alexAubin: Ugh how drunk was I

PR #1192, opened 2021-03-19
Uniformize API routes

alexAubin:
## The problem

Followup of the rewrite using VueJS, @Axolotle identified that our API routes nomenclature is chaotic

## Solution

- `POST/DELETE` to create/delete resources
- `PUT /foos/<id>/action` for actions on resources
- Don't expose routes that are deprecated or not relevant

## PR Status

Yolocommited

## How to test

Try the different routes manually I suppose @_@

I'm a bit concerned about wether ambiguous stuff like having both `/ressource/<foo>` and `/ressource/action_bar` is properly handled but I think it's okay


alexAubin:
Added a super boring commit about supporting `yunohost tools update system|apps` instead of `--system/--apps` ... and same for `tools upgrade` ... with the original goal of being able to have an API route like `PUT /update/<target>` :expressionless:

Axolotle:
Splitted some routes that looks like `<action>/update` and data like `{ add, remove }` by 2 commands `<action>/add` `<action>/remove`.  

Will try to check functions returned values next week !

PR #1196, opened 2021-03-22
Switch from n to nodenv

yalh76:
## The problem

None, just improving

## Solution

Using *env for various language
https://github.com/YunoHost-Apps/Experimental_helpers/pull/82

## PR Status

Ready and tested

## How to test

Like done for https://github.com/YunoHost-Apps/wikijs_ynh/pull/137

Important points:
- This helper doesn't uninstall previous n and nodejs installation
- It will uninstall previous n and all nodejs installation if there is no more app using nodejs => no apps has a `nodejs_version` setting


yalh76:
to check:
disk space difference require

yalh76:
Won't be merged for now, will break nodejs based packaged

PR #1202, opened 2021-03-28
Update log.py

Stylix58:
## The problem
If you try to use the "installation from GitHub" way for install a app and if you don't reply correctly, He throw a `AttributeError` exception who say `'NoneType' object has no attribute 'strftime'`.

## Solution
Import the `strftime` function from standard Python's libray `time` and use this function for fix the problem.

## PR Status
- Waiting for be tested

### DIFF
--- a/src/yunohost/log.py
+++ b/src/yunohost/log.py
@@ -30,6 +30,7 @@ import yaml
 import glob
 import psutil
 
+from time import strftime
 from datetime import datetime, timedelta
 from logging import FileHandler, getLogger, Formatter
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Eeeh good catch for the error but I think the fix ain't the good one ... The issue was apparently introduced by an earlier commit from me: https://github.com/YunoHost/yunohost/commit/37f0c30ddc8cf50a0fc7839026d6e604bff32810

When an error occurs during an unit_operation, we want to inject log_ref which is then used by the webadmin to redirect to the log, c.f. explanation in https://github.com/YunoHost/yunohost/pull/1187

hence `error.log_ref = self.name` in `close()` ... but I didn't realize when writing this that `self.name` is in fact a computed property and not a static attribute that depends on the start_date. So in fact, if there's no start date, we should at all do this operation - because there will be no log to redirect to (the operation did not really start).

So the proper fix should be to add an `and self.start_date` in the if in `close()`

alexAubin:
Fixed in https://github.com/YunoHost/yunohost/commit/cd1f64383bdac3994fec2ca7a39e02e63a7a1372

PR #1203, opened 2021-04-02
Drop support for backups from prior 3.8, introduce hooks in migrations to apply migrations during restore

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1758

## Solution

- Drop support for 3.8 because meh, legacy code
- Introduce a proper mechanism to trigger migrations during backup restore

## PR Status

Yolocommited

## How to test

Zblerg idk run the unit tests


alexAubin:
Merging to move forward because this is somewhat needed for like 3 other pending PRs

PR #1207, opened 2021-04-11
Diagnosis command : use "--human-readable"

lapineige:
As recommended in the documentation https://yunohost.org/fr/install/hardware:vps_debian

## The problem
Diagnosis command advice does not include an "human readable" output option. 

## Solution
Add the option


lapineige:
I have to admit I did not find the courage to change it for every locale 😅

alexAubin:
Cheers !

PR #1209, opened 2021-04-12
add ssh port setting

kay0u:
## The problem

...

## Solution

Close https://github.com/YunoHost/issues/issues/1740

We probably want a migration to handle the current port, IDK if we need to do something for https://github.com/YunoHost/yunohost/blob/ce64a6380abfe0c8b9892172fd72c901c2685d36/data/templates/yunohost/services.yml#L46

## PR Status

...

## How to test

...

### DIFF
--- a/data/hooks/conf_regen/03-ssh
+++ b/data/hooks/conf_regen/03-ssh
@@ -26,6 +26,8 @@ do_pre_regen() {
     # Support different strategy for security configurations
     export compatibility="$(yunohost settings get 'security.ssh.compatibility')"
 
+    export port="$(yunohost settings get 'security.ssh.port')"
+
     export ssh_keys
     export ipv6_enabled
     ynh_render_template "sshd_config" "${pending_dir}/etc/ssh/sshd_config"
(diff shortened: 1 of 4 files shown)


kay0u:
I've found this dirty hack too:
https://github.com/YunoHost/yunohost/blob/ce64a6380abfe0c8b9892172fd72c901c2685d36/src/yunohost/service.py#L683-L688

alexAubin:
Sounds okay because as you saw, during the diagnosis / service stuff we fetch the port directly in the file so that covers the case where it got manually modified (and/or defined through settings)

zamentur:
What about fail2ban, i remember some port should be change in Fail2ban.

kay0u:
I don't fully understand this conf file, but there is no 22 port in this file
https://github.com/YunoHost/yunohost/blob/fc818cc6a7ec9281adc24a3b98fd6c0a6e761a8a/data/templates/fail2ban/jail.conf#L224-L250

zamentur:
yep but ssh mean "default ssh port" so 22...

alexAubin:
Fixed in https://github.com/YunoHost/yunohost/commit/37c0825eed421acadd0a7528dbab8172695eed8f

PR #1211, opened 2021-04-15
[enh] add header to disallow FLoC

Psycojoker:
## The problem

Google is starting to use FLoC and it's a privacy nightmare, see https://www.eff.org/deeplinks/2021/03/google-testing-its-controversial-new-ad-targeting-tech-millions-browsers-heres

## Solution

While destroying google would be a great thing in the meantime we can apparently disallow it using a nginx header, see https://diaspodon.fr/@etienne/106070042112522839

## PR Status

Tested on my prod and used https://securityheaders.com linked in the toot to see if it is working and the website is happy.

## How to test

Apply the patch on your local nginx conf, reload nginx and check the headers or use security headers.

### DIFF
--- a/data/templates/nginx/security.conf.inc
+++ b/data/templates/nginx/security.conf.inc
@@ -33,6 +33,9 @@ more_set_headers "X-Download-Options : noopen";
 more_set_headers "X-Permitted-Cross-Domain-Policies : none";
 more_set_headers "X-Frame-Options : SAMEORIGIN";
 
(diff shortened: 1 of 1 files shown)


randomstuff:
Would it still possible to obtain the cohort ID and display a creepy message accordingly?

Psycojoker:
> Would it still possible to obtain the cohort ID and display a creepy message accordingly?

I don't think so? Maybe on the first request, but you'll need to read the specs of that thing.

PR #1212, opened 2021-04-15
[enh] by default be very strict on the Permissions-Policy header

Psycojoker:
## The problem

I'm not 100% sure about this one.

By setting the Permissions-Policy header we can create a very strict policy for the browser and avoiding leaking a lot of information that are opened by default. This will prevent the app from accessing them including rogue javascript or things like that that could leak them.

On the downside, this will force every app developers for which apps needs some permissions to explicitly allow them thus breaking the feature of some apps that are already working, so : communications, temporary bugs etc ... I'm not sure if it's worth it?

Reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy

## PR Status

Working on my prod.

## How to test

Apply the patch on your prod and reload nginx then look at the headers.

### DIFF
--- a/data/templates/nginx/security.conf.inc
+++ b/data/templates/nginx/security.conf.inc
@@ -33,8 +33,8 @@ more_set_headers "X-Download-Options : noopen";
 more_set_headers "X-Permitted-Cross-Domain-Policies : none";
 more_set_headers "X-Frame-Options : SAMEORIGIN";
 
-# Disable the disaster privacy thing that is FLoC
-more_set_headers "Permissions-Policy : interest-cohort=()";
+# Disable most possible leaking things by default including the horrible FLoC
+more_set_headers "Permissions-Policy : fullscreen=(), geolocation=(), display-capture=(), gyroscope=(), ambient-light-sensor=(), payment=(), accelerometer=(), battery=(), camera=(), microphone=(), magnetometer=(), usb=(), interest-cohort=(), publickey-credentials-get=()";
 
 # Disable gzip to protect against BREACH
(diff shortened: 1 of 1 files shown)


data/templates/nginx/security.conf.inc
```diff
@@ -33,8 +33,8 @@ more_set_headers "X-Download-Options : noopen";
 more_set_headers "X-Permitted-Cross-Domain-Policies : none";
 more_set_headers "X-Frame-Options : SAMEORIGIN";
 
-# Disable the disaster privacy thing that is FLoC
-more_set_headers "Permissions-Policy : interest-cohort=()";
+# Disable most possible leaking things by default including the horrible FLoC
+more_set_headers "Permissions-Policy : fullscreen=(), geolocation=(), payment=(), accelerometer=(), battery=(), camera=(), microphone=(), magnetometer=(), usb=(), interest-cohort=()";
```
alexAubin, line 37: Eeeh maybe we can keep allowing camera+microphone, otherwise that may spontaneously break Nextcloud talk or Galene (at least for my work we're using a Galene instance, but i know some people use Nextcloud talk too ?)

I guess the other ones are okay ?

Psycojoker: Or we modify those headers in Nextcloud talk or Galene? That's a bit the idea of allowing to overwrite them

But yeah, that's the problem that will rise :/

alexAubin:
To be addressed in https://github.com/YunoHost/yunohost/pull/1290

PR #1213, opened 2021-04-15
Add route to fetch app manifest for custom app installs in a forge-agnostic way

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1117

Currently, custom app install work by fetching **from the client** the raw manifest.json. This was good enough so far but recently break after I started advising removing the superflous ask strings ... But also it worked with tight-coupling to raw.githubuserstuff.com, etc

## Solution

Add a `yunohost app manifest foobar`, which can also be fed an url, and is meant to be used by the API such that the backends fetch the manifest in a forge-agnostic way.

## PR Status

Tested and working on my side

## How to test

c.f. corresponding PR on yunohost-admin

### DIFF
--- a/data/actionsmap/yunohost.yml
+++ b/data/actionsmap/yunohost.yml
@@ -608,6 +608,14 @@ app:
                 string:
                     help: Return matching app name or description with "string"
 
+        ### app_manifest()
+        manifest:
+            action_help: Return the manifest of a given app from the manifest, or from a remote git repo
+            api: GET /apps/manifest
+            arguments:
+                app:
+                    help: Name, local path or git URL of the app to fetch the manifest of
+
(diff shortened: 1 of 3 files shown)


data/actionsmap/yunohost.yml
```diff
@@ -608,6 +608,14 @@ app:
                 string:
                     help: Return matching app name or description with "string"
 
+        ### app_manifest()
+        manifest:
+            action_help: Return the manifest of a given app from the manifest, or from a remote git repo
```
kay0u, line 613: `from the manifest` ?

alexAubin: Ugh from the catalog* sry

PR #1216, opened 2021-04-24
Add ssh.app, sftp.app groups to cover my_webapp and borg needing ssh access

alexAubin:
## The problem

https://github.com/YunoHost/issues/issues/1784

## Solution

Introduce new ssh.app and sftp.app groups

Apps can add their system user to this group by adding for example `--groups="ssh.app"` to `ynh_system_user_create`

## PR Status

Tested and working on my side
Though we'll have to create PR to propagate that change on borg, restic, my_webapp

## How to test

Zblerg idk, run the regen-conf, run the unit test for ynh_system_user_create, try to connect with the created user

### DIFF
--- a/data/helpers.d/user
+++ b/data/helpers.d/user
@@ -92,10 +92,11 @@ ynh_system_group_exists() {
 
 # Create a system user
 #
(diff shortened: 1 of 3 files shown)


kay0u (approved):
:+1:

PR #1217, opened 2021-04-24
Add diagnosis section to check that app are in catalog with good quality, check for deprecated practices

alexAubin:
## The problem

There's no feedback provided to the admins if one of their installed app is removed from the catalog, or becomes low quality, or if they didn't upgrade the app in a long time and the app use deprecated helpers.

The latter was found to be super problematic on an old server a while ago, because the app was still using checkurl or i-dont-remember-which-helper, which made stuff almost explode during a backup/restore attempt.

The intention is also that app packagers would get notified that their app is in bad state, if they have this app installed on one of their server.

## Solution

Add a diagnosis section which will report apps that:
- ain't in the catalog
- are "broken" (no level or level 0)
- (warning) low quality, i.e. level <= 4
- the app requires only yunohost 2.x which is indicative that the app didn't get upgrade in a while, already checked by the linter
- the app uses super outdated helpers such as "yunohost app initdb", already checked by the linter

The design is debatable as people will get a warning/error in their diagnosis if a maintained, popular app temporarily goes below level 4 .. (for example mastodon recently) I tried to design the message in such a way that it doesn't create too much panic:
>This may be a temporary issue while the maintainers attempt to fix the issue. In the meantime, upgrading this app is disabled.

## PR Status

Tested and working on my side

## How to test

For example I tried to install [helloworld ynh](https://github.com/YunoHost-Apps/helloworld_ynh) which ain't in the catalog (maybe it should but that's another story, idk), and ran the diagnosis


alexAubin:
(Bumpity bump in case somebody has an opinion about this)

kay0u (approved):
LGTM

PR #1225, opened 2021-05-13
Updating requirements

yalh76:
## The problem

Some requirements are not OK for functions using `ynh_add_confg`

## Solution

Update requirements

## PR Status

Done

## How to test

Nothing to test


alexAubin:
Uuuuh wokay but that's probably linter warnings for apps that are using these and require some version of yunohost prior to 4.1 .. Or is it precisely what's intended ?

yalh76:
Yes and no. I don't know how the linter implement the requirement. It was at least to document it.

PR #1227, opened 2021-05-17
[fix] Email on certificate renewing failed

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/1803
This fix only the tail not the reason why the renewing failed

## Solution

use subprocess to call tail command.

Note we could instead use tail function from services.py

## PR Status

Just tested with a python3 interpreter

## How to test

...

### DIFF
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -949,11 +949,11 @@ def _name_self_CA():
 
 
 def _tail(n, file_path):
-    stdin, stdout = os.popen2("tail -n %s '%s'" % (n, file_path))
-
-    stdin.close()
-
-    lines = stdout.readlines()
-    stdout.close()
-
-    return "".join(lines)
+    p = subprocess.Popen("tail -n %s '%s'" % (n, file_path), close_fds=True, 
+                         shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    
+    p.stdin.close()
+    lines = p.stdout.readlines()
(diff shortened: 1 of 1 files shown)


yalh76 (approved):

src/yunohost/certificate.py
```diff
@@ -949,11 +949,4 @@ def _name_self_CA():
...
 def _tail(n, file_path):
-    stdin, stdout = os.popen2("tail -n %s '%s'" % (n, file_path))
-
-    stdin.close()
-
-    lines = stdout.readlines()
-    stdout.close()
-
-    return "".join(lines)
+    return check_output("tail -n %s '%s'" % (n, file_path))
```
kay0u, line 952:
```suggestion
    from moulinette.utils.process import check_output
    return check_output("tail -n %s '%s'" % (n, file_path))
```

src/yunohost/certificate.py
```diff
@@ -949,11 +949,5 @@ def _name_self_CA():
...
-    stdin, stdout = os.popen2("tail -n %s '%s'" % (n, file_path))
-
-    stdin.close()
-
-    lines = stdout.readlines()
-    stdout.close()
-
-    return "".join(lines)
+    from moulinette.utils.process import check_output
+    return check_output("tail -n %s '%s'" % (n, file_path))
```
alexAubin, line 953:
```suggestion
    return check_output(f"tail -n {n} '{file_path}'")
```

PR #1230, opened 2021-05-20
[fix] Be able to init slapd in a chroot

zamentur:
## The problem

ISO image and debootsrap (and probably arm images) need to be able to install yunohost package in a chroot environement (so without any services running)

## Solution
Use slapadd, this command doesn't need slapd running. However slapmodify is only available in slapd 2.5. So we can't update the cn=admin object. That's why i put this specific part in the adminpw method.

## PR Status

Not completely tested (just partially with a debian chrooted)
Next step i think we should merge this in unstable and build an iso (or use debootstrap repo) with unstable branch to see if it fix things.

## How to test

...


alexAubin (commented):
Naively LGTM on principle, but we currently also use ldap_scheme.yml here: https://github.com/YunoHost/yunohost/blob/dev/src/yunohost/data_migrations/0020_ssh_sftp_permissions.py#L41

data/hooks/conf_regen/06-slapd
```diff
@@ -36,20 +35,31 @@ EOF
...
+      | grep -v "none elapsed\|Closing DB" || true
+  chown -R openldap: /var/lib/ldap
+
+  nscd -i groups
+  nscd -i passwd
 
-  yunohost tools shell -c "from yunohost.tools import tools_ldapinit; tools_ldapinit()"
+  systemctl restart slapd
+
+  mkhomedir_helper admin
```
zamentur, line 62: Not sure this one work inside chroot cause admin user doesn't exists as slapd is down.

PR #1238, opened 2021-05-26
allow cert installation when failing diagnostics are ignored

khimaros:
## The problem

https://github.com/YunoHost/issues/issues/1813

## Solution

- include `ignored` state in diagnosis cache
- check the `ignored` state when deciding whether to throw an error in cert creation

## PR Status

ready for review

## How to test

see linked issue for reproduction steps.

### DIFF
--- a/src/yunohost/diagnosis.py
+++ b/src/yunohost/diagnosis.py
@@ -465,9 +465,9 @@ class Diagnoser:
         new_report = {"id": self.id_, "cached_for": self.cache_duration, "items": items}
 
         self.logger_debug("Updating cache %s" % self.cache_file)
-        self.write_cache(new_report)
         Diagnoser.i18n(new_report)
         add_ignore_flag_to_issues(new_report)
+        self.write_cache(new_report)
 
         errors = [
             item
--- a/src/yunohost/certificate.py
+++ b/src/yunohost/certificate.py
@@ -881,13 +881,13 @@ def _check_domain_is_ready_for_ACME(domain):
     if not dnsrecords.get("status") in [
         "SUCCESS",
         "WARNING",
-    ]:  # Warning is for missing IPv6 record which ain't critical for ACME
+    ] and not dnsrecords.get("ignored"):  # Warning is for missing IPv6 record which ain't critical for ACME
         raise YunohostValidationError(
             "certmanager_domain_dns_ip_differs_from_public_ip", domain=domain
         )
 
     # Check if domain seems to be accessible through HTTP?
-    if not httpreachable.get("status") == "SUCCESS":
+    if not httpreachable.get("status") == "SUCCESS" and not dnsrecords.get("ignored"):
         raise YunohostValidationError(
             "certmanager_domain_http_not_working", domain=domain
         )


src/yunohost/diagnosis.py
```diff
@@ -465,9 +465,9 @@ def diagnose(self):
         new_report = {"id": self.id_, "cached_for": self.cache_duration, "items": items}
 
         self.logger_debug("Updating cache %s" % self.cache_file)
-        self.write_cache(new_report)
         Diagnoser.i18n(new_report)
         add_ignore_flag_to_issues(new_report)
+        self.write_cache(new_report)
```
alexAubin, line 471: Eeeeeh not sure about this change ? The fact that we do the i18n "rendering" with `Diagnoser.i18n(new_report)` *after* we wrote the report is a pretty big deal

Basically we want to save a locale-agnostic diagnosis report because we can't be sure that the locale we'll want to display the report with is the same as the locale used when the report was generated (for example, because the diagnosis is ran from a cron job, etc..)

But I'm assuming that the purpose of this change was related to the `add_ignore_flag_to_issues`  because you want to be able to access the ignore flag in the certificate.py code ... But for the same kind of reason we probably don't want to do that here

I think instead we should either : 
- call add_ignore_flag_to_issues in the relevant certificate.py code
- or maybe rely on diagnosis_show instead of diagnosis_get in the certificate.py code

zyphlar:
This bug is randomly preventing me from accessing my self-hosted resources because letsencrypt renewal is silently failing and I wake up and I can't get into my password manager anymore and my day grinds to a halt.

I get the error "the DNS records for domain xxxxxx is different from this server's IP" which I know and don't care about: my yunohost installation has a private VPS IP that's different from the static public IP and that's how it should be to work reliably. That's never caused SSL renewal issues in the past and it shouldn't be a problem now either. Until this is fixed, I guess I need to set up a cron job to renew my SSL with "--no-checks" so I can keep access to my server.

alexAubin:
Closing because it's been 2 years, the PR is incorrect because creates other issues, and I don't understand the initial issue ... If YunoHost's diagnosis reports that the DNS record IP is different from the server's IP, then ignoring the diagnosis is not gonna magically solve the issue ... If this is an advanced use case then ... yeah, no magic, YunoHost is meant to be one-size fits all, and you should be able to install the cert with `--no-checks` anyway ...

zyphlar:
@alexAubin you "can" install it with --no-checks but pragmatically what you're telling me is that I need to come up with my own cron jobs for auto-renewing my certbot/letsencrypt certs that don't conflict with the YunoHost scheduled jobs, and do my own monitoring/restitution of the renewal if/when they fail.

Which is what I've been doing for over two years, and is a huge pain in the butt, and often fails, which means I'm locked out of my self-hosted Bitwarden... you know, the password manager I use to store my Yunohost SSH password... it's a bunch of no fun.

What we "should" do is be more tolerant of configurations where YunoHost's ability to detect its own IP versus the IP entered into DNS and the fact that they may differ (such as on NATed connections, or on many cloud/VPS hosting providers who give you a static IP but only via NAT and not directly as an interface IP, such as on Amazon and DigitalOcean) because relying on this check and failing without it means anyone who prefers easy management tools like YunoHost and who don't know what "cron" or "ssh" mean will be shit out of luck and stuck with broken SSL, and thus broken self-hosting. It's a big deal.

If you don't understand all that, then the most likely case is that you're lucky enough to be using YunoHost with an interface IP that matches its public DNS IP. Which is the situation with some VPS hosts, but certainly not all.

PR #1244, opened 2021-06-01
[DRAFT] make IP.YUNOHOST.ORG redundant

frju365:
## The problem

If ip.yunohost.org is down, ip verification failed for all servers.

## Solution

Host different instances of ip.yunohost.org and make it redundant.

## PR Status

DRAFT

## How to test

...

## Prerequisites
0-ip.yunohost.org and 0-ip6.yunohost.org exist.

### DIFF
--- a/data/hooks/conf_regen/43-dnsmasq
+++ b/data/hooks/conf_regen/43-dnsmasq
@@ -6,6 +6,9 @@ set -e
 do_pre_regen() {
   pending_dir=$1
 
+  declare -a ip_yuno=(ip.yunohost.org 0-ip.yunohost.org)
+  declare -a ip_yuno6=(ip.yunohost.org 0-ip6.yunohost.org)
+
   cd /usr/share/yunohost/templates/dnsmasq
 
   # create directory for pending conf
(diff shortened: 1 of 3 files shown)


data/hooks/conf_regen/43-dnsmasq
```diff
@@ -6,6 +6,9 @@ set -e
 do_pre_regen() {
   pending_dir=$1
 
+  declare -a ip_yuno=(ip.yunohost.org 0-ip.yunohost.org)
+  declare -a ip_yuno6=(ip.yunohost.org 0-ip6.yunohost.org)
```
kay0u, line 10:
```suggestion
  declare -a ip_yuno6=(ip6.yunohost.org 0-ip6.yunohost.org)
```

frju365: Indeed, thanks. I make it draft because I didn't check it (test). :-)

zamentur:
I know my comments will be a little out of scope. This PR is about fiability and i will speak about security.

## How to check the returned value ?
This ip.yunohost.org is a security risk: couple with the dyndns and the future auto config dns mechanism, an attackers who get control on the php script behind ip.yunohost.org could choose to change IP behind A or AAAA record for all server using this mechanism. So this very small service is very very critical (as a lot of things on our infra).

It could be better if we were able to check the returned value. Several ideas to do that:

- Double check value on 2 ip services: ip.yunohost.org and 0-ip.yunohost.org ?
- Use traceroute to check how many nodes we follow when we traceroute the ip we get with ip.yunohost.org :  
  - major vps/ dedicated server and server with vpn has their ip directly set
  - at home, on some dedicated ip VPN, on some proxmox, we have NAT some times several NAT, but all those IP are on local ranges, so we can count number of nodes with public ips...
  - We might have proxy that redirect on server with dedicated ip. However in this case, why not deactivate this mechanism ?

Note : For major vps/ dedicated server and server with vpn has their ip directly set, we could also analyze ip route and watch ip is set on the default interfaces, if it's a public ip it's a big clue. However if the ip connexion doesn't really work, we could deduce an ip that is not really usable...

I probably should create a ticket with, that, but i say that here as you work pretty close on this part.

I am pretty sure ipv6 will be more difficult about this part

PR #1256, opened 2021-06-15
[enh] use rich.console.{print,format}_exception

Psycojoker:
## The problem

Default python traceback isn't very lisible and kinda sucks.

The [rich](https://github.com/willmcgugan/rich) module offers way better display which will increase UX and lisibility and debugging at not that much cost.

It can also be used for other cool things.

## Solution

Use rich.

Related: https://github.com/YunoHost/moulinette/pull/280

Looks pretty:

image

## PR Status

Tested and working.

It will sadly need a debian package for the rich module as it isn't in debian yet.

## How to test

Use this PR and the [moulinette's one](https://github.com/YunoHost/moulinette/pull/280) and generate and exception (like put "1 / 0" in the source code somewhere) and enjoy pretty tracebacks.


Psycojoker:
Urgh it's a disaster I've found out that [loguru]() uses [better-exception]() which is super interesting too because it display the values of the call which is like super easful for debugging like:

image

But:

- it doesn't show detailed lines around the exceptions
- it's density make it harder to read and understand
- it's not especially meant to be integrated like I had to produce all this code for that:

image

- and since it's not meant for that I fear that we won't have problem already solved by rich like testing if it's in a tty and so one

But gosh having values for debugging :star_struck: 

So I'm not sure which one to integrate... at least rich work is done and we have a lot other interesting features in rich and nothing prevent us from migrating one day to loguru to trash this horrible thing that is the logging module.

Psycojoker:
test

alexAubin:
Closing following discussion in today's meeting / The PR is unfortunately not really in working state : the changes in app_list etc. break the API .. Also it's probably too much of an overhead to add rich wrt to what it brings ... Nevertheless Rich is pretty interesting to have pretty-printing of stuff in CLI so may revisit this at some point

PR #1266, opened 2021-07-07
[enh] Add check for available disk space before app install/upgrade

lelibreauquotidien:
This PR resolve #1833 issue (https://github.com/YunoHost/issues/issues/1833)

## The problem

When installing or updating an application, the system may return a "no space left on disk" error.

## Solution

Add python code to verify disk space left

## PR Status

Works !

## How to test

Install test_ynh app (https://github.com/lelibreauquotidien/test_ynh) and update it with the instructions in README


src/yunohost/app.py
```diff
@@ -517,6 +517,10 @@ def app_upgrade(app=[], url=None, file=None, force=False):
     from yunohost.regenconf import manually_modified_files
 
     apps = app
+    # Check if disk space available
+    size = os.statvfs('/')
+    if (size.f_bavail * size.f_frsize) / 1024 <= 512000:
```
alexAubin, line 522: Sooo ideally it could be nice to use https://github.com/YunoHost/yunohost/blob/dev/src/yunohost/utils/filesystem.py#L24 

(to be checked if this need to be imported at the top of the file)

```suggestion
    if free_space_in_directory("/") <= 512 * 1000 * 1000:
```

src/yunohost/app.py
```diff
@@ -875,6 +879,11 @@ def confirm_install(confirm):
         manifest, extracted_app_folder = _extract_app_from_file(app)
     else:
         raise YunohostValidationError("app_unknown")
+    
+    # Check if disk space available
+    size = os.statvfs('/')
+    if (size.f_bavail * size.f_frsize) / 1024 <= 512000:
```
alexAubin, line 885: (same here)

```suggestion
    if free_space_in_directory("/") <= 512 * 1000 * 1000:
```
```

File: locales/fr.json [hunk 3]
```diff
@@ -44,6 +44,8 @@
     "backup_output_directory_required": "Vous devez spécifier un dossier de destination pour la sauvegarde",
     "backup_running_hooks": "Exécution des scripts de sauvegarde...",
     "custom_app_url_required": "Vous devez spécifier une URL pour mettre à jour votre application personnalisée {app:s}",
+    "disk_space_not_sufficient_install": "Il ne reste pas assez d'espace disque pour installer cette application",
+    "disk_space_not_sufficient_update": "Il ne reste pas assez d'espace disque pour mettre à jour cette application",
```
alexAubin, line 48: It should be okay, but note for next time : we tend to avoid editing locale files manually - apart from `en.json`, because this may create merge conflicts on our translation platform weblate

alexAubin (approved):
LGTM, possibly small improvement to use existing helper function :+1:

src/yunohost/app.py
lelibreauquotidien: OK, thanks for the suggestion.
Why multiply 2 times by 1000 and not 1 time by 2000?

alexAubin: It's mainly cosmetics to try to improve readability ... maybe not so much in this case, but in the diagnosis code we had a similar situation so i defined `MB = 1000 * 1000` so I could write `512 * MB` ... but anyway nevermind, the important stuff here is about using the existing helper function

PR #1269, opened 2021-07-10
[fix] set .ssh folder permissions to 600

Tagadda:
## The problem

`.ssh/` is created with `777` and break publickey authentication.

Fix YunoHost/issues#1770

## Solution

Set permissions to 600.

I use `chmod()` instead of setting `mode=0o600` in `mkdir()` in hope this will magically fix someone's folder.

## PR Status

Tested with ynh-dev. :ok_hand:

## How to test

Use `yunohost user ssh add-key`


alexAubin:
Cheers :+1:

PR #1276, opened 2021-07-22
Force utf8 for char set in mysql database

Josue-T:
Fix issue https://github.com/YunoHost-Apps/seafile_ynh/issues/77

## The problem

https://github.com/YunoHost-Apps/seafile_ynh/issues/77

## Solution

Force char set to UTF8

## PR Status

Not tested but hope that it might fix the issue

## How to test

...

### DIFF
--- a/data/helpers.d/mysql
+++ b/data/helpers.d/mysql
@@ -89,7 +89,7 @@ ynh_mysql_execute_file_as_root() {
 ynh_mysql_create_db() {
     local db=$1
 
-    local sql="CREATE DATABASE ${db};"
+    local sql="CREATE DATABASE ${db} CHARACTER SET = 'utf8';"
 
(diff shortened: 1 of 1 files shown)


zamentur:
What about restore operation ? and upgrade ?

Josue-T:
For the upgrade, I think the we could just leave as it is for the old install. And for the restoration we use the same helper.

zamentur:
Yes but export an SQL database in ISO and import it in utf-8 could create some encoding errors, no ?

kay0u:
I did something like that in this [PR](https://github.com/YunoHost-Apps/nextcloud_ynh/pull/384)

Neustradamus:
@alexAubin: Have you planned to add directly in the YunoHost based on Debian 11?

PR #1289, opened 2021-08-18
[enh] App manifest v2

alexAubin:
## The problem

c.f. https://github.com/YunoHost/issues/issues/1614#issuecomment-898914769

## Solution

- [x] (The following items are supposed to be only effective if manifest is explicitly declared with `packaging_format = 2`. Everything should be backward compatible / transparent for v1 apps)
- [x] Save all install question as settings (except user-provided passwords)
- [x] Load all settings in app scripts's environment (so e.g. directly define `$domain`, etc)
- [x] Auto-enable `set -eu` aka `ynh_abort_if_errors` in appropriate scripts
- [x] Handle the safety-backup-before-upgrade
- [x] Permissions are to be initialized with the appropriate `allowed` before the install script, no need to write a boring `if $is_public ...` to add visitors
- [x] Manifests are now to be written in toml
- [x] New manifest format includes in particular:
    - (Some old keys become irrelevant as they are superseded by new ones)
    - [x] The `upstream` section which is now mandatory
    - [x] An `integration` section containing:
        - [x] The `yunohost` minimum version required
        - [x] The list of supported architectures (or `all` if not relevant)
        - [x] A provided estimate of the disk and ram usage (for build and runtime)
        - [x] `sso` and `ldap` keys to declared if they are supported, or if it's irrelevant for this app
     - [x] A magic script will be provided to automatize (as much as possible) the conversion from v1 to v2
   - [x] A brand new `resource` section which will auto-provision/deprovision:
        - [x] system user
        - [x] app install directory
        - [x] data directory
        - [x] port
        - [x] apt dependencies
        - [x] permissions
        - [x] database (mysql/postgreslq)
        - [x] The resource system is to be declarative and should prevent a lot of boring backward-compat code in the upgrade scripts
- [x] (TODO) Create a new standard to define notifications in the `doc/` folder, to be a replacement for the `ynh_send_readme_to_admin`, c.f. draft in https://github.com/YunoHost/issues/issues/1957   
        - These should correspond to pre-install or post-install messages, or also pre-upgrade / post-upgrade, and possibly other scripts too

## PR Status

Work in progress, everything is broken /o/

## How to test

Zblerg


alexAubin:
Btw don't know where to put this, but created a magic script to convert manifests from v1 to v2 : https://gist.github.com/alexAubin/6a966d0e83717829d1ae12640f6067ec#file-convert_to_packaging_format_v2-py

tituspijean:
What about putting it in YunoHost/apps/tools?

alexAubin:
Yup indeed :+1: To be done once it's fully validated

tituspijean:
We also need to assess *where* would be stored the apps, if we consider establishing our own directory tree (like `/apps`).

alexAubin:
>We also need to assess where would be stored the apps, if we consider establishing our own directory tree (like /apps).

Eh yeah I was thinking about doing this in the v3 because there's already so much going on in v2 and that thing about moving app install folders has a lot of implications

PR #1301, opened 2021-08-27
Proof of concept for multiple-admins auth

alexAubin:
## The problem

Having a single is problematic for many reasons

Also we want to get rid of the weird "admin" special user because it creates confusion, etc

## Solution

Have an "admins" group containing multiple user

So far, this only demonstrates the ability to : 
- have multiple users being to authenticate on the admin/api
- accessing what's the current logged-in username (not super useful here, but useful for other projects such as the-future-portal/user-api-to-replace-part-of-ssowat)

Stuff not covered:
- ... everything else such as all the modifications in other pieces of code, and legacy migration

## PR Status

Tested and working, but much work still to be done

## How to test

- One has to first manually adds yunohost users to the `admins` group. From a yunohost shell: `ldap.update("cn=admins,ou=groups", {"memberUid": ["admin", "johndoe"]})`
- Then in the yunohost admin, login by putting in the password field : `johndoe:The_Actual_Password` (to be replaced later with two different input fields instead of this hack)
- Go to the user list to see the `print()` in `user_list()` demonstrating that we can access the currently logged in username

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -481,7 +481,7 @@
     "migrations_to_be_ran_manually":  "Migration {id} has to be run manually. Please go to Tools → Migrations on the webadmin page, or run `yunohost tools migrations run`.",
     "not_enough_disk_space": "Not enough free space on '{path}'",
     "invalid_number": "Must be a number",
-    "invalid_password": "Invalid password",
+    "invalid_credentials": "Invalid credentials",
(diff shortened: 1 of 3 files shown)


alexAubin:
Superseded by https://github.com/YunoHost/yunohost/pull/1408

PR #1304, opened 2021-08-31
Implement global settings for https redirect

alexAubin:
## The problem

Fix https://github.com/YunoHost/issues/issues/1024

- The fact that every web app has to add its own `if ($scheme = http) rewrite ^ https://$server_name$request_uri? permanent;` is madness...
- One may want to have an easy way to disable HTTPS redirection (for example in the context of yunohost behind a reverse proxy, such as our infrastructure)

## Solution

Have a global setting to control wether HTTPS redirect is enabled or not

(... and then remove the https redirect snippet in every nginx.conf in every app ...)

## PR Status

Tested and working on my side

## How to test

- `yunohost settings set security.nginx.redirect_to_https -v True`
- check that you're redirected to https, except for requests on .well-known stuff (assuming the target exists and the request is explicitly in http, not https)
- `yunohost settings set security.nginx.redirect_to_https -v False` removes the redirection (at least in the nginx conf, then there's the whole HSTS story etc)


alexAubin:
Hmpf it's in fact probably not that simple: 

- if an app installs on the domain's root, it will create a duplicate `location /` and nginx won't be happy (c.f. test not passing on the CI I believe)
- if an app installs on a subpath, nginx won't go inside the `location /` block and therefore won't redirect to https

I'm thinking that we could maybe remove the `include *.conf` if the https redirect is enabled ... ? I don't know if there are apps that need to expose some routes on http ... To cover this case, we could imagine apps adding an additional `*.conf.http` that would be included only in the http conf ?

Psycojoker:
Omg this is finally a reality, I can't believe it DDD:

PR #1317, opened 2021-09-08
[fix] sort_buffer_size too small to make nextcloud work

zamentur:
## The problem

Nextcloud can install with the previous settings, however it display an error page on some real servers.

## Solution

Increase to the config used in debian mariadb package
https://github.com/MariaDB/server/blob/10.3/debian/additions/my.cnf#L56

## PR Status
Ready tested on sans-nuage.fr

## How to test

...

### DIFF
--- a/data/templates/mysql/my.cnf
+++ b/data/templates/mysql/my.cnf
@@ -30,7 +30,7 @@ skip-external-locking
 key_buffer_size = 16K
 max_allowed_packet = 16M
 table_open_cache = 4
-sort_buffer_size = 256K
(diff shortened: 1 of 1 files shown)


ericgaspar (approved):

zamentur:
Needed for https://github.com/YunoHost-Apps/nextcloud_ynh/pull/435

ericgaspar:
Most of the values seem to have been pumped up quite a bit in the mariadb file...

kay0u (approved):

PR #1323, opened 2021-09-14
Fix action

kay0u:
## The problem

`YNH_APP_BASEDIR` is incorrect during the execution of the action
Actions are not executed in a tmp directory

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -1651,25 +1651,35 @@ def app_action_run(operation_logger, app, action, args=None):
     )
     env_dict["YNH_ACTION"] = action
 
-    _, path = tempfile.mkstemp()
+    tmp_workdir_for_app = _make_tmp_workdir_for_app(app=app)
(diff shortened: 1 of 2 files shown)


src/yunohost/app.py
```diff
@@ -1651,25 +1651,35 @@ def app_action_run(operation_logger, app, action, args=None):
...
+        retcode = hook_exec(
+            action_script,
+            env=env_dict,
+            chdir=cwd,
+            user=action_declaration.get("user", "root"),
+        )[0]
+    # Here again, calling hook_exec could fail miserably, or get
+    # manually interrupted (by mistake or because script was stuck)
+    # In that case we still want to proceed with the rest of the
+    # removal (permissions, /etc/yunohost/apps/{app} ...)
```
alexAubin, line 1675:
```suggestion
    # Calling hook_exec could fail miserably, or get
    # manually interrupted (by mistake or because script was stuck)
    # In that case we still want to delete the tmp work dir
```

PR #1328, opened 2021-09-17
add YNH_APP_BASEDIR in the env var

kay0u:
## Solution

Half tested on my side

## PR Status

Done, waiting for the ci

## How to test

Run all possibles actions for an app and see

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ -z "$YNH_APP_BASEDIR" ];
+if [ -z "${YNH_APP_BASEDIR:-}" ];
 then
     echo -e "YNH_APP_BASEDIR is not defined, many helpers use it, please define it!" >&2
     YNH_APP_BASEDIR=$(realpath ..)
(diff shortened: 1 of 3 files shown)


alexAubin (approved):
Super quick review, lgtm

data/helpers.d/utils
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-YNH_APP_BASEDIR=$(realpath $([[ "$(basename $0)" =~ ^backup|restore$ ]] && echo '../settings' || [[ -n "${YNH_ACTION:-}" ]] && echo '.' || echo '..' ))
+if [ -z "${YNH_APP_BASEDIR:-}" ];
+then
+    echo -e "YNH_APP_BASEDIR is not defined, many helpers use it, please define it!" >&2
+    YNH_APP_BASEDIR=$(realpath ..)
+fi
```
kay0u, line 7: After checking that we have put YNH_APP_BASEDIR everywhere we need to, I think we should change this part to :

```suggestion
YNH_APP_BASEDIR=${YNH_APP_BASEDIR:-$(realpath ..)}
```

PR #1338, opened 2021-09-21
Rework prompt() again

alexAubin:
## The problem

c.f. https://github.com/YunoHost/moulinette/pull/303

## Solution

c.f. https://github.com/YunoHost/moulinette/pull/303

## PR Status

...

## How to test

...

### DIFF
--- a/src/yunohost/utils/config.py
+++ b/src/yunohost/utils/config.py
@@ -31,6 +31,7 @@ from moulinette.interfaces.cli import colorize
 from moulinette import Moulinette, m18n
 from moulinette.utils.log import getActionLogger
 from moulinette.utils.filesystem import (
+    read_file,
     write_to_file,
     read_toml,
     read_yaml,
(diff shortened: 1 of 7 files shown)


src/yunohost/utils/config.py
```diff
@@ -558,18 +565,8 @@ def _raise_invalid_answer(self):
             choices=", ".join(self.choices),
         )
 
-    def _format_text_for_user_input_in_cli(self, column=False):
-        text_for_user_input_in_cli = _value_for_locale(self.ask)
-
-        if self.choices:
-            text_for_user_input_in_cli += " [{0}]".format(" | ".join(self.choices))
```
alexAubin, line 565: Originally did this because I thought the new autosuggest mecanism would be okay, but that's not great in terms of UX

We should probably re-add this, but I'm thinking also that we should have a better handling of the case where there's a shitload of cases (typically because many many domains, or many many users such as 100+)

Something like display only the X first entries :  `Pick an admin user [alice | bob | charlie | delphine | ... + N other choices]`

PR #1340, opened 2021-09-24
Argo Tunnels

martinkeat:
## The problem

Setting up DNS is a pain in the backside for amateur users

## Solution

Addition of cloudflared to Yuno so that users can easily create Argo Tunnels to their domain and therefore make like easier and their setup more secure in the process as well. 
...

## PR Status

...

## How to test

...


tituspijean:
PEBCAK 😇

PR #1335, opened 2021-09-20
Fix yunomdns

tituspijean:
## The problem

- [x] yunomdns fails if another `yunohost.local` domain is on the network 
- [x] yunomdns service is... silent.

## Solution

- [x] ~~Do not forcibly add `yunohost.local` in the service, the regen-conf takes care of that, Let's let the user remove its line in `/etc/yunohost/mdns.yml`~~
- [x] Use built-in function to append `-2` (and so on) if another service is already on the network
- [x] Introducing my new best friend, `Environment=PYTHONUNBUFFERED=1` 

## PR Status

Ongoing? Waiting for other users feedback.

## How to test

...

PR #1341, opened 2021-09-26
Fix yunomdns

Salamandar:
## The problem

* The tool was crashing when a bad ip was provided to Zeroconf (a global ipv6 for example)
* We were parsing the output of iproute, but there already exists some good python libraries that do the job.

## Solution

* Use `ifaddr`.
* Also, do some cleanup on the main function:
  * b_ips are not necessary (we can use the `parsed_addresses` argument of ServiceInfo
  * Add Python3 type hints

### DIFF
--- a/bin/yunomdns
+++ b/bin/yunomdns
@@ -4,125 +4,42 @@
 Pythonic declaration of mDNS .local domains for YunoHost
 """
 
-import subprocess
-import re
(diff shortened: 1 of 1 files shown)


tituspijean (approved):
🚀 
Let's continue on #1335

PR #1352, opened 2021-10-05
Update logrotate

ericgaspar:
## The problem
- Rights for log path are not define in logrotate helper

## Solution

- Add Rights


data/helpers.d/logrotate
```diff
@@ -96,6 +96,10 @@ $logfile {
 EOF
     mkdir --parents $(dirname "$logfile")	# Create the log directory, if not exist
     cat ${app}-logrotate | $customtee /etc/logrotate.d/$app > /dev/null	# Append this config to the existing config file, or replace the whole config file (depending on $customtee)
+
+    ynh_user_exists "$app" || chown $app:$app "$logfile"
```
kay0u, line 100:
```suggestion
    ynh_user_exists --username="$app" || chown $app:$app "$logfile"
```

PR #1356, opened 2021-10-06
Lint/autoformat helpers, hooks and debian scripts

alexAubin:
## The problem / solution

idk if that's so relevant but I was digging the possibility to lint / autoformat bash stuff and stumbled upon [shfmt](https://github.com/mvdan/sh) which does a nice job and sounds like "Black for bash". At least I'm hoping it should standardize indentation, remove trailing spaces, and other various stuff.

I'm a bit unpleased by using the syntax `if expr; then` (compared to `then` on a newline) but whatever I guess

## PR Status

Yolocommited

## How to test

¯\\\_(ツ)\_/¯ idk that shouldnt break anything

### DIFF
--- a/data/helpers.d/utils
+++ b/data/helpers.d/utils
@@ -699,6 +699,28 @@ properly with chmod/chown."
     echo $TMP_DIR
 }
 
+_acceptable_path_to_delete() {
+    local file=$1
+
+    local forbidden_paths=$(ls -d / /* /{var,home,usr}/* /etc/{default,sudoers.d,yunohost,cron*})
+
+    # Legacy : A couple apps still have data in /home/$app ...
+    if [[ -n "$app" ]]
+    then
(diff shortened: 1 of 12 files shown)


data/helpers.d/utils
```diff
@@ -743,24 +718,21 @@ ynh_secure_remove () {
...
-        || "$file" =~ ^/[[:alnum:]]+$ \
+    elif [[ "$forbidden_path" =~ "$file" ||
+
+        "$file" =~ ^/[[:alnum:]]+$ ||
+
+        "${file:${#file}-1}" = "/" ]]; then # Match all paths or subpaths in $forbidden_path
         # Match all first level paths from / (Like /var, /root, etc...)
-        || "${file:${#file}-1}" = "/" ]]
         # Match if the path finishes by /. Because it seems there is an empty variable
-    then
```
alexAubin, line 760: Uh that one is a bit odd

PR #1369, opened 2021-11-01
Making mariadb, php and metronome optional + better integration for postgresql + add sury by default

alexAubin:
## The problem

This is an attempt to address https://github.com/YunoHost/issues/issues/1578

## Solution

- Move mariadb / php-stuff / metronome to 'recommends' in `debian/control`, hence they'll be installed by default, but can be removed. So far this is the best design I found without falling into a legacy management hell
   - ideally, all apps using mysql database should now declare an explicit dependency to `mariadb-server` to cover the case where it could have been uninstalled ... Maybe a check to add in the package linter .... Postgresql is already covered (apps explicitly declare it). And PHP-fpm should automatically be handled (c.f. next items)
   - That said, apps should also ideally declare an explicit dependency to packages like `php7.3-ldap`, `php7.3-mysql`, `php7.3-intl`, `php7.3-curl` or `php7.3-gd` ... though sometime this ain't obvious to identify if the app needs those or not
- **Sury is now to be added to every setup** via the regen conf (... because meh it's fine, nowadays the pinning policy doesn't seem to lead to disasters anymore)
- `ynh_install_app_dependencies` will now keep a hard dependency to the `phpx.y-xxx` stuff even for the default version
- Service integration inside yunohost is magically handled by the core (e.g. php8.0-fpm will automatically be added in the list if php8.0-fpm is installed, mariadb/mysql will be hidden if mariadb-server ain't installed, ...)
- Tweak regen-conf hooks for mariadb and metronome so that they won't be doing anything if corresponding packages are not installed.
- Add regen-conf hook for postgresql that replaces `ynh_psql_test_if_first_run`. Also this regen conf is to run automagically inside `ynh_install_app_dependencies` upon installation, removing the need to call `ynh_psql_test_if_first_run`.

## PR Status

Yolocommited

There may be quite a lot of places that need fix / tweaks ... For example, we may want to avoid generating DNS records for metronome if it ain't installed ... But maybe not needed for that first iteration.

Also overall so far there's no forrmal way in the code to check if optional modules are installed apart from grepping `dpkg --list` ... Or no way for users to remove/reinstall packages apart from manually calling apt with the appropriate package names.

## How to test

Zblerg gotta try to install apps using mysql, php, or postgresql. Also try to uninstall metronome and check if things are going okay


data/helpers.d/apt
```diff
@@ -320,34 +299,21 @@ EOF
...
-ynh_add_sury() {
 
-    # Add an extra repository for those packages
-    ynh_install_extra_repo --repo="https://packages.sury.org/php/ $(ynh_get_debian_release) main" --key="https://packages.sury.org/php/apt.gpg" --name=extra_php_version --priority=600
+    # Trigger postgresql regenconf if we may have just installed postgresql
+    local psql_installed2="$(ynh_package_is_installed "postgresql-$PSQL_VERSION" && echo yes || echo no)"
+    if [[ "$psql_installed" != "$psql_installed2" ]]
+    then
+        yunohost tools regen-conf postgresql
+    fi
```
alexAubin, line 313: (We probably gotta do something similar in case mysql was just installed, probably some stuff to configure via the regenconf ? idk, to be checked)

alexAubin:
Zblerg yolomerging because it's hell to test on the CI otherwise >_>

PR #1375, opened 2021-11-07
[fix] Read DNS Blacklist answer and compare it against list of non-BL codes

rthouvenin:
## The problem

For some DNS Spam blacklists (e.g. hostkarma), an answer to the DNS query does not necessarily mean that the IP is blacklisted. The actual answer needs to be read to find out whether it's blacklist, whitelist, other...

## Solution

Add a property `non_blacklisted_return_code` to the blacklist servers that tells which answers should not be interpreted as "back-listed", [as suggested by zamentur](https://github.com/YunoHost/issues/issues/1873#issuecomment-917407489).

## PR Status

I only added the return codes for hostkarma, because it's the server that was affecting my instance. Any other that I should add?

## How to test

c.f. comment linked above

### DIFF
--- a/data/hooks/diagnosis/24-mail.py
+++ b/data/hooks/diagnosis/24-mail.py
@@ -209,8 +209,8 @@ class MailDiagnoser(Diagnoser):
                 query = subdomain + "." + blacklist["dns_server"]
 
                 # Do the DNS Query
-                status, _ = dig(query, "A")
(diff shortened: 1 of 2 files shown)


alexAubin:
Not tested but yolomerging because sounds legit and supposedly matches the design suggestd by @zamentur 

Thanks a lot :+1: :tada:

PR #1376, opened 2021-11-08
Fix dump_script_log_extract_for_debugging 

Salamandar:
## The problem

When a failure occurs during the upgrade, the first ynh_exit is catched. That will be the one during the backup. So the logs abstract does not contain the code execution of the failure.

## Solution
Find the last ynh_exit in the logs by using a reversed iterator of the lines.

## PR Status

UNTESTED only for review.

## How to test

Try to upgrade an app whose upgrade script fails (after backup).


Salamandar:
Well that was efficient.

PR #1384, opened 2021-11-26
[enh] Tweak umask for SFTP

zamentur:
## The problem

Via sftp when you load a file or a directory, the content uploaded has no group permission even if the parent dir has.
It's a bit shame if you want to share an app dir to a user like in this documentation PR https://github.com/YunoHost/doc/pull/1836/files

## Solution

Set an appropriate umask.

Alternatively, we could define umask by user through gcos in ldap... But it's more difficult.

## PR Status

Currently in testing on a true server with a true person. I wait a bit if this person get an issue.

## How to test

...

### DIFF
--- a/conf/ssh/sshd_config
+++ b/conf/ssh/sshd_config
@@ -84,7 +84,7 @@ Subsystem sftp internal-sftp
 
 # Apply following instructions to user with sftp perm only
 Match Group sftp.main,!ssh.main
-    ForceCommand internal-sftp
+    ForceCommand internal-sftp -u 0002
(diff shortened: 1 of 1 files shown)


zamentur:
It's not a complete fix, but it could help a lot of person and be part of the solution (with this method: https://forum.yunohost.org/t/comment-editer-un-fichier-directement-sur-le-serveur/18323 ). 

This PR allows to conserve the parent group permission. Other solutions to test: https://unix.stackexchange.com/questions/7559/set-umask-to-0002-for-sftp-and-scp

PR #1387, opened 2021-12-04
WIP : New portal API to partially replace SSOwat

alexAubin:
## The problem

- SSOwat should be partially reworked to get rid of some lua madness (c.f. cookie management / authentication, ldap interfacing, ..) and allow for more flexibility regarding the portal front end, etc.
- Projects regarding password reset and user self-registration imply new authentication method, and ideally these operations should not rely on an API running as root

## Solution / specs

- Create a new "user" or "portal API" meant to replace some routes of SSOwat such as login/logout, updating user info, etc.
- The API is also meant to handle password reset and user registration
- It' meant to run as an unprivileged user ... though because we need to update user's password or create brand new users, some sort of mecanism shall be designed to interface with a privileged program to order the password reset or user creation.
- Ideally SSOwat should be able to validate the authenticity of the new cookie even though it's delivered from a different process

## PR Status

Very much work in progress

## How to test

So far: 

```bash
# Check that the new portal API is up
systemctl status yunohost-portal-api
```

In your browser, open `https://thelocaldomain/yunohost/sso`

You should get the new raw/crude index.html displayed, which contain super minimal JS to test the API

Enter some user credentials, and you should see the corresponding user info displayed below the Login button


src/portal.py
```diff
@@ -0,0 +1,89 @@
...
+    Keyword argument:
+        username -- Username to get informations
+
+    """
+
+    username = None  # FIXME : this info should come from the authentication layer
+
+    from yunohost.utils.ldap import _get_ldap_interface
+
+    ldap = _get_ldap_interface()
```
alexAubin, line 43: Ah that bit will also assume it can authenticate as root on slapd

conf/nginx/yunohost_api.conf.inc
```diff
@@ -23,3 +23,23 @@ location = /yunohost/api/error/502 {
     add_header Content-Type text/plain;
     internal;
 }
+
+location /yunohost/portalapi/ {
+    proxy_read_timeout 3600s;
```
selfhoster1312, line 28: That's really high timeout for this central part of the system. Of course it should never timeout but just in case receiving error after 5s is better than waiting 1h

PR #1391, opened 2021-12-13
Don't resolve symlinks while transforming backup path to absolute.

Salamandar:
This fixes backup/restore scripts. 

If you're backuping /lib/systemd/system/jellyfin.service but /lib is a symlink to /usr/lib, the path stored in the backup will be /usr/lib/systemd/system/jellyfin.service.
Thus restoring will not find the file and will fail.

THIS MAY BREAK SOME RESTORE SCRIPTS that would use this bug as a feature.

### DIFF
--- a/data/helpers.d/backup
+++ b/data/helpers.d/backup
@@ -113,7 +113,7 @@ ynh_backup() {
 
     # Transform the source path as an absolute path
     # If it's a dir remove the ending /
-    src_path=$(realpath "$src_path")
+    src_path=$(realpath --no-symlink "$src_path")
 
     # If there is no destination path, initialize it with the source path
     # relative to "/".


zamentur (changes requested):
I think it could be dangereous to merge that.
We should reread PR in which we added the realpath.

What about user change like describe in our doc: https://yunohost.org/en/moving_app_folder

Salamandar:
I agree this is a dangerous change. 
I also think the current behaviour is buggy, for cases like my example (jellyfin_ynh is broken because of that.).

IMHO, though, IN PRACTICE (take it with salt) that SHOULDN'T break anything. Let me explain.
* At backup time, we have a path with symlinks in it. Ok, so no issue as we can read / copy files across symlinks
* In the backup archive, we have stored the path contained in the backup script, this path contained symlinks in the real filesystem.
* At restore time, the restore script has ALSO the path with symlinks, and it's the same path that is in the backup files.
* Files are restored, and again, it's done via symlinks if there are, and without symlinks if there aren't.
So nothing should go wrong.

In case the backed up file is ITSELF a symlink (and not one of its parent dirs), we should copy the file instead of the symlink though. Maybe that's the reason of the realpath call.

Also I'm thinking of another (improbable) possible bug in the current code: If a path contained symlinks at backup time and no symlinks at restore time, the restoration is broken !
For example, if `/lib/systemd/` is a symlink to `/my_systemd_stuff` at backup time, the archive will contain `/my_systemd_stuff`. If the symlink is deleted between backup and restore, then `/lib/systemd` won't contain the restored files.

PR #1392, opened 2021-12-17
Message stays compliant in each locale with --apps deprecation

moutonjr:
## The problem

image

## Solution

Just remove the --apps from all locales :)

## How to test

`# yunohst tools upgrade`


zamentur:
Thanks for your contribution
Translation should be made on https://translate.yunohost.org/

alexAubin (approved):
@zamentur : indeed but for general maintenance like this one it's okay

It used to create annoying git conflict but i believe nowadays it works okay, i regularly cleanup stuff in other locale files

PR #1395, opened 2021-12-24
Pyupgrade

Psycojoker:
## The problem

Old style python code.

## Solution

Run https://github.com/asottile/pyupgrade and removed old style classes.

## PR Status

If CI is happy it should be good.

## How to test

...

### DIFF
--- a/data/hooks/diagnosis/10-ip.py
+++ b/data/hooks/diagnosis/10-ip.py
@@ -167,10 +167,7 @@ class IPDiagnoser(Diagnoser):
 
         assert (
             resolvers != []
-        ), "Uhoh, need at least one IPv{} DNS resolver in {} ...".format(
-            protocol,
-            resolver_file,
-        )
+        ), f"Uhoh, need at least one IPv{protocol} DNS resolver in {resolver_file} ..."
 
         # So let's try to ping the first 4~5 resolvers (shuffled)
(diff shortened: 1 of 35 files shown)


kay0u (approved):

data/hooks/diagnosis/10-ip.py
```diff
@@ -220,9 +217,9 @@ def get_public_ip(self, protocol=4):
         try:
             return download_text(url, timeout=30).strip()
         except Exception as e:
-            self.logger_debug(
-                "Could not get public IPv%s : %s" % (str(protocol), str(e))
-            )
+            protocol = str(protocol)
+            e = str(e)
+            self.logger_debug(f"Could not get public IPv{protocol} : {e}")
```
kay0u, line 222: IIRC you can do something like that:

```suggestion
            self.logger_debug(f"Could not get public IPv{str(protocol)} : {str(e)}")
```

alexAubin: Ah yes, in fact just using `{protocol}` probably cast it as a string, but i didn't want to introduce any regression and was too lazy to confirm it so idk wat do é_è

kay0u: it was more a general remark, not specific to this line :)

PR #1400, opened 2021-12-29
Lgtm warnings

kay0u:
## The problem

https://lgtm.com/projects/g/YunoHost/yunohost/alerts/?mode=list

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/src/yunohost/regenconf.py
+++ b/src/yunohost/regenconf.py
@@ -48,7 +48,7 @@ logger = log.getActionLogger("yunohost.regenconf")
 @is_unit_operation([("names", "configuration")])
 def regen_conf(
     operation_logger,
-    names=[],
+    names=None,
     with_diff=False,
     force=False,
     dry_run=False,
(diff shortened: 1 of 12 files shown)


src/yunohost/tools.py
```diff
@@ -1134,7 +1130,7 @@ def __init__(self, id_):
     def description(self):
         return m18n.n("migration_description_%s" % self.id)
 
-    def ldap_migration(run):
+    def ldap_migration(self, run):
```
kay0u, line 1133: I'm really not sure about this one

src/yunohost/regenconf.py
```diff
@@ -48,7 +48,7 @@
 @is_unit_operation([("names", "configuration")])
 def regen_conf(
     operation_logger,
-    names=[],
+    names=None,
```
kay0u, line 51: More information about default value mutable:
https://stackoverflow.com/a/41686973

alexAubin: >Default argument values are evaluated only once at function definition time, which means that modifying the default value of the argument will affect all subsequent calls of the function.

My god that's terrifying o.o

alexAubin (approved):

PR #1406, opened 2022-01-07
Manage default application with DomainConfigPanel

Tagadda:
## The problem

`yunohost app makedefault my_app` is great!

Fix https://github.com/YunoHost/issues/issues/40 

## Solution

`yunohost domain config set domain.tld feature.app.default_app -v myapp` or using the webadmin!

Then it's generated in `app_ssowatconf`, not in `conf.json.persistent` anymore.

The redirection is updated when the app change url or is removed. :sparkles: 

I made `makedefault` use the config system too so it's not broken o/

## PR Status

Tested locally and ready for reviews. :confetti_ball: 

I'm not sure about the webadmin, I don't know what I'm doing with Vue but it-works! :tm: 

## How to test

Using `ynh-dev`, pull this PR and the [yunohost-admin PR](https://github.com/YunoHost/yunohost-admin/pull/422).

`cp /ynh-dev/yunohost/data/other/config_domain.toml /usr/share/yunohost/other/`


src/yunohost/app.py
```diff
@@ -1419,6 +1396,15 @@ def app_ssowatconf():
         redirected_urls.update(app_settings.get("redirected_urls", {}))
         redirected_regex.update(app_settings.get("redirected_regex", {}))
 
+    for domain in domains:
+        default_app = domain_config_get(domain, "feature.app.default_app")
+        if (default_app != "_none"):
+            app_settings = _get_app_settings(default_app)
```
alexAubin, line 1402: There are some boring edge cases to take into account, such as if restoring the domain settings *without* restoring the corresponding default app .. in which case the default app would not be installed

```suggestion
        if default_app != "_none" and _is_installed(default_app):
            app_settings = _get_app_settings(default_app)
```

src/yunohost/app.py
```diff
@@ -1153,36 +1157,9 @@ def app_makedefault(operation_logger, app, domain=None):
...
     operation_logger.start()
 
-    # TODO / FIXME : current trick is to add this to conf.json.persisten
-    # This is really not robust and should be improved
-    # e.g. have a flag in /etc/yunohost/apps/$app/ to say that this is the
-    # default app or idk...
-    if not os.path.exists("/etc/ssowat/conf.json.persistent"):
-        ssowat_conf = {}
-    else:
-        ssowat_conf = read_json("/etc/ssowat/conf.json.persistent")
```
alexAubin, line 1173: I'm wondering what to do with the existing setups ... or maybe we just warn people about it in the next release note

Tagadda: I was thinking about a migration but I don't know if it's right (you have to run it manually anyway...) Is there some mecanism to run some code after a yunohost package upgrade?

Tagadda: Oh didn't know about the legacy utils yet!
Since we have to get rid of the old redirect rules when ssowatconf is run(?), we might as well try to get the redirection right
I'll give it a go

src/yunohost/utils/config.py
```diff
@@ -1123,6 +1123,19 @@ def normalize(value, option={}):
...
+    argument_type = "app"
+
+    def __init__(
+        self, question, context: Mapping[str, Any] = {}, hooks: Dict[str, Callable] = {}
+    ):
+        from yunohost.app import app_list
+
+        super().__init__(question, context, hooks)
+
+        apps = [app["id"] for app in app_list(full=True)["apps"] if app["is_webapp"]]
```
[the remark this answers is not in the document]
Tagadda: I had initially think about a "filters" in the `config_domain.toml` instead of "hard-coding" the condition but I like to build lists this way :(((

PR #1408, opened 2022-01-11
Drop the 'admin' user, have 'admins' be a group of Yunohost users instead

alexAubin:
## The problem

For many reasons, the fact that 'admin' is a special user is annoying

Fix https://github.com/YunoHost/issues/issues/1110

## Solution

Create a special 'admins' group composed of yunohost users

This impacts the following things:
- authentication on the API (and therefore login on the webadmin) requires to specify the username
- the postinstall now asks for the username + first/last names of the first user
- until now, the root password was ~synced with the admin password ... it's not clear what to do with this. It'll be initialized with the password provided during postinstall, and may be editable from the webadmin
- the root@, admin@, abuse@, ... aliases are to be binded to the admins group instead of the very first user

## PR Status

Somewhat working

Still need to do a clean test of everything

Also need to implement the migration which is a tricky one : 
- technically : ldap migration, have to delete the admin user + "guess" at least one user to include in the admins group ... we can use the fact that it should have the root@ alias but it may not because that user got deleted ...)
- UX-wise : having to tell people that the admin user doesnt exists, and they should specify their username upon login on the webadmin...

Also gotta polish a bit the CLI UX for postinstall questions

## How to test

...


src/migrations/0024_new_admins_group.py
```diff
@@ -0,0 +1,86 @@
...
+            "cn=admins,ou=groups",
+            {
+                "cn": ["admins"],
+                "objectClass": ["top", "posixGroup", "groupOfNamesYnh", "mailGroup"],
+                "gidNumber": [4001],
+                "mail": ["root", "admin", "admins", "webmaster", "postmaster", "abuse"],
+            }
+        )
+
+        user_group_update(groupname="admins", add=new_admin_user, sync_perm=True)
```
alexAubin, line 83: FIXME: we should also remove the reserved aliases for root@ etc from that user

src/user.py
```diff
@@ -421,16 +421,16 @@ def user_update(
             change_password = Moulinette.prompt(
                 m18n.n("ask_password"), is_password=True, confirm=True
             )
+
         # Ensure compatibility and sufficiently complex password
         assert_password_is_compatible(change_password)
-        assert_password_is_strong_enough("user", change_password)
+        assert_password_is_strong_enough("user", change_password)    # FIXME FIXME FIXME : gotta use admin profile if user is admin
```
alexAubin, line 427: (to be fixed)

PR #1413, opened 2022-01-19
[enh] Add SNI support for postfix and dovecot

Tagadda:
## The problem

In a multi-domain context, postfix and dovecot always serve the main domain SSL certificates. This cause issue when users try to connect with a secondary domain.
- Fix https://github.com/YunoHost/issues/issues/1301

## Solution

- Enable SNI support in postfix and dovecot.
Configuration from https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#with-client-tls-sni-server-name-indication-support and http://www.postfix.org/postconf.5.html#tls_server_sni_maps

## PR Status

- Works great with Thunderbird o/
- Maybe need testing on other mail clients but idk

## How to test

- Pull this PR with `ynh-dev`
- Add 2 or more domains on YunoHost
- Create a user with a main on a secondary domain
- Try to log from a mail client
- Try to send a mail

### DIFF
--- a/conf/dovecot/dovecot.conf
+++ b/conf/dovecot/dovecot.conf
@@ -21,6 +21,11 @@ ssl = required
 
 ssl_cert = </etc/yunohost/certs/{{ main_domain }}/crt.pem
 ssl_key = </etc/yunohost/certs/{{ main_domain }}/key.pem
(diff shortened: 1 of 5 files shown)


alexAubin:
Not tested but sound great and simple enough, cheers !

Yolomergin'

PR #1425, opened 2022-01-26
Enhance yunomdns service startup

tituspijean:
## The problem

I had the service start before it was able to list any interface.

## Solution

Make it rely on network-online.target instead of network.target

## PR Status

Done, ready to test, tested locally.

## How to test

...


tituspijean:
Suggestion from @alexAubin: make the service into a template and have one instance per interface.

PR #1427, opened 2022-01-28
[fix] Compute choices for the yunohost admin when installing an app

Tagadda:
## The problem

Domains and users appears empty when we install an app from the yunohost admin
This is a side effect of https://github.com/YunoHost/yunohost-admin/pull/426
Now, the admin doesn't compute choices and need them from the API

## Solution

Give `choices` to the yunohost admin through the manifest

## PR Status

Working on my side o/

## How to test

...


alexAubin:
Cheers :partying_face:

PR #1429, opened 2022-01-30
configpanel: filter in AppQuestion

Tagadda:
`filter` is now a simple expression, this allow filtering apps by id.
In the case of https://github.com/YunoHost-Apps/mautrix_whatsapp_ynh, the manifest.json would look like this
```json
{
"arguments": {
		"install": [
			{
				"name": "synapsenumber",
				"type": "app",
				"ask": {
					"en": "Choose the local Synapse instance to communicate with mautrix_whatsapp.",
					"fr": "Choisissez l'instance Synapse qui doit communiquer avec mautrix_whatsapp."
				},
				"filter": "match(id, 'synapse')"
			},
		]
}
```
and render in the webadmin like this:
image

### DIFF
--- a/share/config_domain.toml
+++ b/share/config_domain.toml
@@ -14,7 +14,7 @@ i18n = "domain_config"
     [feature.app]
         [feature.app.default_app]
         type = "app"
-        filters = ["is_webapp"]
+        filter = "is_webapp"
         default = "_none"
     
     [feature.mail]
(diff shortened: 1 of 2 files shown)


alexAubin:
Oh nice :o Didnt check exactly how those simple_js_expression work, but the code looks legit :+1:

PR #1432, opened 2022-02-01
Removes nginx Content-Security-Policy overrides

DougReeder:
## The problem
Different web apps use different features of the browser environment, so each web app must have its own Content Security Policy.  These directives override the CSP set be the app server itself - they will break some web apps and open security holes in others.  https://github.com/openresty/headers-more-nginx-module#more_set_headers

Setting default_src or any of the other *_src directives to data: or 'unsafe-inline' disables key XSS protection. It's hardly a security policy at all.  Content-Security-Policy-Report-Only is useless without a report-uri (and probably a report-to) directive, pointing to a host that accepts violation reports for your domain: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

...

## Solution

Remove the nginx directives, and let each app server set its own CSP.

...

## PR Status

...

## How to test

Install Armadietto, and with the current nginx config, observe that 
1. The Content-Security-Policy header is set by the nginx config
2. Some browsers complain about the Content-Security-Policy-Report-Only header without a report-uri directive.

Remote the nginx directives, restart and observe
1. Armadietto sets a strict Content-Security-Policy
2. No browser complaints about a Content-Security-Policy-Report-Only header without a report-uri directive.

...

### DIFF
--- a/conf/nginx/security.conf.inc
+++ b/conf/nginx/security.conf.inc
@@ -45,7 +45,6 @@ header_filter_by_lua_block {
     ngx.header["Content-Security-Policy"] = "upgrade-insecure-requests"
   end
 }
-more_set_headers "Content-Security-Policy-Report-Only : default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'";
 {% endif %}
 
 # Disable the disaster privacy thing that is FLoC
--- a/conf/nginx/security.conf.inc
+++ b/conf/nginx/security.conf.inc
@@ -31,6 +31,23 @@ more_set_headers "X-Download-Options : noopen";
 more_set_headers "X-Permitted-Cross-Domain-Policies : none";
 more_set_headers "X-Frame-Options : SAMEORIGIN";
 
+# Here we add a generic CSP headers only if the app doesn't define it.
+# NB: if the <meta tag is used the most restrictive rules will be used
+{% if experimental == "True" %}
+header_filter_by_lua_block {
+  if ngx.header["Content-Security-Policy"] == nil or ngx.header["Content-Security-Policy"] == "" then
+    ngx.header["Content-Security-Policy"] = "upgrade-insecure-requests; default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'"
+  end
+}
+{% else %}
+header_filter_by_lua_block {
+  if ngx.header["Content-Security-Policy"] == nil or ngx.header["Content-Security-Policy"] == "" then
+    ngx.header["Content-Security-Policy"] = "upgrade-insecure-requests"
+  end
+}
+more_set_headers "Content-Security-Policy-Report-Only : default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'";
+{% endif %}
+
 # Disable the disaster privacy thing that is FLoC
 {% if experimental == "True" %}
 more_set_headers "Permissions-Policy : fullscreen=(), geolocation=(), payment=(), accelerometer=(), battery=(), magnetometer=(), usb=(), interest-cohort=()";
--- a/conf/nginx/security.conf.inc
+++ b/conf/nginx/security.conf.inc
@@ -25,12 +25,6 @@ ssl_dhparam /usr/share/yunohost/ffdhe2048.pem;
 # Follows the Web Security Directives from the Mozilla Dev Lab and the Mozilla Obervatory + Partners
 # https://wiki.mozilla.org/Security/Guidelines/Web_Security
 # https://observatory.mozilla.org/
-{% if experimental == "True" %}
-more_set_headers "Content-Security-Policy : upgrade-insecure-requests; default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'";
-{% else %}
-more_set_headers "Content-Security-Policy : upgrade-insecure-requests";
-more_set_headers "Content-Security-Policy-Report-Only : default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'";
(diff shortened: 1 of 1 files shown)


DougReeder:
I don't want to seem harsh - it's good to look at security headers.  The other security headers are a good idea - few modern web apps need the functionality they prohibit.

alexAubin:
Ah yes it's good that you're bringing this up ... I always wondered how this interacted with the CSP provided by the apps

With that said, isn't it good to have a "default" CSP policy for apps that provide none ?

I would suggest an alternative battle plan, which makes seems more complicated though ... :
- (lets remove the Report-Only header because it's pretty useless indeed)
- we could have the app CI test if each app returns CSP headers
    - if app doesnt handle any CSP itself, advise packages to add a keyword such as `__DEFAULT_CSP_POLICY__` in the app nginx conf, which the core would replace with the default CSP policy
    - for other apps: no action required
- once a decent number of apps implements the previous stuff, remove the CSP from the core ?

But it also depends on what fraction of apps currently implement a CSP or not ...? Do you have any insight on this ? Do apps typically implement a CSP nowadays ? Or is it only "high profile" apps like Nextcloud, Mastodon, etc that would do so and the majority doesnt ?

conf/nginx/security.conf.inc
```diff
@@ -25,18 +25,29 @@ ssl_dhparam /usr/share/yunohost/ffdhe2048.pem;
...
 more_set_headers "X-XSS-Protection : 1; mode=block";
 more_set_headers "X-Download-Options : noopen";
 more_set_headers "X-Permitted-Cross-Domain-Policies : none";
 more_set_headers "X-Frame-Options : SAMEORIGIN";
 
+# Here we add a generic CSP headers only if the app doesn't define it.
+# NB: if the <meta tag is used the most restrictive rules will be used
+{% if experimental == "True" %}
+header_filter_by_lua_block {
+  if ngx.header["Content-Security-Policy"] == nil or ngx.header["Content-Security-Policy"] == "" then
```
zamentur, line 38: I am not totally sure about `ngx.header["Content-Security-Policy"] == ""`

zamentur: I have tested it adds the field if the header is not set by the app, but i have not tested it doesn't if the header (from reverse proxy) exists...

zamentur: I have now test in the 2 use case. For me this patch is ok

conf/nginx/security.conf.inc
```diff
@@ -25,18 +25,29 @@ ssl_dhparam /usr/share/yunohost/ffdhe2048.pem;
...
+    ngx.header["Content-Security-Policy"] = "upgrade-insecure-requests; default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'"
+  end
+}
+{% else %}
+header_filter_by_lua_block {
+  if ngx.header["Content-Security-Policy"] == nil or ngx.header["Content-Security-Policy"] == "" then
+    ngx.header["Content-Security-Policy"] = "upgrade-insecure-requests"
+  end
+}
+more_set_headers "Content-Security-Policy-Report-Only : default-src https: data: blob: ; object-src https: data: 'unsafe-inline'; style-src https: data: 'unsafe-inline' ; script-src https: data: 'unsafe-inline' 'unsafe-eval'";
```
DougReeder, line 48: Content-Security-Policy-Report-Only is useless without a report-uri (and probably a report-to) directive, pointing to a host that accepts violation reports for your domain: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

zamentur: Indeed feel free to remove it

conf/nginx/security.conf.inc
alexAubin: Imho this is tweak is still debatable because, as far as I understand, apps that *do* implement CSP may provide it via an HTML head tags, not via a CSP HTTP header (though i don't have any quantitative data on what's the most widespread practice). So this lua tweak still overrides legit CSP headers and therefore reduce the security in that case ...

DougReeder: That's a good point - headers set by nGinX would also override http-equiv directives in the HTML.  And no security directives should be set by nGinX if any of the security directives are set by the app, either as headers or http-equiv directives.

klisza1993:
it would be great to have an option to apply Security Headers by custom menu during app deployment, some will be always present like xss, iframe, permissions etc. but some should be always disabled like, x-powered-by oraz "server" - those two are information leaking.
additional to that is, once you modify ngnx conf file, then you cant apply "experimental security" - error says that "faile has been modified"

alexAubin:
Some development regarding this:

yesterday I realized with the help of a friend that multiple CSP headers may be returned by a web server, each one containing a CSP snippet. However, in the context of nginx, this relates to [the infamous `add_header` vs `more_set_headers` thing](https://www.peterbe.com/plog/be-very-careful-with-your-add_header-in-nginx)

Doing some tests to reconfirm the behavior:


### 1. The original "add_header" issue

```
add_header Foo "foo";
add_header Bar "bar";

location /bar {
     add_header Baz "baz";
}
```

-> results in only Baz being sent

### 2. Using more_set_headers in the hope to send multiple CSP headers

```
more_set_headers "Content-Security-Policy: foo";

location /bar {
     more_set_headers "Content-Security-Policy: bar";
}
```

-> results in only the "bar" version being sent

### 3. Using `more_set_headers` in global context, `add_header` in location

Turns out that you can write:

```
more_set_headers "Content-Security-Policy: foo";

location /bar {
     add_header Content-Security-Policy "bar";
     add_header Content-Security-Policy "baz";
}
```

and end up with the three versions (foo, bar, baz) of the header being sent

### Conditional behavior depending if the app does provide a CSP header

It's not clear to me if the previous stuff really solves the issue of "what about apps that do provide their own headers"

However I'm also learning about way in nginx to refer to headers from the reverse-proxied app, cf https://stackoverflow.com/questions/12431496/nginx-read-custom-header-from-upstream-server

In particular we probably want to check `$upstream_http_content_security_policy`, and maybe we can use it in an `if` block

alexAubin:
Soooooo apparently a `-a` option for `more_set_headers` landed recently and does exactly what we want in terms of "not overriding what the proxy would set but only append an additional header with the same name ...

https://github.com/openresty/headers-more-nginx-module/commit/db45361256a665e4e076465249628bb4631c3354

but it's only available in version 0.35~0.36 of the module, and bullseye only ships 0.33 ...

alexAubin:
To be continued in https://github.com/YunoHost/issues/issues/2301

PR #1435, opened 2022-02-06
Update n to 8.0.2

ericgaspar:
- Upgrade n to 8.0.2


tituspijean:
Slightly off topic: should we establish an auto-updater for that? 😇

alexAubin:
@tituspijean : yes indeed, feel free to propose something :+1:

PR #1436, opened 2022-02-07
Config panel actions + applied to domain cert management

alexAubin:
## The problem

- We need actions in the config panel for many reasons

## Solution

- Add an action mechanism @_@

## PR Status

- This is mainly a POC for now
- The idea is to have a new component type `button`
- Any section containing a `button` becomes an "action section" which is not displayed by default during `config get`
- `yunohost app config list` lists all the `buttons` ids

## How to test

Example dummy toml snippet in an app :

```toml
[actions]
name = "Actions example"

    [actions.foo]
    name = "Trigger foo action"

        [actions.foo.bar]
        ask = "I can haz bar ?"
        type = "number"
	min = 1
	max = 100
        default = 5

        [actions.foo.force]
        ask = "Force it ?"
        type = "boolean"
        default = false
	visible = "bar > 10"

        [actions.foo.foo]
        ask = "Foo!"
        type = "button"
        icon = "rocket"
        style = "success"
        help = "This should trigger function run__foo"
        enabled = "bar <= 10 or force"  # This doesnt do anything for now but could be useful for condition button (enabled/disabled)
        visible = "bar <= 10 or force"   # This doesnt do anything for now but could be useful for condition button (visible/invisible)

        [actions.foo.foo2]
        ask = "Foo2?"
        type = "button"
        icon = "rocket"
        style = "warning"
        help = "This should trigger function run__foo2"
        enabled = "bar > 50"   # This doesnt do anything for now but could be useful for condition button (enabled/disabled)
        visible = "bar > 50"    # This doesnt do anything for now but could be useful for condition button (visible/invisible)
```

and corresponding controller in `config` script:

```bash
run__foo() {
    ynh_print_info "I am running the action foo"
    touch /tmp/foo_$(date -Iseconds)
}
```

Then run `yunohost app action list config_app` and `yunohost app action run actions.foo.foo` (yeah the app is the full triplet for non-trivial reasons gni)


share/config_domain.toml
```diff
@@ -58,3 +58,60 @@ i18n = "domain_config"
...
+
+    [cert.status]
+
+        [cert.status.cert_summary]
+        type = "alert"
+        # Automatically filled by DomainConfigPanel
+
+        [cert.status.cert_validity]
+        type = "number"
+        readonly = true
```
Axolotle, line 73:
```suggestion
        readonly = true
        visible = "false"
```

share/config_domain.toml
```diff
@@ -58,3 +58,60 @@ i18n = "domain_config"
...
+        # Automatically filled by DomainConfigPanel
+
+        [cert.cert.acme_eligible]
+        type = "boolean"
+        readonly = true
+        visible = "false"
+        # Automatically filled by DomainConfigPanel
+
+        [cert.cert.acme_eligible_explain]
+        type = "alert"
```
Axolotle, line 91:
```suggestion
        type = "alert"
        style = "warning"
```

src/app.py
```diff
@@ -1609,8 +1539,10 @@ def _call_config_script(self, action, env=None):
         if ret != 0:
             if action == "show":
                 raise YunohostError("app_config_unable_to_read")
-            else:
+            elif action == "show":
                 raise YunohostError("app_config_unable_to_apply")
```
alexAubin, line 1543: uuuuuh wut ?

src/utils/config.py
```diff
@@ -1397,6 +1545,19 @@ def ask_questions_and_parse_answers(
         question_class = ARGUMENTS_TYPE_PARSERS[raw_question.get("type", "string")]
         raw_question["value"] = answers.get(raw_question["name"])
         question = question_class(raw_question, context=context, hooks=hooks)
+        if question.type == "button":
+            if (
+                not question.enabled
```
alexAubin, line 1550:
```suggestion
                question.enabled is None
```

PR #1439, opened 2022-02-09
Only include multimedia system folder in system backup/restore

autra:
## The problem

Here is a proposed fix for https://github.com/YunoHost/issues/issues/1981

### DIFF
--- a/hooks/backup/18-data_multimedia
+++ b/hooks/backup/18-data_multimedia
@@ -14,4 +14,7 @@ if [ -e "/home/yunohost.multimedia/.nobackup" ]; then
 fi
 
 # Backup multimedia directory
-ynh_backup --src_path="/home/yunohost.multimedia" --dest_path="${backup_dir}" --is_big --not_mandatory
+ynh_backup --src_path="/home/yunohost.multimedia/Music" --dest_path="${backup_dir}/Music" --is_big --not_mandatory
+ynh_backup --src_path="/home/yunohost.multimedia/Picture" --dest_path="${backup_dir}/Picture" --is_big --not_mandatory
+ynh_backup --src_path="/home/yunohost.multimedia/Video" --dest_path="${backup_dir}/Video" --is_big --not_mandatory
+ynh_backup --src_path="/home/yunohost.multimedia/eBook" --dest_path="${backup_dir}/eBook" --is_big --not_mandatory
--- a/hooks/restore/18-data_multimedia
+++ b/hooks/restore/18-data_multimedia
@@ -8,4 +8,7 @@ source /usr/share/yunohost/helpers
 
 backup_dir="data/multimedia"
 
-ynh_restore_file --origin_path="${backup_dir}" --dest_path="/home/yunohost.multimedia" --not_mandatory
+ynh_restore_file --origin_path="${backup_dir}/Music" --dest_path="/home/yunohost.multimedia/Music" --not_mandatory
+ynh_restore_file --origin_path="${backup_dir}/Picture" --dest_path="/home/yunohost.multimedia/Picture" --not_mandatory
+ynh_restore_file --origin_path="${backup_dir}/Video" --dest_path="/home/yunohost.multimedia/Video" --not_mandatory
+ynh_restore_file --origin_path="${backup_dir}/eBook" --dest_path="/home/yunohost.multimedia/eBook" --not_mandatory


alexAubin (commented):
(Gotta discuss this with @zamentur but apriori LGTM)

alexAubin:
Bump @zamentur ... cf also the related issue to fully understand where this is coming from

alexAubin:
Closing because after discussing this today we concluded that this wasnt the right way to go ... The whole multimedia setup stuff is clearly funky and to be reworked/clarified I think (eg it's not clear what are the practical use cases)

In the meantime I pushed https://github.com/YunoHost/yunohost/commit/94f21ea20e47cae27fc71cf3375b5912453549aa which should fix boring crashes because of broken symlinks

autra:
@alexAubin 

> (eg it's not clear what are the practical use cases)

I can answer that with my only practical use case, which is a killer feature in my opinion: downloading something with the transmission application, and have it immediately available in the minidlna server. For ex, after download is complete, a video will be immediately playable with vlc from my personal computer. That is the only use case I really care about.

> The whole multimedia setup stuff is clearly funky and to be reworked/clarified I think 

I'd probably agree :-) 

> Closing because after discussing this today we concluded that this wasnt the right way to go

Can you elaborate a bit more on why? What is the right way to do it then? I wasn't in the discussion, so I think a summary here would be a good idea :-)

In the meantime, I propose another fix: make [these setfacl calls](https://github.com/YunoHost/yunohost/blob/dev/helpers/multimedia#L47) work (either by ignoring the error or by checking the validity of each links and use setfacl only on those that are valid). What do you think?

alexAubin:
>I can answer that with my only practical use case, which is a killer feature in my opinion: downloading something with the transmission application, and have it immediately available in the minidlna server. For ex, after download is complete, a video will be immediately playable with vlc from my personal computer. That is the only use case I really care about.

Yes indeed that's a major use case, thanks for reminding about it :+1: 

I think in the discussion the main confusion was about the whole "Share" folder thingy between users and the multimedia folder

>Can you elaborate a bit more on why? What is the right way to do it then? I wasn't in the discussion, so I think a summary here would be a good idea :-)

Eeeer yeah I don't remember the full discussion because it was late x_x I think it was mainly : 

1) "we should really re-review the whole multimedia folder thingy because we don't understand what's going on" ... for example I think I regularly see people encountering similar issues because some /home/$user doesnt exist anymore (not to mention the fact that implicitly displaying the list of user in logs is not great) 

and 2) as you suggest, ignoring the error seems OK, ultimately we don't want to miserably crash just because some symlink is broken, which can happen for a good variety of reasons ... hence adding this trick : https://github.com/YunoHost/yunohost/commit/94f21ea20e47cae27fc71cf3375b5912453549aa ... though re-reading it I'm wondering if we shouldnt also add it to the two previous lines. Note that we also quickly explored the option to "checking the validity of each links and use setfacl only on those that are valid", but the exact option was not easily available in setfacl's man

autra:
> hence adding this trick : [94f21ea](https://github.com/YunoHost/yunohost/commit/94f21ea20e47cae27fc71cf3375b5912453549aa) ... though re-reading it I'm wondering if we shouldnt also add it to the two previous lines. Note that we also quickly explored the option to "checking the validity of each links and use setfacl only on those that are valid", but the exact option was not easily available in setfacl's man

Yep I think the 2 lines above should be changed as well indeed.

I'd try something like (not tested):

```bash
    # I removed the -R below
    ## Application des droits étendus sur le dossier multimedia.
    # Droit d'écriture pour le groupe et le groupe multimedia en acl et droit de lecture pour other:
    setfacl -nL -m g:$MEDIA_GROUP:rwX,g::rwX,o:r-X "$MEDIA_DIRECTORY"
    # Application de la même règle que précédemment, mais par défaut pour les nouveaux fichiers.
    setfacl -nL -m d:g:$MEDIA_GROUP:rwX,g::rwX,o:r-X "$MEDIA_DIRECTORY"
    # Réglage du masque par défaut. Qui garantie (en principe...) un droit maximal à rwx. Donc pas de restriction de droits par l'acl.
    setfacl -L -m m::rwx "$MEDIA_DIRECTORY"
   
   # check subdirs and use setfacl in the link is valid
   for $subdir in $(ls $MEDIA_DIRECTORY); do
       if [ -d "$subdir" ]; then
             # Droit d'écriture pour le groupe et le groupe multimedia en acl et droit de lecture pour other:
             setfacl -nL -m g:$MEDIA_GROUP:rwX,g::rwX,o:r-X "$subdir"
             # Application de la même règle que précédemment, mais par défaut pour les nouveaux fichiers.
             setfacl -nL -m d:g:$MEDIA_GROUP:rwX,g::rwX,o:r-X "$subdir"
             # Réglage du masque par défaut. Qui garantie (en principe...) un droit maximal à rwx. Donc pas de restriction de droits par l'acl.
             setfacl -L -m m::rwx "$subdir"
       fi

  done
```
But I'm not really sure how setfacl behaves with symlink (does it dereference them?). Also maybe this MEDIA_DIRECTORY can contains regular files and dir, this proposition does not test for that (though it is easy to add other tests).

PR #1440, opened 2022-02-13
Avoid grep issue with vmstat command

ewilly:
due to localized vmstat in bullseye

## The problem
`vmstat --stats --unit M | grep "free memory" | awk '{print $1}'` return nothing
`vmstat --stats --unit M` give expected values but with translated descriptions

## Solution
force running the command in LANG C

## PR Status

...

## How to test

...

### DIFF
--- a/helpers/hardware
+++ b/helpers/hardware
@@ -30,8 +30,8 @@ ynh_get_ram() {
         ram=0
     # Use the total amount of ram
     elif [ $free -eq 1 ]; then
-        local free_ram=$(vmstat --stats --unit M | grep "free memory" | awk '{print $1}')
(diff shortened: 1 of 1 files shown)


ewilly:
should solve [https://forum.yunohost.org/t/beta-stage-testing-for-yunohost-11-0-bullseye-and-buster-bullseye-migration/18531/44](https://forum.yunohost.org/t/beta-stage-testing-for-yunohost-11-0-bullseye-and-buster-bullseye-migration/18531/44)
`/usr/share/yunohost/helpers.d/php: ligne 408: / 2 / 20 : erreur de syntaxe : opérande attendu (le symbole erroné est « / 2 / 20 »)`

alexAubin:
Ooooh thanks !

PR #1441, opened 2022-02-16
Several FIXME fix :rocket: 

Tagadda:
## The problem

I got distracted by some FIXME :fearful: 

## Solution

Some small fixes...okwell that was small at some point
* `yunohost domain list -output-as plain` also print `#main` and `#domain`, use the command in metronome regen-conf create some unnecessary files. => Create a new env `YNH_MAIN_DOMAINS` for regen-conf.
* Add `default` and `pattern` in the `yunohost <thing> config get --full`. `default` is used in the webadmin, especially in `select`s. I don't think we use `pattern` yet but this could be usefull
* Move the `_list_upgradable_apps` logic from `tools.py` to `app.py`. Also added `yunohost app list --upgradable` in the actionmap because why not ? :rocket: 

## PR Status

Tested locally

## How to test

```
yunohost tools regen-conf metronome --debug

yunohost domain config get domain.tld --full

# Install some old app version
yunohost app install https://github.com/YunoHost-Apps/20euros_ynh/tree/f8c5f11894b7d0579bea6d583a776ada9de4a6b8
yunohost tools update apps
yunohost app list --upgradable
yunohost tools upgrade apps
```

### DIFF
--- a/hooks/conf_regen/12-metronome
+++ b/hooks/conf_regen/12-metronome
@@ -49,12 +49,8 @@ do_post_regen() {
     # retrieve variables
     main_domain=$(cat /etc/yunohost/current_host)
 
-    # FIXME : small optimization to do to avoid calling a yunohost command ...
-    # maybe another env variable like YNH_MAIN_DOMAINS idk
-    domain_list=$(yunohost domain list --exclude-subdomains --output-as plain --quiet)
-
     # create metronome directories for domains
-    for domain in $domain_list; do
+    for domain in $YNH_MAIN_DOMAINS; do
         mkdir -p "/var/lib/metronome/${domain//./%2e}/pep"
         # http_upload directory must be writable by metronome and readable by nginx
         mkdir -p "/var/xmpp-upload/${domain}/upload"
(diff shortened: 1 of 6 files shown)


src/diagnosers/10-ip.py
```diff
@@ -176,7 +176,7 @@ def is_default_route(r):
         def ping(protocol, target):
             return (
                 os.system(
-                    "ping%s -c1 -W 3 %s >/dev/null 2>/dev/null"
+                    "ping%s -c1 %s >/dev/null 2>/dev/null"
```
alexAubin, line 179: Heuarg I'm not sure about that one ... I believe it's important to have a timeout otherwise the program may get stuck here ? (Or maybe there's a default timeout anyway ?) I'm confused about `ping6` not having a `-W` option ... there's also `ping -[4|6]`, I'm not sure why that one wasn't used ... But if I remember correctly there may be several versions of the `ping` command hence why it's a mess @_@

alexAubin: Sooo I'd be curious to know how you noticed that bug ? And on that corresponding machine, check if `ping -6` exists and wether or not it's the same as `ping6` or what are the differences / from what deb package it is from

Tagadda: Uuuhh I can't reproduce it :shrug:
I will revert this commit

PR #1443, opened 2022-02-27
[fix] Avoid to delete --is-big unbackuped directories

zamentur:
## The problem

Some apps like bozon declare the data_dir with --is-big in backup and remove this path in the remove script (without --purge option)
However, it doesn't make sense for bozon or similar apps (like lufi) to remove conditionnaly with the --purge option, cause if we remove the db, we have no interest to keep the files.

## Solution

I suggest to add here a REMOVE_CORE_ONLY option to use like BACKUP_CORE_ONLY. With this option if a user remove the app, unreusable big files will be deleted, but if it's the ynh_restore_upgradebackup helper that trigger the remove, we will keep the files to avoid to lose them before the restore operation...

## PR Status

Ready

## How to test

...


alexAubin:
Closing as discussed in yunocamp

PR #1444, opened 2022-03-02
Always activate time when running inside CI tests

pagiraud:
## The problem

To adjust the weight parameter, you need to use the --time argument of ynh_script_progression and then need to remove it, and that for every occurrence. Sure, using an alias or a regular expression may help, but configuring the script may provide a more efficient way to alleviate this task.

## Solution

The time is now always displayed when the scripts (install, backup, upgrade, etc.) are run in the CI environment while it can still be manually set while running outside.

## How to test

I did not test my commit.

### DIFF
--- a/helpers/logging
+++ b/helpers/logging
@@ -250,7 +250,7 @@ ynh_script_progression() {
     weight=${weight:-1}
     
     # Always activate time when running inside CI tests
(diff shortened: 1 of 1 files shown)


alexAubin (approved):

tituspijean (approved):

alexAubin:
Yolomergin'

PR #1445, opened 2022-03-03
Implement install and removal of YunoHost apps

yalh76:
## The problem

Improve management of YunoHost apps installation 

## Solution

Adding a dedicated helper

## PR Status

Ready

## How to test

Done some implementation on:

- https://github.com/YunoHost-Apps/peertube_ynh/pull/309
- https://github.com/YunoHost-Apps/jitsi_ynh/pull/65 include a full test with prosody already installed


yalh76:
seems quite ready

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+	
+	# For each app
+	for i in "${apps_list[@]}"
+	do
+		# Retrieve the name of the app (part before _ynh)
+		local oneapp=$(echo "$i" | awk -F'_ynh' '{print $1}')
+		[ -z "$oneapp" ] && ynh_die --message="You didn't provided a YunoHost app to install"
+
+		# Retrieve the arguments of the app (part after ?)
+		local oneargument=$(echo "$i" | awk -F'?' '{print $2}')
```
alexAubin, line 28: Note that this is not resilient to `?` being in the args (such as a password value)

Same goes for `"` later

But maybe this doesn't matter too much because in actual use cases (such as installing prosody etc) there's no need to define a human password for these apps

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+		[ ! -z "$oneargument" ] && oneargument="--args \"$oneargument\""
+
+		if ! yunohost app list | grep -q "$oneapp"
```
alexAubin, line 31: Ideally it is more robust to use `yunohost app list --output-as json | jq ...` with the appropriate jq expression / filter :stuck_out_tongue_winking_eye:  (not trivial but it's worth learning to craft such stuff). There may be similar examples in the other helpers to get inspiration from

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+		then
+			yunohost tools update
```
alexAubin, line 33: - `yunohost tools update` is called both in the `if` and `else`, so might as well factorize it outside the if ?
- maybe add `--apps` (or is it `apps` nowadays?) to speed things up and not run the `apt update` which can take significant time on small arm boards

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+	ynh_handle_getopts_args "$@"
+
+	# Split the list of apps in an array
+	local apps_list=($(echo $apps | tr " " "\n"))
+	
+	# For each app
+	for i in "${apps_list[@]}"
+	do
+		# Retrieve the name of the app (part before _ynh)
+		local oneapp=$(echo "$i" | awk -F'_ynh' '{print $1}')
```
alexAubin, line 66:
```suggestion
		local oneapp=$(echo "$i" | awk -F'?' '{print $1}')
```

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+		fi
+		ynh_app_setting_set --app=$app --key=require_$oneapp --value="1"
+	done
+}
+
+# Remove other YunoHost apps
+#
+# apps will be removed only if no other apps need them.
+#
+# usage: ynh_remove_apps --apps="a_ynh b_ynh?domain=DOMAIN&path=PATH&admin=USER&language=fr&is_public=1&password=pass&port=666"
```
alexAubin, line 47: Is it really meaningful to provide the args for that case, and not just the app ids ?

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+	for i in "${apps_list[@]}"
+	do
+		# Retrieve the name of the app (part before _ynh)
+		local oneapp=$(echo "$i" | awk -F'_ynh' '{print $1}')
+		[ -z "$oneapp" ] && ynh_die --message="You didn't provided a YunoHost app to remove"
+
+		ynh_app_setting_delete --app=$app --key=require_$oneapp
+
+		# List apps requiring $oneapp
+		local installed_apps=$(yunohost app list | grep -oP 'id: \K.*$')
```
alexAubin, line 72: (Same here about parsing app ids, it's better to parse `--output-as json` with jq ? ... Or alternatively `ls /etc/yunohost/apps` ?)

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+			local installed_app_required_by=$(ynh_app_setting_get --app=$installed_app --key="require_$oneapp")
+			if [[ $installed_app_required_by ]]
+			then
+				required_by="${installed_app_required_by}"
+			fi
+			installed_app_required_by=""
+		done
+
+		# If $oneapp is no more required
+		if [[ ! $required_by ]]
```
alexAubin, line 86:
```suggestion
		if [[ -z "$required_by" ]]
```

helpers/app
```diff
@@ -0,0 +1,93 @@
...
+		ynh_app_setting_delete --app=$app --key=require_$oneapp
+
+		# List apps requiring $oneapp
+		local installed_apps=$(yunohost app list | grep -oP 'id: \K.*$')
+		local required_by=""
+		local installed_app_required_by=""
+		for installed_app in $installed_apps
+		do
+			local installed_app_required_by=$(ynh_app_setting_get --app=$installed_app --key="require_$oneapp")
+			if [[ $installed_app_required_by ]]
```
alexAubin, line 78:
```suggestion
			if [[ -n "$installed_app_required_by" ]]
```

helpers/app
yalh76: found it:
`yunohost app list --output-as json --quiet | jq -e --arg id $oneapp '.apps[] | select(.id == $id)'`

helpers/app
yalh76: done ;)

helpers/app
yalh76: found it too.
To have a lits of all installer apps `yunohost app list --output-as json --quiet | jq -r .apps[].id`

helpers/app
yalh76: fixed using `one_argument=$(cut -d "?" -f2- <<< "$one_app_and_its_args")`

yalh76:
Add a test for that helper:
like
https://github.com/YunoHost/yunohost/blob/dev/tests/test_helpers.d/ynhtest_apt.sh

PR #1449, opened 2022-03-23
[fix] Config Panel edits partially config of the first instance

zamentur:
## The problem

> hey, do you know if config_panel could be multi_instance? 😬
> the bookstack config_panel edit the config of the first installed bookstack instance, so not the instance I would like to configure

## Solution

It seems config panel have the final_path loaded from the first instance instead the good one !

Use self.entity that contains the full app name (like nextcloud__2) instead of the app_id (the app name without the number, like nextcloud). 

## PR Status

Untested

## How to test

...

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -293,8 +293,8 @@ def app_map(app=None, raw=False, user=None):
     permissions = user_permission_list(full=True, absolute_urls=True, apps=apps)[
         "permissions"
     ]
-    for app_id in apps:
-        app_settings = _get_app_settings(app_id)
+    for app in apps:
+        app_settings = _get_app_settings(app)
         if not app_settings:
             continue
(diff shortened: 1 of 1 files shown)


src/app.py
```diff
@@ -1592,15 +1592,16 @@ def _call_config_script(self, action, env=None):
 
         # Call config script to extract current values
         logger.debug(f"Calling '{action}' action from config script")
-        app_id, app_instance_nb = _parse_app_instance_name(self.entity)
-        settings = _get_app_settings(app_id)
+        app = self.entity
+        app_id, app_instance_nb = _parse_app_instance_name(app)
+        settings = _get_app_settings(app)
```
zamentur, line 1597: The true fix is this line. The other are just semantic fixes to avoid it happens again.

PR #1450, opened 2022-03-23
Fix flag case sensitivity in dovecot and rspamd sieve filter

zeroheure:
AFAIK this value is case sensitive. It is written as "Yes" in rspamd/milter_headers.conf

### DIFF
--- a/conf/dovecot/dovecot.sieve
+++ b/conf/dovecot/dovecot.sieve
@@ -1,4 +1,4 @@
 require "fileinto"; 
-    if header :contains "X-Spam-Flag" "YES" { 
+    if header :contains "X-Spam-Flag" "Yes" { 
         fileinto "Junk"; 
     }
--- a/conf/rspamd/rspamd.sieve
+++ b/conf/rspamd/rspamd.sieve
@@ -1,4 +1,4 @@
 require ["fileinto"];
-if header :is "X-Spam" "yes" {
+if header :is "X-Spam" "Yes" {
     fileinto "Junk";
 }


alexAubin:
Eeeeh, is it ? I see several results on search engine where it's written with all cases ? For example Dovecot official doc (but not only):

https://doc.dovecot.org/configuration_manual/sieve/examples/#direct-filtering-using-message-header

zeroheure:
Well I've browsed many pages in dovecot doc where there is warning about case sensitivity (for example https://doc.dovecot.org/search/?q=case+sensitivity). And case sensitivity is usual. 

Maybe I'm wrong, but I can see no reason of someone implemented case insensitivity for analysing a value on Unix servers.

zeroheure:
unrelated, but are you sure that the Junk folder is allways created ?

zeroheure:
See [RFC 5228](https://www.ietf.org/rfc/rfc5228.txt) "Sieve: An Email Filtering Language". 

Look at the 2.7 section. As it left me very confused, I quote below some other paragraphs . **NB: If you want to search in the RFC for case (in)sensitivity the search string should be "sensitiv" because of various spelling used** 


> 2.1. Form of the Language 
Tokens other than strings are considered case-insensitive.

Does this implie that strings are case sensitive ?

> 2.7 String Comparison
In order to provide for matches between character sets and case insensitivity, [snip]
Header comparisons are always  done with the "i;ascii-casemap" operator, i.e., case-insensitive comparisons

> 2.7.3.  Comparators
In order to allow for language-independent, case-independent matches, the match type may be coupled with a comparator name.  [snip]
"i;ascii-casemap" comparator (which treats uppercase and lowercase characters in the US-ASCII subset of UTF-8 as the same).  If left unspecified, the default is "i;ascii-casemap".

PR #1454, opened 2022-04-15
Enforce chmod 644(-rw-r--r--) on logrotate configuration files

Salamandar:
## The problem

Some logrotate files are created as `-rw-rw-rw-`. This is really bad.

Also, obviously, logrotate doesn't want to use a configuration file that anyone in the world can write. Good boi.

## Solution

We can just ensure the files are correctly chmod in the logrotate helper.

### DIFF
--- a/helpers/logrotate
+++ b/helpers/logrotate
@@ -90,6 +90,7 @@ $logfile {
 EOF
     mkdir --parents $(dirname "$logfile")                              # Create the log directory, if not exist
     cat ${app}-logrotate | $customtee /etc/logrotate.d/$app >/dev/null # Append this config to the existing config file, or replace the whole config file (depending on $customtee)
+    chmod 644 /etc/logrotate.d/$app
 }
 
 # Remove the app's logrotate config.


Tagadda:
What about setting a `umask` for default permissions ?

Salamandar:
@Tagadda There's a risk it has an impact on the rest of the script.

Buuuut I guess that's not a bad idea to set a umask for the whole package scripts. That's another issue though. What do you think @tituspijean ? Is it worth creating an issue ? :D

helpers/logrotate
```diff
@@ -88,8 +88,12 @@ $logfile {
     $su_directive
 }
 EOF
-    mkdir --parents $(dirname "$logfile")                              # Create the log directory, if not exist
-    cat ${app}-logrotate | $customtee /etc/logrotate.d/$app >/dev/null # Append this config to the existing config file, or replace the whole config file (depending on $customtee)
+    # Create the log directory, if not exist
+    mkdir --parents $(dirname "$logfile")
+    chmod 644 $(dirname "$logfile")
```
alexAubin, line 93: Shouldnt `755` be more appropriate for a directory (c.f the `x` permission)

I'm not sure to understand if permissions for others really are necessary ? (Maybe that's because logrotate runs with a specific user)

PR #1458, opened 2022-04-26
[Fix] 1980 extract correct domain basename

DiesDasJenes:
## The problem

The TLD is removed from the domain basename in the line 71 by the function check_domain in the file 12-dnsrecords.pz by the following line:
`basename = domain.replace(base_dns_zone, "").rstrip(".") or "@"`

If you have a domain like “trucfrmuch.fr” the diagnosis tool will complain about not having a record for "trucmuch" if you are in France. I am located in Germany and got the same error with the domain “philippludewig” with “de” so I can expect that there are similar problems for different countries.

https://github.com/YunoHost/issues/issues/1980

## Solution

I introduced the package python3-tldextract and the following lines 
`basename = tldextract.extract(domain).domain
        if len(basename) == 0:
            basename = "@"`

## PR Status

Initialised

## How to test

Depending on your tld of your country. Example is using France.

-     *Go to the yunohost admin page to the menu domain
-     *Click on create new domain
-     Create a new domain “trucfrmuch.fr”
-     Run the diagnosis tool for it
-     See error


src/diagnosers/12-dnsrecords.py
```diff
@@ -68,7 +69,9 @@ def check_domain(self, domain, is_main_domain):
             return
 
         base_dns_zone = _get_dns_zone_for_domain(domain)
-        basename = domain.replace(base_dns_zone, "").rstrip(".") or "@"
+        basename = tldextract.extract(domain).domain
```
zamentur, line 72: Are you sure the goal is to remove the TLD ?

If i understand the goal of `_get_dns_zone_for_domain` is to get the DNS zone (so not the tld), the test seems confirm the semantic: https://github.com/YunoHost/yunohost/blob/140e50253fac0d3c9aa6fcab9e392a462c914e98/src/tests/test_dns.py#L30

So `basename` should be the part at the left in the DNS configuration.

zamentur:
Thanks for your contribution :)

src/diagnosers/12-dnsrecords.py
zamentur: So i confirm there is 2 issues: the way we do the replace (that you have spotted) and the 

```
# yunohost tools shell
Warning: The ldap interface is available in this context
Warning: You don't have IPython installed, consider installing it as it is way better than the standard shell.
Warning: Falling back on the standard shell.
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from yunohost.dns import _get_dns_zone_for_domain
>>> _get_dns_zone_for_domain("phillipludewig.de")
'de'
```

For me the result should be "phillipludewig.de"

zamentur: I note that the domain name `philipludewig.de` don't return a NS record and is not reserved. So the bug appears when the domain is not recorded OR if there is no NS record configured.

zamentur:
I produces an alternative PR to solve this issues, thanks for your contribution and explanation about this issues :)

src/diagnosers/12-dnsrecords.py
DiesDasJenes: btw your fix, when its merged into the next release, will enable me to use my yunohost instance again properly ^^
so I am really grateful for your support. :D

PR #1459, opened 2022-04-26
[enh] Refactor global settings to use the new config panel framework

Tagadda:
## The problem

...

## Solution

...

## PR Status

Mostly done, needs testing and unit tests.

## How to test

...


locales/en.json
```diff
@@ -367,33 +367,41 @@
...
-    "global_settings_setting_smtp_relay_password": "SMTP relay host password",
-    "global_settings_setting_smtp_relay_port": "SMTP relay port",
-    "global_settings_setting_smtp_relay_user": "SMTP relay user account",
-    "global_settings_setting_ssowat_panel_overlay_enabled": "Enable SSOwat panel overlay",
+    "global_settings_reset_success": "Reset global settings",
+    "global_settings_setting_admin_strength": "Admin password strength",
+    "global_settings_setting_backup_compress_tar_archives": "Compress backups",
+    "global_settings_setting_backup_compress_tar_archives_help": "When creating new backups, compress the archives (.tar.gz) instead of uncompressed archives (.tar). N.B. : enabling this option means create lighter backup archives, but the initial backup procedure will be significantly longer and heavy on CPU.",
+    "global_settings_setting_nginx_compatibility": "Compatibility",
```
alexAubin, line 374:
```suggestion
    "global_settings_setting_nginx_compatibility": "NGINX Compatibility",
```

locales/en.json
```diff
@@ -367,33 +367,41 @@
...
+    "global_settings_setting_nginx_compatibility_help": "Compatibility vs. security tradeoff for the web server NGINX. Affects the ciphers (and other security-related aspects)",
+    "global_settings_setting_nginx_redirect_to_https": "Force HTTPS",
+    "global_settings_setting_nginx_redirect_to_https_help": "Redirect HTTP requests to HTTPs by default (DO NOT TURN OFF unless you really know what you're doing!)",
+    "global_settings_setting_pop3_enabled": "Enable POP3",
+    "global_settings_setting_pop3_enabled_help": "Enable the POP3 protocol for the mail server",
+    "global_settings_setting_postfix_compatibility": "Compatibility",
```
alexAubin, line 380:
```suggestion
    "global_settings_setting_postfix_compatibility": "Postfix Compatibility",
```

locales/en.json
```diff
@@ -367,33 +367,41 @@
...
+    "global_settings_setting_smtp_allow_ipv6_help": "Allow the use of IPv6 to receive and send mail",
+    "global_settings_setting_smtp_relay_enabled": "Enable SMTP relay",
+    "global_settings_setting_smtp_relay_enabled_help": "Enable the SMTP relay to use in order to send mail instead of this yunohost instance. Useful if you are in one of this situation: your 25 port is blocked by your ISP or VPS provider, you have a residential IP listed on DUHL, you are not able to configure reverse DNS or this server is not directly exposed on the internet and you want use an other one to send mails.",
+    "global_settings_setting_smtp_relay_host": "Relay host",
+    "global_settings_setting_smtp_relay_password": "Relay password",
+    "global_settings_setting_smtp_relay_port": "Relay port",
+    "global_settings_setting_smtp_relay_user": "Relay user",
+    "global_settings_setting_ssh_allow_deprecated_dsa_hostkey": "Allow DSA hostkey",
+    "global_settings_setting_ssh_allow_deprecated_dsa_hostkey_help": "Allow the use of (deprecated) DSA hostkey for the SSH daemon configuration",
+    "global_settings_setting_ssh_compatibility": "Compatibility",
```
alexAubin, line 394:
```suggestion
    "global_settings_setting_ssh_compatibility": "SSH Compatibility",
```

locales/en.json
```diff
@@ -478,6 +486,9 @@
     "log_user_permission_reset": "Reset permission '{}'",
     "log_user_permission_update": "Update accesses for permission '{}'",
     "log_user_update": "Update info for user '{}'",
+    "log_settings_set": "Apply settings",
+    "log_settings_reset": "Reset setting",
+    "log_settings_reset_all": "Reset all setting",
```
alexAubin, line 491:
```suggestion
    "log_settings_reset_all": "Reset all settings",
```

share/config_settings.toml
```diff
@@ -0,0 +1,145 @@
...
+
+        [security.ssh.ssh_port]
+        type = "number"
+        default = 22
+
+        [security.ssh.ssh_password_authentication]
+        type = "boolean"
+        yes = "True"
+        no = "False"
+        default = "False"
```
alexAubin, line 31: I guess all those are related to the fact that in bash scripts we later check `if [[ $value == "True" ]]` (or similarly in jinja templates) ? I wonder what happens if we just set `default = true`, will `yunohost setting get foo.bar` output `true` ? I guess we could 'clean' the bash scripts and jinja template, but maybe that's not worth it ... (I'm just thinking out loud)

share/config_settings.toml
```diff
@@ -0,0 +1,145 @@
...
+        no = "False"
+        default = "True"
+    
+    [misc.backup]
+    name = "Backup"
+        [misc.backup.backup_compress_tar_archives]
+        type = "boolean"
+        yes = "True"
+        no = "False"
+        default = "False"
```
alexAubin, line 145: So I guess in the mid/long-term we'll probably want a `portal` and `backup` entire sections (or is that a panel ?). But maybe migrating the name later is not really a big deal (thinking out loud again)

src/settings.py
```diff
@@ -2,157 +2,67 @@
...
-    """
-    Ensure a string value is intended as a boolean
+SETTINGS_PATH = "/etc/yunohost/settings.yml"
 
-    Keyword arguments:
-        arg -- The string to check
+BOOLEANS = {
+    "True": True,
+    "False": False,
+}
```
alexAubin, line 21:
```suggestion
```

(c.f. related suggestion before/after)

src/settings.py
```diff
@@ -161,222 +71,122 @@ def settings_set(key, value):
...
-    # modification inside of it and take advantage of its git history
-    old_settings_backup_path = (
-        SETTINGS_PATH_OTHER_LOCATION % datetime.utcnow().strftime("%F_%X")
-    )
-    _save_settings(settings, location=old_settings_backup_path)
 
-    for value in settings.values():
-        value["value"] = value["default"]
+class SettingsConfigPanel(ConfigPanel):
+    entity_type = "settings"
```
alexAubin, line 109:
```suggestion
    entity_type = "global"
```

Should be less confusing ? Since usually we talk about "domain settings" or "app settings" or "global settings", and entity_type is already "domain" and "app" for the other two ?

src/settings.py
```diff
@@ -161,222 +71,122 @@ def settings_set(key, value):
...
+        if mode == "full":
+            for panel, section, option in self._iterate():
+                if m18n.key_exists(self.config["i18n"] + "_" + option["id"] + "_help"):
+                    option["help"] = m18n.n(self.config["i18n"] + "_" + option["id"] + "_help")
+            return self.config
 
-def _get_settings():
+        # Dirty hack to let settings_get() to work from a python script
+        if isinstance(result, str) and result in BOOLEANS:
+            result = BOOLEANS[result]
```
alexAubin, line 140:
```suggestion
        # Dirty hack to let settings_get() to work from a python script
        if isinstance(result, str) and result in ["True", "False"]:
            result = bool(result)
```

Sounds easier to read that way ?

src/settings.py
```diff
@@ -404,48 +214,48 @@ def trigger_post_change_hook(setting_name, old_value, new_value):
...
-@post_change_hook("security.postfix.compatibility")
+@post_change_hook("smtp_allow_ipv6")
+@post_change_hook("smtp_relay_host")
+@post_change_hook("smtp_relay_port")
+@post_change_hook("smtp_relay_user")
+@post_change_hook("smtp_relay_password")
+@post_change_hook("postfix_compatibility")
 def reconfigure_postfix(setting_name, old_value, new_value):
     if old_value != new_value:
         regen_conf(names=["postfix"])
```
alexAubin, line 255: It feels like it would be more elegant to have some way to tell directly in the settings' toml that we want to regen-conf some stuff when these gets changed ... but that wouldnt cover more complex cases such as `pop3_enabled` where we trigger and entire `apt install` etc ... maybe that's no big deal and we can keep this as is for now

hooks/conf_regen/19-postfix
```diff
@@ -22,17 +22,19 @@ do_pre_regen() {
...
-        relay_port="$(yunohost settings get 'smtp.relay.port')"
-        relay_user="$(yunohost settings get 'smtp.relay.user')"
-        relay_password="$(yunohost settings get 'smtp.relay.password')"
+    export relay_host=""
+    export relay_enabled="$(yunohost settings get 'email.smtp.smtp_relay_enabled')"
+    if [ "${relay_enabled}" == "True" ]; then
+        relay_host="$(yunohost settings get 'email.smtp.smtp_relay_host')"
+        relay_port="$(yunohost settings get 'email.smtp.smtp_relay_port')"
+        relay_user="$(yunohost settings get 'email.smtp.smtp_relay_user')"
+        relay_password="$(yunohost settings get 'email.smtp.smtp_relay_password')"
```
alexAubin, line 37: Maybe if `yunohost settings get 'email.smtp'` works that could be a smarter way to fetch those settings with only one CLI call

alexAubin (approved):
Quick review, LGTM (only minor comments / thinking out loud about possible improvements later)

alexAubin (approved):
LGTM I think, but still have to test on the webadmin

PR #1460, opened 2022-04-27
[fix] Bad DNS conf suggestion

zamentur:
## The problem

Try to fix the problem explain in
https://github.com/YunoHost/yunohost/pull/1458

## Solution

...

## PR Status

Untested

## How to test

...

### DIFF
--- a/src/dns.py
+++ b/src/dns.py
@@ -183,8 +183,7 @@ def _build_dns_conf(base_domain, include_empty_AAAA_if_no_ipv6=False):
         # foo.sub.domain.tld #       domain.tld  #  foo.sub  # .foo.sub #
         #     sub.domain.tld #   sub.domain.tld  #        @  #          #
         # foo.sub.domain.tld #   sub.domain.tld  #      foo  # .foo     #
-
-        basename = domain.replace(base_dns_zone, "").rstrip(".") or "@"
+        basename = _get_relative_name_for_dns_zone(domain, base_dns_zone)
         suffix = f".{basename}" if basename != "@" else ""
 
         # ttl = settings["ttl"]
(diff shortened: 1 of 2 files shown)


src/dns.py
```diff
@@ -482,11 +488,15 @@ def _get_dns_zone_for_domain(domain):
...
 
     logger.warning(
-        f"Could not identify the dns zone for domain {domain}, returning {zone}"
+        f"Could not identify correctly the dns zone for domain {domain}, returning {zone}"
     )
     return zone
 
 
+def _get_relative_name_for_dns_zone(domain, base_dns_zone):
+    return re.sub("\.?" + base_dns_zone.replace(".", "\.") + "$", "", domain.strip(".")) or "@"
```
alexAubin, line 497:
```suggestion
def _get_relative_name_for_dns_zone(domain, base_dns_zone):
    # Strip the base dns zone name from a domain such that it's suitable for DNS manipulation relative to a defined zone
    # For example, assuming base_dns_zone is "example.tld":
    #    example.tld -> @
    #    foo.example.tld -> foo
    #    .foo.example.tld -> foo
    #    bar.foo.example.tld -> bar.foo
    return re.sub("\.?" + base_dns_zone.replace(".", "\.") + "$", "", domain.strip(".")) or "@"
```

alexAubin (approved):

src/dns.py
alexAubin, line 497:
```
src/dns.py:497:20: W605 invalid escape sequence '\.'
src/dns.py:497:55: W605 invalid escape sequence '\.'
```

Uuuh not sure why the linter is unhappy about that

zamentur:
```suggestion
    return re.sub(r"\.?" + base_dns_zone.replace(".", r"\.") + "$", "", domain.strip(".")) or "@"
```

Tagadda (approved):

PR #1466, opened 2022-05-19
Make changed variables' old values available in config panels scripts

tituspijean:
## The problem

https://github.com/YunoHost-Apps/my_webapp_ynh/pull/82

PHP versions management is hindered partially because the variables' new values are already set when a config panel script is launched.

## Solution

In any ways, we should have the old variable values available at old_${variable}.

## PR Status

"Meh"

## How to test

Just. Do. It.


tituspijean:
Closed for the time being, we can simply use a setter for that.

PR #1471, opened 2022-06-29
[enh] Use multiple servers behind get_public_ip to prevent a single point of failure when ip.yunohost.org is down

theo-is-taken:
## The problem

Currently, If ip.yunohost.org is down, IP verification will fail for all servers.

## Solution

We need to have a (preferably configurable) list of similar services to check. 
I used https://github.com/YunoHost/yunohost/pull/1244 as a base, fixed the different merge conflicts, and created two new options ( using the global settings config panel https://github.com/YunoHost/yunohost/pull/1459 )

## PR Status

Ready for review

## How to test

...


zamentur (changes requested):

zamentur (approved):
LGTM

src/utils/network.py
```diff
@@ -76,19 +64,75 @@ def is_default_route(r):
...
+        return ip_list[0]
 
-    try:
-        return download_text(url, timeout=30).strip()
-    except Exception as e:
-        logger.debug(f"Could not get public IPv{protocol} : {e}")
-        return None
+    return None
+
+def get_public_ips(protocol=4):
```
alexAubin, line 78: This is getting confusing because we end up with `get_public_ip` calling `get_public_ip_from_remote_server` calling `get_public_ips` ... I would tend to keep `get_public_ip_from_remote_server` and `get_public_ips` merged together as one big function (yeah) because it's easier to read and understand. Maybe renaming the function into `get_public_ip_from_mirrors` in the process

alexAubin (changes requested):
It would really be helpful to properly review this if the commits were probably rebased on top of the appropriate branch to avoid the whole diff mess x_x (which also mean we'd get duplicated commits if merged as is..)

alexAubin:
Closing because it's been a year and the git history is still a huge mess .. personally I'm not going to clean it up ...

PR #1475, opened 2022-07-15
Dyndns recovery password

zamentur:
Just to trigger tests on the other PR...


alexAubin:
Copypaste from https://github.com/YunoHost/yunohost/pull/1472#issuecomment-1175338452

---

## 



## TODO
### CLI
- [ ] Unit test
- [x] Rename TSIG public_key into tsig_shared_key
- [x] Add a Domain settings to activate or deactivate the automatic push for dynette domain (Caution: we should think the place in web admin where this should be displayed)
- [ ] Give a way to download/replace the shared key from web admin, for example via the "registrar config" ?

### Image and install process
- [x] Adapt the internet cube and clic install process:
   - [x] https://github.com/labriqueinternet/install/blob/master/install_procedure.py#L36 
   - [x] https://github.com/alexAubin/clic/blob/main/install_procedure.py#L39

### Webadmin
- [x] Adapt webadmin postinstall and domain add
- [ ] Find a way to display unsubscribed domains and subscribed domains in webadmin ?
- [x] Add a question about unsubscribe/no-unsubscribe in the remove domain popup
- [x] Same for domain add

### Documentation
?

### Need an API dynette change

- [ ] Allowing to unsubscribe with the shared key
- [ ] Allowing to change the passphrase `yunohost domain dyndns changepw`
- [ ] Allowing to renew the shared key with the passphrase or the old shared-key 
   - [ ] Make `postinstall` or `domain add` with the good password renew the key
   - [ ] If the domain already exist, ask if the user own the domain and ask the passphrase
   - [ ] Make the restoration on an other server change the key 
- [ ] Get the per server domain limit from dynette API

src/domain.py
```diff
@@ -130,21 +140,25 @@ def _get_parent_domain_of(domain):
...
+def domain_add(operation_logger, domain, subscribe=None, no_subscribe=False):
     """
     Create a custom domain
 
     Keyword argument:
         domain -- Domain name to add
         dyndns -- Subscribe to DynDNS
-
+        subscribe -- Password used to later unsubscribe from DynDNS
+        no_unsubscribe -- If we want to just add the DynDNS domain to the list, without subscribing
```
alexAubin, line 152: These option names are hella confusing x_x

If `subscribe` is the recovery password .. why not name it `recovery_password` ?

alexAubin: Also the `no_(un)subscribe` is confusing ... the previous `dyndns` option was already handling that ... changing the name is going to cause boring backward compatibility issues (at least on the postinstall, I'm pretty sure some scripts have some `--ignore-dyndns` or similar options somewhere)

src/dyndns.py
```diff
@@ -154,14 +179,18 @@ def dyndns_subscribe(operation_logger, domain=None, key=None):
             error = f'Server error, code: {r.status_code}. (Message: "{r.text}")'
         raise YunohostError("dyndns_registration_failed", error=error)
 
+    # Set the domain's config to autopush
+    from yunohost.domain import domain_config_set
+    domain_config_set(domain, key="dns.zone.autopush", value=1)
```
alexAubin, line 184: Something needs to be designed for this to be applied to all dyndns domains on existing instances somehow

hooks/conf_regen/01-yunohost
```diff
@@ -116,8 +116,9 @@ SHELL=/bin/bash
 #   - (sleep random 60 is here to spread requests over a 1-min window)
 #   - if ip.yunohost.org answers ping (basic check to validate that we're connected to the internet and yunohost infra aint down)
 #   - and if lock ain't already taken by another command
-#   - trigger yunohost dyndns update
-*/10 * * * * root : YunoHost DynDNS update; sleep \$((RANDOM\\%60)); ! ping -q -W5 -c1 ip.yunohost.org >/dev/null 2>&1 || test -e /var/run/moulinette_yunohost.lock || yunohost dyndns update >> /dev/null
+#	- check if some domains are flagged as autopush
+#   - trigger yunohost domain dns push --auto
+*/10 * * * * root : YunoHost DynDNS update ; sleep \$((RANDOM\\%60)); ! ping -q -W5 -c1 ip.yunohost.org >/dev/null 2>&1 || test -e /var/run/moulinette_yunohost.lock || !(grep -nR "autopush: 1" /etc/yunohost/domains/*.yml > /dev/null) || yunohost domain dns push --auto >> /dev/null
```
alexAubin, line 121: There's some ambiguity here : as far as I understand, the new `domain dns push --auto`  is about pushing all domains, not just the "yunohost dyndns" ones.

But this cron job is only created if some dyndns key exist

We probably want to get rid of the legacy `yunohost-dyndns` cron job and have a new `yunohost-autopushdns` (or similar) instead

src/dyndns.py
zamentur: all dyndns with a cron on the system

PR #1476, opened 2022-07-15
[fix] Be able to redo postinstall after 128+ chars password

zamentur:
## The problem
When someone do the postinstall with a 128+ char password, the person gets a message about the need for a 127- char password.
However when the person retry to postinstall, the postinstall is half made.So the second time it says  "The domain already exists" so the person have to reinstall or use the unknown reset script on https://github.com/YunoHost/yunoScripts/blob/master/resetPostinstall.sh

## Solution

Do the check before to make changes.

In more i have added the checks on user account (not just admin) cause user account could have ssh permission too now.

## PR Status

Untested

## How to test

...

### DIFF
--- a/src/tools.py
+++ b/src/tools.py
@@ -71,7 +71,10 @@ def tools_adminpw(new_password, check_strength=True):
 
     """
     from yunohost.user import _hash_user_password
-    from yunohost.utils.password import assert_password_is_strong_enough
+    from yunohost.utils.password import (
+        assert_password_is_strong_enough,
+        assert_password_is_compatible
+    )
     import spwd
 
     if check_strength:
--- a/src/tools.py
+++ b/src/tools.py
@@ -50,7 +50,7 @@ from yunohost.utils.packages import (
     _list_upgradable_apt_packages,
     ynh_packages_version,
 )
-from yunohost.utils.error import YunohostError, YunohostValidationError
+from yunohost.utils.error import yunohosterror, yunohostvalidationerror
 from yunohost.log import is_unit_operation, OperationLogger
 
 MIGRATIONS_STATE_PATH = "/etc/yunohost/migrations.yaml"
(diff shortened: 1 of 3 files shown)


src/tools.py
```diff
@@ -50,7 +50,7 @@
     _list_upgradable_apt_packages,
     ynh_packages_version,
 )
-from yunohost.utils.error import YunohostError, YunohostValidationError
+from yunohost.utils.error import yunohosterror, yunohostvalidationerror
```
alexAubin, line 53: Typo?

src/tools.py
```diff
@@ -77,10 +77,7 @@ def tools_adminpw(new_password, check_strength=True):
     if check_strength:
         assert_password_is_strong_enough("admin", new_password)
 
-    # UNIX seems to not like password longer than 127 chars ...
-    # e.g. SSH login gets broken (or even 'su admin' when entering the password)
-    if len(new_password) >= 127:
-        raise YunohostValidationError("admin_password_too_long")
+    assert_password_is_compatible(new_password)
```
alexAubin, line 80: That function should be imported (cf import just a few lines before)

src/user.py
```diff
@@ -146,7 +146,8 @@ def user_create(
     from yunohost.utils.password import assert_password_is_strong_enough
     from yunohost.utils.ldap import _get_ldap_interface
 
-    # Ensure sufficiently complex password
+    # Ensure compatibility and sufficiently complex password
+    assert_password_is_compatible(password)
```
alexAubin, line 150: That function should be imported (cf import just a few lines before)

src/user.py
```diff
@@ -414,7 +415,8 @@ def user_update(
             change_password = Moulinette.prompt(
                 m18n.n("ask_password"), is_password=True, confirm=True
             )
-        # Ensure sufficiently complex password
+        # Ensure compatibility and sufficiently complex password
+        assert_password_is_compatible(password)
```
alexAubin, line 419: That function should be imported (cf import just a few lines before)

alexAubin (changes requested):

src/tools.py
zamentur:
```suggestion
from yunohost.utils.error import YunohostError, YunohostValidationError
```

PR #1478, opened 2022-07-20
App Config Panel : Don't restrict choices if there's no choices specified

theo-is-taken:
## The problem

Currently, if you have an app config panel that uses tags without restricting choices ( for example, `borg`'s one, with it's `app` field ), the web-admin will still use the component `TagsSelectizeItem` instead of `TagsItem` , making the option unchangeable from the web-admin.

This is due to the fact that the web-admin request the `--full` version of the config panel, and this returns by default an empty list for the choices, instead of `None`.

## Solution

Simply change the default from an empty list to `None`

## PR Status

Ready to merge ? I was trying to think of edge cases where this could break something, but couldn't find one.

### DIFF
--- a/src/utils/config.py
+++ b/src/utils/config.py
@@ -703,7 +703,8 @@ class Question:
         self.default = question.get("default", None)
         self.optional = question.get("optional", False)
         self.visible = question.get("visible", None)
-        self.choices = question.get("choices", [])
+        # Don't restrict choices if there's none specified
+        self.choices = question.get("choices", None)
         self.pattern = question.get("pattern", self.pattern)
         self.ask = question.get("ask", {"en": self.name})
         self.help = question.get("help")


alexAubin:
I guess we could yolomerge but naively to me it sounded consistent to return empty list when there's no choices ...

What I understand is that there are three different kind of "widgets" ?
- The classic simple select where you select one option among N
- The multiple select (equivalent to checkboxes) where you select N tags among N
- (the one causing trouble:) One where the user may enter arbitrary tags ?

This really seems super tight-coupled with the way `borg` works with the field that coresponds to a list of app, but also has special behaviors for `all` and `exclude:` so it can't be just a list of apps ...

alexAubin (approved):
Following discussions IRL, this sounds legit, yolomergin o/

PR #1479, opened 2022-07-22
Saving app's venv in a requirements file, in order to regenerate it

theo-is-taken:
## The problem

After migrating, the old virtual env for python apps ( located in /opt/ ) is no longer valid

## Solution

Before upgrading, the migration script now creates a requirements file listing the different dependancies.
After upgrading the packages, the script deletes and recreates the virtual env from the generated requirements file

## PR Status

Draft

### DIFF
--- a/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
+++ b/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
@@ -71,7 +71,7 @@ def _rebuild_venvs():
             if status!=0:
                 logger.warning(m18n.n("venv_regen_failed", venv=venv))
             else:
-                rm(venv+VENV_REQUIREMENTS_SUFFIX, recursive=True)
+                rm(venv+VENV_REQUIREMENTS_SUFFIX)
         else:
             logger.warning(m18n.n("venv_regen_failed", venv=venv))
 
--- a/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
+++ b/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
@@ -30,7 +30,6 @@ N_CURRENT_YUNOHOST = 4
 N_NEXT_DEBAN = 11
 N_NEXT_YUNOHOST = 11
 
-VENV_BACKUP_SUFFIX = "_BACKUP_VENV"
 VENV_REQUIREMENTS_SUFFIX = "_req.txt"
 VENV_IGNORE = "ynh_migration_no_regen"
 
@@ -64,15 +63,7 @@ def _rebuild_venvs():
 
     venvs = _get_all_venvs("/opt/")+_get_all_venvs("/var/www/")
     for venv in venvs:
-        # Create a backup of the venv, in case there's a problem
-        if os.path.isdir(venv+VENV_BACKUP_SUFFIX):
-            rm(venv+VENV_BACKUP_SUFFIX, recursive=True)
-        backup = True
-        try:
-            cp(venv, venv+VENV_BACKUP_SUFFIX, recursive=True)
-        except:
-            backup = False
-        if backup and os.path.isfile(venv+VENV_REQUIREMENTS_SUFFIX):
+        if os.path.isfile(venv+VENV_REQUIREMENTS_SUFFIX):
             # Recreate the venv
             rm(venv, recursive=True)
             os.system(f"python -m venv {venv}")
--- a/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
+++ b/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
@@ -35,6 +35,7 @@ VENV_REQUIREMENTS_SUFFIX = "_req.txt"
 VENV_IGNORE = "VENVNOREGEN"
 
 def _get_all_venvs(dir,level=0,maxlevel=2):
+    # Using os functions instead of glob, because glob doesn't support hidden folders, and we need recursion with a fixed depth
     result = []
     for file in os.listdir(dir):
         path = os.path.join(dir,file)
--- a/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
+++ b/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
@@ -58,7 +58,7 @@ def _get_all_venvs(dir, level=0, maxlevel=3):
     return result
 
 
-def _generate_requirements():
+def _backup_pip_freeze_for_python_app_venvs():
     """
         Generate a requirements file for all python virtual env located inside /opt/ and /var/www/
     """
@@ -112,7 +112,7 @@ class MyMigration(Migration):
         # Get requirements of the different venvs from python apps
         #
 
-        _generate_requirements()
+        _backup_pip_freeze_for_python_app_venvs()
 
         #
         # Run apt update
--- a/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
+++ b/src/yunohost/data_migrations/0021_migrate_to_bullseye.py
@@ -5,7 +5,7 @@ from moulinette import m18n
 from yunohost.utils.error import YunohostError
 from moulinette.utils.log import getActionLogger
 from moulinette.utils.process import check_output, call_async_output
-from moulinette.utils.filesystem import read_file, rm, write_to_file
+from moulinette.utils.filesystem import read_file, rm, write_to_file, cp
 
 from yunohost.tools import (
     Migration,
@@ -29,6 +29,52 @@ N_CURRENT_YUNOHOST = 4
 
 N_NEXT_DEBAN = 11
 N_NEXT_YUNOHOST = 11
+
+VENV_BACKUP_PREFIX= "BACKUP_VENV_"
+VENV_REQUIREMENTS_SUFFIX= "_req.txt"
+
+def _get_all_venvs():
+    result = []
+    exclude = glob.glob(f"/opt/{VENV_BACKUP_PREFIX}*")
+    for x in glob.glob('/opt/*'):
+        if x not in exclude and os.path.isdir(x) and os.path.isfile(f"{x}/bin/activate"):
+            content = read_file(f"{x}/bin/activate")
+            if "VIRTUAL_ENV" and "PYTHONHOME" in content:
+                result.append(x)
+    return result
+
+def _generate_requirements():
+
+    venvs = _get_all_venvs()
+    for venv in venvs:
+        # Generate a requirements file from venv
+        os.system(f"bash -c 'source {venv}/bin/activate && pip freeze > {venv}{VENV_REQUIREMENTS_SUFFIX} && deactivate'")
+
+
+def _rebuild_venvs():
+
+    venvs = _get_all_venvs()
+    for venv in venvs:
+        venvdirname = venv.split("/")[-1]
+        # Create a backup of the venv, in case there's a problem
+        if os.path.isdir(f"/opt/{VENV_BACKUP_PREFIX}{venvdirname}"):
+            rm(f"/opt/{VENV_BACKUP_PREFIX}{venvdirname}", recursive=True)
+        backup = True
+        try:
+            cp(venv, f"/opt/{VENV_BACKUP_PREFIX}{venvdirname}", recursive=True)
+        except:
+            backup = False
+        if backup and os.path.isfile(venv+VENV_REQUIREMENTS_SUFFIX):
+            # Recreate the venv
+            rm(venv, recursive=True)
+            os.system(f"python -m venv {venv}")
+            status = os.system(f"bash -c 'source {venv}/bin/activate && pip install -r {venv}{VENV_REQUIREMENTS_SUFFIX} && deactivate'")
+            if status!=0:
+                logger.warning(m18n.n("venv_regen_failed", venv=venv))
+            else:
+                rm(venv+VENV_REQUIREMENTS_SUFFIX, recursive=True)
+        else:
+            logger.warning(m18n.n("venv_regen_failed", venv=venv))
 
 
 class MyMigration(Migration):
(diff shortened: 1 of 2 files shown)


zamentur:
Venv could be in /opt /opt/yunohost /var/www (or in sub subdir of those paths)

Here are some specific stuff:
- homesassistant use a specific python version
  - https://github.com/YunoHost-Apps/homeassistant_ynh/blob/master/scripts/_common.sh#L66
- jupiterlab use pipenv (don't know if it's important)
  - https://github.com/YunoHost-Apps/jupyterlab_ynh/blob/master/scripts/install#L103
- kresus use virtualenv
  - https://github.com/YunoHost-Apps/kresus_ynh/blob/master/scripts/install#L129
- seafile doesn't use a virtualenv x_x
  - https://github.com/YunoHost-Apps/seafile_ynh/blob/master/scripts/install
- weblate do strange things
  - https://github.com/YunoHost-Apps/weblate_ynh/blob/master/scripts/install#L158

tituspijean:
> seafile doesn't use a virtualenv x_x 

Erf, it should though. If I recall correctly, it already induced issues with system-level packages. :/

> weblate do strange things 

It does strange things, but they look like to be documented upstream, and they are done in a [virtualenv]
(https://github.com/YunoHost-Apps/weblate_ynh/blob/5ea9dc1ede5faf0b4eefe2f69898238b56ad2d83/scripts/install#L145), so it seems alright. 😇

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,59 @@
 N_NEXT_DEBAN = 11
 N_NEXT_YUNOHOST = 11
 
+VENV_BACKUP_SUFFIX = "_BACKUP_VENV"
+VENV_REQUIREMENTS_SUFFIX = "_req.txt"
+VENV_IGNORE = "VENVNOREGEN"
```
zamentur, line 35:
```suggestion
VENV_IGNORE = "ynh_migration_no_regen"
```

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,59 @@
...
+            if level<maxlevel:
+                result+=_get_all_venvs(path,level=level+1)
+    return result
+
+def _generate_requirements():
+
+    venvs = _get_all_venvs("/opt/")+_get_all_venvs("/var/www/")
+    for venv in venvs:
+        # Generate a requirements file from venv
+        os.system(f"bash -c 'source {venv}/bin/activate && pip freeze > {venv}{VENV_REQUIREMENTS_SUFFIX} && deactivate'")
```
zamentur, line 59: I am not totally sure, but i think this is better:
```suggestion
        os.system(f"{venv}/bin/pip freeze > {venv}{VENV_REQUIREMENTS_SUFFIX}")
```

Missing try except clause ?

zamentur: What if we have no space left on device ?

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,59 @@
...
+    result = []
+    for file in os.listdir(dir):
+        path = os.path.join(dir,file)
+        if os.path.isdir(path):
+            if os.path.isfile(os.path.join(path,VENV_IGNORE)):
+                continue
+            activatepath = os.path.join(path,"bin","activate")
+            if os.path.isfile(activatepath):
+                content = read_file(activatepath)
+                if "VIRTUAL_ENV" and "PYTHONHOME" in content:
```
zamentur, line 47:
```suggestion
                if "VIRTUAL_ENV" and "PYTHONHOME" in content:
```
bool("VIRTUAL_ENV") -> True, so if content is equals to "PYTHONHOME" this condition is true

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,59 @@
...
+                    result.append(path)
+                    continue
+            if level<maxlevel:
+                result+=_get_all_venvs(path,level=level+1)
+    return result
+
+def _generate_requirements():
+
+    venvs = _get_all_venvs("/opt/")+_get_all_venvs("/var/www/")
```
zamentur, line 56: And what about /opt/yunohost/APP/venv ? I am not sure it works with a max_level =2

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,59 @@
 N_NEXT_DEBAN = 11
 N_NEXT_YUNOHOST = 11
 
+VENV_BACKUP_SUFFIX = "_BACKUP_VENV"
+VENV_REQUIREMENTS_SUFFIX = "_req.txt"
+VENV_IGNORE = "ynh_migration_no_regen"
+
+def _get_all_venvs(dir,level=0,maxlevel=2):
```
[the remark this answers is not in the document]
theo-is-taken: Added in https://github.com/YunoHost/yunohost/pull/1479/commits/a440afe8ebc8d9fe3ce02f3f6da40dfcae505027

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
zamentur: To be sure we keep the unchange version

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,60 @@
...
+
+    venvs = _get_all_venvs("/opt/")+_get_all_venvs("/var/www/")
+    for venv in venvs:
+        # Create a backup of the venv, in case there's a problem
+        if os.path.isdir(venv+VENV_BACKUP_SUFFIX):
+            rm(venv+VENV_BACKUP_SUFFIX, recursive=True)
+        backup = True
+        try:
+            cp(venv, venv+VENV_BACKUP_SUFFIX, recursive=True)
+        except:
```
zamentur, line 73: You probably should declare which errot you try catch

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,60 @@
...
+            backup = False
+        if backup and os.path.isfile(venv+VENV_REQUIREMENTS_SUFFIX):
+            # Recreate the venv
+            rm(venv, recursive=True)
+            os.system(f"python -m venv {venv}")
+            status = os.system(f"bash -c 'source {venv}/bin/activate && pip install -r {venv}{VENV_REQUIREMENTS_SUFFIX} && deactivate'")
+            if status!=0:
+                logger.warning(m18n.n("venv_regen_failed", venv=venv))
+            else:
+                rm(venv+VENV_REQUIREMENTS_SUFFIX, recursive=True)
```
zamentur, line 83:
```suggestion
                rm(venv+VENV_REQUIREMENTS_SUFFIX)
```

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
theo-is-taken: Fixed in https://github.com/YunoHost/yunohost/pull/1479/commits/fc0266a62e8c1e9658b3829f15b3f5dbca03e06e

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
theo-is-taken: Fixed in https://github.com/YunoHost/yunohost/pull/1479/commits/b7b4dbfcdff2264011e1b247991bf15a2587a362

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
theo-is-taken: Fixed in https://github.com/YunoHost/yunohost/pull/1479/commits/cdd579080833bc77e78246f78dbb16762c91fe14

zamentur:
> It does strange things, but they look like to be documented upstream, and they are done in a [virtualenv](https://github.com/YunoHost-Apps/weblate_ynh/blob/5ea9dc1ede5faf0b4eefe2f69898238b56ad2d83/scripts/install#L145), so it seems alright. innocent

Yes indeed, but the issue is our code can't manage this kind of things properly i think. So it could be possible for jupiterlab and weblate to need to force upgrade like today.

Josue-T:
Hello,
At my point of view it's quite really difficult to manage this like that. In the case of my app there are many specific case which can't be managed by the core automatically. By example in pgadmin:
- In ARM we use a prebuild venv because if we build the venv from scratch it took about 3 hours to install all dependency.
- For some venv we need to install rust and we need to pass the correct user and environment variable to build correctly the venv.
- For some venv we need to install/upgrade some package before the other one because of a build dependency.

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,67 @@
...
+
+def _generate_requirements():
+    """
+        Generate a requirements file for all python virtual env located inside /opt/ and /var/www/
+    """
+
+    venvs = _get_all_venvs("/opt/") + _get_all_venvs("/var/www/")
+    for venv in venvs:
+        # Generate a requirements file from venv
+        os.system(f"bash -c 'source {venv}/bin/activate && pip freeze > {venv}{VENV_REQUIREMENTS_SUFFIX} && deactivate'")
```
alexAubin, line 72:
```suggestion
        os.system(f"{venv}/bin/pip freeze > {venv}{VENV_REQUIREMENTS_SUFFIX}")
```

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,67 @@
...
+        After the update, recreate a python virtual env based on the previously generated requirements file
+    """
+
+    venvs = _get_all_venvs("/opt/") + _get_all_venvs("/var/www/")
+    for venv in venvs:
+        if os.path.isfile(venv + VENV_REQUIREMENTS_SUFFIX):
+            # Recreate the venv
+            rm(venv, recursive=True)
+            os.system(f"python -m venv {venv}")
+            status = os.system(f"bash -c 'source {venv}/bin/activate && pip install -r {venv}{VENV_REQUIREMENTS_SUFFIX} && deactivate'")
```
alexAubin, line 86:
```suggestion
            status = os.system(f"{venv}/bin/pip install -r {venv}{VENV_REQUIREMENTS_SUFFIX}")
```

alexAubin: Also ideally, would be nice to run this with `subprocess` somehow (or even `call_async_output` like we do for app scripts /hooks if i remember correctly), otherwise the stdout/stderr of the command is not gonna be logged and this is gonna be hell to debug

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,67 @@
...
+
+def _rebuild_venvs():
+    """
+        After the update, recreate a python virtual env based on the previously generated requirements file
+    """
+
+    venvs = _get_all_venvs("/opt/") + _get_all_venvs("/var/www/")
+    for venv in venvs:
+        if os.path.isfile(venv + VENV_REQUIREMENTS_SUFFIX):
+            # Recreate the venv
```
alexAubin, line 83: Possibly let's add a `logger.info` here

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,47 @@
...
+                content = read_file(activatepath)
+                if ("VIRTUAL_ENV" in content) and ("PYTHONHOME" in content):
+                    result.append(path)
+                    continue
+            if level < maxlevel:
+                result += _get_all_venvs(path, level=level + 1)
+    return result
+
+
+def _generate_requirements():
```
alexAubin, line 64:
```suggestion
def _backup_pip_freeze_for_python_app_venvs():
```

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -70,6 +111,12 @@ def run(self):
                 'wget --timeout 900 --quiet "https://packages.sury.org/php/apt.gpg" --output-document=- | gpg --dearmor >"/etc/apt/trusted.gpg.d/extra_php_version.gpg"'
             )
 
+        #
+        # Get requirements of the different venvs from python apps
+        #
+
+        _generate_requirements()
```
alexAubin, line 118:
```suggestion
        _backup_pip_freeze_for_python_app_venvs()
```

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,47 @@
 N_NEXT_DEBAN = 11
 N_NEXT_YUNOHOST = 11
 
+VENV_REQUIREMENTS_SUFFIX = ".requirements_backup_for_bullseye_upgrade.txt"
+VENV_IGNORE = "ynh_migration_no_regen"
```
alexAubin, line 34:
```suggestion
```

src/yunohost/data_migrations/0021_migrate_to_bullseye.py
```diff
@@ -30,6 +30,47 @@
...
+            maxlevel - the depth of the recursion
+            level - do not edit this, used as an iterator
+    """
+    # Using os functions instead of glob, because glob doesn't support hidden folders, and we need recursion with a fixed depth
+    result = []
+    for file in os.listdir(dir):
+        path = os.path.join(dir, file)
+        if os.path.isdir(path):
+            if os.path.isfile(os.path.join(path, VENV_IGNORE)):
+                continue
```
alexAubin, line 52:
```suggestion
```

PR #1487, opened 2022-08-10
[enh] Add level 0 bullseye apps to disclaimer

zamentur:
## The problem

It seems ffsync is not displayed in disclaimer for bullseye migration

## Solution

Check if the apps as a level 0 and warn in disclaimer.

## PR Status

Done (untested)

## How to test

...

### DIFF
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -2509,7 +2509,7 @@ def unstable_apps():
         if not infos.get("from_catalog") or infos.get("from_catalog").get("state") in [
             "inprogress",
             "notworking",
-        ] or infos["id"] in deprecated_apps:
+        ] or infos["id"] in deprecated_apps or infos.get("from_catalog").get("level") == 0:
             output.append(infos["id"])
 
     return output
--- a/src/yunohost/app.py
+++ b/src/yunohost/app.py
@@ -2506,10 +2506,10 @@ def unstable_apps():
 
     for infos in app_list(full=True)["apps"]:
 
-        if not infos.get("from_catalog") or infos.get("from_catalog").get("state") in [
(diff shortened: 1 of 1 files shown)


src/yunohost/app.py
```diff
@@ -2506,10 +2506,10 @@ def unstable_apps():
 
     for infos in app_list(full=True)["apps"]:
 
-        if not infos.get("from_catalog") or infos.get("from_catalog").get("state") in [
+        if not infos.get("from_catalog") or infos.get("from_catalog", {}).get("state") in [
             "inprogress",
             "notworking",
-        ] or infos["id"] in deprecated_apps:
+        ] or infos["id"] in deprecated_apps or infos.get("from_catalog", {}).get("level") == 0:
```
alexAubin, line 2512: Ugh yeah but sometimes some apps just get level 0 temporarily on the CI but are still actively maintained and alive, we need to differentiate those two ... In this code, as far as I remember I assumed that actually non-maintained app would get flagged as "notworking" in the catalog. The code implicitly check this because only working (and maybe inprogress? not sure) apps are listed in the actual "built" catalog.

Why not just add `ffsync` in the `deprecated_apps` list just a few lines before ?

zamentur: This `unstable_apps` function is used only by buster and bullseye migration. So if an app is level  in bullseye it could be a good idea to discourage migrating until the issues has been fix (in the case the person had to reinstall or force upgrade the app).

alexAubin:
Closing because as explained, the intended message is really about apps flagged as (longterm) notworking, not about apps that may temporarily be level 0 in the catalog, which doesn't really block the upgrade in itself. The initial issue is about ffsync, which should just be flagged as notworking in the catalog

PR #1502, opened 2022-09-04
11.1

alexAubin:
Currently: 

- [x] https://github.com/YunoHost/yunohost/pull/1477
- [x] https://github.com/YunoHost/yunohost/pull/1459
- [x] https://github.com/YunoHost/yunohost/pull/1289


src/utils/system.py
```diff
@@ -1,35 +1,105 @@
...
+        raise YunohostError(f"Failed to convert size {size} to float")
+
+    return int(size_ * factor[suffix])
+
+
+def binary_to_human(n: int) -> str:
+    """
+    Convert bytes or bits into human readable format with binary prefix
+    """
+    symbols = ("K", "M", "G", "T", "P", "E", "Z", "Y")
```
zamentur, line 86: The previous binary_to_human was supporting GiB notation. What should we do about that ?

zamentur:
- [ ] It seems that running `yunohost tools regenconf` trigger several "Warning: No config panel found." when we come from an instance without /etc/yunohost/settings.json even after the global settings migrations

alexAubin:
>It seems that running yunohost tools regenconf trigger several "Warning: No config panel found." 

Are you sure you copied the proper config_domain.toml and config_global.toml in your dev env ?

src/utils/system.py
alexAubin: Is that really important x_x

PR #1507, opened 2022-09-13
edits to locales/en.json

mooseyboots:


alexAubin:
Thanks, this is great !

As you figured out, that string is in the yunohot-admin repo, not the "core" yunohost repo ;) (cf your other PR)

PR #1509, opened 2022-10-01
Able to replace sources

yalh76:
## The problem

Sometimes during upgrade of an app we need to completly remove previous sources before installing the new ones

Some examples:

- https://github.com/YunoHost-Apps/agora_ynh/blob/3ce3213030bcc533205259c7737df4c213bbd6f7/scripts/upgrade#L71-L83
- https://github.com/YunoHost-Apps/chuwiki_ynh/blob/098937aadc6495bed4ff5858cd09bcbde3f3050e/scripts/upgrade#L74-L90
- https://github.com/YunoHost-Apps/shaarli_ynh/blob/445198c4c555d6a5adedfc54b9112264f2559328/scripts/upgrade#L74-L96
- https://github.com/YunoHost-Apps/discourse_ynh/blob/ec974723113fa8f984af8ab47d1cc99765ae6c2f/scripts/upgrade#L108-L174
- https://github.com/YunoHost-Apps/focalboard_ynh/blob/76ae13bb2e4c5bc4efde3d3081c92282c35ce405/scripts/upgrade#L84-L104

And as --keep is already implemented, we have a simple way to do so

## Solution

add an option for the ynh_setup_source helper to remove previous installed sources

## PR Status

Ready

## How to test

...

### DIFF
--- a/helpers/utils
+++ b/helpers/utils
@@ -63,10 +63,11 @@ ynh_abort_if_errors() {
 
 # Download, check integrity, uncompress and patch the source from app.src
 #
-# usage: ynh_setup_source --dest_dir=dest_dir [--source_id=source_id] [--keep="file1 file2"]
+# usage: ynh_setup_source --dest_dir=dest_dir [--source_id=source_id] [--keep="file1 file2"] [--replace]
 # | arg: -d, --dest_dir=    - Directory where to setup sources
 # | arg: -s, --source_id=   - Name of the source, defaults to `app`
 # | arg: -k, --keep=        - Space-separated list of files/folders that will be backup/restored in $dest_dir, such as a config file you don't want to overwrite. For example 'conf.json secrets.json logs/'
+# | arg: -r, --replace=     - Remove previous sources before installing new sources
 #
 # This helper will read `conf/${source_id}.src`, download and install the sources.
 #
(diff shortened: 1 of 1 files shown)


alexAubin (approved):

alexAubin:
@yalh76 : if my previous commit is ok for you i guess we can merge in 11.0 for the next minor release

helpers/utils
```diff
@@ -172,6 +175,10 @@ ynh_setup_source() {
         done
     fi
 
+    if [ "$full_replace" -eq 1 ]; then
+        ynh_secure_remove --file="$dest_dir"
```
zamentur, line 179: Should we test if $dest_dir is a mount point ?

If we delete the mount point and recreate the dir, the sources could be installed on the bad disk/partition.

alexAubin: I suppose, but that doesn't really look like a regression from this specific commit.

We have similar behavior when eg. failed upgrade will trigger a remove + backup restore, and the remove itself also include an `ynh_secure_remove` ...

yalh76:
> @yalh76 : if my previous commit is ok for you i guess we can merge in 11.0 for the next minor release

I'm fully ok with your commit

tituspijean (approved):

Tagadda (approved):

PR #1514, opened 2022-10-07
call _get_parent_domain_of to ensure getting any level parent

Axolotle:
also set default `return_self` to `False` since only one call needs this

### DIFF
--- a/src/dns.py
+++ b/src/dns.py
@@ -41,6 +41,7 @@ from yunohost.domain import (
     _get_domain_settings,
     _set_domain_settings,
     _list_subdomains_of,
+    _get_parent_domain_of,
 )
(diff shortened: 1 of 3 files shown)


src/dns.py
```diff
@@ -512,17 +513,19 @@ def _get_registrar_config_section(domain):
 
     from lexicon.providers.auto import _relevant_provider_for_domain
 
-    registrar_infos = {}
+    registrar_infos = {
+        "name": "Registrar infos",
```
alexAubin, line 517:
```suggestion
        "name": m18n.n('registrar_infos'),   # This is meant to name the config panel section, for proper display in the webadmin
```

alexAubin (approved):

PR #1518, opened 2022-10-12
Not need to force $YNH_APP_BASEDIR

yalh76:
## The problem

Able to define a specific path for some files ie outside of `/conf`

## Solution

ynh_add_config directly handle the `$YNH_APP_BASEDIR/conf/` at https://github.com/YunoHost/yunohost/blob/551095aa7098c2ad4a8dbc9348cf80e4b5400f83/helpers/utils#L364-L370

and allow to define a specific path

So removing it from the helpers

## PR Status

Ready

## How to test

...


Tagadda (approved):
:rocket:

PR #1523, opened 2022-10-27
Fix variable not found when using appv2

ericgaspar:
## The problem

when using `ynh_add_fpm_config` and `ynh_install_composer` we get the error `$finalpath` and `$final_path` not found.

## Solution

Replace by the new variable used in manifest v2: `$install_dir`

## PR Status

...

## How to test

...

### DIFF
--- a/helpers/php
+++ b/helpers/php
@@ -156,7 +156,7 @@ ynh_add_fpm_config() {
 user = __APP__
 group = __APP__
 
-chdir = __FINALPATH__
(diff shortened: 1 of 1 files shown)


alexAubin:
Eeeeer unfortunately that's a bit more complex than this because for things to be backward compatible we need to use `$install_dir` only in app v2 context, but v1 apps will still use/need `$final_path` @_@

ericgaspar:
Yes, I knew it was too easy...

PR #1526, opened 2022-11-03
Enh apps v2

Axolotle:
## The problem
The web-admin needs to access some new app v2 data and behavior

## Solution

- `app_catalog` with new arg `--with-antifeatures|-a` to get antifeatures list
- inject `quality`, `antifeatures`, `potential_alternative_to` values from the catalog in `_extract_app_from_gitrepo()`
- add requirements checks in `app_install` and `app_upgrade`:
  - fails if not met except for ram that will ask for confirmation
- add `requirements` dict with requirements tests result ("ok" or "{error}" in cli and more detailed in web-admin) to `app_manifest`
- also add a base64 screenshot for api in `app_manifest`
- display notifications in `app_install` and `app_upgrade`:
  - add simple confirm for acknowledgment  in cli for `pre` and `post`
  - return post_install notification hydrated with setting in api
- override manifest["name"] with given label in `app_install`

## PR Status
Testable i guess!

`post_upgrade` notifications for api is hydrated with before upgrade settings though, since it is loop.
Could we go for a single app arg only in `app_upgrade`?

## How to test
run `yunohost app catalog -af` 
run `yunohost app manifest <app>`
run `yunohost tools update apps`
run `yunohost tools upgrade apps` or `yunohost app upgrade <app>`

and web-admin equivalent with code from https://github.com/YunoHost/yunohost-admin/pull/486


src/app.py
```diff
@@ -783,11 +821,40 @@ def app_manifest(app):
...
+        import base64
+
+        manifest["image"] = None
+        screenshots_folder = os.path.join(extracted_app_folder, "doc", "screenshots")
+
+        if os.path.exists(screenshots_folder):
+            with os.scandir(screenshots_folder) as it:
+                for entry in it:
+                    ext = os.path.splitext(entry.name)[1].replace(".", "").lower()
+                    if entry.is_file() and ext in ("png", "jpg", "jpeg", "webp"):
```
alexAubin, line 838:
```suggestion
                    if entry.is_file() and ext in ("png", "jpg", "jpeg", "webp", "gif"):
```

There's also a bunch of apps shipping a .gif ... do you think those may work too on the webadmin ?

src/app.py
```diff
@@ -783,11 +821,40 @@ def app_manifest(app):
...
 
-    shutil.rmtree(extracted_app_folder)
-
     raw_questions = manifest.get("install", {}).values()
     manifest["install"] = hydrate_questions_with_choices(raw_questions)
 
+    # Add a base64 image to be displayed in web-admin
+    if Moulinette.interface.type == "api":
+        import base64
+
```
alexAubin, line 830: Sooo I was thinking about maybe having another route `GET /app/screenshots/whatever` to fetch screenshot asynchronously on the webadmin, otherwise this may be kinda slow on bad internet connection ... but after thinking about it, this would probably mean that YunoHost would download the app *twice* unless we implement some smarter cache system :|

I did try to add a check in the package linter to recommend to limit the size of the screenshots/ folder to 512k, though this was only an "Info" and I see quite a lot of apps with kinda large files. I'm switching this to a Warning now ... + add a check on allowed file types

I would still add a `with_screenshot` option to `app_manifest()` though, because it's likely to be called by other functions in the code and you don't want to spend time encoding base64 screenshots for these, that's only useful for when the webadmin displays the install form / catalog view. Watcha think ? :grimacing:

src/app.py
```diff
@@ -883,7 +939,17 @@ def app_install(
     app_id = manifest["id"]
 
     # Check requirements
-    _check_manifest_requirements(manifest, action="install")
+    for name, check, values, err in _check_manifest_requirements(
+        manifest, action="install"
+    ):
+        if not check:
```
alexAubin, line 945:
```suggestion
    for name, passed, values, err in _check_manifest_requirements(
        manifest, action="install"
    ):
        if not passed:
```

#SemanticObsession

src/app.py
```diff
@@ -783,11 +821,40 @@ def app_manifest(app):
...
+
+    shutil.rmtree(extracted_app_folder)
+
+    manifest["requirements"] = {}
+    for name, check, values, err in _check_manifest_requirements(
+        manifest, action="install"
+    ):
+        if Moulinette.interface.type == "api":
+            manifest["requirements"][name] = {
+                "pass": check,
```
alexAubin, line 852:
```suggestion
    for name, passed, values, err in _check_manifest_requirements(
        manifest, action="install"
    ):
        if Moulinette.interface.type == "api":
            manifest["requirements"][name] = {
                "pass": passed,
```

#SemanticObsession

src/app.py
```diff
@@ -783,11 +821,40 @@ def app_manifest(app):
...
+                "values": values,
+            }
+        else:
+            manifest["requirements"][name] = "ok" if check else m18n.n(err, **values)
```
alexAubin, line 856:
```suggestion
            manifest["requirements"][name] = "ok" if passed else m18n.n(err, **values)
```

#SemanticObsession

src/app.py
```diff
@@ -587,7 +588,29 @@ def app_upgrade(app=[], url=None, file=None, force=False, no_safety_backup=False
                     upgrade_type = "UPGRADE_FULL"
 
         # Check requirements
-        _check_manifest_requirements(manifest, action="upgrade")
+        for name, check, values, err in _check_manifest_requirements(
+            manifest, action="upgrade"
+        ):
+            if not check:
```
alexAubin, line 594:
```suggestion
        for name, passed, values, err in _check_manifest_requirements(
            manifest, action="upgrade"
        ):
            if not passed:
```

#SemanticObsession

src/app.py
```diff
@@ -783,11 +821,40 @@ def app_manifest(app):
...
-    shutil.rmtree(extracted_app_folder)
-
     raw_questions = manifest.get("install", {}).values()
     manifest["install"] = hydrate_questions_with_choices(raw_questions)
 
+    # Add a base64 image to be displayed in web-admin
+    if Moulinette.interface.type == "api":
+        import base64
+
+        manifest["image"] = None
```
alexAubin, line 831:
```suggestion
        manifest["screenshots"] = None
```

Sounds odd to to name this image instead of screenshots ? :thinking: 

(In the future we may have other images such as the logo or icon so can be nice to prevent that confusion)

src/app.py
```diff
@@ -783,11 +821,40 @@ def app_manifest(app):
...
+        screenshots_folder = os.path.join(extracted_app_folder, "doc", "screenshots")
+
+        if os.path.exists(screenshots_folder):
+            with os.scandir(screenshots_folder) as it:
+                for entry in it:
+                    ext = os.path.splitext(entry.name)[1].replace(".", "").lower()
+                    if entry.is_file() and ext in ("png", "jpg", "jpeg", "webp"):
+                        with open(entry.path, "rb") as img_file:
+                            data = base64.b64encode(img_file.read()).decode("utf-8")
+                            manifest["image"] = f"data:image/{ext};charset=utf-8;base64,{data}"
```
alexAubin, line 841:
```suggestion
                            manifest["screenshots"] = f"data:image/{ext};charset=utf-8;base64,{data}"
```

alexAubin (commented):
Sounds pretty cool :+1: 

I have a bunch of minor comments

My biggest interrogation is with the screenshot stuff, but the current implementation is probably fine modulo adding a `--with-screenshots` option to `app_manifest` ?

src/app.py
```diff
@@ -2737,3 +2853,67 @@ def replace_alias(service):
...
+    Keyword argument:
+        question -- m18n key or string
+        params -- dict of values passed to the string formating
+        kind -- "hard": ask with "Yes, I understand", "soft": "Y/N", "simple": "press enter"
+        force -- Will not ask for confirmation
+
+    """
+    if force or Moulinette.interface.type == "api":
+        return
+
```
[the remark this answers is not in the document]
alexAubin: (This bit is pretty important because in many context apps may be installed in a non-interactive context and you don't want to miserably crash because of this)

src/app_catalog.py
```diff
@@ -65,28 +65,38 @@ def app_catalog(full=False, with_categories=False):
                 "description": infos["manifest"]["description"],
                 "level": infos["level"],
             }
-        else:
-            infos["manifest"]["install"] = _set_default_ask_questions(
-                infos["manifest"].get("install", {})
-            )
```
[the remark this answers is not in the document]
Axolotle: Yes, app_catalog is only used to display the simple card list, app_manifest is used to display almost everything now

src/tools.py
```diff
@@ -365,6 +372,21 @@ def is_legit_warning(m):
...
+            current_version = version.parse(app["current_version"])
+            app["notifications"] = {
+                type_: _filter_and_hydrate_notifications(
+                    manifest["notifications"][type_], current_version, app["settings"]
+                )
+                for type_ in ("pre_upgrade", "post_upgrade")
+            }
+            # FIXME Post-upgrade notifs should be hydrated with post-upgrade app settings
+            del app["settings"]
+            shutil.rmtree(extracted_app_folder)
```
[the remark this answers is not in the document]
Axolotle: I put this in a `_list_upgradable_apps()` with pre_upgrade notif only!

src/app.py
```diff
@@ -2352,72 +2442,98 @@ def _get_all_installed_apps_id():
...
 
-    # Yunohost version requirement
+    # Packaging format
+    yield (
+        "packaging_format",
+        manifest["packaging_format"] in (1, 2),
+        {},
+        "app_packaging_format_not_supported",  # i18n: app_packaging_format_not_supported
+    )
+
```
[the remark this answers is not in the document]
Axolotle: Hum, didn't understand if there was something to do about this?

src/app.py
Axolotle: Added `with_screenshot` in actionmap!

src/app.py
Axolotle: renamed to `screenshot` singular

Axolotle:
Thx! Should have fixed most of your comments and reflect changes to the web-admin.
I did the ugly return of post_upgrade notif in API mode

Remaining:
- disk space edge case
- markdown for CLI (maybe later)
- include_swap thing

src/app.py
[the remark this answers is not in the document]
alexAubin: Yeah imho we should keep the initial code with `raise YunohostValidationError`, because if `packaging_format` is for example 3, it's safe to assume that the rest of the code *is* gonna crash because there's no guarantee that for example the `integration` key will still exist (or other keys/code assumptions)

alexAubin (approved):

PR #1529, opened 2022-11-07
Add support for Webgo Registrar DNS

mod242:
## The problem

Recently added Provider "webgo" in lexicon-api leads to an stacktrace in yunohost (testing) when adding a domain which is registered there, since it is found by the lexicn-api, but is not in the yunohost registrar_list.
(I am also the author of the webgo provider for the lexicon-api)

## Solution

Add webgo Provider into registrar_list

## PR Status

Done

## How to test

I tested the whole implementation with about 20 Domains on webgo without any issue.


alexAubin:
Sounds legit, cheers !

PR #1532, opened 2022-11-16
Fix docker-image-extract script name

tituspijean:
## The problem

"File not found error" when using the `ynh_setup_source` helper in docker mode.

## Solution

Rename the vendor script file appropriately.
And now I realize the current script is quite different as the source material, though the parameters usage fits.

## PR Status

...

## How to test

...


alexAubin:
Ah my bad, but shouldnt we fix it in the other way around (= fix the call of that script) ? The idea was to keep the folder to be pretty much equal to a git clone

tituspijean:
Yeah buuuuuut... https://github.com/jjlin/docker-image-extract/commit/95b3db8af8e10b852d1d5121426d9bf91b8aae45

¯\_(ツ)_/¯

alexAubin:
Uuuuh indeed ? How drunk was I D:

PR #1533, opened 2022-11-17
Create sysctl.conf

jookk:
I enabled this in systl.conf for better server security.

## The problem

...

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/debian/sysctl.conf
+++ b/debian/sysctl.conf
@@ -1,0 +1,68 @@
+#
+# /etc/sysctl.conf - Configuration file for setting system variables
+# See /etc/sysctl.d/ for additional system variables.
+# See sysctl.conf (5) for information.
+#
+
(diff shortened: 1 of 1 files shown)


alexAubin (changes requested):
In its current state, this PR does nothing as the new `sysctl.conf` file is not actually installed anywhere

Also we probably dont want to just yoloreplace any existing `sysctl.conf` file so a smart strategy should be designed

It would also be nice to have more info on the enabled setting (I see `rp_filter`, `tcp_syncookies`, `log_martians`) and their implications (in particular any possible drawbacks to enabling these?) and explanations on why enabling these for ipv4 but not v6

alexAubin:
Closing due to lack of feedback and cf previous comments

PR #1538, opened 2022-11-24
be able to change the loginShell of a user

kay0u:
## The problem

We can't use another shell like `sh` or `zsh`.

## Solution

Add the possibility to change it

## PR Status

Yolotested in prod.

We should validate the shell before setting it.
Maybe we should manage the moment when the shell is removed from the system,  but meh, it's not simple.

## How to test

...

### DIFF
--- a/src/user.py
+++ b/src/user.py
@@ -134,6 +134,7 @@ def user_create(
     lastname=None,
     mailbox_quota="0",
     admin=False,
+    loginShell="/bin/bash",
     from_import=False,
 ):
 
@@ -253,7 +254,7 @@ def user_create(
         "gidNumber": [uid],
         "uidNumber": [uid],
         "homeDirectory": ["/home/" + username],
-        "loginShell": ["/bin/bash"],
+        "loginShell": [loginShell],
     }
 
     try:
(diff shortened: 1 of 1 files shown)


src/user.py
```diff
@@ -122,6 +122,27 @@ def display_default(values, _):
...
+    libc = ctypes.CDLL(ctypes.util.find_library("c"))
+    getusershell = libc.getusershell
+    getusershell.restype = ctypes.c_char_p
+    libc.setusershell()
+    while True:
+        shell = getusershell()
+        if not shell:
+            break
+        yield shell.decode()
+    libc.endusershell()
```
alexAubin, line 139: Uuuuh I guess that works but that looks pretty complex ? :thinking: 

Why not just read `/etc/shells` directly ? (I guess we just need a trick to ignore comments / empty lines ?)

kay0u: done :)

src/user.py
```diff
@@ -123,6 +123,23 @@ def display_default(values, _):
...
+def list_shells():
+    with open("/etc/shells", "r") as f:
+        content = f.readlines()
+
+    shells = []
+    for line in content:
+        if line.startswith("/"):
+            shells.append(line.replace("\n",""))
+    return shells
+
```
alexAubin, line 135:
```suggestion
    return [line.strip() for line in content if line.startswith("/")]
```

PR #1546, opened 2022-12-06
[fix] hotspot config panel fails in webadmin

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/2099

## Solution

1. Use future_values instead of new_values (new_values includes only changed new values, instead of future_values, that includes all values updated with the already answered questions)
2. After hydrating the questions, make sure to overwrite each self.values[OPTION_ID] by a simple value and not a structured data

## PR Status

Ready (just tested on a server by copying the change manually not with ynh-dev)

## How to test

Install hotspot and try to activate an hotspot

### DIFF
--- a/src/utils/config.py
+++ b/src/utils/config.py
@@ -651,10 +651,19 @@ class ConfigPanel:
                         raw_msg=True,
                     )
             value = self.values[option["name"]]
+            
+            # Allow to use value instead of current_value in app config script.
+            # For example hotspot used it...
+            # See https://github.com/YunoHost/yunohost/pull/1546
+            if isinstance(value, dict) and "value" in value and "current_value" not in value:
+                value["current_value"] = value["value"]
+            
             # In general, the value is just a simple value.
             # Sometimes it could be a dict used to overwrite the option itself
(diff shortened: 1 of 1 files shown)


zamentur:
SO the issue seems data are given like `{'no_antenna': {'value': False}}` instead of `{'no_antenna': False` AND i don't know exactly why for the moment.

zamentur:
SO the reason of this notation is {'value': XXXX} appears when a config script return a dict in order to change other property of the question. That's the case of hotspot and probably of vpnclient.

So just give self.new_values or self.future_values is quite risky i guess o_O.

src/utils/config.py
```diff
@@ -651,10 +651,19 @@ def _hydrate(self):
                         raw_msg=True,
                     )
             value = self.values[option["name"]]
+            
+            # Allow to use value instead of current_value in app config script.
```
alexAubin, line 655:
```suggestion
            # Allow to use value instead of current_value in app config script.
            # e.g. apps may write `echo 'value: "foobar"'` in the config file (which is more intuitive that `echo 'current_value: "foobar"'`
```

PR #1552, opened 2022-12-20
Add postfix SASL login failure to a fail2ban jail

Ddataa:
in order to prevent those 
```
Aug 31 22:23:52 hostxyz postfix/smtpd[38697]: warning: unknown[192.168.xx.xx]: SASL LOGIN authentication failed: authentication failure
Aug 31 22:23:52 hostxyz postfix/smtpd[38697]: lost connection after AUTH from unknown[192.168.xx.xx]
```

## The problem

Too many login failures not taken into accounts by fail2ban

## Solution

Apply jail conf and filter conf

## PR Status

...

## How to test

...

### DIFF
--- a/conf/fail2ban/postfix-sasl.conf
+++ b/conf/fail2ban/postfix-sasl.conf
@@ -1,0 +1,6 @@
+# Fail2Ban filter for postfix authentication failures
+[INCLUDES]
+before = common.conf
+[Definition]
+_daemon = postfix/smtpd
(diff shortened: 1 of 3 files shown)


conf/fail2ban/postfix-sasl.conf
```diff
@@ -0,0 +1,6 @@
+# Fail2Ban filter for postfix authentication failures
```
alexAubin, line 1: Cheers ! :+1: 

I think for this fact to be effectively added to the configuration, you should add a similar line to what's here:

https://github.com/YunoHost/yunohost/blob/dev/hooks/conf_regen/52-fail2ban#L16

Ddataa: yep that is exactly what I was missing !!

Ddataa: thanks

alexAubin:
Cheers :clap:  :+1:

PR #1558, opened 2023-01-04
Update yunopaste

FabianWilkens:
## The problem

yunopaste is not working correctly.
When I try to use yunopaste I get no output.

## PR Status

...

## How to test

...

### DIFF
--- a/bin/yunopaste
+++ b/bin/yunopaste
@@ -19,7 +19,7 @@ paste_data() {
   [[ -z "$json" ]] && _die "Unable to post the data to the server."
 
   key=$(echo "$json" \
-    | python -c 'import json,sys;o=json.load(sys.stdin);print(o["key"])' \
+    | python3 -c 'import json,sys;o=json.load(sys.stdin);print(o["key"])' \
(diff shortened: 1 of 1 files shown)


tituspijean (approved):
```
# echo 'test' | yunopaste
https://paste.yunohost.org/asixewomiz
```
🚀

bin/yunopaste
```diff
@@ -19,7 +19,7 @@ paste_data() {
   [[ -z "$json" ]] && _die "Unable to post the data to the server."
 
   key=$(echo "$json" \
-    | python -c 'import json,sys;o=json.load(sys.stdin);print o["key"]' \
+    | python -c 'import json,sys;o=json.load(sys.stdin);print(o["key"])' \
```
alexAubin, line 22:
```suggestion
    | python3 -c 'import json,sys;o=json.load(sys.stdin);print(o["key"])' \
```

Let's be explicit and call python3 (just in case for some reason some people have `python-is-python2` installed idk)

alexAubin:
Cheers !

PR #1559, opened 2023-01-05
[fix] Enable pop3 settings make dovecot failed

zamentur:
## The problem

When we enable dovecot pop3 on sans-nuage.fr, the dovecot fail with error:
```
Fatal: service(pop3) access(/usr/lib/dovecot/pop3) failed: No such file or directory
```

## Solution

Install `dovecot-pop3d` package

## PR Status

Ready

## How to test

...

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -23,7 +23,7 @@ Depends: ${python3:Depends}, ${misc:Depends}
  , slapd, ldap-utils, sudo-ldap, libnss-ldapd, unscd, libpam-ldapd
  , dnsmasq, resolvconf, libnss-myhostname
  , postfix, postfix-ldap, postfix-policyd-spf-perl, postfix-pcre
- , dovecot-core, dovecot-ldap, dovecot-lmtpd, dovecot-managesieved, dovecot-antispam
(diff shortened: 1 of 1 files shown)


alexAubin:
The thing is that it's supposed to be installed here é_è : https://github.com/YunoHost/yunohost/blob/dev/src/settings.py#L350

Gredin67:
I also had to open the 995 port manually.
Here is the log of our chat :
```

sudo yunohost settings set pop3.enabled -v True
gives
Le paquet « dovecot-pop3d » n'est pas installé, et ne peut donc être supprimé
(still on 4.4.2.14, but might still be a bug)
gredin67
also port 995 does not seem to open
@Alekswag:matrix.org
zblerg ben le code est la je capte pas comment ça peut passer là ou y'a le remove ... https://github.com/YunoHost/yunohost/blob/buster/src/yunohost/settings.py#L458
a moins que la valeur soit convertie en autre chose entre temps
gredin67
ça a pas un lien avec le fait que True=yes et False=no ?
Aleks (he/him/il/lui)
peut-être
m'enfin apriori non parce qu'en buster les settings étaient pas géré par les config panel
mais ptete y'a un truc similaire
```

alexAubin:
Closing due to inactivity and this is not the right way to fix the issue ...

PR #1564, opened 2023-01-10
[fix] Missing migration

zamentur:
## The problem

SSOwat conf is not regen after the upgrade and `remote_user_var_in_nginx_conf` are missing

## Solution

Trigger once with a migration

Alternative solution not implemented: regen at each debian package upgrade ?

## PR Status

Opinion needed
Untested (i have an issue with my ynh-dev)

## How to test

...


alexAubin:
Fixed directly in debian's posinst because it was also needed to backport this on older versions

PR #1578, opened 2023-01-31
Output checksums if ynh_setup_source fails

tituspijean:
## The problem

Output checksums if ynh_setup_source fails during their verification, to help maintainers check if it is a bug or persistent source change.

## Solution

Make the error message a bit more verbose.

## PR Status

Untested in situ, but command works.

## How to test

...

### DIFF
--- a/helpers/utils
+++ b/helpers/utils
@@ -167,7 +167,7 @@ ynh_setup_source() {
             || ynh_die --message="$out"
         # Check the control sum
         echo "${src_sum} ${src_filename}" | ${src_sumprg} --check --status \
-            || ynh_die --message="Corrupt source for ${src_filename}: Expected ${src_sum} but got $(${src_sumprg} ${src_filename} | cut --delimiter=' ' --fields=1)."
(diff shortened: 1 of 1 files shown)


tituspijean:
It could be a bit smarter, here the checksum is calculated a second time to print it.

helpers/utils
```diff
@@ -167,7 +167,7 @@ ynh_setup_source() {
             || ynh_die --message="$out"
         # Check the control sum
         echo "${src_sum} ${src_filename}" | ${src_sumprg} --check --status \
-            || ynh_die --message="Corrupt source"
+            || ynh_die --message="Corrupt source for ${src_filename}: Expected ${src_sum} but got $(${src_sumprg} ${src_filename} | cut --delimiter=' ' --fields=1)."
```
alexAubin, line 170:
```suggestion
            || ynh_die --message="Corrupt source for ${src_filename}: Expected ${src_sum} but got $(${src_sumprg} ${src_filename} | cut --delimiter=' ' --fields=1) (size : $(du -hs ${src_filename} | cut --delimiter=' ' --fields=1))."
```

Proposing to also add the file size

tituspijean: Ah yeah nice, with that it's easy to see if we actually got a 404 or other error code instead of the actual file.

PR #1586, opened 2023-02-03
Fix composer workdir variable for package v2

ericgaspar:
## The problem

https://ci-apps-dev.yunohost.org/ci/job/5463

## Solution

Replace `$final_path` by `$install_dir`

## PR Status

...

## How to test

...


helpers/php
```diff
@@ -516,7 +516,7 @@ ynh_install_composer() {
     local composerversion
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
-    workdir="${workdir:-$final_path}"
+    workdir="${workdir:-$install_dir}"
```
alexAubin, line 519:
```suggestion
    if dpkg --compare-versions ${YNH_APP_PACKAGING_FORMAT:-0} lt 2; then
        workdir="${workdir:-$final_path}"
    else
        workdir="${workdir:-$install_dir}"
    fi
```

PR #1588, opened 2023-02-06
[fix] Allow to do replace string with @

zamentur:
## The problem
This command doesn't work:
```
ynh_replace_string "Toto" "Toto@titi"
```

## Solution
Replace the delimiter by something generally unused $'\001'

## PR Status

Waiting automatic test

## How to test

...


alexAubin:
Noice :O

PR #1589, opened 2023-02-06
[fix] Write var in files with redundants keys

zamentur:
## The problem

With file like:
```
{
    "tic": {
        "toto": "titi"
    },
    "tac": {
        "toto": "tutu"
    },
```
This command fails
```
ynh_write_var_in_file --file=config.json --key=toto --value=tata --after="tic"
```

Here the kind of error messages:
```
sed: -e expression #1, char 173: unterminated `s' command
```

It's because `head -n1` is deactivated by the `|| echo YNH_NULL` just before... So we find several value in the key and get newline inside, so the sed command doesn't finish cause we insert a newline...

## Solution

Use parenthesis to avoid that

## PR Status

Waiting automatic test
Tested manually on mattermost config json file

## How to test

...

### DIFF
--- a/helpers/utils
+++ b/helpers/utils
@@ -568,7 +568,7 @@ ynh_read_var_in_file() {
     var_part+='\s*'
 
     # Extract the part after assignation sign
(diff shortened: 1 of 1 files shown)


alexAubin:
Hmmmmokay not sure to understand the subtleties but sounds legit I guess :sweat_smile:

PR #1593, opened 2023-02-07
enh-PostgreSQL helpers for multidatabases handling

oleole39:
## The problem

Facing the case of the broken app package [Noalyss](https://github.com/YunoHost-Apps/noalyss_ynh) (accounting solution), it happens that this app would need more than one PostgreSQL database to work. As a matter of fact this particular app is designed to require: 
- 1 database to store global settings
- 1 database per accounting ruleset (2 by default, total number up to the user)
- 1 database per accounting project (total number up to the user)

All databases are created by the app via the dedicated rolename assigned to it (e.g. here "noalyss"). However, a script is needed to correctly backup, restore, and delete those multiple databases which only have in common the fact of being owned by the rolename associated to the app.

## Solution

4 helpers were created:
- `ynh_psql_list_user_dbs()` is an internal helper (meant to be used by the 3 others helpers below) which returns the list of databases associated to the specified PostgreSQL rolename as a string combining names separated with a space (it is assumed that no database name will ever contain space).
- `ynh_psql_remove_all_user_dbs()` loops within the list of databases associated to the given rolename and drops them all (it is assumed that no database name will ever contain space or asterisk).
- `ynh_psql_dump_all_user_dbs()` loops within the list of databases associated to the given rolename and dumps them all in the current folder, naming each file as a combination of app_id, db_name, and a "-dump.sql" suffix (it is assumed that no database name will ever contain space or asterisk).
- `ynh_psql_restore_all_app_dbs_dumps()` loops among all the files in the current folder with a filename ending with the suffix "-dump.sql" and restore the ones matching the app_id of the app being restored (it is assumed that the dumps to be restored follow the filename template defined in the previous helper).

## PR Status

Helpers are ready and [already implemented](https://github.com/oleole39/noalyss_ynh/blob/standard-install-fixed/scripts/_common.sh#L75) as functions limited to the Noalyss app's scope in the fixed package [currently being reviewed](https://github.com/YunoHost-Apps/noalyss_ynh/pull/39) for this app. As such they have been tested on a VPS running Debian and YNH 11.6 stable and worked as expected.

**The goal of this PR is to turn these functions into helpers so that they could be used by other apps.**
They have not been tested as such in a Yunohost custom build, since trying to create the required build environment happened to be complicated for my dev machine (no snap, limited disk space for a Virtualbox machine).

This PR would also take the [docs on helpers](https://yunohost.org/en/packaging_apps_helpers) to be regenerated. 
 
## How to test

Functions can be tested easily installing [this branch](https://github.com/oleole39/noalyss_ynh/tree/standard-install-fixed) of Noalyss package as a custom app - following install instructions till the end, then backup the app, remove it and eventually restore it and see that after restoration you still have access to the app with the credentials defined during the initial installation.  
Between each step you can list PostgreSQL databases on the test YNH server to note what is happening to them (not knowing YNH server's postgres user password, I have to log first as root as indicated below):
```
su -
sudo -u -i postgres
psql -l 
```
As global helpers, a custom build is required and call examples can be found in the backup/restore/remove scripts of the fixed Noalyss package's branch indicated hereinabove.

### DIFF
--- a/helpers/postgresql
+++ b/helpers/postgresql
@@ -143,6 +143,41 @@ ynh_psql_dump_db() {
     sudo --login --user=postgres pg_dump "$database"
 }
 
+# Dump all existing PostgreSQL databases associated with a given user
+#
+# usage: ynh_psql_dump_all_user_dbs --db_user=db_user [--app=app]
+# | arg: -u, --db_user=    - the PostgreSQL role/user of which to remove all owned databases
+# | arg: -a, --app=     - the application id to tag the dump with
+#
+# Requires YunoHost version 3.5.0 or higher.
+ynh_psql_dump_all_user_dbs() {
+    # Declare an array to define the options of this helper.
+    local legacy_args=ua
+    local -A args_array=([u]=db_user= [a]=app=)
+    local db_user
+    local app
+    # Manage arguments with getopts
+    ynh_handle_getopts_args "$@"
+    app="${app:-}"
+	
+	local dbs_to_dump=$(ynh_psql_list_user_dbs $db_user)
+	if [ -n "$dbs_to_dump" ]; then	   # Check that the list of database(s) is not empty 
+		
+		local db_name
+		for db_name in $dbs_to_dump   # Iterate through the list of database(s) to dump. Note: this would fail in case databases names would contain space character or characters such as "*". IFS parsing method would then be required. 
+		do
+			if ynh_psql_database_exists --database=$db_name; then # Check if the database exists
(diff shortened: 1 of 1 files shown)


zamentur:
> The goal of this PR is to turn these functions into helpers so that they could be used by other apps.

Thanks for this contribution. However, i suggest to wait for at least an other apps that need this specific behaviour.

Maybe we can set those helpers into experimental ones https://github.com/YunoHost-Apps/Experimental_helpers

alexAubin:
Been almost a year, sorry, closing because this is way too specific and so far there's only one example of app requiring multiple databases...

PR #1602, opened 2023-02-21
feat: add '--continue-on-failure' to 'yunohost app upgrade

Psycojoker:
## The problem

The problem is describe here https://github.com/YunoHost/issues/issues/2072

The user complains that if an app failed to upgrade on "yunohost app upgrade" you have to upgrade all apps by hand if you still want to upgrade the other ones.

## Solution

I've added a flag `--continue-on-failure`

The output looks like this (this is from the tests)

image

## PR Status

Ready to be review, I've wrote tests for it and the tests seems to demonstrate that this is working as expected.

Those test heavily uses the mock library to only test the expected workflow and not a real life testing as we don't care about really upgrading installed applications here but how the workflow behave when it thinks an application failed to upgrade.

## How to test

Launch the tests (`pytest src/tests/test_apps.py::TestMockedAppUpgrade`) or actually recreate a situation where you have several apps to upgrade and one failed to upgrade.


src/app.py
```diff
@@ -820,20 +821,26 @@ def app_upgrade(app=[], url=None, file=None, force=False, no_safety_backup=False
             # If upgrade failed or broke the system,
             # raise an error and interrupt all other pending upgrades
             if upgrade_failed or broke_the_system:
```
alexAubin, line 823: I'm wondering if we shouldnt still *not* continue if `broke_the_system is True` ... like, this means that either dpkg is broken, or nginx is broken, which has a 95% change of breaking the next upgrade (whereas the `upgrade_failed is True and broke_the_system is False` scenario is more likely to be specific to the app)

locales/en.json
```diff
@@ -27,6 +27,7 @@
     "app_config_unable_to_apply": "Failed to apply config panel values.",
     "app_config_unable_to_read": "Failed to read config panel values.",
     "app_extraction_failed": "Could not extract the installation files",
+    "app_failed_to_upgrade_but_continue": "App {failed_app} failed to upgrade, continue to next upgrades as requested. Run 'yunohost log show {operation_logger_name}' to see failure log",
```
alexAubin, line 30:
```suggestion
    "app_failed_to_upgrade_but_continue": "Application {failed_app} failed to upgrade, but YunoHost will continue with next app upgrades as requested. Run 'yunohost log show {operation_logger_name}' to see failure log",
```

locales/en.json
```diff
@@ -75,6 +76,8 @@
     "apps_catalog_obsolete_cache": "The app catalog cache is empty or obsolete.",
     "apps_catalog_update_success": "The application catalog has been updated!",
     "apps_catalog_updating": "Updating application catalog...",
+    "apps_failed_to_upgrade": "Those applications failed to upgrade:{apps}",
```
alexAubin, line 79:
```suggestion
    "apps_failed_to_upgrade": "Those applications failed to upgrade: {apps}",
```

locales/en.json
```diff
@@ -75,6 +76,8 @@
...
+    "apps_failed_to_upgrade_line": "\n * {app_id} (to see corresponding log do a 'yunohost log show {operation_logger_name}')",
```
alexAubin, line 80:
```suggestion
    "apps_failed_to_upgrade_line": "\n * {app_id} (to see corresponding error logs, use 'yunohost log show {operation_logger_name}')",
```

alexAubin (approved):

src/app.py
Psycojoker: Ah yeah make totally sens :)

locales/en.json
Psycojoker: I'm don't think you want that because this is merged with the next line en en.json that will add a "\n" (but tbf it won't changed anything accepting or not your modification)

This is this output:

image

1 - is this docstring
2 - is `"apps_failed_to_upgrade_line": "\n * {app_id} (to see corresponding log do a 'yunohost log show {operation_logger_name}')",`

Psycojoker:
I've taken into account the "broke_the_system" situation and added a new test for it

image

I've also added custom messages for UX reason.

PR #1605, opened 2023-02-21
regenconf: Add a new ssh security setting to hide the banner

milouse:
## The problem

I’m not very confident about advertizing the whole world what system is running on my server. Maybe it may look as a pedantic or useless feature, but if others are like me, this MR will avoid us to always be warned about configuration not managed by yunohost.

## Solution

Add a new `security.ssh.ssh_banner` as a boolean to comment or not the display of `/etc/issue.net`. By default it’s still yes to keep the current behavior.

## PR Status

...

## How to test

After installation and regenconf (and thus by default) the `sshd_config` file must still contain `Banner /etc/issue.net`

    $ grep Banner /etc/ssh/sshd_config
    Banner /etc/issue.net

Then enter `yunohost settings set security.ssh.ssh_banner -v no`. You must now have

    $ grep Banner /etc/ssh/sshd_config
    #Banner none

Finally ensure the "yes" value with `yunohost settings set security.ssh.ssh_banner -v yes`. You must have again

    $ grep Banner /etc/ssh/sshd_config
    Banner /etc/issue.net

I’ve added a comment as for the other ssh settings in the file to ease migration path for others like me willing to hide this banner.

### DIFF
--- a/conf/ssh/sshd_config
+++ b/conf/ssh/sshd_config
@@ -64,7 +64,13 @@ PasswordAuthentication no
 {% endif %}
 
 # Post-login stuff
+# PLEASE: if you wish to hide the system name and version when attempting to connect to your server, run this command:
+# yunohost settings set security.ssh.ssh_banner -v no
+{% if banner == "False" %}
+#Banner none
+{% else %}
 Banner /etc/issue.net
+{% endif %}
 PrintMotd no
 PrintLastLog yes
 ClientAliveInterval 60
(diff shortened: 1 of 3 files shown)


alexAubin:
Hmmm yeah, I'm actually wondering if we shouldn't simply have `Banner none` as default ... Like, I don't see the point of the default SSH banners just mentioning the OS's version ... I can't even find a proper answer on the internet about why SSH banners exists in the first place, except about displaying a legal warning to try to discourage attackers x_X

alexAubin:
Superseded by https://github.com/YunoHost/yunohost/commit/432a9ab544800782dcdaa0bef9ae84480c7d77ba, followup of previous discussion

PR #1610, opened 2023-02-24
Add default_character_set option for MySQL database

ericgaspar:
Add default_character_set option to MySQL database cf. kayou addition on Nextcloud: https://github.com/YunoHost-Apps/nextcloud_ynh/pull/384

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -334,7 +334,7 @@ class SourcesResource(AppResource):
 
     `autoupdate.strategy` is expected to be one of :
     - `latest_github_tag` : look for the latest tag (by sorting tags and finding the "largest" version). Then using the corresponding tar.gz url. Tags containing `rc`, `beta`, `alpha`, `start` are ignored, and actually any tag which doesn't look like `x.y.z` or `vx.y.z`
-    - `latest_github_release` : similar to `latest_github_tags`, but starting from the list of releases. Pre- or draft releases are ignored. Releases may have assets attached to them, in which case you can define:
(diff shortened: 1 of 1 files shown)


helpers/mysql
```diff
@@ -2,28 +2,37 @@
...
     local password
     local database
+    local default_character_set
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
     database="${database:-}"
+    default_character_set="${default_character_set:-}"
 
-    mysql --user="$user" --password="$password" --batch "$database"
+    if [ -n "$default_character_set" ]
```
zamentur, line 28: is there a way to get the default_charset from database itself in order to add this change in a generic way in the official helper ?

helpers/mysql
```diff
@@ -118,22 +127,38 @@ ynh_mysql_drop_db() {
...
+    local -A args_array=( [d]=database= [c]=default_character_set= )
     local database
+    local default_character_set
     # Manage arguments with getopts
     ynh_handle_getopts_args "$@"
+    default_character_set="${default_character_set:-}"
+    MYSQL_ROOT_PWD_FILE=/etc/yunohost/mysql
 
-    mysqldump --single-transaction --skip-dump-date --routines "$database"
+    if [ -n "$default_character_set" ]
```
zamentur, line 149: Same here i think we should find a generic way in the official helper to do this. Other apps or custom instance might have the issue (cause users have seen the utf8 warning, so they might have made the change in their DB)

helpers/mysql
ericgaspar: I don't know how to do that.

PR #1612, opened 2023-02-24
[fix] Pop3 is not enable in config file

zamentur:
## The problem

Dovecot config file is not change when activating pop3

Comparaison with"True" intead of "1"

## Solution

Replace "True" by "1"

## PR Status

...

## How to test

...


alexAubin:
Fixed in b40c0de3 which is the proper fix, cf other jinja files using True/False .. I don't trust the settings get output, it always seems to sometimes yield randomly either bool-like string or integers ...

PR #1613, opened 2023-02-24
fix: add missing translatable string log_remove_on_failed_restore

Psycojoker:
## The problem

Missing translatable string from log

## Solution

Add it

## PR Status

Honestly idk if that's the exact message we want to have? I have no idea of what it's supposed to be

## How to test

Have a log message with this key

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -516,6 +516,7 @@
     "log_permission_url": "Update URL related to permission '{}'",
     "log_regen_conf": "Regenerate system configurations '{}'",
     "log_remove_on_failed_install": "Remove '{}' after a failed installation",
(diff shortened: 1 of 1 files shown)


alexAubin:
Uuuh I was confused because I removed it on purpose because of 7be7eb115497da6b0f92d618baa5fd86ca8fd261 (and in fact we have an ad-hoc script in the CI that autodetects which string should exist / should not exist) ... But I guess the original issue is that there are existing instance with such messages for old logs and it wants to find the correponding name :/

alexAubin:
Closing because as tried to explain in previous comment, it's supposed to not exist and yes we may have some legacy logs which therefore end up with no i18n string but shouldnt be a huge deal

PR #1615, opened 2023-02-26
helpers: add support for a 'sources' app resources to modernize and replace app.src format

alexAubin:
## The problem

cf https://github.com/YunoHost/issues/issues/1978

- The whole key-value thing is not standard super (except that it's a source-able file, but is not used as such)
- It would be nice to have all sources declared in a single file rather than gazillion files for each arch or extra module
- YunoHost should automatically switch to the appropriate amd64 / armhf / i386 / ... source rather than the packager taking care of this
- The "SOURCE_FILENAME" option is confusing
- We are in the process of designing the future of the app's auto-upgrade.yml/sh hooks and having some additional info close to the app sources info could be useful to define "how to watch for new releases"

## Solution

- Add support for a `sources` app resource in the manifest that is meant to declare the external assets to fetch to install/upgrade the app
- syntax is expected to look like:

```toml
   [resources.sources]
       [resources.sources.main]
       url = "https://github.com/foo/bar/archive/refs/tags/v1.2.3.tar.gz"
       sha256 = "01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b"
       # format = "tar.gz"   #  (only needed if not obvious from the url itself)
       in_subdir = false  # or true which is the default
```

ooooor for different-source-per-arch use case:

```toml
   [resources.sources]

       [resources.sources.main]
       in_subdir = false
       amd64.url = "https://github.com/foo/bar/archive/refs/tags/v1.2.3.amd64.tar.gz"
       amd64.sha256 = "01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b"
       i386.url = "https://github.com/foo/bar/archive/refs/tags/v1.2.3.amd64.tar.gz"
       i386.sha256 = "53c234e5e8472b6ac51c1ae1cab3fe06fad053beb8ebfd8977b010655bfdd3c3"
       armhf.url = "https://github.com/foo/bar/archive/refs/tags/v1.2.3.armhf.tar.gz"
       armhf.sha256 = "4355a46b19d348dc2f57c046f8ef63d4538ebb936000f3c9ee954a27460dd865"

       [resources.sources.zblerg]
       url = "https://zblerg.com/download/zblerg"
       sha256 = "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2"
       format = "script"
       rename = "zblerg.sh"
```

- During the provisioning phase, the relevant assets are pre-fetched (unless `prefetch = false` is specified for that asset)
- You can also just call `ynh_setup_source --dest_dir=$install_dir` without specificying the arch anymore, yunohost will automatically pick up the corresponding arch if there's no `url` defined but `amd64.url` or `armhf.url` or `i386.url` or ... exits
 
- Tried have some autosmart detection of tar.gz / zip archive, in which case `extract` is true, otherwise it's set to `false`
- The `SOURCE_FILENAME` now corresponds to `rename = "foobar.ext"` which should be more clear what it's useful to (only relevant when extract = false) and doesn't need to be defined in many cases...

## PR Status

Tested on my side, though not in every possible usecase

- [ ] Would be nice to have a magic script to autoconvert old app.src etc into the new format
- [ ] Still need to think about what exact info we would need to cover the "watch-for-new-update" usecase

## How to test

...

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -352,8 +352,6 @@ class SourcesResource(AppResource):
         if context.get("action") == "restore":
             return
 
-        import pdb; pdb.set_trace()
-
         for source_id, infos in self.sources.items():
 
             if not infos["prefetch"]:
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -292,7 +292,7 @@ class SourcesResource(AppResource):
 
         [resources.sources.zblerg]
         url = "https://zblerg.com/download/zblerg"
-        sha256sum = "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2"
+        sha256 = "1121cfccd5913f0a63fec40a6ffd44ea64f9dc135c66634ba001d10bcf4302a2"
         format = "script"
         rename = "zblerg.sh"
 
@@ -384,8 +384,7 @@ class SourcesResource(AppResource):
         if returncode != 0:
             if os.path.exists(filename):
                 rm(filename)
-            out = out.decode()
-            raise YunohostError(f"Failed to download asset {source_id} ({url}) for {self.app}: {out}", raw_msg=True)
+            raise YunohostError("app_failed_to_download_asset", source_id=source_id, url=url, app=self.app, out=out.decode())
 
         assert os.path.exists(filename), f"For some reason, wget worked but {filename} doesnt exists?"
 
@@ -393,7 +392,7 @@ class SourcesResource(AppResource):
         if computed_sha256 != expected_sha256:
             size = check_output(f"du -hs {filename}").split()[0]
             rm(filename)
-            raise YunohostError(f"Corrupt source for {url} : expected to find {expected_sha256} as sha256sum, but got {computed_sha256} instead ... (file size : {size})", raw_msg=True)
+            raise YunohostError("app_corrupt_source", source_id=source_id, url=url, app=self.app, expected_sha256=expected_sha256, computed_sha256=computed_sha256, size=size)
 
 
 class PermissionsResource(AppResource):
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -320,7 +320,7 @@ class SourcesResource(AppResource):
     - For elements with `prefetch = true`, will download the asset (for the appropriate architecture) and store them in `/var/cache/yunohost/download/$app/$source_id`, to be later picked up by `ynh_setup_source`. (NB: this only happens during install and upgrade, not restore)
 
     ##### Deprovision:
-    - Nothing
+    - Nothing (just cleanup the cache)
     """
 
     type = "sources"
@@ -345,7 +345,6 @@ class SourcesResource(AppResource):
     def deprovision(self, context: Dict = {}):
         if os.path.isdir(f"/var/cache/yunohost/download/{self.app}/"):
             rm(f"/var/cache/yunohost/download/{self.app}/", recursive=True)
-        pass
 
     def provision_or_update(self, context: Dict = {}):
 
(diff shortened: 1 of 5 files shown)


zamentur:
Is there a reason to not put this in manifest.toml directly ?

alexAubin:
I dunno, that's on the table, this was kind of about keeping the manifest "not too long" but that's more like a cosmetic thing ... like for apps with many modules like mediawiki ... but there shouldnt be that many ...? (Well there's Nextcloud too maybe which needs url to past versions)

In the future packaging v2 I was also planning to probably have a section like `configurations` maybe, to declare stuff about the nginx, php, systemd, fail2ban, logorate, etc confs

Sooo yeah idk maybe we can embrace the whole thing about "manifest stores basically all the metadata about the app"

Whatcha think about all this @YunoHost/apps ?

ericgaspar:
I am personally in favor of sources being part of the manifest.toml file

src/utils/resources.py
```diff
@@ -257,6 +258,146 @@ def _run_script(self, action, script, env={}):
...
+        - `whatever`: whatever arbitrary value, not really meaningful except to imply that the file won't be extracted (eg because it's a .deb to be manually installed with dpkg/apt, or a script, or ...)
+    - `in_subdir`: `true` (default) or `false`, depending on if there's an intermediate subdir in the archive before accessing the actual files. Can also be `N` (an integer) to handle special cases where there's `N` level of subdir to get rid of to actually access the files
+    - `extract` : `true` or `false`. Defaults to `true` for archives such as `zip`, `tar.gz`, `tar.bz2`, ... Or defaults to `false` when `format` is not something that should be extracted. When `extract = false`, the file will only be `mv`ed to the location, possibly renamed using the `rename` value
+    - `rename`: some string like `whatever_your_want`, to be used for convenience when `extract` is `false` and the default name of the file is not practical
+    - `platform`: for exampl `linux/amd64` (defaults to `linux/$YNH_ARCH`) to be used in conjonction with `format = "docker"` to specify which architecture to extract for
```
ericgaspar, line 316:
```suggestion
    - `platform`: for example `linux/amd64` (defaults to `linux/$YNH_ARCH`) to be used in conjonction with `format = "docker"` to specify which architecture to extract for
```

ericgaspar (changes requested):

src/utils/resources.py
```diff
@@ -257,6 +258,143 @@ def _run_script(self, action, script, env={}):
...
+            out = out.decode()
+            raise YunohostError(f"Failed to download asset {source_id} ({url}) for {self.app}: {out}", raw_msg=True)
+
+        assert os.path.exists(filename), f"For some reason, wget worked but {filename} doesnt exists?"
+
+        computed_sha256 = check_output(f"sha256sum {filename}").split()[0]
+        if computed_sha256 != expected_sha256:
+            size = check_output(f"du -hs {filename}").split()[0]
+            rm(filename)
+            raise YunohostError(f"Corrupt source for {url} : expected to find {expected_sha256} as sha256sum, but got {computed_sha256} instead ... (file size : {size})", raw_msg=True)
```
kay0u, line 396: Don't we want to translate all these `YunohostError`?

alexAubin: Zblerg yeah I was a bit lazy, but yeah, at least the most "probable" ones, such as this one

alexAubin: Done :+1:

tituspijean (approved):
I approve the general approach. ❤️

zamentur (approved):
LGTM (for the approach, don't have time to review the code)

kay0u (approved):

PR #1628, opened 2023-03-15
POC to add a route to fetch what's the current running action, cf plans to improve the webadmin UX

alexAubin:
## The problem

cf https://github.com/YunoHost/issues/issues/2055

We should have a mechanism to know what's the current operation running on the server when the lock is active

## Solution

Add a `yunohost log current-operation` (and similar API route)

## PR Status

Tested during an app install, working-ish ...

## How to test

Trigger an app install and run `yunohost log current-operation` in parallel ... Should return `operation_description: 'app_install'` while you answer the questions, and then a proper human-readable description during the actual app install

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -1911,6 +1911,11 @@ log:
                 path:
                     help: Log file to share
 
+        ### log_current-operation()
+        current-operation:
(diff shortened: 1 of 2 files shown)


alexAubin:
... or is it ? :thinking:

PR #1629, opened 2023-03-22
[tests] Rework questions/options tests

Axolotle:
Rework the questions/options tests and add multiple missing option types tests scenarios.

There's some tests that currently passes but probably shouldn't, some that do not but probably should.
You can search for `xfail`, `xpass`, and some `FIXME`s to have a quick look at those.
Plan is to fix the unexpected results on the pydantic migration if you agree with my comments.
There's a bunch of problems with nones () where things are not very clear what final None value we want.

Also added some quick fixes in `utils/config.py` to ease tests. @zamentur do you see any potential problems with those modifications? There might be magic stuff in bash helpers that I'm not aware of.

## PR Status

ok?

## How to test

```bash
pytest src/tests/test_questions.py
```

### DIFF
--- a/src/utils/config.py
+++ b/src/utils/config.py
@@ -856,7 +856,9 @@ class Question:
         # Don't restrict choices if there's none specified
         self.choices = question.get("choices", None)
         self.pattern = question.get("pattern", self.pattern)
-        self.ask = question.get("ask", {"en": self.name})
+        self.ask = question.get("ask", self.name)
+        if not isinstance(self.ask, dict):
+            self.ask = {"en": self.ask}
         self.help = question.get("help")
         self.redact = question.get("redact", False)
         self.filter = question.get("filter", None)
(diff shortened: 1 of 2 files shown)


src/utils/config.py
```diff
@@ -1481,8 +1505,12 @@ def _prevalidate(self):
 
         super()._prevalidate()
 
+        # Validation should have already failed if required
+        if self.value in (None, ""):
```
alexAubin, line 1509:
```suggestion
        if self.value in [None, ""]:
```

src/utils/config.py
```diff
@@ -1481,8 +1505,12 @@ def _prevalidate(self):
...
+            return self.value
+
         if Moulinette.interface.type != "api":
-            if not self.value or not os.path.exists(str(self.value)):
+            if not os.path.exists(str(self.value)) or not os.path.isfile(str(self.value)):
```
alexAubin, line 1513: hmm what's the motivation for this change here ? Not a huge deal but why not keep the `if not self.value` ? And Why enforce `isfile` ? (`exists` should be enough ... I guess people *could* maybe want to use this kind of arg for a directory or a symlink ... ?). Though as said, not a huge deal but I'm confused :sweat_smile:

Axolotle: Hum, I removed the `if not self.value` because it was raising the error even if it was optional in cli, hence the check for it before. For the enforcing I don't remember lel, i'll run the tests without to check

PR #1631, opened 2023-03-22
[DEL] remove args parameter from actionsmap

hbrunn:
## The problem

when adding a button to a config panel like in https://github.com/moshchot/libreerp_ynh/blob/update_oca_and_gitaggregate/config_panel.toml#L10, I can run my action perfectly fine from the terminal with

```
yunohost app action run libreerp custom_code.components.gitaggregate
```

But calling this via the admin interface, I receive `argument @args: expected one argument` without this change, with the request posting an empty args parameter:

image

## Solution

As passing parameters to actions doesn't really work currently anyways, I figure we can remove this parameter entirely from the map.

## PR Status

Dubious. Probably the correct solution would be not to post the args parameter at all from the admin interface?

## How to test

Create a button in `config_panel.toml` similar to the one linked above, observe you can call the action from the CLI as well as via the admin interface.

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -1032,9 +1032,6 @@ app:
                         help: App name
                     action:
                         help: action id
-                    -a:
-                        full: --args
-                        help: Serialized arguments for app script (i.e. "domain=domain.tld&path=/path")
 
       config:
           subcategory_help: Applications configuration panel


Axolotle:
Hey! Sorry but i think this will break action sections. There's no action parameters per se but you can get extra data by putting other options in your action's section.

For example:
```toml
[custom_code]
name = "Manage custom code"

    [custom_code.components]
    name = "Actions"

        [custom_code.components.action_number]
        ask = "gimme number for the action"
        type = "number"

        [custom_code.components.gitaggregate]
        ask = "Run gitaggregate"
        type = "button"
        help = "This assumes you have clone a repository in custom-addons that contains a repos.yml file"
        enabled = "action_number > 2"
```
the `action_number` can be used here to evaluate the `enabled` prop, but you can also expect it to be passed alongside the action and retrieve it in your bash action function so we need `args` to pass the extra data.

You are in a special case, there's no inputs in the config panel, this is not yet handled. I'll check on the web-admin side what can be done ; )

hbrunn:
thank you for the workaround! Irony is I was planning to add configuration vars later, but started with just the button.

So I'll PR a few lines about buttons and how they can't be on an empty panel to https://github.com/YunoHost/example_ynh/blob/master/config_panel.toml.example which I used as de facto documentation for this feature?

PR #1647, opened 2023-04-17
[enh] exclude .well-known subpaths from conflict checks

tituspijean:
Attempts to solve:
Fixes https://github.com/YunoHost/issues/issues/684
Fixes https://github.com/YunoHost/issues/issues/2016

It basically adds an exception for app conflicts if an app is installed at the root of a domain and another app tries to add a `./well-known/foo` subpath, but keeps returning a conflict if any two apps try to register the same `./well-known/foo` subpath.

## PR Status

Finished? Untested.

## How to test

...

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -2869,7 +2869,7 @@ def _get_conflicting_apps(domain, path, ignore_app=None):
         for p, a in apps_map[domain].items():
             if a["id"] == ignore_app:
                 continue
-            if path == p or path == "/" or p == "/":
+            if path == p or ( not path.startswith("/.well-known/") and path == "/" ) or ( not path.startswith("/.well-known/") and p == "/" ):
                 conflicts.append((p, a["id"], a["label"]))
 
     return conflicts
--- a/src/app.py
+++ b/src/app.py
@@ -2869,7 +2869,7 @@ def _get_conflicting_apps(domain, path, ignore_app=None):
         for p, a in apps_map[domain].items():
             if a["id"] == ignore_app:
                 continue
-            if path == p or ( not path.startswith("/.well-known/") and path == "/" ) or ( not path.startswith("/.well-known/") and p == "/" ):
+            if path == p or ( not path.startswith("/.well-known/") and ( path == "/" or p == "/" ) ):
                 conflicts.append((p, a["id"], a["label"]))
 
     return conflicts


src/app.py
```diff
@@ -2869,7 +2869,7 @@ def _get_conflicting_apps(domain, path, ignore_app=None):
         for p, a in apps_map[domain].items():
             if a["id"] == ignore_app:
                 continue
-            if path == p or path == "/" or p == "/":
+            if path == p or ( not path.startswith("/.well-known/") and path == "/" ) or ( not path.startswith("/.well-known/") and p == "/" ):
```
tituspijean, line 2872:
```suggestion
            if path == p or ( not path.startswith("/.well-known/") and ( path == "/" or p == "/" ) ):
```

A bit of factorization

tituspijean:
It's been a while, I cannot remember why I implemented that. Basically this PR allows apps to publish something in `/.well-known` even though an app is installed at the root of the domain.

(and you misread the description, it states we still return an error if apps try to register the same `./well-known/foo` path ;) )

rodneyrod:
Thanks for the contribution, just to mention from testing I've done on a testbed server, it looks like the full_domain flag may need to have extra logic to prevent scripts for install and upgrade failing with the full_domain error.
I believe the step that's causing my tests to fail is this one (not a programmer by trade so I could be reading this wrong):
`def _assert_no_conflicting_apps(domain, path, ignore_app=None, full_domain=False):
    conflicts = _get_conflicting_apps(domain, path, ignore_app)

    if conflicts:
        apps = []
        for path, app_id, app_label in conflicts:
            apps.append(f" * {domain}{path} → {app_label} ({app_id})")

        if full_domain:
            raise YunohostValidationError("app_full_domain_unavailable", domain=domain)
        else:
            raise YunohostValidationError(
                "app_location_unavailable", apps="\n".join(apps)
            )
`

For testing, I've installed Friendica on example.com, then installed Synapse on m.example.com but configuring accounts to use user@example.com using the option in the setup script for that.
This works to install those apps if done in that order.
However, upgrading Friendica fails as example.com is detected as in use.
Additionally, if Synapse is installed first, Friendica will fail to install as the domain is detected as in use.

tituspijean:
Thank you for your testing! Can you share the failed installation log?

centralscrutinizer21:
It worked for me. Bludit on / `domain.com` Synapse on `matrix.domain.com` and server name `domain.com`.
Recoverd Synapse from a backup, and then changed URL for Bludit from the temporary `site.domain.com` I had move it to, back to its original `domain.com`

tituspijean:
After further consideration, I would push for a dedicated "well-known path" resource, rather than patching the current "path" one.

alexAubin:
(Smol bump because we're discussing this with somebody today ... Maybe to be discussed during the Yunocamp ... It feels like we could just merge the PR in its current state)

gwen-lg:
Hello, I like to be able to configure the matrix server (Conduit by example) to use the root domain with a simple `my_webapp` installed on this root domain.

rodneyrod:
Hey, just saw this again, turns out I copied the commit before @tituspijean did the refactor, all working on my end.

Tagadda:
Lets merge~

PR #1650, opened 2023-04-19
ConfigPanel: Before pydantic (structure changes) 2/3

Axolotle:
PR 2/3 for the pydantic migration of config panel.
This is a separated PR to hopefully ease reviewing.

- add `OptionType` enum
- Add intermediate base class for display only types (`alert`, `button`, etc.) and input types to restrict some properties only to input types like `default`, `optional`, etc.
- rename `Option.name` to `Option.id`
- remove `choices` from BaseOption, only apply to types `select` and `tags` (and `app|domain|user|group` but without control over it). So type `string` can no longer have `choices`
- fix `Option.readonly` so a readonly option's value can't be overriden + update tests
- rename `Option._format_text_for_user_input_in_cli()` to `Option._get_prompt_message()`
- extract `Option._ask_if_needed()` to an external function `prompt_or_validate_form` that iterate over Options directly

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1099,7 +1099,7 @@ def app_install(
     raw_questions = manifest["install"]
     questions = ask_questions_and_parse_answers(raw_questions, prefilled_answers=args)
     args = {
-        question.name: question.value
+        question.id: question.value
         for question in questions
         if question.value is not None
     }
@@ -1147,7 +1147,7 @@ def app_install(
             if question.type == "password":
                 continue
 
-            app_settings[question.name] = question.value
+            app_settings[question.id] = question.value
 
     _set_app_settings(app_instance_name, app_settings)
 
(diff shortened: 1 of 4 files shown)


alexAubin (commented):
>remove `choices` from BaseOption, only apply to types `select` and `tags` (and `app|domain|user|group` but without control over it). So type `string` can no longer have `choices`

This sound like a good thing, but I'm pretty sure we can find lots of apps in the wild using choices for string-type questions and we have to be careful about this :/ What are your thoughts on this ?

Axolotle:
> > remove `choices` from BaseOption, only apply to types `select` and `tags` (and `app|domain|user|group` but without control over it). So type `string` can no longer have `choices`
> 
> This sound like a good thing, but I'm pretty sure we can find lots of apps in the wild using choices for string-type questions and we have to be careful about this :/ What are your thoughts on this ?

I've added a default type to "select" in the `pydantic` branch if there's `choices` in the option (but only if the type is not explicitly set) which cover some cases of some apps, but [searching `choices` in apps](https://grep.app/search?q=choices&filter[repo.pattern][0]=YunoHost-Apps&filter[lang][0]=JSON&filter[lang][1]=TOML) shows explicit `type = "string"` with `choices` indeed. Can make a bunch of PR or add some legacy check in option parsing (checking explicit `string` type with choices and change it to `select`?

alexAubin:
>(checking explicit string type with choices and change it to select?

Yup that sounds good, I wouldn't bother with making PR to apps, just displaying a warning when finding such case (+ fixing it on the fly) should be enough imho

Axolotle:
> Yup that sounds good, I wouldn't bother with making PR to apps, just displaying a warning when finding such case (+ fixing it on the fly) should be enough imho

added a warning + hard set to `select` in https://github.com/YunoHost/yunohost/pull/1651/commits/21e3aafce227ded28323f888818d507b60757930 + a test

PR #1652, opened 2023-04-20
Add certificate as postinstall tip

ericgaspar:
Proposing to add let's Encrypt command line as postinstall tip.


tituspijean:
I'm guessing you faced an issue with the certificate when reinstalling your server due to HSTS cache. You modification implies that the command line is the only way to generate a certificate, I would still refer to the webadmin too :)

PR #1656, opened 2023-05-03
Offer a shell into an app environment

tituspijean:
## The problem

Closes https://github.com/YunoHost/issues/issues/2041 

TODO:
- [x] check that sourced `EnvironmentFiles` are actually propagated to the subshell
- [x] check that `$app` user actually exists
- [x] do we need to fail if no `$install_dir` is set (non-webapp)
- [x] how to handle e.g. Synapse, whose service is not `synapse.service` but `matrix-synapse.service`?
- [x] `alias` `php` command to `php$phpversion`

## PR Status

To be extensively tested.

## How to test

...

### DIFF
--- a/helpers/apps
+++ b/helpers/apps
@@ -164,6 +164,14 @@ ynh_spawn_app_shell() {
         [ -n "$env_var" ] && export $env_var;
         export HOME=$install_dir;
 
+        # Force `php` to its intended version
+        local phpversion=$(ynh_app_setting_get --app=$app --key=phpversion)
+        if [ -n "$phpversion" ]
+        then
+            eval "php() { php${phpversion} \"\$@\"; }"
+            export -f php
+        fi
+
         # Source the EnvironmentFiles from the app's service
         local env_files=(`systemctl show $app.service -p "EnvironmentFiles" --value`)
         if [ ${#env_files[*]} -gt 0 ]
--- a/helpers/apps
+++ b/helpers/apps
@@ -112,16 +112,19 @@ ynh_remove_apps() {
 	fi
 }
 
-# Load an app environment in the current Bash shell
+# Spawn a Bash shell with the app environment loaded
 #
-# usage: ynh_install_apps --app="app"
+# usage: ynh_spawn_app_shell --app="app"
 #     | arg: -a, --app=     - the app ID
 #
 # examples:
 #   ynh_spawn_app_shell --app="APP" <<< 'echo "$USER"'
 #   ynh_spawn_app_shell --app="APP" < /tmp/some_script.bash
 #
-# Requires YunoHost version 11.0.* or higher.
+# Requires YunoHost version 11.0.* or higher, and that the app relies on packaging v2 or higher.
+# The spawned shell will have environment variables loaded and environment files sourced
+# from the app's service configuration file (defaults to $app.service, overridable by the packager with `service` setting).
+# If the app relies on a specific PHP version, then `php` will be aliased that version.
 ynh_spawn_app_shell() {
         # Declare an array to define the options of this helper.
         local legacy_args=a
--- a/helpers/apps
+++ b/helpers/apps
@@ -167,7 +167,7 @@ ynh_spawn_app_shell() {
         [ -z "$service" ] && service=$app;
 
         # Load the Environment variables from the app's service
-        local env_var=`systemctl show $service.service -p "Environment" --value`
+        local env_var=$(systemctl show $service.service -p "Environment" --value)
         [ -n "$env_var" ] && export $env_var;
         export HOME=$install_dir;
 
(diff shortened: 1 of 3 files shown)


helpers/apps
```diff
@@ -111,3 +111,68 @@ ynh_remove_apps() {
...
+            set -a
+            for file in ${env_files[*]}
+            do
+                [[ $file = /* ]] && source $file
+            done
+            set +a
+        fi
+
+        # Open the app shell
+        su -s /bin/bash $app
```
Tagadda, line 177:
```suggestion
        su -s /bin/bash $app && cd $(systemctl show $app.service -p "WorkingDirectory" --value)
```

tituspijean: Good idea :) Why not defaulting to `$install_dir` if there is no service too, like for `mywebapp_ynh`?

Tagadda: Good idea!
```suggestion
        local env_dir = $(systemctl show $app.service -p "WorkingDirectory" --value)
        if [[ $env_dir = "" ]];
        then
            env_dir = $install_dir
        fi

        cd $env_dir
        su -s /bin/bash $app
```

helpers/apps
```diff
@@ -111,3 +111,68 @@ ynh_remove_apps() {
...
 	fi
 }
+
+# Load an app environment in the current Bash shell
+#
+# usage: ynh_install_apps --app="app"
+#     | arg: -a, --app=     - the app ID
+#
+# Requires YunoHost version 11.0.* or higher.
+ynh_load_app_environment() {
```
fflorent, line 121: The naming suggests that we only load the app environment, but we rather log as the app user (using the `su` command).

What about: `ynh_spawn_app_shell`? `ynh_start_app_shell`?

(Also thanks so muuuch for the PR, I really look forward using the command :heart:)

helpers/apps
```diff
@@ -111,3 +111,68 @@ ynh_remove_apps() {
 		done
 	fi
 }
+
+# Load an app environment in the current Bash shell
+#
+# usage: ynh_install_apps --app="app"
+#     | arg: -a, --app=     - the app ID
+#
+# Requires YunoHost version 11.0.* or higher.
```
fflorent, line 120: If I am correct, we could also suggest using this command as follow:

```suggestion
# examples:
#   ynh_load_app_environment --app="APP" <<< 'echo "$USER"'
#   ynh_load_app_environment --app="APP" < /tmp/some_script.bash
# 
# Requires YunoHost version 11.0.* or higher.
```

tituspijean: Absolutely! :)

tituspijean:
Ready to test. :)

helpers/apps
```diff
@@ -111,3 +111,92 @@ ynh_remove_apps() {
...
+
+        # Make sure the app is installed
+        local installed_apps_list=($(yunohost app list --output-as json --quiet | jq -r .apps[].id))
+        if [[ " ${installed_apps_list[*]} " != *" ${app} "* ]]
+        then
+            ynh_print_err --message="$app is not in the apps list"
+            exit 1
+        fi
+
+        # Make sure the app is installed
```
fflorent, line 151: It's already stated above. Probably better changed to:
```suggestion
        # Make sure the app has its own user
```

helpers/apps
```diff
@@ -111,3 +111,92 @@ ynh_remove_apps() {
...
+        # Load the Environment variables from the app's service
+        local env_var=`systemctl show $service.service -p "Environment" --value`
+        [ -n "$env_var" ] && export $env_var;
+        export HOME=$install_dir;
+
+        # Force `php` to its intended version
+        local phpversion=$(ynh_app_setting_get --app=$app --key=phpversion)
+        if [ -n "$phpversion" ]
+        then
+            eval "php() { php${phpversion} \"\$@\"; }"
```
fflorent, line 178: Nitpicking: I suggest this for more concision and simplicity:
```suggestion
            alias php="php${phpversion}"
```

fflorent (approved):

helpers/apps
tituspijean: Ooops, bad copy and paste. :)

helpers/apps
tituspijean: `alias` is not propagated to subshells. ;) The `eval+export` way is the most concise way I found.

fflorent: Interesting! Even `alias+export`?

BTW, if I attempt on a local shell to run `eval "foo() { echo \"$@\"; }"` and then `su -s /bin/bash $USER`, I cannot invoke `foo WHATEVER_ARG`. Wonder what is actually missing to make it work.

tituspijean: Even. Tested it earlier this morning, even with `set -a`.

You forgot to escape `$`:

```
bash
eval "foo() { echo \"\$@\"; }"
export -f foo
su -s /bin/bash 
type foo # returns the function definition
foo testme  # returns testme
```

alexAubin (approved):
LGTM, I don't have time to test until next week, but considering it's a power-user feature anyway, we can probably yolomerge it ?

PR #1661, opened 2023-05-16
keep fail2ban rules on firewall reload

aya:
## The problem

Reloading firewall flushes all iptables rules to create new ones, dropping fail2ban rules in the same time.

## Solution

Restarting fail2ban restores its iptables rules.

## PR Status

Proposal.
I'm new to yunohost, I don't know if/how it should be fixed.

## How to test

With fail2ban activated, check the firewall status

```
# iptables -nL
Chain INPUT (policy DROP)
target     prot opt source               destination
f2b-recidive  tcp  --  anywhere             anywhere
f2b-sshd   tcp  --  anywhere             anywhere             multiport dports ssh
[...]
```

Reload yunohost firewall

```
# yunohost firewall reload
```

The fail2ban rules are gone
```
# iptables -nL
Chain INPUT (policy DROP)
target     prot opt source               destination
[...]
```

### DIFF
--- a/hooks/post_iptable_rules/52-fail2ban
+++ b/hooks/post_iptable_rules/52-fail2ban
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-set -e
-
-systemctl restart fail2ban
--- a/hooks/post_iptable_rules/52-fail2ban
+++ b/hooks/post_iptable_rules/52-fail2ban
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+systemctl restart fail2ban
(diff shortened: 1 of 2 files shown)


alexAubin:
Hmmmokay but it sounds a bit funky to add an entire bash hook just for this

There's already this line : https://github.com/YunoHost/yunohost/blob/dev/src/firewall.py#L334 which should do the trick but only does a "reload"

Does this means that "reload" doesn't tell fail2ban to recreate the rules and a restart is mandatory ?

aya:
Yes, restarting is mandatory.
Thank you for the review !

PR #1666, opened 2023-05-24
[fix] Remove deprecated domain_dns_conf

eldertek:
## The problem

This issue https://github.com/YunoHost/issues/issues/2090 exists since 22 nov 2022, I think we could delete the deprecation and close the issue.

## Solution

Delete actionmap and the corresponding function

## PR Status

Done

## How to test

Clone then, use with ynh-dev.

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -537,17 +537,6 @@ domain:
                     full: --force
                     help: Do not ask confirmation to remove apps
                     action: store_true
-
-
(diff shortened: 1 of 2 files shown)


alexAubin:
I guess we'll ultimately do this, but "6 month ago" is still somewhat fresh for removing a deprecated feature .. the original issue was more about the misleading doc and lack of pointer to the new command

alexAubin:
Merging in bookworm

PR #1667, opened 2023-05-24
[fix/enh] Rewrite of yunopaste CLI tool

eldertek:
## The problem

As mentionned is this issue https://github.com/YunoHost/issues/issues/1291 , yunohost cli was doing weird.

## Solution

A new CLI tool written in python, ip and domain protect is activated by default and can be disabled by parameter.
image

Here is an example: https://paste.yunohost.org/ubezadequh
image

## PR Status

Done

## How to test

Replace yunopaste with the one of the pull request.

### DIFF
--- a/bin/yunopaste
+++ b/bin/yunopaste
@@ -3,7 +3,6 @@
 import sys
 import requests
 import json
-import re
 
 SERVER_URL = "https://paste.yunohost.org"
 TIMEOUT = 3
@@ -19,43 +18,14 @@ def create_snippet(data):
         print("\033[31mError: {}\033[0m".format(e))
         sys.exit(1)
 
-def anonymize_output(output):
-    # Replace IPv4 addresses
-    output = re.sub(r"\b(?:\d{1,3}\.){3}\d{1,3}\b", "[IPv4]", output)
-    # Replace IPv6 addresses
-    output = re.sub(r"\b(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}\b", "[IPv6]", output)
-    # Replace domain names
-    output = re.sub(r"\b(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}\b", "[DOMAIN]", output)
-
-    return output
-
-def print_usage():
-    print("\033[33mUsage:")
-    print("\033[34m  yunopaste [-na|--no-anonymize]\033[0m")
-    print("\033[33mDescription:")
-    print("\033[36m  Reads the command output from stdin and pastes it to the YunoHost Haste server.")
-    print("  Prints the URL of the created snippet.\033[0m")
-    print("\033[33mOptions:")
-    print("\033[35m  -na, --no-anonymize\t\033[0mDo not anonymize the output")
-    print("\033[33mExample:")
-    print("\033[35m  yunohost diagnosis show | yunopaste\033[0m")
-    print("\033[35m  yunohost diagnosis show | yunopaste -na\033[0m")
 
 def main():
-    if "-na" in sys.argv or "--no-anonymize" in sys.argv:
-        anonymize_output_flag = False
-    else:
-        anonymize_output_flag = True
-
     output = sys.stdin.read()
 
     if not output:
         print("\033[31mError: No input received from stdin.\033[0m")
         sys.exit(1)
 
-    if anonymize_output_flag:
-        output = anonymize_output(output)
-        
     url = create_snippet(output)
 
     print("\033[32mURL: {}\033[0m".format(url))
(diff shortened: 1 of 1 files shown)


tituspijean:
The aforementioned issue was actually solved by https://github.com/YunoHost/yunohost/pull/1558, with the pipe `|` working too. Sorry!
Let's keep it open while core dev assess if it's worth further rewriting the tool.

eldertek:
> The aforementioned issue was actually solved by #1558, with the pipe `|` working too. Sorry!
> Let's keep it open while core dev assess if it's worth further rewriting the tool.

I tested before doing my PR and it was not working especially with journal (without sudo) or with yunohost cli. This PR allow command to be visible during execution (so user can input) and save them output into yunopaste. 

Please consider that

bin/yunopaste
```diff
@@ -1,77 +1,98 @@
...
+
+    command_args = sys.argv[1:]
+    anonymize_output_flag = True
+
+    if "-na" in command_args or "--no-anonymize" in command_args:
+        command_args = [arg for arg in command_args if arg not in ("-na", "--no-anonymize")]
+        anonymize_output_flag = False
+
+    command = " ".join(command_args)
+    output = execute_command(command)
```
alexAubin, line 83: Hmmmokay but I would refrain from supporting the syntax `yunopaste "yunohost foo bar"` : 
- It's not really UNIX-ish in the sense that you would expect this to work rather using pipes : `yunohost foo bar | yunopaste`
- It's not yunopaste's job to run the command itself, there are many ways this can go wrong
- It breaks backward compatibility because in the past it was working using pipes and there's not technical reason for us to change this behavior

Imho you should just use `sys.stdin.read()` as the input (or rather what you called `output`)

eldertek: Okay I modified to allow only pipe, but it don't allow to be interactive

bin/yunopaste
```diff
@@ -1,77 +1,98 @@
...
+
+    if not output:
+        print("\033[31mError: Command didn't produce any output.\033[0m")
+        sys.exit(1)
+
+    if anonymize_output_flag:
+        anonymized_output = anonymize_output(output)
+        url = create_snippet(anonymized_output)
+    else:
+        url = create_snippet(output)
```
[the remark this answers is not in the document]
eldertek: Commited your suggestion !

alexAubin (approved):
LGTM, still gotta find 5 minutes to quickly actually test it :+1:

eldertek:
Perfect, do you know if there is a server where I can talk to yunohost team ? Like discord

alexAubin:
cf https://yunohost.org/chat_rooms

bin/yunopaste
```diff
@@ -1,77 +1,64 @@
...
+
+def main():
+    if "-na" in sys.argv or "--no-anonymize" in sys.argv:
+        anonymize_output_flag = False
+    else:
+        anonymize_output_flag = True
+
+    command_output = sys.stdin.read()
+
+    if not command_output:
```
alexAubin, line 52:
```suggestion
    output = sys.stdin.read()

    if not output:
```

(otherwise `output` doesnt exist after ..)

bin/yunopaste
```diff
@@ -1,77 +1,64 @@
...
+        print("\033[31mError: {}\033[0m".format(e))
+        sys.exit(1)
+
+def anonymize_output(output):
+    # Replace IPv4 addresses
+    output = re.sub(r"\b(?:\d{1,3}\.){3}\d{1,3}\b", "[IPv4]", output)
+    # Replace IPv6 addresses
+    output = re.sub(r"\b(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}\b", "[IPv6]", output)
+    # Replace domain names
+    output = re.sub(r"\b(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}\b", "[DOMAIN]", output)
```
alexAubin, line 28: This regex is a bit too brutal ... It finds anything looking like `foo.bar` as a domain, such as for example `some_app.conf` or `manifest.json`

Resulting in confusing lines such as : `chown tiki:tiki /etc/nginx/conf.d/[DOMAIN].d/[DOMAIN]`

or : `manifest=/var/cache/yunohost/app_tmp_work_dirs/app_6k1a4ef4/[DOMAIN]`

To a lesser extend, also the apt install stuff: `Get:1 http://[DOMAIN]/debian stable/main amd64 libgif7 amd64 5.1.9-2 [45.1 kB]`

Personally I would tend to only replace domains registered in YunoHost ... the list can be obtained in various ways, but for example finding all occurences of folders named `/etc/nginx/conf.d/DOMAIN.d/`

alexAubin:
Due to lack of feedback, and cf previous discussion, I removed the anonymization stuff because anyway I'm not convinced it's Yunopaste's job to anonymize stuff ... feel free to open follow-up PR or issues regarding the topic though ...

PR #1670, opened 2023-06-10
Allow passing a list in the manifest.toml for the apt resource packages

Salamandar:
## The problem

We use Toml and Python. Let's use the types they provide.

## Solution

Use a list instead of a string. You can split it into lines, add comments etc.

The code is expected to retain compatibility with current manifests with strings.

## PR Status

Untested

## How to test

...


alexAubin (commented):
(sounds legit, will try to test/merge in a few days once we're past the recent security release)

PR #1674, opened 2023-06-12
Add logrotate resource

Salamandar:
Actually it should not be merged : i think this should be called *after* the install script… not before. But still, here's the idea.

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -22,7 +22,7 @@ import shutil
 import random
 import tempfile
 import subprocess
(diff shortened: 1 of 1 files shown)


Salamandar:
Yeah I just wanted to try to see if it was possible. I'm OK if it doesn't get merged… but it's done if you want it :D

> i.e. for the nginx conf you want to install it "in the middle" of the current install script, which has many implication on the app design ...

Why in the middle ? I would say at the very end.

Salamandar:
👍

PR #1675, opened 2023-06-13
Magically keeping all custom-config panel settings during the conf upgrade via ynh_add_config without having to define all of them as yunohost settings

Tagadda:
## The problem

- YunoHost/issues#1973

## Solution

In `ynh_add_config`:
- Backup config_panel values with `yunohost app config get $app --export`
- *usual ynh_add_config stuff...*
- Restore the config_panel values with `yunohost app config set $app -f $file`
- Recompute and store the file hash

## PR Status

Yolocommited, working with a single .env file
Need to be tested with a config_panel with multiples files

## How to test

```
sudo yunohost app install https://github.com/Tagadda/helloworld_ynh/tree/enh/ynh_add_config/keep_config_panel_settings_during_upgrade
sudo yunohost config set helloworld max_poll_option_chars -v 666
sudo yunohost app upgrade helloworld -u https://github.com/Tagadda/helloworld_ynh/tree/enh/ynh_add_config/keep_config_panel_settings_during_upgrade
sudo yunohost config get helloworld max_poll_option_chars
```

### DIFF
--- a/helpers/utils
+++ b/helpers/utils
@@ -496,6 +496,11 @@ ynh_add_config() {
         ynh_die --message="The provided template $template doesn't exist"
     fi
 
+    # Backup variables handled by the config panel
+    local configpanel_backup_path
+    configpanel_backup_path="$(mktemp)"
+    yunohost app config get $app --export > "$configpanel_backup_path"
+
     ynh_backup_if_checksum_is_different --file="$destination"
 
     # Make sure to set the permissions before we copy the file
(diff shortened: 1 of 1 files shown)


zamentur:
This draft is fine, but I wonder if we should consider what happens if there are more complex custom setter. 

Technically, there could also be several configuration files, so each call would reload the settings, which could be a bit cumbersome.

We might need an option to deactivate this behaviour ?

zamentur:
Paswrod question type could also be a problem if the getter return nothing, the export won't export the password value i think.

zamentur:
I was just thinking about an issue: what if the config panel is broken ? We probably become unable to upgrade the app no ?

zamentur:
https://github.com/YunoHost/issues/issues/1973

PR #1679, opened 2023-06-20
Simplify handling of ssh_password_authentication 

npalix:
The current template use if/else/endif which introduce spurious empty lines. As the setting value is "yes" or "no", as expected by the configuration file, the value is directly use.

All uses of passwordauthentication are addressed. This adds the one used for the sftp group.

Finally, the global configuration sets the yes and no values to "yes" and "no" respectively.

## The problem

sftp group doesn't use the ssh password setting.
Imperative style introduce useless complexity with it's control-flow.

## Solution

- Propagate the setting
- Use config_global default values

## PR Status

...

## How to test

...

### DIFF
--- a/conf/ssh/sshd_config
+++ b/conf/ssh/sshd_config
@@ -57,11 +57,7 @@ UsePAM yes
 
 # PLEASE: if you wish to force everybody to authenticate using ssh keys, run this command:
 # yunohost settings set security.ssh.ssh_password_authentication -v no
-{% if password_authentication == "False" %}
-PasswordAuthentication no
-{% else %}
-#PasswordAuthentication yes
-{% endif %}
+PasswordAuthentication {{ password_authentication }}
 
 # Post-login stuff
 Banner /etc/issue.net
@@ -103,7 +99,7 @@ Match Group sftp.app,!ssh.app
     AllowStreamLocalForwarding no
     PermitTunnel no
     PermitUserRC no
-    PasswordAuthentication yes
+    PasswordAuthentication {{ password_authentication }}
 
 # root login is allowed on local networks
 # It's meant to be a backup solution in case LDAP is down and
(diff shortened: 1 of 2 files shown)


conf/ssh/sshd_config
```diff
@@ -103,7 +99,7 @@ Match Group sftp.app,!ssh.app
     AllowStreamLocalForwarding no
     PermitTunnel no
     PermitUserRC no
-    PasswordAuthentication yes
+    PasswordAuthentication {{ password_authentication }}
```
alexAubin, line 102: Dunno what to think about this, the point of it being `yes` is for the use-case of, for example, `my_webapp`, which you may want to access via SFTP and only a password is provided ... (+ dunno how convenient/easy it is to configure Filezilla to use an SSH key rather than a password ...)

alexAubin: Right, but if "consistency" equals "nobody understands how to use some features, such as connecting to my_webapp using SFTP because it's not properly documented or is overly complex" then it's still an issue

It's not just about configuring FileZilla, it's also about adding the SSH key to the `my_webapp` user, which is not a trivial operation for not-so-tech-savvy users

npalix: Then, this people will stick with password authentication anyway, or learn how to do with an SSH key if they really want. Most of cloud providers allow key authentication. It's even sometimes the only authentication method
to avoid having (encrypted/hashed) password on the cloud side.

npalix: Another alternative will be to have something similar to webadmin_allowlist for SSH to defined trusted IPs with PermitRootLogin and PasswordAuthentication. This will solved the point of #1551.

alexAubin:
Closing because honestly "meh" ... If we really want to improve SSH security, the first thing to work on would be to integrate the SSH key management in the webadmin

PR #1683, opened 2023-07-03
Allow installation from gitea

orhtej2:
[Gitea](https://about.gitea.com/) has branch URL in form `https://domain.tld/gitea/path/<owner>/<repo>_ynh/src/branch/<branch_name>`.

## The problem
It's impossible to install branch from non-default branch if repo is hosted in Gitea.
...

## Solution
Adjust `APP_REPO_URL` to allow Gitea-specific syntax.
...

## PR Status
Ready I guess :)

...

## How to test
Try installing apps from repo mirrored to Gitea. Should work with default branch, arbitrary branch, tag and commit hash.
...

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -84,7 +84,7 @@ re_app_instance_name = re.compile(
 )
 
 APP_REPO_URL = re.compile(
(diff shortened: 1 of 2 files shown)


alexAubin:
Sounds legit, cheers ! (and thanks for adding tests o/)

orhtej2:
Ooops, it's half-baked as it lacks changes [here](https://github.com/YunoHost/yunohost/blob/42d74bfa3d31983eac19e38c268f685f0c882916/src/app.py#L2478) and proper support for tags/commits (?) [here](                if "ref: refs/heads/" in line:). 

Will set up proper dev env and fix.

PR #1697, opened 2023-08-11
feature: new global setting options to enable TLS passthrough and reverse proxy to external domains

alexAubin:
## The problem

For various scenarios, it would be nice to be able to have SNI-based forwarding of HTTPS traffic (ie without decrypting the traffic) to another machine. For example:
- hosting several Yunohost behind a single IP (or a single YunoHot + other stuff like NAS / whatever, behind a single IP)
- more advanced case such as having a VPN on a YunoHost server on whatever provider, but which serves a purpose of "exposing" a Yunohost server which otherwise cannot be exposed because behind too many layers of networks etc

## Solution

Introduce a new setting to "enable SNI based forwarding" + configure a mapping of domain:IPv4

The whole thing tweaks nginx's configuration and use nginx's ssl_preread stuff : https://nginx.org/en/docs/stream/ngx_stream_ssl_preread_module.html

2023-08-11-230653_1366x768_scrot

## PR Status

Somewhat tested but should be propreply re-tested

Also we should : 
- check that the SNI-forwarded domains are not YunoHost domains
- technical tweak to add in nginx regenconf to forget about old .forward80.conf (cf FIXME in code)
- ??? think about other funky cases idk

## How to test

...


conf/nginx/sni_forward_server.conf
```diff
@@ -0,0 +1,38 @@
...
+        proxy_pass http://{{ sni_forward_ip }};
+
+        proxy_set_header Host $host;
+        proxy_set_header X-Original-URL $scheme://$http_host$request_uri;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Host $http_host;
+        proxy_set_header X-Forwarded-Uri $request_uri;
+        proxy_set_header X-Forwarded-Ssl on;
+        proxy_set_header X-Forwarded-For $remote_addr;
+        proxy_set_header X-Real-IP $remote_addr;
```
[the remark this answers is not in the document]
zamentur: Could be important to know it, cause fail2ban could ban the reverse proxy server...

share/config_global.toml
```diff
@@ -169,3 +169,18 @@ name = "Other"
...
+    name = "SNI-based forwarding"
+
+        [misc.sni_forward.sni_forward_enabled]
+        type = "boolean"
+        default = false
+
+        [misc.sni_forward.sni_forward_list]
+        type = "tags"
+        # Regex is just <domain regex>:<ip regex>
+        pattern.regexp = '^([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+((xn--)?[^\W_]{2,}):((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$'
```
zamentur, line 183: What the problem to accept IPv6 too ?

conf/nginx/sni_forward.conf
```diff
@@ -0,0 +1,27 @@
+{% set domain_ip_map = sni_forward_list.split(',') %}
+stream {
+
+    map $ssl_preread_server_name $name {
+        {% for domain_ip in domain_ip_map %}
+        {{ domain_ip.split(":")[0] }} {{ domain_ip.split(":")[0].replace('.', '') }};
```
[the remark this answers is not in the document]
zamentur: So a way to reduce the risk is to replace by `_` instead of nothing.

However it's not enough
```
domain_eu.org -> domain_eu_org 
domain.eu.org -> domain_eu_org
```

share/config_global.toml
```diff
@@ -169,3 +169,18 @@ name = "Other"
...
+        pattern.error = "You should specify a list of items formatted as DOMAIN:IPv4, such as yolo.test:12.34.56.78"
```
zamentur, line 184: Seems utf8 domains are not supported here (they are supported a bit in the domain part).

zamentur (commented):
As Aleks ping me on the chat, i find some times to make a review.

Feel free to decide to ignore some of my comments...

conf/nginx/server.tpl.conf
```diff
@@ -39,8 +39,12 @@ server {
 }
 
 server {
+    {% if sni_forward_enabled != "True" %}
     listen 443 ssl http2;
     listen [::]:443 ssl http2;
+    {% else %}
+    listen 127.0.0.1:443 ssl http2;
```
zamentur, line 46: Use another port than 443 to avoid conflict with 0.0.0.0:443 ?

zamentur: Or maybe use a socket ?

share/config_global.toml
alexAubin: These are the same regex as our actionsmap.yml

zamentur: indeed ! So i close

share/config_global.toml
zamentur: We also could imagine someone who rent IPv6 VPS (some provider do it for free) and wants to assemble all the thing on a VPS with IPV4

conf/nginx/sni_forward.conf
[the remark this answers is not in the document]
Tagadda: I think you can't register a domain with a `_`.
Usually `_` are used in "special" subdomains (SRV records...)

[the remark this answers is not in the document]
alexAubin:
```suggestion
        {{ domain_ip.split(":")[0] }} {{ domain_ip.split(":")[0].replace('.', '_') }};
```

conf/nginx/sni_forward.conf
```diff
@@ -0,0 +1,27 @@
...
+        {% endfor %}
+        default https_default_backend;
+    }
+    
+    {% for domain_ip in domain_ip_map %}
+    upstream {{ domain_ip.split(":")[0].replace('.', '') }} {
```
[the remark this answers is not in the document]
alexAubin: Marking as resolved because this is clearly an advanced use case and let's focus and having something that works in the nominal case first

share/config_global.toml
alexAubin: Marking as resolved because this is an advanced use case and let's focus on nominal / classic use case first

conf/nginx/server.tpl.conf
alexAubin: (moved to port 444, flagging as resolved)

alexAubin:
(The remaining blocker is to properly configure the IP header thingy)

Tagadda:
IP header thingy, wat dat iz??

alexAubin:
cf https://github.com/YunoHost/yunohost/pull/1697#discussion_r1292130209 and the X-Real-IP header

Basically if you enable SNI-forward, all traffic on the "server B" appears as coming from "server A"'s IP ... which means fail2ban, if triggered, will ban server A entirely.

Nginx has some option somewhere to assume that the real IP is not the IP from the raw packets, but a specific HTTP header, typically `X-Real-IP`

Salamandar:
"Tested" (actually copy pasted) manually on tartare, seems to kinda work ? I think we can actually use 443 instead of 444 because we bind on 127.0.0.1

alexAubin:
Small todos before merging (relevant for the new Yunohost infra)
- rename to TLS passthrough which is a more common description for this ?
- make it visible only if experimental security features are on
- tweak the description to warn that fail2ban won't work on the proxied servers

Tagadda:
Does the FIXME prevent the merge ?

share/config_global.toml
```diff
@@ -180,3 +180,19 @@ name = "Other"
...
+    visible = "security_experimental_enabled"
+
+        [misc.tls_passthrough.tls_passthrough_enabled]
+        type = "boolean"
+        default = false
+
+        [misc.tls_passthrough.tls_passthrough_list]
+        type = "tags"
+        # Regex is just <domain regex>:<ip regex>
+        pattern.regexp = '^([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+((xn--)?[^\W_]{2,}):((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$'
```
Salamandar, line 195: IMHO we should allow domain names (not only ipv4), and maybe port.

Salamandar, line 195: Having : as separator forbids ipv6. Maybe ; or , or space or pipe would be more permissive.

alexAubin: cf https://github.com/YunoHost/yunohost/pull/1697#discussion_r1421757429, it's an experimental feature, the point is having something that works that we can iterate on, and in the nominal case there's an IPv4 available

alexAubin: cf https://github.com/YunoHost/yunohost/pull/1697#discussion_r1421757429, it's an experimental feature, the point is having something that works that we can iterate on, and in the nominal case there's an IPv4 available

Salamandar: cf #1970

Salamandar: cf #1970 ;)

Tagadda:
Issue: when one is redirected to login page by ssowat, port 444 is used in the Location header.

PR #1705, opened 2023-08-25
[fix]  Diagnosis: reverse DNS check should be case-insensitive #2235

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/2235

## Solution

Use lower and assume that r["helo"] is always a string

## PR Status

Untested, but micro

## How to test

...

### DIFF
--- a/src/diagnosers/24-mail.py
+++ b/src/diagnosers/24-mail.py
@@ -43,7 +43,7 @@ class MyDiagnoser(Diagnoser):
     dependencies: List[str] = ["ip"]
 
     def run(self):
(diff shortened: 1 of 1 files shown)


src/diagnosers/24-mail.py
```diff
@@ -132,7 +132,7 @@ def check_ehlo(self):
                     summary=summary,
                     details=[summary + "_details"],
                 )
-            elif r["helo"] != self.ehlo_domain:
+            elif r["helo"].lower() != self.ehlo_domain:
```
zamentur, line 135: Are we sure r["helo"] is always a string ?

PR #1711, opened 2023-09-08
Update fail2ban jail conf

kay0u:
## The problem

`jail.conf` is an old version

## Solution

Update it to the [version used on bookworm](https://sources.debian.org/src/fail2ban/1.0.2-2/config/jail.conf/)

## PR Status

Yolocommitted, not tested

## How to test

...

### DIFF
--- a/conf/fail2ban/jail.conf
+++ b/conf/fail2ban/jail.conf
@@ -105,7 +105,7 @@ bantime  = 10m
 findtime  = 10m
 
 # "maxretry" is the number of failures before a host get banned.
-maxretry = 5
+maxretry = 10
 
 # "maxmatches" is the number of matches stored in ticket (resolvable via tag <matches> in actions).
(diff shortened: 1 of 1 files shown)


conf/fail2ban/jail.conf
```diff
@@ -56,14 +98,17 @@ ignoreip = 127.0.0.1/8
...
+bantime  = 10m
 
 # A host is banned if it has generated "maxretry" during the last "findtime"
 # seconds.
-findtime  = 600
+findtime  = 10m
 
 # "maxretry" is the number of failures before a host get banned.
-maxretry = 10
+maxretry = 5
```
alexAubin, line 108: I think `10` was a feature maybe, like 5 is too harsh, too many people end up being banned and complain about it on the support ...

kay0u, line 108:
```suggestion
maxretry = 10
```

PR #1712, opened 2023-09-10
Add a zsh completion file (automatically generated)

getzze:
## The problem

There is no auto-completion suggestions for ZSH.

## Solution

Follow up on PR #1034.

Add auto-completion to ZSH, by generating that completion file from `share/actionsmap.yml`

As `yunohost` needs to be called with `sudo`, completion may ask for password to autocomplete.

## PR Status

- updated PR #1034 to the latest commit
- tested all completions

## How to test

Run the script with:
```
python3 doc/generate_zsh_completion.py
```
It will create a file `yunohost_zsh_completion` in the `doc` folder.
Move and rename the created file to `/usr/local/share/zsh/site-functions/_yunohost` (or `/usr/share/zsh/vendor-completions/_yunohost`, that is where yunohost installation will install it).
Restart zsh to test the completions.

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -53,6 +53,16 @@ user:
                --fields:
                     help: fields to fetch (username, fullname, mail, mail-alias, mail-forward, mailbox-quota, groups, shell, home-path)
                     nargs: "+"
+                    choices:
+                        - username
+                        - fullname
+                        - mail
+                        - mail-alias
+                        - mail-forward
+                        - mailbox-quota
+                        - groups
+                        - shell
+                        - home-path
 
         ### user_create()
         create:
@@ -91,6 +101,10 @@ user:
                         pattern: &pattern_domain
                             - !!str ^([^\W_A-Z]+([-]*[^\W_A-Z]+)*\.)+((xn--)?[^\W_]{2,})$
                             - "pattern_domain"
+                        autocomplete: &domains_list
+                            ynh_selector: domain list
+                            jq_selector: '.domains[]'
+                            use_cache: false
                 -q:
                     full: --mailbox-quota
                     help: Mailbox size quota
@@ -115,6 +129,10 @@ user:
                     help: Username to delete
                     extra:
                         pattern: *pattern_username
+                        autocomplete: &users_list
+                            ynh_selector: user list --fields username
+                            jq_selector: '.users[].username'
+                            use_cache: false
                 --purge:
                     help: Purge user's home and mail directories
                     action: store_true
@@ -129,6 +147,8 @@ user:
             arguments:
                 username:
                     help: Username to update
+                    extra:
+                        autocomplete: *users_list
                 -F:
                     full: --fullname
                     help: The full name of the user. For example 'Camille Dupont'
@@ -189,6 +209,8 @@ user:
             arguments:
                 username:
                     help: Username or email to get information
+                    extra:
+                        autocomplete: *users_list
 
         ### user_export()
         export:
@@ -203,6 +225,9 @@ user:
                 csvfile:
                     help: "CSV file with columns username, firstname, lastname, password, mail, mailbox-quota, mail-alias, mail-forward, groups (separated by coma)"
                     type: open
+                    extra:
+                        autocomplete:
+                            zsh_completion: _files
                 -u:
                     full: --update
                     help: Update all existing users contained in the CSV file (by default existing users are ignored)
@@ -252,6 +277,10 @@ user:
                             help: Name of the group to be deleted
                             extra:
                                 pattern: *pattern_groupname
+                                autocomplete: &user_groups_list
+                                    ynh_selector: user group list -s
+                                    jq_selector: '.groups[]'
+                                    use_cache: false
 
                 ### user_group_info()
                 info:
@@ -261,7 +290,8 @@ user:
                         groupname:
                             help: Name of the group to fetch info about
                             extra:
-                                pattern: *pattern_username
+                                pattern: *pattern_groupname
+                                autocomplete: *user_groups_list
 
                 ### user_group_add()
                 add:
@@ -272,12 +302,14 @@ user:
                             help: Name of the group to add user(s) to
                             extra:
                                 pattern: *pattern_groupname
+                                autocomplete: *user_groups_list
                         usernames:
                             help: User(s) to add in the group
                             nargs: "*"
                             metavar: USERNAME
                             extra:
                                 pattern: *pattern_username
+                                autocomplete: *users_list
 
                 ### user_group_remove()
                 remove:
@@ -288,12 +320,14 @@ user:
                             help: Name of the group to remove user(s) from
                             extra:
                                 pattern: *pattern_groupname
+                                autocomplete: *user_groups_list
                         usernames:
                             help: User(s) to remove from the group
                             nargs: "*"
                             metavar: USERNAME
                             extra:
                                 pattern: *pattern_username
+                                autocomplete: *users_list
 
                 ### user_group_add_mailalias()
                 add-mailalias:
@@ -304,6 +338,7 @@ user:
                             help: Name of the group to add user(s) to
                             extra:
                                 pattern: *pattern_groupname
+                                autocomplete: *user_groups_list
                         aliases:
                           help: Mail aliases to add
                           nargs: "+"
@@ -323,6 +358,7 @@ user:
                             help: Name of the group to add user(s) to
                             extra:
                                 pattern: *pattern_groupname
+                                autocomplete: *user_groups_list
                         aliases:
                           help: Mail aliases to remove
                           nargs: "+"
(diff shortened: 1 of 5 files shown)


tituspijean:
In a testing directory, I have copied the appropriate files in `doc/generate_zsh_completion.py` and `share/actionsmap.yml`

I'm ending up with the following error:

```zsh
python3 doc/generate_zsh_completion.py
  File "/root/apps/doc/generate_zsh_completion.py", line 720, in <module>
    main()
  File "/root/apps/doc/generate_zsh_completion.py", line 73, in main
    output += make_category(key, value)
  File "/root/apps/doc/generate_zsh_completion.py", line 159, in make_category
    output += make_action(action, action_details).replace("COMMAND", category_name)
  File "/root/apps/doc/generate_zsh_completion.py", line 223, in make_action
    args = make_argument_list(action_map, spaces)
  File "/root/apps/doc/generate_zsh_completion.py", line 273, in make_argument_list
    if argument_details["nargs"] == "+":
KeyError: 'nargs'
```

getzze:
Ah sorry, I made some modifications after running the linters and I left a few bugs. I tested it and it should be working now.

tituspijean:
Thank you so much, it is so great. :)

Upon testing everything:
- [ ] `yunohost user create` completion is empty, but maybe it is expected not to show the optional flags?
- [ ] `yunohost user delete` lists both usernames and their email, making it a bit redundant
- [ ] `yunohost user import` completion outputs `_arguments:463: command not found: _file`
- [ ] `yunohost domain action-run <domain> ` does not suggest `action` nor `-a ARGS`
- [ ] `yunohost domain config get <domain>` lacks further completions (-f, -e, ...)
- [ ] `yunohost domain config set <domain>` lacks further completions (-v, -a, ...)
- [ ] `yunohost app manifest <app> --with-screenshot` should be hidden if possible (API only)
- [ ] `yunohost app install` completion lists already installed apps, I suggest nothing
- [ ] `yunohost app setting <app>`, I do not know how complex it would be to list said settings (from `/etc/yunohost/apps/<app>/settings.yml`)
- [ ] Same thing for `yunohost app config get <app>` and `set`
- [ ] ` yunohost backup download` should be hidden if possible (API only)

The following completions seem broken, though they all fit the same pattern: it looks like they do not have any args, only flags.
- [ ] `yunohost user update <user> -`
- [ ] `yunohost user group list -`
- [ ] `yunohost domain main-domain -`
- [ ] `yunohost domain dyndns set-recovery-password <domain> -`
- [ ] `yunohost app catalog -`
- [ ] `yunohost app list -`
- [ ] `yunohost app info <app> -`
- [ ] `yunohost app map -`
- [ ] `yunohost app change-url <app> -`
- [ ] `yunohost backup create -`
- [ ] `yunohost backup list -`
- [ ] `yunohost backup info <backup> -`
- [ ] `yunohost settings list -`
- [ ] `yunohost settings get <setting> -` and `set`
- [ ] `yunohost firewall list -`
- [ ] `yunohost dyndns subscribe -` and `update`
- [ ] `yunohost tools <args> -`
- [ ] `yunohost hook callback <hook> -`
- [ ] `yunohost log list -`
- [ ] `yunohost log show <log> -`

For reference, here is an example output:

```zsh
yunohost log list -
--limit
--with-details
--with-suboperations
-l
-d
-s
-- Maximum number of operations to list (default to 50)
-- Show additional infos (e.g. operation success) but may significantly increase command tim
-- Include metadata about operations that are not the main operation but are sub-operations
--limit
--with-details
--with-suboperations
-l
-d
-s
-- Maximum number of operations to list (default to 50)
-- Show additional infos (e.g. operation success) but may significantly increase command tim
-- Include metadata about operations that are not the main operation but are sub-operations
--limit
--with-details
--with-suboperations
-l
-d
-s
-- Maximum number of operations to list (default to 50)
-- Show additional infos (e.g. operation success) but may significantly increase command tim
-- Include metadata about operations that are not the main operation but are sub-operations
```

Everything else seems to be working, which is already a lot. ❤️

getzze:
Thanks for testing!
First, about autocompleting or not the flags (starting with `-`), it seems it is the default behavior with zsh:
 - if the next word is an argument (not starting with `-`), only provide arguments autocompletion.
 - if there is no argument left to autocomplete, provide the flags.
 - if you start writing `-`, it provides the flags.

I don't know how to change this default behavior, neither if we want to change it.

Regarding the other points:

> * `yunohost user create` completion is empty, but maybe it is expected not to show the optional flags?
> 
Default behavior of zsh.
> * `yunohost user delete` lists both usernames and their email, making it a bit redundant
>
The documentation says you can use both, but it is indeed redundant for completion.
> * `yunohost user import` completion outputs _arguments:463: command not found: _file
>
I missplelled the command.
> * `yunohost domain action-run <domain>` does not suggest action nor -a ARGS
>
I don't know how to retrieve the list of actions. Is there a function to list them? Or if it's a fixed list, I can add the list of choices to `actionsmap.yml`.
> * `yunohost domain config get <domain>` lacks further completions (-f, -e, ...)
> * `yunohost domain config set <domain>` lacks further completions (-v, -a, ...)
>
I don't know how to retrieve the list of allowed keys. Same as above, is there a function to get this list?
> * `yunohost app manifest <app> --with-screenshot` should be hidden if possible (API only)
>
I can add a flag in `actionsmap.yml` to ignore it for completion.
> * `yunohost app install` completion lists already installed apps, I suggest nothing
>
Sometimes you want to install the same app twice, but I agree it can be confusing.
> * `yunohost app setting <app>`, I do not know how complex it would be to list said settings (from /etc/yunohost/apps/<app>/settings.yml). Same thing for yunohost app config get <app> and set
>
I didn't find a way of completing the next word depending on a previous one, in this case we need to know which app was given in the first argument to know which file to read. But it would be great to be able to do that.
> * `yunohost backup download` should be hidden if possible (API only)
>
Same as above.

Also, for `yunohost hook list <action>`, I ended up writing the list of allowed hooks in `actionsmap.yml` because I didn't find a command to list the allowed hook names, but maybe this function exists.

share/actionsmap.yml
```diff
@@ -37,6 +37,32 @@ _global:
...
+        --debug:
+            help: Log and print debug messages
+            action: store_true
+        --quiet:
+            help: Don't produce any output
+            action: store_true
+        --timeout:
+            help: Number of seconds before this command will timeout because it can't acquire the lock (meaning that another command is currently running), by default there is no timeout and the command will wait until it can get the lock
+            type: int
+
```
alexAubin, line 65: Hmmm I'm confused about this ... I understand that you need some declarative way of finding about these for the zsh autocompletion, but I doubt that `moulinette` really does handle these ? It's gonna be pretty confusing when debugging the action map later ... Also I see that you have a `GLOBAL_ARGUMENTS` which kinda hard-code those in `generate_zsh_completion.py` so i'm confused x_X

alexAubin:
Regarding auto-completing the `-` flag thingy: 

>First, about autocompleting or not the flags (starting with -), it seems it is the default behavior with zsh:
>
>    if the next word is an argument (not starting with -), only provide arguments autocompletion.
>    if there is no argument left to autocomplete, provide the flags.
>    if you start writing -, it provides the flags.
>
>I don't know how to change this default behavior, neither if we want to change it.

It seems that titus is more pointing the issue of rendering : each flag is displayed on a newline instead of all available flags on a single line, plus the output is duplicated/triplicated

share/actionsmap.yml
getzze: Hey, thanks for reviewing. This section was in the actionsmap before and it was removed. I didn't test with `moulinette` actually. I let the `GLOBAL_ARGUMENTS` in the code in case the section is missing in actionsmap, which can happen if the actionsmap file is still an old version. I don't know what's best... I think having the arguments in the `_global` section in actionsmap makes is easier to update them because they are all together. But if it is not working with `moulinette` we can remove the section.

getzze:
> It seems that titus is more pointing the issue of rendering : each flag is displayed on a newline instead of all available flags on a single line, plus the output is duplicated/triplicated

Ah ok, I also have this problem sometimes but it disappear when I start a new session of `zsh`. Can you check if you still see the glitch @tituspijean ?

getzze:
Hey, I just realized this was never merged. It's working fine on my server, what else is needed to merge it?

getzze:
@alexAubin

Salamandar:
I will test it on my setup and make comments afterwards.

Anyways we might want to prevent the code duplication with the bash completion autogeneration code.

getzze:
Great! Let me know if I can help. 
I went to check the bash autogeneration script and it looks very different I don't know if we can share a lot of code. Maybe more with https://github.com/YunoHost/issues/issues/2309

getzze:
I rebased the PR using jinja template (following #2061). It simplifies the script quite a lot.

@Salamandar it's ready for review.

share/actionsmap.yml
getzze: The global arguments were removed from actionsmap, they are now in the jinja template of zsh completion.

Salamandar:
I rebased the branch and squashed some commits.

Salamandar:
Eh, tested, LGTM. Mergeit ™.

Thanks a lot @getzze .

If some completion is missing we can decide to fix it later.

PR #1715, opened 2023-09-25
enh: Handle dyndns subscribe/unsubscribe in web admin

Axolotle:
## Solution

Adapt a bit the subscription for the web admin.  
Most of the changes while adding a domain is about trying to unsubscribe the domain if it is not available and a recovery password is set, before trying to subscribe to it again.

## PR Status

Needs moar tests.
We may have to change dynette's requests limiter value. One is allowed to subscribe/unsubscribe 5 times per hour only.

## How to test

- add a dyndns domain with a recovery password
- check that a password is set in the domain's config panel
- delete the domain without subscribing to it
- try to readd the domain with the recovery password

- add a dyndns domain without recovery password
- set one with the config panel
- delete the domain without subscribing to it
- try to readd the domain with the recovery password


src/domain.py
```diff
@@ -317,6 +317,11 @@ def domain_add(
             pass
         raise e
 
+    if dyndns and dyndns_recovery_password:
+        domain_config_set(
+            domain, "dns.registrar.recovery_password", dyndns_recovery_password
+        )
```
alexAubin, line 323: I'm just starting my review but i'm not sure I get the intent here, the dyndns_recovery_password is already defined during the previous call to `domain_dyndns_subscribe` ?

src/domain.py
```diff
@@ -469,6 +466,11 @@ def domain_remove(
             domain=domain, recovery_password=dyndns_recovery_password
         )
 
+    rm(f"/etc/yunohost/certs/{domain}", force=True, recursive=True)
+    for key_file in glob.glob(f"/etc/yunohost/dyndns/K{domain}.+*"):
+        rm(key_file, force=True)
+    rm(f"{DOMAIN_SETTINGS_DIR}/{domain}.yml", force=True)
```
alexAubin, line 472: Hmmm I'm skeptical about moving this after the regenconf ... or was the intent to run the `dyndns_unsubscribe` before the dyndns key is deleted ?

alexAubin (commented):
LGTM but a few things to clarify

src/domain.py
```diff
@@ -707,6 +709,14 @@ def _apply(self):
...
                     other_app=app_map(raw=True)[self.entity]["/"]["id"],
                 )
+        if (
+            "recovery_password" in self.future_values
+            and self.future_values["recovery_password"]
+            != self.values["recovery_password"]
+        ):
+            domain_dyndns_set_recovery_password(
+                self.entity, self.future_values["recovery_password"]
+            )
```
alexAubin, line 719: Hmmm it looks like in the current code, the recovery password is effectively saved to the file, but we dont want this. It's only meant to be hashed and sent to the API and stored (hashed *again*) on the client side

I don't think we want to validate that the `future_values` is different from the current(?) `values` ...

this is similar to the global settings panel where people can now change their root password but we don't want to save it in cleartext etc ... 

cf https://github.com/YunoHost/yunohost/blob/dev/src/settings.py#L216-L227

I got a bit paranoid in that bit of code, also because there can be some confusion between `new_values`, `future_values` etc...

PR #1716, opened 2023-09-28
Fix ynh_safe_rm: check if target is not a broken symlink before erorring out.

Salamandar:
## The problem

[yunohost/issues#2253](https://github.com/YunoHost/issues/issues/2253)

## Solution

Before erroring out before file doesn't exist, we need to check two things:
* ! test -e target (target doesn't exist, or target is a broken symlink)
* ! test -L target (target is NOT a symlink)

## PR Status

...

## How to test

...

### DIFF
--- a/helpers/helpers.v2.1.d/utils
+++ b/helpers/helpers.v2.1.d/utils
@@ -168,7 +168,7 @@ ynh_safe_rm() {
 
     if [[ -z "$target" ]]; then
         ynh_print_warn "ynh_safe_rm called with empty argument, ignoring."
-    elif [[ ! -e $target ]]; then
+    elif [[ ! -e "$target" ]] && [[ ! -L "$target" ]]; then
         ynh_print_info "'$target' wasn't deleted because it doesn't exist."
     elif ! _acceptable_path_to_delete "$target"; then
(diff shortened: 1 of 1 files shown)


helpers/utils
```diff
@@ -854,33 +854,43 @@ _acceptable_path_to_delete() {
...
+        file=${file%;*}
     fi
-
     if [[ -z "$file" ]]; then
         ynh_print_warn --message="ynh_secure_remove called with empty argument, ignoring."
-    elif [[ ! -e $file ]]; then
+        return
+    fi
+    if [[ "$follow_symlink" -eq 1 ]]; then
+        file=$(readlink -f "$file")
```
kay0u, line 880: So we don't remove the link?

chri2:
A symlink is just a file. My expectation (beside an additional `--follow-symlinks`) would be that the file is deleted no matter whether the symlink is broken or not, points to a socket, directory or whatever. I stumbled over this already...

I started using `rm` for symlinks: https://codeberg.org/flohmarkt/flohmarkt_ynh/issues/52

Salamandar:
@chri2 Yes, I fully agree. The issue described here is not that.
The current issue is that if a symlink is broken, trying to remove it from the helper doesn't work.

PR #1717, opened 2023-09-28
Create versionned directories of the helpers

Salamandar:
## The problem

Having a single helpers directory completely freezes its API, we don't want to change it by fear of breaking old or unmaintained apps.
But sometimes we see bugs, deprecated functions, want to change the overall API, want to introduce new behaviours.

## Solution

This MR introduces a new variable, `YNH_APP_HELPERS_VERSION`.

This variable can be passed as an env var, or as the first arg of the script  (`source /usr/share/yunohost/helpers 2.1`). It defaults to `1`.

Directories are named `helpers.v${VERSION}.d`.
A symlink `helpers.d -> helpers.v1.d` keeps the "legacy" directory.
The version `1` is the current one, so old packages won't break.

## PR Status

Completely untested.

## How to test

...


kay0u (approved):
I like the idea, but I'm not sure that the package version must follow the helper version.
To me it's completely uncorrelated, and we could have a turbo package V5 with helper v2 because it doesn't evolve as fast as the package needs etc.

We could use a key in the manifest, and if it's not declared, use helperV1.

Salamandar:
> I'm not sure that the package version must follow the helper version. To me it's completely uncorrelated, and we could have a turbo package V5 with helper v2 because it doesn't evolve as fast as the package needs etc.

If i understand your idea correctly, you would just like to decorrelate versionning of the 2 ? That i'm fine with. Correlating the two numbers was just a simplification.

> We could use a key in the manifest, and if it's not declared, use helperV1.
Yes that's what I did :)

selfhoster1312:
Great idea!

Why not add a `helpers_version` to manifest.toml so that we can easily detect repo-wide which apps need to be updated for newer versions?

Salamandar:
I think I prefer the first suggestion, because it's less about the format and more about "dependencies" like yunohost version.

PR #1718, opened 2023-09-28
Add home.arpa as special TLD

tituspijean:
OK, I concede it is technically not a TLD, but it fits the special case list for domains.

As per https://www.rfc-editor.org/rfc/rfc8375.html, domains ending by `home.arpa` are expected to solely be used locally.


zamentur (approved):
LGTM

PR #1723, opened 2023-10-03
Add ynh_exec_and_print_stderr_only_if_error that only prints stderr when command fails

Salamandar:
## The problem

npm install triggers this error:

> Error: There's quite a lot of warnings in the output ! If those warnings are coming from some app build step and ain't actual warnings, please redirect them to the standard output instead of the error output ...!

and it's right : it's some stderr we don't care about… Except when it fails!

## Solution

a helper that prints the stderr when the command fails.

## PR Status

untested

### DIFF
--- a/helpers/logging
+++ b/helpers/logging
@@ -186,6 +186,26 @@ ynh_exec_fully_quiet() {
     fi
 }
 
+# Execute a command and redirect stderr in /dev/null. Print stderr on error.
+#
+# usage: ynh_exec_stderr_on_error your command and args
+# | arg: command - command to execute
+#
+# Note that you should NOT quote the command but only prefix it with ynh_exec_stderr_on_error
+#
+# Requires YunoHost version 11.2 or higher.
+ynh_exec_stderr_on_error() {
+    logfile="$(mktemp)"
+    rc=0
+    # Note that "$@" is used and not $@, c.f. https://unix.stackexchange.com/a/129077
+    "$@" 2> "$logfile" || rc="$?"
+    if (( rc != 0 )); then
+        ynh_exec_warn cat "$logfile"
+        ynh_secure_remove "$logfile"
+        return "$rc"
+    fi
+}
+
 # Remove any logs for all the following commands.
 #
 # usage: ynh_print_OFF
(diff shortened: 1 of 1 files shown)


helpers/logging
```diff
@@ -186,6 +186,35 @@ ynh_exec_fully_quiet() {
...
 }
 
+# Execute a command and redirect stderr in /dev/null. Print stderr on error.
+#
+# usage: ynh_exec_stderr_on_error your command and args
+# | arg: command - command to execute
+#
+# Note that you should NOT quote the command but only prefix it with ynh_exec_stderr_on_error
+#
+# Requires YunoHost version 3.2.0 or higher.
```
alexAubin, line 196:
```suggestion
# Requires YunoHost version 11.2 or higher.
```

helpers/logging
```diff
@@ -186,6 +186,35 @@ ynh_exec_fully_quiet() {
...
+ynh_exec_stderr_on_error() {
+    logfile="$(mktemp)"
+    rc=0
+    # Boring legacy handling for when people calls ynh_exec_* wrapping the command in quotes,
+    # (because in the past eval was used) ...
+    # we detect this by checking that there's no 2nd arg, and $1 contains a space
+    if [[ "$#" -eq 1 ]] && [[ "$1" == *" "* ]]
```
alexAubin, line 203: Imho there's no legacy to be handled here because it's a new command, and we should just assume people will use it in the appropriate way, ie just calling `ynh_exec_stderr_on_error foo --bar`

helpers/logging
```diff
@@ -186,6 +186,26 @@ ynh_exec_fully_quiet() {
...
 
+# Execute a command and redirect stderr in /dev/null. Print stderr on error.
+#
+# usage: ynh_exec_stderr_on_error your command and args
+# | arg: command - command to execute
+#
+# Note that you should NOT quote the command but only prefix it with ynh_exec_stderr_on_error
+#
+# Requires YunoHost version 11.2 or higher.
+ynh_exec_stderr_on_error() {
```
alexAubin, line 197:
```suggestion
# Execute a command and redirect stderr in /dev/null. Print stderr on error.
#
# usage: ynh_exec_and_print_stderr_only_if_error your command and args
# | arg: command - command to execute
#
# Note that you should NOT quote the command but only prefix it with ynh_exec_and_print_stderr_only_if_error
#
# Requires YunoHost version 11.2 or higher.
ynh_exec_and_print_stderr_only_if_error() {
```

PR #1726, opened 2023-10-13
form: add options for multiple users and groups

Tagadda:
## The problem

- Sometimes we want to select multiple users and groups

## Solution

Add options to be rendered as tags in the webadmin:
- AppsOption
- UsersOption
- GroupsOption
- UsersAndGroupsOption

## PR Status

Yolocommit

## How to test

...


zamentur:
Multiple attribute are implemented in pydantic v2 PR #2005 .SO we decided to close this one.

PR #1732, opened 2023-10-22
fix typos

rndmh3ro:
## The problem

...

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -61,10 +61,10 @@ class AppResourceManager:
             try:
                 if todo == "deprovision":
                     # FIXME : i18n, better info strings
-                    logger.info(f"Deprovisionning {name}...")
+                    logger.info(f"Deprovisioning {name}...")
                     old.deprovision(context=context)
                 elif todo == "provision":
-                    logger.info(f"Provisionning {name}...")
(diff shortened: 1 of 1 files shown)


src/utils/resources.py
```diff
@@ -90,14 +90,14 @@ def apply(
...
+                        logger.info(f"Reprovisioning {name}...")
                         old.provision_or_update(context=context)
                     elif todo == "provision":
-                        logger.info(f"Deprovisionning {name}...")
+                        logger.info(f"Deprovisioning {name}...")
                         new.deprovision(context=context)
                     elif todo == "update":
                         logger.info(f"Reverting {name}...")
-                        old.provision_or_update(context=context)
+                        old.sion_or_update(context=context)
```
alexAubin, line 100:
```suggestion
                        old.provision_or_update(context=context)
```

ericgaspar:
```suggestion
                        old.provision_or_update(context=context)
```

ericgaspar (changes requested):

src/utils/resources.py
ericgaspar: I am getting old

PR #1733, opened 2023-10-27
Update docker-image-extract

Thovi98:
## The problem

It seems that the current docker-image-extract is not compatible with some more recent docker images.
e.g. https://ci-apps-dev.yunohost.org/ci/job/10596

`No layers returned. Verify that the image and ref are valid.`

A quick search and I see that a fix is made https://github.com/jjlin/docker-image-extract/issues/2

## Solution

Update the docker-image-extract script with the latest version made by `jjlin`

TBH, I don’t know why we took the script by `efrecon`, as it seems more at stale than the proposed one. Therefore I took the liberty to change it.

## PR status

Ready to review. But I don’t know if I did the thing right. I have no idea if some changes have to be made with the `resources` permissions.

### DIFF
--- a/helpers/vendor/docker-image-extract/LICENSE
+++ b/helpers/vendor/docker-image-extract/LICENSE
@@ -1,21 +1,19 @@
-MIT License
+Copyright (c) 2020-2023, Jeremy Lin
 
-Copyright (c) 2021 Emmanuel Frecon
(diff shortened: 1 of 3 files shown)


Tagadda (approved):
LGTM! Yoloreview

PR #1736, opened 2023-11-02
logrotate: fix generated config file perms

OniriCorpe:
## The problem

Logrotate generated config files could have wrong permissions and in this case the log files are never rotated
See:
https://github.com/YunoHost/issues/issues/2242 and https://github.com/YunoHost-Apps/gotosocial_ynh/issues/110

## Solution

Add a chmod to ensure that the generated config file permissions are correct

## PR Status

Could be reviewed and maybe merged
Not tested tho 

## How to test

...

### DIFF
--- a/helpers/fail2ban
+++ b/helpers/fail2ban
@@ -53,6 +53,8 @@
 # You can find some more explainations about how to make a regex here :
 # https://www.fail2ban.org/wiki/index.php/MANUAL_0_8#Filters
 #
+# Note that the logfile need to exist before to call this helper !!
+#
 # To validate your regex you can test with this command:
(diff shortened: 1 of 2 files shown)


Tagadda (approved):

OniriCorpe:
btw this line is useless if the log provided is not a folder but a file:
```bash
mkdir --parents $(dirname "$logfile") # Create the log directory, if not exist
```

if the log is a file, the helper will return an error because logrotate WANTS the file to already exist

so in my packages, I do a `touch LOGFILE` before the helper to get around the problem

should I solve this problem directly in the helper?

PR #1737, opened 2023-11-02
fail2ban: make sure the logfile exists

OniriCorpe:
## The problem

if the logfile doesn't exists (as in a fresh install), the helper will return an error because fail2ban WANTS the file to already exist

so in my packages, I do a `touch LOGFILE` before the helper to get around the problem

but it's better to fix that issue in the core

## Solution

verify if the logfile exists, if not, create it:

```bash
# Create the logfile if it doesn't exist, as fail2ban require an existing file
if [ ! -f "$logpath" ]; then
    touch "$logpath"
fi
```

## PR Status

ready but not tested

## How to test

...

### DIFF
--- a/helpers/fail2ban
+++ b/helpers/fail2ban
@@ -102,6 +102,11 @@ ignoreregex =
     ynh_add_config --template="f2b_jail.conf" --destination="/etc/fail2ban/jail.d/$app.conf"
     ynh_add_config --template="f2b_filter.conf" --destination="/etc/fail2ban/filter.d/$app.conf"
 
+    # Create the logfile if it doesn't exist, as fail2ban require an existing file
+    if [ ! -f "$logpath" ]; then
+        touch "$logpath"
+    fi
+
     ynh_systemd_action --service_name=fail2ban --action=reload --line_match="(Started|Reloaded) Fail2Ban Service" --log_path=systemd
 
     local fail2ban_error="$(journalctl --no-hostname --unit=fail2ban | tail --lines=50 | grep "WARNING.*$app.*")"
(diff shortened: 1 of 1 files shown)


Salamandar:
Dangerous, as this file will then be owned by root:root. The process might not have the permissions to write to it.

OniriCorpe:
> Dangerous, as this file will then be owned by root:root. The app process/service might not have the permissions to write to it.

Thanks, I fixed it :)

helpers/fail2ban
```diff
@@ -102,6 +100,15 @@ ignoreregex =
     ynh_add_config --template="f2b_jail.conf" --destination="/etc/fail2ban/jail.d/$app.conf"
     ynh_add_config --template="f2b_filter.conf" --destination="/etc/fail2ban/filter.d/$app.conf"
 
+    # Create the folder and logfile if they doesn't exist,
+    # as fail2ban require an existing logfile before configuration
+    mkdir -p "/var/log/$app"
+    if [ ! -f "$logpath" ]; then
+        touch "$logpath"
+        chown -R "$app:$app" "$logpath"
+        chmod -R u=rwX,g=rX,o= "$logpath"
```
alexAubin, line 109:
```suggestion
        chown "$app:$app" "$logpath"
        chmod u=rw,g=r,o= "$logpath"
```

i'm a bit uncomfortable with the `-R` and folder-related options if `logpath` is meant to be a file ?

OniriCorpe: ah yes, in fact I foolishly copied the @Salamandar code in https://github.com/YunoHost/example_ynh/pull/216

OniriCorpe: I made the appropriate fix

Salamandar: That is not an issue :
* X (caps) only applies to directories
* -R does not fail if it's a file

OniriCorpe: yes but I made a mistake by indicating only the file and not the folder!
with the folder, the configuration will also be applied to the file (created just before)

PR #1738, opened 2023-11-04
Don't make so many queries for user info when building UserOption form

selfhoster1312:
Solution for https://github.com/YunoHost/issues/issues/2280

Tested on my system. Hopefully works everywhere


Tagadda (approved):

selfhoster1312:
Saves over 1 minute of app installation time on a system with ~150 users :) :)

PR #1739, opened 2023-11-04
Only warn that mailbox is disabled for admins

selfhoster1312:
Possible solution for https://github.com/YunoHost/issues/issues/2226


Tagadda (approved):

alexAubin:
Not sure why we should display this only for admin, actually i don't really understand why this is a warning in the first place ... requalified as debug

PR #1741, opened 2023-11-10
Fix backup error message if archive already exists

mh4ckt3mh4ckt1c4s:
## The problem

When the backup of an app fails for some reason because a backup file with the same name already exists, the error message `A backup archive with this name already exists.`. This error message does not give the name of the already existing backup, preventing the user to try and fix the error by himself (by deleting or moving the already existing backup file for example). 

## Solution

Modify the error message to give the name of the already existing file inside it.

## PR Status

The code is ready to be merged. I modified the french and english translations, I don't know if we need the other translations to be modified before this PR is merged or if translate later will do.

## How to test

- `touch /home/yunohost.backup/archives/test-backup-fix.tar`
- `yunohost backup create -n test-backup-fix  --app adguardhome`
And see the modified error message.


Tagadda (approved):
:rocket:

PR #1743, opened 2023-11-26
fix the fail2ban helper when using using --use_template arg

OniriCorpe:
## The problem

when the fail2ban helper is used with the --use_template argument, the `logpath` variable is not assigned, so there's an unbound error 

thanks to @Psycojoker who helped me to understand the bug ^w^

## Solution

if `logpath` doesn't exist, get the logpath from the template and assign it to the `logpath` variable
replace any `__VAR__` in the logpath by their real variable using ynh_replace_vars

## PR Status

yolo

## How to test

...

### DIFF
--- a/helpers/fail2ban
+++ b/helpers/fail2ban
@@ -103,10 +103,15 @@ ignoreregex =
     # if "$logpath" doesn't exist (as if using --use_template argument), assign
     # "$logpath" using the one in the package's conf/f2b_jail.conf template
     if [ -z "$logpath" ]; then
-        # the first sed deleteds possibles spaces and the second one extract the path
+        # the first sed deletes possibles spaces and the second one extract the path
         logpath=$(grep logpath "$YNH_APP_BASEDIR/conf/f2b_jail.conf" | sed "s/ //g" | sed "s/logpath=//g")
-        # replace the '__APP__' by the real app name in the path
-        logpath=${logpath//__APP__/$app}
+        # replace any '__VAR__' by their real variable using ynh_replace_vars
+        # in some hacky way because ynh_replace_vars supports only files ^^'
+        tempdir="$(mktemp -d)"
+        echo "$logpath" > "$tempdir/process-logpath"
+        ynh_replace_vars --file="$tempdir/process-logpath"
+        logpath="$(cat "$tempdir/process-logpath")"
+        rm -r "$tempdir"
     fi
 
     # Create the folder and logfile if they doesn't exist,
@@ -114,9 +119,10 @@ ignoreregex =
     mkdir -p "/var/log/$app"
     if [ ! -f "$logpath" ]; then
         touch "$logpath"
-        chown -R "$app:$app" "/var/log/$app"
-        chmod -R u=rwX,g=rX,o= "/var/log/$app"
     fi
+    # Make sure log folder's permissions are correct
+    chown -R "$app:$app" "/var/log/$app"
+    chmod -R u=rwX,g=rX,o= "/var/log/$app"
 
     ynh_systemd_action --service_name=fail2ban --action=reload --line_match="(Started|Reloaded) Fail2Ban Service" --log_path=systemd
 
--- a/helpers/fail2ban
+++ b/helpers/fail2ban
@@ -104,14 +104,7 @@ ignoreregex =
     # "$logpath" using the one in the package's conf/f2b_jail.conf template
     if [ -z "$logpath" ]; then
         # the first sed deletes possibles spaces and the second one extract the path
-        logpath=$(grep logpath "$YNH_APP_BASEDIR/conf/f2b_jail.conf" | sed "s/ //g" | sed "s/logpath=//g")
-        # replace any '__VAR__' by their real variable using ynh_replace_vars
-        # in some hacky way because ynh_replace_vars supports only files ^^'
-        tempdir="$(mktemp -d)"
-        echo "$logpath" > "$tempdir/process-logpath"
-        ynh_replace_vars --file="$tempdir/process-logpath"
-        logpath="$(cat "$tempdir/process-logpath")"
-        rm -r "$tempdir"
+        logpath=$(grep logpath "/etc/fail2ban/jail.d/$app.conf" | sed "s/ //g" | sed "s/logpath=//g")
     fi
 
     # Create the folder and logfile if they doesn't exist,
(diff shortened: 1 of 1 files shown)


OniriCorpe:
poke @alexAubin

helpers/fail2ban
```diff
@@ -100,6 +100,15 @@ ignoreregex =
     ynh_add_config --template="f2b_jail.conf" --destination="/etc/fail2ban/jail.d/$app.conf"
     ynh_add_config --template="f2b_filter.conf" --destination="/etc/fail2ban/filter.d/$app.conf"
 
+    # if "$logpath" doesn't exist (as if using --use_template argument), assign
+    # "$logpath" using the one in the package's conf/f2b_jail.conf template
+    if [ -z "$logpath" ]; then
+        # the first sed deleteds possibles spaces and the second one extract the path
+        logpath=$(grep logpath "$YNH_APP_BASEDIR/conf/f2b_jail.conf" | sed "s/ //g" | sed "s/logpath=//g")
+        # replace the '__APP__' by the real app name in the path
+        logpath=$(sed "s/__APP__/$app/g")
```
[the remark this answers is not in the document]
OniriCorpe: uh 😔
what do we do so? :<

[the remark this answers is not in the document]
alexAubin: maybe we could instead check that `$logpath` is defined (i mean in the 'if' later), with eg `if [ -n "${logpath:-}"] && [[ -f "$logpath" ]]` 

and simply ignore the case where logpath is not defined, because those 3 apps should probably not use the "external" template and use the args instead just so we have a single consistent way to define fail2ban stuff accross all apps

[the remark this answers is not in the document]
OniriCorpe: yes, but that would make two different behaviors:
when using `--use_template`, the log file would not be created if needed, whereas when using `--logpath=log_file --failregex=filter`, it would be created.

helpers/fail2ban
```diff
@@ -100,6 +100,15 @@ ignoreregex =
     ynh_add_config --template="f2b_jail.conf" --destination="/etc/fail2ban/jail.d/$app.conf"
     ynh_add_config --template="f2b_filter.conf" --destination="/etc/fail2ban/filter.d/$app.conf"
 
+    # if "$logpath" doesn't exist (as if using --use_template argument), assign
+    # "$logpath" using the one in the package's conf/f2b_jail.conf template
+    if [ -z "$logpath" ]; then
+        # the first sed deleteds possibles spaces and the second one extract the path
+        logpath=$(grep logpath "$YNH_APP_BASEDIR/conf/f2b_jail.conf" | sed "s/ //g" | sed "s/logpath=//g")
+        # replace the '__APP__' by the real app name in the path
+        logpath=${logpath//__APP__/$app}
```
OniriCorpe, line 109: maybe we could use something like [ynh_replace_vars](https://github.com/YunoHost/yunohost/blob/5746b94dff4db4fc7acb12586f1a8e4fab685a18/helpers/utils#L520) there? so that it works with any app variable?

OniriCorpe: I tried something in https://github.com/YunoHost/yunohost/pull/1743/commits/5f7a207f112bae91cf61192cbeae3c9e631db6b0

helpers/fail2ban
```diff
@@ -86,28 +86,43 @@ port = __PORTS__
...
+    if [ -z "$logpath" ]; then
+        # the first sed deletes possibles spaces and the second one extract the path
+        logpath=$(grep logpath "$YNH_APP_BASEDIR/conf/f2b_jail.conf" | sed "s/ //g" | sed "s/logpath=//g")
+        # replace any '__VAR__' by their real variable using ynh_replace_vars
+        # in some hacky way because ynh_replace_vars supports only files ^^'
+        tempdir="$(mktemp -d)"
+        echo "$logpath" > "$tempdir/process-logpath"
+        ynh_replace_vars --file="$tempdir/process-logpath"
+        logpath="$(cat "$tempdir/process-logpath")"
+        rm -r "$tempdir"
```
alexAubin, line 114: In fact I realize that we can probably `grep` directly the destination file, which is rendered right before with the vars and stuff ?

```suggestion
        logpath=$(grep logpath "/etc/fail2ban/jail.d/$app.conf" | sed "s/ //g" | sed "s/logpath=//g")
```

OniriCorpe: hm yes but we should use ynh_replace_vars (or any equivalent) to get a proper logpath, otherwise the path used to `touch` will be incorrect :/

OniriCorpe: maybe something like this?

```bash
tempdir="$(mktemp -d)"
grep logpath "$YNH_APP_BASEDIR/conf/f2b_jail.conf" | sed "s/ //g" | sed "s/logpath=//g" > "$tempdir/process-logpath"
ynh_replace_vars --file="$tempdir/process-logpath"
logpath="$(cat "$tempdir/process-logpath")"
rm -r "$tempdir"
```

alexAubin: Eeeh but the point is that `/etc/fail2ban/jail.d/$app.conf` is generated with `ynh_add_config` which already does "hydrate" the template with the proper `__FOOBAR__` stuff replaced

helpers/fail2ban
```diff
@@ -86,28 +86,36 @@ port = __PORTS__
...
-" >$YNH_APP_BASEDIR/conf/f2b_filter.conf
+" >"$YNH_APP_BASEDIR/conf/f2b_filter.conf"
     fi
 
     ynh_add_config --template="f2b_jail.conf" --destination="/etc/fail2ban/jail.d/$app.conf"
     ynh_add_config --template="f2b_filter.conf" --destination="/etc/fail2ban/filter.d/$app.conf"
 
+    # if "$logpath" doesn't exist (as if using --use_template argument), assign
+    # "$logpath" using the one in the previously generated fail2ban conf file
+    if [ -z "$logpath" ]; then
```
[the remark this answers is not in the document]
alexAubin: Eeeeh but I just tried this script:

```bash
set -eu

function foo() {
    local bar

    if [ -z "$bar" ]; then echo OK; fi
}

foo
```

and got

`test.sh: line 6: bar: unbound variable`

#bashisamess x_x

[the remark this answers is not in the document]
OniriCorpe: wtf 🙃

PR #1745, opened 2023-11-29
apps: add mailbox/IMAP support for apps that declared a system user with mail enabled

Josue-T:
## The problem

Mail reply from app don't work

## Solution

Add missing account registration on postfix for apps

## PR Status

Tested local and it worked

## How to test

Install an app which support fully mail stack and replay by mail.

### DIFF
--- a/conf/dovecot/dovecot.conf
+++ b/conf/dovecot/dovecot.conf
@@ -57,7 +57,7 @@ userdb {
 
 userdb {
   driver = passwd-file
-  args = /etc/dovecot/app-senders-passwd
+  args = username_format=%n /etc/dovecot/app-senders-passwd
 }
 
 protocol imap {
--- a/src/app.py
+++ b/src/app.py
@@ -3267,7 +3267,7 @@ def regen_mail_app_user_config_for_dovecot_and_postfix(only=None):
         if dovecot:
             hashed_password = _hash_user_password(settings["mail_pwd"])
             dovecot_passwd.append(
-                f"{app}:{hashed_password}::::::allow_nets=::1,127.0.0.1/24,local"
+                f"{app}:{hashed_password}::::::allow_nets=::1,127.0.0.1/24,local,mail={mail_user}@{mail_domain}"
             )
         if postfix:
             mail_user = settings.get("mail_user", app)
--- a/src/app.py
+++ b/src/app.py
@@ -3264,14 +3264,15 @@ def regen_mail_app_user_config_for_dovecot_and_postfix(only=None):
         if "domain" not in settings or "mail_pwd" not in settings:
             continue
 
+        mail_user = settings.get("mail_user", app)
+        mail_domain = settings.get("mail_domain", settings["domain"])
+
         if dovecot:
             hashed_password = _hash_user_password(settings["mail_pwd"])
             dovecot_passwd.append(
                 f"{app}:{hashed_password}::::::allow_nets=::1,127.0.0.1/24,local,mail={mail_user}@{mail_domain}"
             )
         if postfix:
-            mail_user = settings.get("mail_user", app)
-            mail_domain = settings.get("mail_domain", settings["domain"])
             postfix_map.append(f"{mail_user}@{mail_domain} {app}")
 
     if dovecot:
--- a/conf/postfix/main.cf
+++ b/conf/postfix/main.cf
@@ -93,14 +93,14 @@ recipient_delimiter = +
 inet_interfaces = all
 
 #### Fit to the maximum message size to 25mb, more than allowed by GMail or Yahoo ####
-# /!\ This size is the size of the attachment in base64. 
+# /!\ This size is the size of the attachment in base64.
 # BASE64_SIZE_IN_BYTE = ORIGINAL_SIZE_IN_MEGABYTE * 1,37 *1024*1024 + 980
 # See https://serverfault.com/questions/346895/postfix-mail-size-counting
 message_size_limit = 35914708
 
 # Virtual Domains Control
 virtual_mailbox_domains = ldap:/etc/postfix/ldap-domains.cf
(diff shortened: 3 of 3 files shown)


alexAubin:
>Install an app which support fully mail stack and replay by mail.

Can you elaborate what app for example ? x_X

It's also not clear to me how this `/etc/postfix/app_accounts_maps` file is supposed to be filled

Josue-T:
By example when we want to be able to replay by email on an issue on gitea or we want to answer by email on discourse. But I seeing that there still be an issue that need a fix.

Josue-T:
So now it works with gitea.

src/app.py
```diff
@@ -3267,7 +3267,7 @@ def regen_mail_app_user_config_for_dovecot_and_postfix(only=None):
         if dovecot:
             hashed_password = _hash_user_password(settings["mail_pwd"])
             dovecot_passwd.append(
-                f"{app}:{hashed_password}::::::allow_nets=::1,127.0.0.1/24,local"
+                f"{app}:{hashed_password}::::::allow_nets=::1,127.0.0.1/24,local,mail={mail_user}@{mail_domain}"
             )
         if postfix:
             mail_user = settings.get("mail_user", app)
```
alexAubin, line 3273: (gotta move that line mail_user and the next one (mail_domain) prior to the 'if' block so it's available in the `if dovecot` too then)

conf/dovecot/dovecot.conf
```diff
@@ -57,7 +57,7 @@ userdb {
 
 userdb {
   driver = passwd-file
-  args = /etc/dovecot/app-senders-passwd
+  args = username_format=%n /etc/dovecot/app-senders-passwd
```
alexAubin, line 60: Hmokay this is very obscure but apparently from the documentation ... 

>username_format=<s>: Look up usernames using this format instead of the full username (%u). If you want to enable [user@domain](mailto:user%40domain) logins but have only user in the file, set this to %n.

Are we sure this ain't going to create issues for app that are already configured to use the username and not user@domain ?

Josue-T: Well, normally this should work. Tested with gitea with this config: https://github.com/Salamandar/gitea_ynh/blob/manifestv2/conf/app.ini#L36-L54

Salamandar: Dovecot is an IMAP server. The config you're linking to is the SMTP config.

For the IMAP config of gitea, please see : https://docs.gitea.com/administration/config-cheat-sheet#incoming-email-emailincoming

Josue-T: > Dovecot is an IMAP server. The config you're linking to is the SMTP config.
> 
> For the IMAP config of gitea, please see : https://docs.gitea.com/administration/config-cheat-sheet#incoming-email-emailincoming

Well I know that dovecot is IMAP server and I configured gitea to use dovecot as you linked the doc (and it worked greatly with this fix and the config that I make on gitea). But you justed droped anything that I did by force pushing on the PR. So yes now linked config is quite not valid...

alexAubin (commented):
I'm still confused about the status of this PR

As far as i understand, there are two different thing:
- the first one adding mailbox maps in postfix, which seems straightforward
- the second one about tweaking app authentication in dovecot to, as far as i understand, use `app@domain` instead of just `app` as login.  This is not clear wether or not this is bacward compatible for apps that currently log using `app` ...

Josue-T:
>  * the second one about tweaking app authentication in dovecot to, as far as i understand, use `app@domain` instead of just `app` as login.  This is not clear wether or not this is bacward compatible for apps that currently log using `app` ...

No using also `$app` as username also work. Anyway with this patch and the following gitea config it work well. But without the patch nothing work.

```

[email.incoming]
ENABLED = true
REPLY_TO_ADDRESS = __APP__+%{token}@__DOMAIN__
HOST = __DOMAIN__
PORT = 993
USE_TLS = true
USERNAME = __APP__
PASSWORD = __MAIL_PWD__
```

I know that it could be just strange but following my test this way the only way to make it working correctly. And in all case imap currently just don't work for apps so there can't by any breaking change...

alexAubin:
> in all case imap currently just don't work for apps so there can't by any breaking change...

Yes there can be regressions, because dovecot is actually the service handling the auth part for SMTP submission (which is the entire point of the `userdb`  using `/etc/dovecot/app-senders-passwd` in dovecot.conf)

I'm still confused because the snippet you're sharing shows `USERNAME = __APP__`, not `USERNAME = __APP__@__DOMAIN__`, but who knows, maybe that just means that internally gitea uses the full email address as username, but that's still pretty confusing and raising eyebrows ...

Josue-T:
Hello,

Well did again more test to be sure about this. So the setting `username_format=%n` are missing I have this error on the mail log when I send email to `gitea@domain.tld`:
```
postfix/pipe[9868]: 15279468124: to=<gitea@domain.tld>, relay=dovecot, delay=0.16, delays=0.11/0/0/0.05, dsn=5.1.1, status=bounced (user unknown)
```

And to be sure that the authentication work (also with `$app`) you can apply the patch and after test with theses command (on my side it worked well):

```bash
openssl s_client -connect localhost:993 -crlf
a login gitea gitea_pwd
a select inbox
a fetch 1 full
a logout
```

Normally you should see the email sent on gitea user. And on my side when I applied the patch it work. But without the patch I've many issues.

alexAubin:
Mokay so after some testing on my side, can confirm that this works and for some reason the `username_format=%n` thing seems to have no impact on SMTP login and instead is necessary for the IMAP mailbox to work. That doesn't make any sense to me neither in term of intuition and documentation, which also seemed to suggest this was to allow login using `gitea@domain`... But whatever, as always the mail stack is just plain madness I guess

PR #1746, opened 2023-11-29
helpers: Further simplify the change url helper

Josue-T:
## The problem

In change url script we need workaround like [this](https://github.com/YunoHost-Apps/seafile_ynh/blob/master/scripts/change_url#L14-L15) because path and domain are the old one and so somethings are set with old values.

## Solution

Set correctly the domain and path with new values.

## PR Status

Patch on helper was tested here: https://github.com/YunoHost-Apps/seafile_ynh/blob/master/scripts/change_url#L27-L52

### DIFF
--- a/helpers/nginx
+++ b/helpers/nginx
@@ -53,19 +53,6 @@ ynh_change_url_nginx_config() {
     local old_nginx_conf_path=/etc/nginx/conf.d/$old_domain.d/$app.conf
     local new_nginx_conf_path=/etc/nginx/conf.d/$new_domain.d/$app.conf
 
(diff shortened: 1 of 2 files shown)


alexAubin:
I'm further simplifying the helper

The whole thing with the trick of "if only the domain change, we only `mv` the config" was supposedly design around the fact that the user may have manually modified the conf

Ironically, in this case we savagely overwrite the checksum, which means that next time `ynh_add_nginx_config`, the file will appear as **not** manually modified and be overwritten anyway

So let's reduce the complexity considering that this the current trick doesn't solve the problem anyway ...

PR #1750, opened 2023-12-09
Allow dots in usernames

alexAubin:
cf https://github.com/YunoHost/issues/issues/2287


Tagadda (approved):

zamentur:
What about the consequences for apps doesn't supporting dot in their internal database ? In general, apps (with ldap integration) maps LDAP uid with the username field.

alexAubin:
Do we know of such apps that would not allow for dots in username ? x_x

zamentur:
> Do we know of such apps that would not allow for dots in username ? x_x
I don't

I succeed to create a user with dot in nextcloud UI but from cli i have "User ID used to login (must only contain a-z, A-Z, 0-9, -, _ and @)" :/

I succeed too on limesurvey.

Anyway, i think on 500 apps, we probably having several apps which doesn't support it... However, all apps doesn't support SSO or LDAP or BASIC AUTH...

alexAubin:
So nextcloud does allow to create user with a dot from its webui (and it works?), but not from the cli ? x_X

zamentur:
> So nextcloud does allow to create user with a dot from its webui (and it works?), but not from the cli ? x_X
i didn't really try on cli, just read the `occ user:create --help` doc...

Tagadda:
At $dayjob we have OwnCloud working with LDAP and users with dots in the username :D

PR #1751, opened 2023-12-14
App resources: be able to use __APP__, __YNH_ARCH__ and __YNH_DEBIAN_VERSION__ in properties

Salamandar:
## The problem

Can't use debian version (bullseye/bookworm) in manifest.toml (for example for apt.extras)

## Solution

Provide more variables to the manifest.toml "environment"

I would like a better way to do that in the future: Every resource should add to the "environment" its variables, to be available to the next resources (although not sure it's the best idea…)

## PR Status

The logic is tested, but not the resource.py

## How to test

...


src/utils/resources.py
```diff
@@ -146,15 +147,22 @@ class AppResource:
...
 
-        for key, value in self.default_properties.items():
-            if isinstance(value, str):
-                value = value.replace("__APP__", self.app)
-            setattr(self, key, value)
+        replacements: dict[str, str] = {
+            "__APP__": self.app,
+            "__YNH_ARCH__": system_arch(),
+            "__YNH_DEBIAN_VERSION__": debian_version(),
+        }
```
alexAubin, line 156: Wokay but this just feels super convoluted to achieve this ... Can't we stick to the current code and add a `.replace()` for `__YNH_ARCH__` and another for `__YNH_DEBIAN_VERSION__` ? x_x

Salamandar: Yes indeed we could.

The idea of my first comment was to have a dict that would be populated resource after resource.

PR #1752, opened 2023-12-15
[fix] First source patches failure doesn't stop app script

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/2296

## Solution
I simply suggest here to die the script on the first patch failure

## PR Status

Tested

## How to test

 1. Install nextcloud with the oldstable branch
 2. Upgrade nextcloud with the master branch
 3. The upgrade should failed due to `Unable to apply patches` and not silently continue to upgrade the app


alexAubin:
So yeah that's great but now it's released in stable and now Nextcloud is broken x_x ... https://forum.yunohost.org/t/installation-nextcloud/27770/13

PR #1753, opened 2023-12-19
Allow dots for username also in import code

YapWC:
Allowed dots to be used in creating username

## The problem
User was not allowed to include "." in creating username like "firstname.lastname"
...

## Solution
Just added a "." in yunohost/src/user.py in line 41
...

## PR Status
Currently figuring out how to test, will update when test are done.
...

## How to test
Create a username with firstname.lastname in yunohost API
...


alexAubin:
Ugh, there's already https://github.com/YunoHost/yunohost/pull/1750 and https://github.com/YunoHost/yunohost-admin/pull/539 x_x

alexAubin:
Though in fact that PR is complementary because i forgot to edit that file for consistency with the other changes :sweat_smile:

PR #1765, opened 2024-01-13
[enh] Add help on ssh port

zamentur:
## The problem

1025+ port could be listen as non root, so an evil user could overload the server to make ssh killed by oom_reaper and next run it's own fake daemon to gain access to the server by collecting password.

## Solution

Advice to use a port under 1024.

## PR Status

Ready

## How to test

...


Salamandar:
Couldn't we just forbid ports above 1024 ?

zamentur:
We could, however a lot of instances already have configured a 1025+ port ...

PR #1772, opened 2024-02-06
Add generales var in config panel like domain, path, etc

Josue-T:
## The problem

- In config panel not all app variables are available and it limit some config file to edit correctly

## Solution

- Add env var as same as for others scripts.

## PR Status

Tested on my side and it provide the necessary variables.

## How to test

Run the config panel script (`config`) and check with `env` command that `domain`, `path`... are provided.


Gredin67:
ping @zamentur

Psycojoker (approved):
Lgtm

PR #1777, opened 2024-02-16
Update sury apt key

Tagadda:
## The problem

- https://forum.yunohost.org/t/something-went-wrong-while-updating-the-cache-of-apt-due-to-sury/15024
- https://github.com/oerdnj/deb.sury.org/issues/2074

## Solution

probably wrong since apt-key is depreciated or something

## PR Status

yolocommit

## How to test

...

### DIFF
--- a/hooks/conf_regen/10-apt
+++ b/hooks/conf_regen/10-apt
@@ -69,6 +69,12 @@ do_post_regen() {
         wget --timeout 900 --quiet "https://packages.sury.org/php/apt.gpg" --output-document=- | gpg --dearmor >"/etc/apt/trusted.gpg.d/extra_php_version.gpg"
     fi
 
+    # Update sury apt key if 95BD4743 is present
+    if apt-key list | grep -q "95BD4743"; then
+        echo "Updating sury apt key..."
+        apt-key del 95BD4743; wget -nv -O - "https://packages.sury.org/php/apt.gpg" | apt-key add -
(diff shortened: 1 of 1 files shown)


hooks/conf_regen/10-apt
```diff
@@ -69,6 +69,12 @@ do_post_regen() {
         wget --timeout 900 --quiet "https://packages.sury.org/php/apt.gpg" --output-document=- | gpg --dearmor >"/etc/apt/trusted.gpg.d/extra_php_version.gpg"
     fi
 
+    # Update sury apt key if 95BD4743 is present
+    if apt-key list | grep -q "95BD4743"; then
+        echo "Updating sury apt key..."
+        apt-key del 95BD4743; wget -nv -O - "https://packages.sury.org/php/apt.gpg" | apt-key add -
+    fi
+
```
alexAubin, line 77:
```suggestion
    # Purge expired keys (such as sury 95BD4743)
    EXPIRED_KEYS="$(LC_ALL='en_US.UTF-8' apt-key list 2>/dev/null | grep -A1 'expired:' | grep -v 'expired\|^-' | sed 's/\s//g')"
    for KEY in $EXPIRED_KEYS; do apt-key del $KEY 2>/dev/null; done
```

Proposing this more general method because my old system has many expired keys (and then we let the other snippet recreate sury's key if needed)

PR #1779, opened 2024-02-21
helpers: ynh_remove_systemd_config: Also remove the systemd service from YunoHost.

Salamandar:
Every app or almost will do that, because it doesn't make sense to remove the systemd config but not the associated yunohost configuration.

This will clean up a bit the remove scripts.

### DIFF
--- a/helpers/systemd
+++ b/helpers/systemd
@@ -29,7 +29,7 @@ ynh_add_systemd_config() {
     systemctl daemon-reload
 }
 
-# Remove the dedicated systemd config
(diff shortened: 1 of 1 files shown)


Tagadda:
We may need for an autopatcher when we merge this PR

Salamandar:
Not really, the "if" test kinda protects from "double calls" to yunohost service remove.

Although agreed, the autopatcher would accelerate the cleanup :D

Salamandar:
(Let's not forget about app version dependencies to yunohost)

PR #1780, opened 2024-02-22
Fix app mail - Add main domain into hosts file

Josue-T:
## The problem

Some apps like synapse need to send email but don't have option to allow invalid certificate. In this case the target server defined on the app must match with the certificate provided by postfix (or dovecot in reception case). But currently it's not possible as postfix (and dovecot) are configured to only accept app connexion on local IP (so 127.0.0.1 or ::1).

So for instance we have a server with main domain `yolo.xyz`.

In synapse to send email we have mainly 2 solutions:

- smtp_server: "localhost" -> this case don't work because the postfix will provide a certificate for `yolo.xyz` but the SMTP client expect to receive a certificate for the configured domain, so `localhost` in this case.
- smtp_server: "yolo.xyz" -> this case don't work too because the connection from SMTP client is made into the external interface (like `eth0`) and postfix don't allow app connection from external interface.

## Solution

To solve this I thought about 2 potential solutions:

1. Add on the hosts file the main domain so the SMTP client configured to point to `yolo.xyz` will resolve this to 127.0.0.1 and so the connection will be with a matching certificate to the configured domain. And on postfix side the connection will be on local interface so it will be happy.
2. Allow app connection from external IP. So we can configure the SMTP client to point to `yolo.xyz`. The provided certificate will match with the configured domain. And if postfix is configured to allow external connection from app there will be no problem.

This PR implement the first solution which I think is the best one.

## PR Status

Tested on my side and it work well.

## How to test

Run regen-conf for dnsmasq and ping main domain and see that ping go directly to `127.0.0.1`.

### DIFF
--- a/hooks/conf_regen/43-dnsmasq
+++ b/hooks/conf_regen/43-dnsmasq
@@ -83,6 +83,11 @@ do_post_regen() {
     short_hostname=$(hostname -s)
     grep -q "127.0.0.1.*$short_hostname" /etc/hosts || echo -e "\n127.0.0.1\t$short_hostname" >>/etc/hosts
 
+    # For SMTP connexion from apps wich request valid certificate we need to force main domain to point to localhost ip
+    # because connexion on external ip is forbinden by mail server for apps.
+    main_domain=$(yunohost domain main-domain --output-as plain)
+    grep -q "127.0.0.1.*$main_domain" /etc/hosts || echo -e "\n127.0.0.1\t$main_domain" >>/etc/hosts
+
     [[ -n "$regen_conf_files" ]] || return 0
 
     # Remove / disable services likely to conflict with dnsmasq


OniriCorpe:
humm yes, but what if an app uses the domain name where it is hosted? and not the ynh main domain?

i think we have to do the same thing for all the installed doamins on ynh

Josue-T:
Well, we don't must use the app domain. We must use a domain which has a valid certificate.

By example:

main_domain: `yolo.xyz`
other domain:
- `toto.net`
- `my.yolo.xyz`
- `matrix.yolo.xyz`

Let's say that synapse is installed on `matrix.yolo.xyz`. We can easily configure synapse to use `yolo.xyz` for SMTP server and it work.

OniriCorpe:
Yes, but it’s a bit less explicit for the user who receives the mail

Josue-T:
> Yes, but it’s a bit less explicit for the user who receives the mail

Well for users it won't change anything as the can still come from `synapse@matrix.yolo.xyz` (if I follow my last example). It's only for packager which will need to configure mail server to main domain.

zamentur:
I don't understand why it's not currently enough. When i ping my main domain i already have ::1 or 127.0.0.1 as answer (and this domain is not listed in `/etc/hosts`).

Are you sure dnsmasq is correctly configured on your test instance ?

> Let's say that synapse is installed on `matrix.yolo.xyz`. We can easily configure synapse to use `yolo.xyz` for SMTP server and it work.

It works, but it probably brokes nextcloud as it's not configured to accept local request  (by default, nextcloud consider request with local ip as potential security attack).

And if the main domain is changed ? This means package should implement hook mechanism to change that...

Josue-T:
> I don't understand why it's not currently enough. When i ping my main domain i already have ::1 or 127.0.0.1 as answer (and this domain is not listed in `/etc/hosts`).
> 
> Are you sure dnsmasq is correctly configured on your test instance ?

Investiguated and on my test intance I didn't had `127.0.0.1` in my resolv.conf. If it set to 127.0.0.1 it work. But if it's not set to this it won't work. But I think we can't expect that all yunohost instance have the `/etc/resolv.conf` set to `127.0.0.1`.

> 
> > Let's say that synapse is installed on `matrix.yolo.xyz`. We can easily configure synapse to use `yolo.xyz` for SMTP server and it work.
> 
> It works, but it probably brokes nextcloud as it's not configured to accept local request (by default, nextcloud consider request with local ip as potential security attack).

Don't really understand as on your setup as you said it won't change anything.

> And if the main domain is changed ? This means package should implement hook mechanism to change that...

Well yes you are true.

So anyway for now let's leave it as it is.

alexAubin:
> I think we can't expect that all yunohost instance have the /etc/resolv.conf set to 127.0.0.1.

The diagnosis checks precisely for this and complains about it if that's not the case

Josue-T:
Closing for now, I'll reopen if I find a good reason to have it.

PR #1785, opened 2024-02-28
[enh] add xz decompression for sources

grosmanal:
## The problem
Forgejo application is a single executable file. It is released compressed in xz format (https://codeberg.org/forgejo/forgejo/releases)
Yunohost does not allow to extract such archives with `ynh_setup_source` helper. The extraction has to be done in `install` script.
Since Yunohost supports `tar.xz` format (and many others), it should supports `xz` format as well.

## Solution
Update `ynh_setup_source` helper to introduce the `xz` format.

## PR Status
Ready to test

## How to test
- get forgejo package source (https://github.com/YunoHost-Apps/forgejo_ynh)
- remove `format` and `extract` configuration from `[resources.sources.main]`
- set `rename` configuration to «forgejo» in `[resources.sources.main]`
- remove extraction line in `scripts/install` (`xz -d "$install_dir/forgejo.xz"`)
- install the app

### DIFF
--- a/helpers/utils
+++ b/helpers/utils
@@ -201,6 +201,9 @@ ynh_setup_source() {
             elif [[ "$src_url" =~ ^.*\.tar\.bz2$ ]]
             then
                 src_format="tar.bz2"
+            elif [[ "$src_url" =~ ^.*\.xz$ ]]
+            then
+                src_format="xz"
             elif [[ -z "$src_extract" ]]
             then
                 src_extract="false"
@@ -329,6 +332,13 @@ ynh_setup_source() {
             unzip -quo $src_filename -d "$dest_dir"
         fi
         ynh_secure_remove --file="$src_filename"
+    elif [[ "$src_format" == "xz" ]]; then
+        if [[ -z "$src_rename" ]]; then
+            xz -d --stdout $src_filename > "$dest_dir/$(basename $src_filename)"
+        else
+            xz -d --stdout $src_filename > "$dest_dir/$src_rename"
+        fi
+
     else
         local strip=""
         if [ "$src_in_subdir" != "false" ]; then


zamentur:
Documentation should be change too:
https://github.com/YunoHost/yunohost/blob/dev/src/utils/resources.py#L318

grosmanal:
> `tar.xz` is already supported but not `xz` alone, should we add `gzip` and `bzip2` (without tar) too ?
I can add it too. It seems a good idea for Yunohost to supports theses formats.

grosmanal:
I'll make a PR for documentation soon (I hope :smile: )

grosmanal:
Alongside xz / gz / bz2 extraction, I propose in this PR to rename the extracted file automaticaly (if `rename` is not set) to : 
- the app name if it's the `main` source
- the source name otherwise

What do you think about this ?

alexAubin:
@grosmanal : it's pretty much already what it does (except that "main" is kept to "main"), because unless you use the `rename` option of the resource, the assumption is that you don't care what the name is because it's gonna be extracted automagically anyway in 99% of cases

helpers/utils
```diff
@@ -329,6 +333,22 @@ ynh_setup_source() {
             unzip -quo $src_filename -d "$dest_dir"
         fi
         ynh_secure_remove --file="$src_filename"
+    elif [[ "$src_format" =~ ^gz|xz|bz2$ ]]; then
+        if [[ -z "$src_rename" ]]; then
+            if [[ "$source_id" == "main" ]]; then
+                local src_rename=$app
```
alexAubin, line 339: I'm confused by this because this will lead to inconsistent behavior between this case and other case (where the main source is just kept as "main" and not "$app") x_X ... What's the rationale for this ?

grosmanal: The goal is to allow the omission of `rename` in resource definition. (I just remember, writing this line, I've mentioned to set `rename` in the «how to test» section of this PR 🥴)

In the the forgejo app case, I thought it was logical to extract the archive to the same name as the application one.

alexAubin:
Closing because support for xz has been implemented since then ... got confused about the "rename" story ... only one app seems to be using bz2 (restic) and can probably handle it manually ... no app found using just .gz ... Feel free to reopen a similar PR with the similar behavior as the existing code for .xz if needed

PR #1800, opened 2024-03-09
ci(autoblacks): switch from using gitlab CI to github actions for autoblacking code

Psycojoker:
## The problem

We've moved all our code base to the autoblacks.yml github action for autoblacking our code but yunohost is still using a gitlab CI job for that.

The gitlab CI job obviously works and there is nothing wrong about it but uniformising everything with one tool is easier to maintain.

## Solution

Use autoblack instead.

## PR Status

Works as expected but I'm waiting for an answer from @kay0u to see if we completely drop the gitlab job or if we convert it into a linting job that fails when black isn't happy.

## How to test

It's a bit hard but you can change the target branch of the script to this MR and push it again and see if a PR is generated but it has already been tested on other projects since quite some time already so testing isn't really needed.

### DIFF
--- a/.github/workflows/autoblack.yml
+++ b/.github/workflows/autoblack.yml
@@ -1,0 +1,35 @@
+name: Check / auto apply Black
+on:
+  push:
+      branches:
+          - dev
+jobs:
+  black:
+    name: Check / auto apply black
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check files using the black formatter
+        uses: psf/black@stable
+        id: black
+        with:
+          options: "."
+        continue-on-error: true
+      - shell: pwsh
+        id: check_files_changed
+        run: |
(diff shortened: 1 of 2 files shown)


kay0u (approved):
LGTM:

I think we can completely remove black from gitlab CI scripts.
We did it only on tags to avoid spam, idk if we want to keep this behavior?

https://github.com/YunoHost/yunohost/blob/415e444f130adc0129c2728adb16068e838a33b6/.gitlab/ci/lint.gitlab-ci.yml#L6-L19

We already a lint (but I'm not sure if it still relevant to allow it fail?) and a invalidcode the check if there is any error in the code

Psycojoker:
Ok, let's totally remove black then!

Psycojoker:
> We did it only on tags to avoid spam, idk if we want to keep this behavior?

I think we can but I'm not sure we want? Idk honestly, it seems fine on other repositories without using it on tags '-'

PR #1805, opened 2024-03-21
[enh] Implement 'yunohost log show last' to display the last log file.

Abperron:
Resolves YunoHost/issues#2047 

## The problem

As said in YunoHost/issues#2047, `yunohost log show` is annoying to use because it's required to know the name of the log file.

## Solution

The path argument can be replaced by the "magic" keywords `last` or `last-X` to display the last or last but X log file (`last-0` means last).

`log_show` now uses the `log_list` function.

We didn't internationalize the error message if 'X' is too big because it can only appear in the CLI for now, which seems not internationalized either.

## PR Status

Done.

## How to test

- Run `yunohost log show last` or `yunohost log show last-1`.


src/log.py
```diff
@@ -173,6 +173,20 @@ def log_show(
         share
     """
 
+    # Set up path with correct value if 'last' or 'last-X' magic keywords are used
+    last = re.match(r"last(?:-(?P<position>[0-9]{1,6}))?$", path)
```
kay0u, line 177: We probably want to match the beginning of the line (with a `^`) to avoid the case where the path ends with "last" for some reason?
```suggestion
    last = re.match(r"^last(?:-(?P<position>[0-9]{1,6}))?$", path)
```

(Not tested yet)

Abperron: `re.match` already searches if the correspondence is at the beggining of the string unlike `re.search`. The `^` isn't necessary for `re.match` but we can use `re.search` instead and add it if you think it's more suitable.

kay0u: No it's perfect like that then

Psycojoker:
I don't understand why you aren't using CLI arguments like "--last/-l" and "--number-of-logs/-N" and parsing a string instead, is there a reason for that?

manor-tile:
We did this because the issue seemed to propose to do this, and it makes sense because the `path` argument already means "which log file to show?".

I also think it's easier to read, remember and write, compare:
```shell
yunohost log show last-1
yunohost log show -l -N 1
yunohost log show --last --number-of-logs 1
```

On the other side using dedicated CLI arguments as your propose is more standard and avoids using a regex, so there is a trade-off between our solution and yours and we are not sure about which is best.

manor-tile:
Would an alternative pull request using dedicated CLI arguments be appreciated? Is there something else we could do to enhance our pull request or resolve the related issue ?

kay0u:
Sorry about the response time, we just need volunteers to test the code and press the green button 🙂

alexAubin:
Cheers and sorry for the delay, bonus point for the regex thing, the named "position" group thing is neat :+1: :ok_hand:

PR #1808, opened 2024-03-25
helpers: run composer as the app user

OniriCorpe:
## The problem

https://github.com/YunoHost/issues/issues/2367
we can't use `ynh_exec_as "$app" ynh_composer_exec` and composer cries about running as root

## Solution

edit the `ynh_composer_exec` to run composer as the `$app` user
we can't choose the user but idk if that would be a good thing: an app has to stay in its own box 

## PR Status

yolodone

## How to test

...

### DIFF
--- a/helpers/php
+++ b/helpers/php
@@ -533,7 +533,7 @@ ynh_composer_exec() {
     fi
 
     COMPOSER_HOME="$workdir/.composer" COMPOSER_MEMORY_LIMIT=-1 \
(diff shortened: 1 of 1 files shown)


alexAubin:
In the meantime the simple solution sounds like just introducing a new option to the helper so that it's backward compatible etc

Salamandar:
@alexAubin do we close that ?

alexAubin:
Zblerg yeah i guess

PR #1809, opened 2024-03-26
[bookworm] Return auth cookie valid for whole main domain.

orhtej2:
## The problem
If app specify private path it's impossible to access as `yunohost.portal` cookie is not forwarded when app is installed on subdomain.

## Solution
Return cookie valid for `.domain.tld`, just like old `SSOWat` did. Come to think about it we should return cookies valid for all domains?

## PR Status
It worked once.

## How to test
Install LimeSurvey from https://github.com/orhtej2/limesurvey_ynh/tree/patch-2 and see if you can access admin.


tituspijean (approved):
Tested and approved

tituspijean:
> Come to think about it we should return cookies valid for all domains?

I think it's to prepare the move to "one domain, one portal" down the line. (https://forum.yunohost.org/t/call-for-feedback-sso-portal-in-multi-domain-context-how-should-it-work/7491)

PR #1811, opened 2024-04-05
helpers: Do not wait for dpkg lock when calling ynh_package_is_installed

Josue-T:
## The problem

Waiting on `ynh_wait_dpkg_free` could be really slow

On a quick machine
```
# time ynh_wait_dpkg_free

real    0m0.183s
user    0m0.059s
sys     0m0.093s
```

On slow machine (like raspi):
```
# time ynh_wait_dpkg_free

real    0m2.223s
user    0m0.602s
sys     0m1.484s
```

When we have multiple call  of `ynh_package_is_installed`  on the same package it could be really slow.

## Solution

From my analyse `ynh_wait_dpkg_free` was added into `ynh_package_is_installed` probably just by generality to ensure that all apt helper wait on dpkg before to run but without a specific need.

I think we can remove this by default and add a specific option to enable it if specific need.

## PR Status

Tested with monitorix https://github.com/YunoHost-Apps/monitorix_ynh/pull/55 and it work well

## How to test

Just run the helper `ynh_package_is_installed` with the option that you want to test.

### DIFF
--- a/helpers/apt
+++ b/helpers/apt
@@ -45,25 +45,19 @@ ynh_wait_dpkg_free() {
 #
 # example: ynh_package_is_installed --package=yunohost && echo "installed"
 #
-# usage: ynh_package_is_installed --package=name [--wait_dpkg_free]
-# | arg: -p, --package=          - the package name to check
-# | arg: -l, --wait_dpkg_free=   - wait for dpkg to be free.
-# |                                Note that waiting on dpkg free could take about 0.2s on quick platform
-# |                                and about 2 seconds on slow platform so in case of multiple call it could be slow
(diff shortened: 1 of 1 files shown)


alexAubin (commented):
Yes indeed, I don't even understand why this is here in the first place ... Naively querying the dpkg database shouldn't require the lock, and this fact it can be confirmed: running an `apt install` in one terminal (hanging asking Y/N), running an `apt install` in another terminal shows apt complaining about the lock, but `dpkg-query` runs fine

I would rather simply remove the call to `ynh_wait_dpkg_free` entirely than adding an option we will never need

Josue-T:
It's ok for me this way, so I remove the option.

PR #1822, opened 2024-05-06
Allow users to access their own domain portal without app permission

selfhoster1312:
Fix for #2384

Removed some indentation levels and added some comments for clarity.

EDIT: tested

### DIFF
--- a/src/authenticators/ldap_ynhuser.py
+++ b/src/authenticators/ldap_ynhuser.py
@@ -41,10 +41,15 @@ SESSION_VALIDITY = 3 * 24 * 3600  # 3 days
 URI = "ldap://localhost:389"
 USERDN = "uid={username},ou=users,dc=yunohost,dc=org"
 
+# Cache on-disk settings to RAM for faster access
 DOMAIN_USER_ACL_DICT: dict[str, dict] = {}
 PORTAL_SETTINGS_DIR = "/etc/yunohost/portal"
 
-
+# Should a user have *minimal* access to a domain?
+# - if the user has permission for an application with a URI on the domain, yes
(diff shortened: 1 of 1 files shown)


src/authenticators/ldap_ynhuser.py
```diff
@@ -54,11 +58,12 @@ def user_is_allowed_on_domain(user: str, domain: str) -> bool:
...
-        else:
-            parent_domain = domain.split(".", 1)[-1]
-            return user_is_allowed_on_domain(user, parent_domain)
-
+        parent_domain = domain.split(".", 1)[-1]
+        return user_is_allowed_on_domain(user, parent_domain)
+    
     mtime = portal_settings_path.stat().st_mtime
+    # TODO: Why would that stored mtime EVER NOT be < time.time()??
+    # What is this condition supposed to check/prevent?
```
selfhoster1312, line 66: I'd really like to have an explanation for this :)

alexAubin (approved):
LGTM unless you want to test if and/or fix the weird `if` mtime thing

selfhoster1312:
Ok i fixed a typo and tested this is all good on my side.

> LGTM unless you want to test if and/or fix the weird if mtime thing

Well i don't understand what it's supposed to do / check ? But i don't understand how a time set in the past could be further into the future than now's time :)

What is the entire condition supposed to check/prevent?

alexAubin:
>What is the entire condition supposed to check/prevent?

It's supposed to cache the relevant info from `{domain}.json` instead of reading it every time. It's generated by `app_ssowatconf()` so it's not likely to change very often

selfhoster1312:
Good for me.

PR #1827, opened 2024-05-08
added SSH key mention in failed ssh connection warning message

OniriCorpe:
## The problem

we advertise changing the ssh port, but using an ssh key is better for the security of the server

## Solution

mention SSH key authentication in failed ssh connection warning message

maybe we want to remove the "use a custom port for SSH" part because it's a pain to use and in 2 seconds you can find the right port by scanning...

tbh that's not good advice

## PR Status

done

## How to test

...

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -234,7 +234,7 @@
     "diagnosis_found_errors": "Found {errors} significant issue(s) related to {category}!",
     "diagnosis_found_errors_and_warnings": "Found {errors} significant issue(s) (and {warnings} warning(s)) related to {category}!",
     "diagnosis_found_warnings": "Found {warnings} item(s) that could be improved for {category}.",
-    "diagnosis_high_number_auth_failures": "There's been a suspiciously high number of authentication failures recently. You may want to make sure that fail2ban is running and is correctly configured, or use a custom port for SSH as explained in https://yunohost.org/security.",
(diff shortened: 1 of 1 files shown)


alexAubin:
>maybe we want to remove the "use a custom port for SSH" part because it's a pain to use and in 2 seconds you can find the right port by scanning...

Well strictly-speaking this exact message is about an usual number of failed auth attempts, which typically comes from bots brute-forcing SSH, and using SSH keys ain't gonna reduce that number, but classically changing the SSH port will, because the vast majority of bots won't scan ports

alexAubin:
Yes and no, because it still means the auth log is unecessarily flooded and growing large, and the mecanism also spots e.g. failed sudo attempts which could be symptomatic of a malicious user or pwned app brute-forcing stuff

OniriCorpe:
hm maybe it would be better to distinguish between "failed ssh auth attempts" and "failed sudo auth attempts" if it's possible?

because for now the message doesn't warn about the sudo thing

alexAubin:
Closing because, yes, there's a whole topic of properly being able to define SSH keys for users and disable password authentication, but it's kind of unrelated to bots brute-forcing SSH (or sudo in the less-likely case where an attacker broke in somehow...) ... changing the port is the classical way of preventing this and also has the side effect to reduce by a good 90~99% the amount of log generated by brute force attacks

PR #1831, opened 2024-05-11
Handle metronome as an app

alexAubin:
## The problem

The way metronome is integrated has always been kind of confusing because it was integrated on every domain including subdomain ... nowadays it's better because one can enable/disable it on specific domain but it remains funky and overkill for people who don't care about XMPP. Also metronome is always installed on every system but it should just be an app

## Solution

Remove metronome, to be replaced by https://github.com/YunoHost-Apps/metronome_ynh once it's ready etc

## PR Status

Draft

## How to test

...


Josue-T:
It would be grate to have a clean feature to manage subdomain for an app. Currently some app use a workaround like [cryptpad](https://github.com/YunoHost-Apps/cryptpad_ynh/blob/master/scripts/install#L41-L42) which is not really perfect.

Ideally we should have in the manifest a way to declare a subdomain. The hook system will work but is probably less documented and  the packager will end by using the hackish `yunohost domain add subdomain.tld` command which is not really clean in the app script.

And to me the hook should be used for specific customization which can't be standardized. On this feature side to me it's a general need which can be standardized.

Note that currently we already have theses app which need this feature and maybe some more that I don't know:
- metronome
- synapse
- crpytpad

alexAubin:
Agreed, but "adding a subdomain" is a fuzzy concept, like "do you want a totally separate nginx config or should it be shared between subdomain and parents", "do you want a shared certificate", "is the name of the subdomain fixed or can it be a variable", etc ... Right now i'm not familiar with the whole thing about cryptpad and synapse so i'm handling the metronome situation with a regenconf hook + dns hook + alternate domain names hook, and it's totally not clear to me wether or not we can standardize it...

Josue-T:
Well,

For the certificate I don't see really a reason to have a separate certificate for subdomain.

For the nginx conf maybe a separate config would be needed in edge case for security, but I think in most of case having the same config would do the job.

And having a fixed subdomain would also be useful in most of cases.

In synapse case, as I'm working on, the need is that we want a subdomain as we have different components which all need the whole domain (which can't be installed in sub path).

For synapse I will probably just copy the hook that you will add on metronome app and adapt it.

Anyway, still be a suggestion, removing metronome from the core still be a big improvement 🎉👍

PR #1836, opened 2024-05-13
fix ldap write access for admin users

aya:
Admin users are unable to modify ldap entries, getting following error: 0x32 (LDAP_INSUFFICIENT_ACCESS)


alexAubin:
Ugh indeed, cheers

PR #1838, opened 2024-05-15
Add redis and ruby helpers

ericgaspar:
This Redis helper seems to be incomplete: no back up

All are vanilla helpers from the repo https://github.com/YunoHost-Apps/Experimental_helpers

PR #1844, opened 2024-05-19
Add mongodb helpers

alexAubin:
Follow-up of https://github.com/YunoHost/yunohost/pull/1838

@orhtej2 was pointing to https://github.com/orhtej2/wekan_ynh/tree/no_keep_version that might be relevant to include in the core

### DIFF
--- a/helpers/mongodb
+++ b/helpers/mongodb
@@ -1,0 +1,355 @@
+#!/bin/bash
+
+# Execute a mongo command
+#
+# example: ynh_mongo_exec --command='db.getMongo().getDBNames().indexOf("wekan")'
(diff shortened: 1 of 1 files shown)


Salamandar:
Should these helpers only be installed by the mongodb app ?

orhtej2:
I hate mongodb app and wish it died in fire.
On more serious note that requires rewriting packages that rely on MongoDB to, instead of managing MongoDB themselves, rely on extra YNH app. Is that what we want?

Salamandar:
> I hate mongodb app and wish it died in fire.

Why is that ? :D

Salamandar:
Considering those helpers are already used all over the place, i'm OK with merging them into the core.

PR #1845, opened 2024-05-19
Remove useless dict lookup, as the keys are always the same two

selfhoster1312:
## The problem

It's not clear what dict += list is.

## Solution

dict += list == dict + list.keys()
(but dict + list is TypeError)

But in fact, we don't need this operation at all because the keys don't change so just remove that part.


selfhoster1312:
This code was rewritten for bookworm so this clarification is useless

PR #1846, opened 2024-05-20
provision: reload the firewall only once

OniriCorpe:
## The problem

for now, during the provisionning/deprovisionning, the firewall is reloaded for each added or removed port 

for example:
```text
2100 INFO Provisioning ports...
3459 SUCCESS Firewall reloaded
4904 SUCCESS Firewall reloaded
```

## Solution

check if the port list has changed at the end of the port provisionning/deprovisionning, and if so: reload the firewall

## PR Status

done

## How to test

do a port provisionning/deprovisionning
not tested tho

### DIFF
--- a/src/firewall.py
+++ b/src/firewall.py
@@ -75,14 +75,11 @@ def firewall_allow(
             "ipv6",
         ]
 
-    changed = False
-
     for p in protocols:
         # Iterate over IP versions to add port
         for i in ipvs:
             if port not in firewall[i][p]:
                 firewall[i][p].append(port)
-                changed = True
             else:
                 ipv = "IPv%s" % i[3]
                 logger.warning(m18n.n("port_already_opened", port=port, ip_version=ipv))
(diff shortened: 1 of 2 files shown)


Salamandar (approved):

src/firewall.py
```diff
@@ -159,20 +149,14 @@ def firewall_disallow(
...
             if port in firewall[i][p]:
                 firewall[i][p].remove(port)
-                changed = True
             else:
                 ipv = "IPv%s" % i[3]
-                if not reload_only_if_change:
-                    logger.warning(
-                        m18n.n("port_already_closed", port=port, ip_version=ipv)
-                    )
+                logger.warning(m18n.n("port_already_closed", port=port, ip_version=ipv))
```
alexAubin, line 159: Ugh actually we probably want to have an option to disable this warning, otherwise everytime a port is not to be "exposed" (= the vast majority of the apps ?), yunohost will display a warning during install/upgrade/... when provision ports

Similar stuff for `port_already_open` too

alexAubin:
Closing because it seems to have been implicitly fixed by the firewall refactoring, cf for example [prosody on the CI](https://ci-apps.yunohost.org/ci/job/23263) which requires several ports, and there's a single message during install about nftable reload, and no message during upgrade. Assuming it's fixed

PR #1849, opened 2024-05-23
do not enable a service disabled by the user

OniriCorpe:
## The problem

if the admin disabled a service using `yunohost service disable $app`, an upgrade will enable it again

## Solution

check in the systemd helpers if the service has been deactivated before activating or before starting it up 

## PR Status

done

## How to test

...

### DIFF
--- a/helpers/systemd
+++ b/helpers/systemd
@@ -25,7 +25,10 @@ ynh_add_systemd_config() {
 
     ynh_add_config --template="$template" --destination="/etc/systemd/system/$service.service"
 
-    systemctl enable $service --quiet
+    # do not enable a service disabled by the user
+    if ! yunohost service status $service | grep -q "start_on_boot: disabled"; then
+        systemctl enable $service --quiet
+    fi
     systemctl daemon-reload
 }
 
@@ -85,6 +88,11 @@ ynh_systemd_action() {
 
     # Manage case of service already stopped
     if [ "$action" == "stop" ] && ! systemctl is-active --quiet $service_name; then
+        return 0
+    fi
+
+    # do not start a service disabled by the user
+    if [ "$action" == "start" ] && ! yunohost service status $service_name | grep -q "start_on_boot: disabled"; then
         return 0
     fi
 


helpers/systemd
```diff
@@ -25,7 +25,10 @@ ynh_add_systemd_config() {
 
     ynh_add_config --template="$template" --destination="/etc/systemd/system/$service.service"
 
-    systemctl enable $service --quiet
+    # do not enable a service disabled by the user
+    if ! yunohost service status $service | grep -q "start_on_boot: disabled"; then
+        systemctl enable $service --quiet
+    fi
```
alexAubin, line 31: Hmmm I think that part should be entirely removed ... the "start_on_boot" thing is precisely the systemd's enabled/disabled flag, cf https://github.com/YunoHost/yunohost/blob/bookworm/src/service.py#L433

so this code is basically "if the service is enabled, enable it"


```suggestion
```

helpers/systemd
```diff
@@ -88,6 +91,12 @@ ynh_systemd_action() {
         return 0
     fi
 
+    # do not start a service disabled by the user
+    if [ "$action" == "start" ] || [ "$action" == "reload_or_restart" ] || [ "$action" == "restart" ] && \
+    ! yunohost service status $service_name | grep -q "start_on_boot: disabled"; then
+        return 0
+    fi
```
alexAubin, line 98:
```suggestion
    # do not start a service disabled by the user
    if ([ "$action" == "start" ] || [ "$action" == "reload_or_restart" ] || [ "$action" == "restart" ]) && \
    ! systemctl --quiet is-enabled $service_name; then
        echo "$service_name is disabled, therefore skipping action $action"
        return 0
    fi
```

helpers/systemd
OniriCorpe: So how can we do to not re-enable an intentionally disabled service on update?

alexAubin: We just remove the original `systemctl enable $service` like you already did ? ;P

OniriCorpe: hummm yes x)
why is he here in the first place?

alexAubin: Because we did not really think about the use case of "the user wants the service to be disabled" i think

OniriCorpe: ah, but now if a packages uses ynh_add_systemd_config at installation, the service will not be enabled :/

OniriCorpe: the issue is during upgrade, not install

alexAubin:
Closing because it's too ambiguous ... there may be cases where the service is disabled but started, or enabled but broken, and the admin (or maybe even the script) expects the service to be restarted (and fixed if it was broken)

We have no clear mechanism to distinguish between "the service was not enabled by the system" versus "the admin do not want the system to be started at boot" ... we'd probably need some sort of setting somehow that would unambiguously state "i'm the admin and do not want this service to start at boot" such that in this case it would skip the "enable" instruction in the helper

But closing because the PR is inactive for more than a year and the current implementation doesn't seem appropriate x_x

It would also help to get the full story behind the initial issue / PR

PR #1851, opened 2024-05-26
helpers: Add a --jinja option to ynh_add_config

alexAubin:
## The problem

cf https://github.com/YunoHost-Apps/Experimental_helpers/pull/111

## Solution

add a `--jinja` option to `ynh_add_config` to render the template using jinja. All bash variables are exported in a subshell. The rendering is done using `j2cli` which is available in debian since bullseye

## PR Status

Tested using new unit tests

## How to test

Try rendering a template using `ynh_add_config --template=foo --destination=bar --jinja`

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -27,7 +27,7 @@ Depends: ${python3:Depends}, ${misc:Depends}
  , rspamd, opendkim-tools, postsrsd, procmail, mailutils
  , redis-server
  , acl
- , git, curl, wget, cron, unzip, jq, bc, at, procps
+ , git, curl, wget, cron, unzip, jq, bc, at, procps, j2cli
  , lsb-release, haveged, fake-hwclock, equivs, lsof, whois
 Recommends: yunohost-admin
  , ntp, inetutils-ping | iputils-ping
--- a/helpers/utils
+++ b/helpers/utils
@@ -489,13 +489,15 @@ ynh_local_curl() {
 # Requires YunoHost version 4.1.0 or higher.
 ynh_add_config() {
     # Declare an array to define the options of this helper.
-    local legacy_args=tdv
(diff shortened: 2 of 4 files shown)


Josue-T (approved):
Pushed some improvements on doc side.

Tested on one of my apps it look good.

Thanks

OniriCorpe (approved):

Salamandar:
Actually I got a request, here in pseudocode :
```
if filename.endswith(".j2"):
    jinja = True
```
That would help using jinja2 for nginx/systemd/etc without having to add --jinja arguments to all the relevant helpers.

alexAubin:
>That would help using jinja2 for nginx/systemd/etc without having to add --jinja arguments to all the relevant helpers.

Weeeeell yeah but on the other hand we should be careful and think about what we want exactly ... The current templating is okay-ish and I'm not sure there's a huge gain in moving to jinja for the vast majority of cases, and I'm not sure the syntax is gonna be "cleaner". For example for nginx conf, implementing the whole thing about `{% if path != "/" %}` and replacing `__PATH__/` with either `$path/` or just `/` depending of the install-on-root vs install-on-subpath etc is not going to be way more elegant or whatever.

Or to rephrase, I don't see the gain of reworking all the configs of every app. A lot of energy went into homogeneizing things, I don't feel like seeing a bunch of apps switching every conf to jinja2 just for the sake of it and ending up with heterogenous practices with no benefits

To me this PR is about specific app configurations which are too complex to be handled by the "simplified templating" mechanism

Josue-T:
Hello,

Well, in any case I'm to merge this PR as it is. It still be already a good improvement. The discussion about this can be later, more on packaging v3 side.

PR #1853, opened 2024-05-27
[WIP] Add new versions of helpers. Based on bash-modules.

Salamandar:
## The problem

helpers v1

## Solution

helpersv2

## PR Status

WIP

## How to test

...

### DIFF
--- a/helpers/helpers.v2.d/apps
+++ b/helpers/helpers.v2.d/apps
@@ -1,0 +1,169 @@
+#!/bin/bash
+
+ynh::apps::list() {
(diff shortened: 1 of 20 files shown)


alexAubin:
Closing because the idea is interesting but it's mainly a POC / discussion about what syntax we want for helpers ... But sounds like we ain't gonna have a "helpers 2.2" just for switching from `foo_bar` to `foo::bar` syntax ... The goal with packaging v3 is to severely reduce the bash codebase and we need to iterate on this after we have a clear view of how many helpers remain etc. I linked this PR in the packaging v3 todolist in https://github.com/YunoHost/yunohost/pull/2070

PR #1858, opened 2024-05-30
Implement config panel for swap management

tituspijean:
## The problem

Swap management is non existent on YunoHost, and we historically let apps set their own, which *has made many core dev very angry and has been widely regarded as a bad move.*

## Solution

- Shamelessly take and adapt the experimental swap helpers
- Implement a config panel

## PR Status

Tested, works for me. Please try :)

## How to test

...


tituspijean:
Some annoying warnings appear in the logs, does anyone know how to mask them?

image

zamentur:
I have had a swappiness configuration to avoid to write on ssd if not necessary (in order to extend its service life).

I wonder if we should adapt the code in order to support do_pre_regen and be able to --dry-run or to detect this file is managed by yunohost. However, if someone change the swappiness value, i guess it's for good reasons and we don't need to remember the user that it's configured like that.

hooks/conf_regen/53-swap
```diff
@@ -0,0 +1,31 @@
...
+do_post_regen() {
+    swapfile_enabled="$(yunohost settings get 'swap.swapfile.swapfile_enabled')"
+    swapfile_size="$(yunohost settings get 'swap.swapfile.swapfile_size')"
+
+    if [ ${swapfile_enabled} -eq 1 ]; then
+        # If a swapfile is requested
+        if [ -f /swap_file ] && [ $(stat -c%s /swap_file) -ne $swapfile_size ]; then
+            # Let's delete it first if it is not of the requested size
+            ynh_print_info --message="Deleting swap first..."
+            ynh_del_swap
```
Salamandar, line 18: That's really dangerous. Let's imagine this situation:
* RAM is 1GB, usage 900MB, Swapfile is 1GB, usage 500MB
* you want a 2GB swap file now, so we fall into this code
* ynh_del_swap deletes the original swap file, making the server (or services) most probably crash from OOM (1.4GB required on a 1GB available ram)

helpers/helpers.v1.d/hardware
Salamandar, line 1: I completely rewrote those helpers, see https://github.com/YunoHost/yunohost/pull/1859, I'll change my PR's target branch to yours :)

helpers/helpers.v1.d/hardware
```diff
@@ -103,3 +103,109 @@ ynh_require_ram() {
         return 0
     fi
 }
+
+# Add swap
+#
+# usage: ynh_add_swap --size=SWAP in MiB
+# | arg: -s, --size= - Amount of SWAP to add in MiB.
+ynh_add_swap () {
```
Salamandar, line 111: We need an additional helper, that would "just" call `yunohost settings {get,set} 'swap.swapfile.swapfile_size'` and make `new_swap_size = current_swap_size + requested_app_swap_size`.

Salamandar (changes requested):

hooks/conf_regen/53-swap
Salamandar: Two better solutions would be :
* just create a second swap file of size = difference (in my example, 2GB(total) - 1GB(existing) = 1GB new swap file)
* create a second swap file of the total requested swap size, then afterwards delete the old file. The danger being the disk space but eh, i think we are okay, ram/swap is way smaller than disks usually…

Salamandar:
All in all… I would prefer if swap file creation was NOT linked to apps requesting swap especially, but rather to the RAM usage declared in the apps manifest.

hooks/conf_regen/53-swap
tituspijean: Ah good, thank you for giving some thoughts into this. :)
I had added a very lazy "Beware, if you reduce its current size, you may crash your server." in the config panel, but indeed there are other ways to crash everything.
I think your proposal is good!

tituspijean:
I completely agree, that's why this is a system config panel, not a general apps config panel.
Maybe we could add a warning that's displayed after we calculate the sum of all requested RAM?

Salamandar:
> I completely agree, that's why this is a system config panel, not a general apps config panel. Maybe we could add a warning that's displayed after we calculate the sum of all requested RAM?

Yeah, something like a button on the app installation page redirecting to the swap config panel you're adding. We already have a warning if the RAM usage reported by the manifest is too high.

helpers/helpers.v1.d/hardware
```diff
@@ -103,3 +103,127 @@ ynh_require_ram() {
...
+    local usable_space_kb=$(( free_space_kb / 2 ))
+
+    # Compare the available space with the size of the swap.
+    # And set a acceptable size from the request
+    size_space_ratio=$(( swap_max_size_kb / usable_space_kb + 1 ))
+    if (( size_space_ratio > 4)); then
+        ynh_print_warn --message="Not enough space for a swap file at $swapfile_path."
+        return
+    fi
+    swap_size_kb=$(( swap_max_size_kb / size_space_ratio ))
```
zamentur, line 161:
```suggestion

    local free_space_kb
    free_space_kb=$(df --block-size=K --output=avail "$swapfile_dir" | sed 1d | sed -e 's/K$//')
    # We don't want to fill the disk with a swap file.
    local usable_space=$(( free_space_kb / 1024 ))

    # Compare the available space with the size of the swap.
    if (( size > usable_space )); then
        ynh_print_warn --message="Not enough space for a swap file at $swapfile_path."
        return
    fi
    swap_size_kb=$(( size * 1024 ))
```

PR #1863, opened 2024-06-03
helpers: goenv is broken when checking out latest master commit.

Salamandar:
Instead, checkout the latest commit. Same for xxenv-latest.


Salamandar:
Tested, apps install properly now. @alexAubin i'm not sure of the "release process" and regeneration of the lxd/incus images ?

PR #1870, opened 2024-06-19
Helpers v2.1 : download composer globally in /opt/yunohost/composer

Salamandar:
* download composer globally in /opt/yunohost/composer
* ~~replace --full_replace with --no_clean~~ (disappeared?)
* ~~ynh_in_ci~~ (included in helpers 2.1)
* ~~ynh_setup_source: change the path to the patches~~
* ~~ynh_setup_source: patch failure is now fatal in CI.~~

### DIFF
--- a/helpers/helpers.v1.d/sources
+++ b/helpers/helpers.v1.d/sources
@@ -267,8 +267,11 @@ ynh_setup_source() {
         for patchfile in "$patches_folder/"*.patch; do
             echo "$patchfile"
             if ! patch --strip=1 < "$patchfile"; then
-                ynh_print_warn --message="Warn your packagers /!\\ patch $patchfile failed to apply"
-            fi
+                if ynh_in_ci_tests; then
+                    ynh_die --message"Patch $patchfile failed to apply!"
+                else
+                    ynh_print_warn --message="Warn your packagers /!\\ patch $patchfile failed to apply"
+                fi            fi
         done
         popd
     fi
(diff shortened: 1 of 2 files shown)


alexAubin:
>download composer globally in /opt/yunohost/composer

Why not but maybe let's use `/opt/composer` to be consistent with nodejs, ruby, go ?

Also I'm a bit perplexed because I do think some app call composer from some intermediate tool such as roundcube ? (https://github.com/YunoHost-Apps/roundcube_ynh/blob/master/scripts/upgrade#L190) soooo @_@ 

>change the path to the patches

+1 but i need to adapt the autopatch script to handle this properly

Salamandar:
> Why not but maybe let's use `/opt/composer` to be consistent with nodejs, ruby, go ?

That would be fine with me.
> 
> Also I'm a bit perplexed because I do think some app call composer from some intermediate tool such as roundcube ? (https://github.com/YunoHost-Apps/roundcube_ynh/blob/master/scripts/upgrade#L190) soooo @_@
> 

their update.sh contains this:
```
function find_composer()
{
    if (is_file(INSTALL_PATH . 'composer.phar')) {
        return 'php composer.phar';
    }

    foreach (['composer', 'composer.phar'] as $check_file) {
        $which = trim(system("which $check_file"));
        if (!empty($which)) {
            return $which;
        }
    }

    return null;
}
```
So I could change this helper to put composer/composer.phar in the PATH.

helpers/helpers.v1.d/sources
```diff
@@ -261,16 +261,19 @@ ynh_setup_source() {
     fi
 
     # Apply patches
-    if [ -d "$YNH_APP_BASEDIR/sources/patches/" ]; then
```
alexAubin, line 264: (done in `dev` but this change is on the .v1 helper, not 2.1)

alexAubin:
Closing because hmpf it's not a worth change and it's fine to have the composer file in the app folder, there's no specific gain in making it a system-wide common thing

PR #1875, opened 2024-06-21
Mongo support for Bookworm & avoid reinstall

gcollin:
## The problem

Only Mongo v7.0 is available for Bookworm. So all applications installing Mongo fails the Bookworm install.
As well can install an older version over a newer one if requested by an application.

## Solution
Forced installation of v7 of Mongo if bookworm is detected.
Detects the version of Mongo installed and ensure no regression in installation.
...

## PR Status

...

## How to test

...


alexAubin:
Related notes from @orhtej2 : https://github.com/YunoHost-Apps/wekan_ynh/pull/284#issuecomment-2106337856

but it's not clear to me in the end if an explicit migration is need to go from 4.x to 5 / 6 / 7.x or not ? Possibly we could somewhat magically upgrade mongo and run the appropriate migration stuff similar to what we already do for postgresql

gcollin:
As per the official upgrade documentation, when upgrading Mongo, you just need to ensure step by step upgrade.
https://www.mongodb.com/docs/manual/release-notes/7.0-upgrade-standalone/

And then set the features compatibility to the latest version **only if you want** by running 
db.adminCommand( { setFeatureCompatibilityVersion: "6.0" } )

So for me it's quite safe to upgrade the mongo version whenever it's not available in the target debian release.

What I can do is check that we're not trying to upgrade more than 2 versions deep, and fail the install if so.

zamentur:
Dear contributors,

PR #1882, opened 2024-06-28
[helpers v2.1]: ynh_remove_systemd_config: Also remove the systemd service f…

Salamandar:
…rom YunoHost.

Every app or almost will do that, because it doesn't make sense to remove the systemd config but not the associated yunohost configuration.

This will clean up a bit the remove scripts.

## The problem

...

## Solution

...

## PR Status

...

## How to test

...


alexAubin:
Closing because hmpf it makes the helper not symetric with respect to the "add" version and it's not really worth it compared to moving forward to packaging v3

PR #1886, opened 2024-06-29
automatically ignore the service in diagnosis if it has been deactivated with the ynh cli

OniriCorpe:
## The problem

if the user deactivate some services, the diagnosis will raises errors because the service is not 	active

that's dumb, if the user deactivated a service using the ynh cli, ynh should know the service is deactivated on purpose

## Solution

automatically ignore the service in diagnosis if it has been deactivated with the ynh cli

## PR Status

done

## How to test

deactivate a service, see if the diagnosis raises an error, then re-activate the service

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -26,6 +26,7 @@ from glob import glob
 from datetime import datetime
 
 from moulinette import m18n
+from yunohost.utils.diagnosis import diagnosis_ignore, diagnosis_unignore
 from yunohost.utils.error import YunohostError, YunohostValidationError
 from moulinette.utils.process import check_output
 from moulinette.utils.log import getActionLogger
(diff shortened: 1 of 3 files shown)


OniriCorpe:
don't know if it's worth adding a `logger.success` for this??

src/service.py
```diff
@@ -296,6 +297,9 @@ def service_enable(names):
         names = [names]
     for name in names:
         if _run_service_command("enable", name):
+            services = _get_services()
+            if name in services:
+                diagnosis_unignore({"services": [{"service": name}]})
```
alexAubin, line 302: Eh apparently it will raise an exception if the filter does not exists ... https://github.com/YunoHost/yunohost/blob/6ed167bfafc4c4f27fa463790fcfd76960b96564/src/diagnosis.py#L325 maybe we should tweak the original code because it feels unecessary to raise an exception in that case. The "This filter already exists." one only shows a warning

OniriCorpe: done ^w^
[`20741c6` (#1886)](https://github.com/YunoHost/yunohost/pull/1886/commits/20741c63aafe11f4443319a35dbfb4b00732c6ca)

PR #1888, opened 2024-06-30
Add a workaround for warnings issued for yarn when adding extra repo …

Salamandar:
## The problem

> `21582 WARNING W: Target Packages (main/binary-amd64/Packages) is configured multiple times in /etc/apt/sources.list.d/fab-manager.list:1 and /etc/apt/sources.list.d/yarn.list:1`

## Solution

don't install the repository when it's yarn

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -1214,6 +1214,15 @@ class AptDependenciesAppResource(AppResource):
             key: values for key, values in self.extras.items() if values["packages"]
         }
 
+        # Yarn repository is now provided by the core.
+        # Let's "move" any extra apt resources depending on yarn to the standard packages list.
+        for key in list(self.extras.keys()):
+            if self.extras[key]["repo"] == "deb https://dl.yarnpkg.com/debian/ stable main" \
+            and self.extras[key]["packages"] == [ "yarn" ]:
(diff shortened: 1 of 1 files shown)


src/utils/resources.py
```diff
@@ -1214,6 +1214,15 @@ def __init__(self, properties: Dict[str, Any], *args, **kwargs):
             key: values for key, values in self.extras.items() if values["packages"]
         }
 
+        # Yarn repository is now provided by the core.
+        # Let's "move" any extra apt resources depending on yarn to the standard packages list.
+        for name, values in self.extras.items():
+            if values["repo"] == "deb https://dl.yarnpkg.com/debian/ stable main" \
+            and values["packages"] == [ "yarn" ]:
+                self.packages.append("yarn")
+                del self.extras[name]
```
alexAubin, line 1223:
```
>>> D = {"foo": "foo", "bar": "bar"}
>>> for k, v in D.items():
...    print(k, v)
...    del D[k]
... 
foo foo
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
```

:grimacing:

Salamandar: Aaaaah we need to iterate over the keys instead…

PR #1913, opened 2024-07-17
Fix "log list" : use root_dir for iglob / make sure we use absolute paths

kay0u:
## The problem

https://github.com/YunoHost/issues/issues/2421

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/src/log.py
+++ b/src/log.py
@@ -89,7 +89,8 @@ def _update_log_parent_symlinks():
 
     logs = glob.iglob("*" + METADATA_FILE_EXT, root_dir=OPERATIONS_PATH)
     for log_md in logs:
-        if os.path.getctime(os.path.join(OPERATIONS_PATH, log_md)) < one_year_ago:
+        log_file = os.path.join(OPERATIONS_PATH, log_md)
+        if os.path.getctime(log_file) < one_year_ago:
             # Let's ignore files older than one year because hmpf reading a shitload of yml is not free
             continue
 
@@ -100,7 +101,7 @@ def _update_log_parent_symlinks():
 
         try:
             metadata = (
-                read_yaml(log_md) or {}
+                read_yaml(log_file) or {}
             )  # Making sure this is a dict and not  None..?
         except Exception as e:
             # If we can't read the yaml for some reason, report an error and ignore this entry...
--- a/src/log.py
+++ b/src/log.py
@@ -96,7 +96,7 @@ def _update_log_parent_symlinks():
 
         name = log_md[: -len(METADATA_FILE_EXT)]
         parent_symlink = os.path.join(OPERATIONS_PATH, f"{name}.parent.yml")
-        if not os.path.exists(parent_symlink) or os.path.islink(parent_symlink):
+        if (os.path.islink(log_file) and os.path.realpath(log_file) == "/dev/null") or os.path.islink(parent_symlink):
             continue
 
         try:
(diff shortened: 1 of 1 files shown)


src/log.py
```diff
@@ -87,20 +87,21 @@ def _update_log_parent_symlinks():
...
+        log_file = os.path.join(OPERATIONS_PATH, log_md)
+        if os.path.getctime(log_file) < one_year_ago:
             # Let's ignore files older than one year because hmpf reading a shitload of yml is not free
             continue
 
         name = log_md[: -len(METADATA_FILE_EXT)]
-        parent_symlink = os.path.join(OPERATIONS_PATH, f".{name}.parent.yml")
-        if os.path.islink(parent_symlink):
+        parent_symlink = os.path.join(OPERATIONS_PATH, f"{name}.parent.yml")
+        if not os.path.exists(parent_symlink) or os.path.islink(parent_symlink):
```
kay0u, line 99: We must remove this check, because the purpose of this function is to create it^^.

I still have errors during `yunohost log list`:

> 88   ERROR   The YAML metadata file associated with logs is damaged: '20240715-030330-backup_create.parent.yml
Error: File does not exist: '/var/log/yunohost/operations/20240715-030330-backup_create.parent.yml''

kay0u: Something like that
```suggestion
        if (os.path.islink(log_file) and os.path.realpath(log_file) == "/dev/null) or os.path.islink(parent_symlink):
```

kay0u:
I still have an issue, the webadmin display the `.parent` files, and when i click on it, i get an empty log view:

image

image

src/log.py
```diff
@@ -87,20 +87,21 @@ def _update_log_parent_symlinks():
...
-        if os.path.getctime(log_md) < one_year_ago:
+        log_file = os.path.join(OPERATIONS_PATH, log_md)
+        if os.path.getctime(log_file) < one_year_ago:
             # Let's ignore files older than one year because hmpf reading a shitload of yml is not free
             continue
 
         name = log_md[: -len(METADATA_FILE_EXT)]
-        parent_symlink = os.path.join(OPERATIONS_PATH, f".{name}.parent.yml")
-        if os.path.islink(parent_symlink):
+        parent_symlink = os.path.join(OPERATIONS_PATH, f"{name}.parent.yml")
```
alexAubin, line 98: the issue with parents being displayed might be related to the fact that the file was meant to be hidden (prefixed with a dot) and maybe "glob" doesn't list hidden files by default, if it behaves the same as `ls`, not sure

kay0u: Hoo that dot was intended

kay0u: `include_hidden=True` param has been added to the glob function in python 3.11

src/log.py
```diff
@@ -87,20 +87,21 @@ def _update_log_parent_symlinks():
...
-        if os.path.getctime(log_md) < one_year_ago:
+        log_file = os.path.join(OPERATIONS_PATH, log_md)
+        if os.path.getctime(log_file) < one_year_ago:
             # Let's ignore files older than one year because hmpf reading a shitload of yml is not free
             continue
 
         name = log_md[: -len(METADATA_FILE_EXT)]
         parent_symlink = os.path.join(OPERATIONS_PATH, f".{name}.parent.yml")
-        if os.path.islink(parent_symlink):
+        if (os.path.islink(log_file) and os.path.realpath(log_file) == "/dev/null") or os.path.islink(parent_symlink):
```
kay0u, line 99: We probably don't need this change then?

alexAubin: Hmmyeah i'm not sure to understand what was the intent ... naively we're good to go with just the stuff that fixes the confusion about relative vs absolute paths

PR #1916, opened 2024-07-21
[Fix] Make slapd listen also on ipv6

Josue-T:
## The problem

Some app are configured to connect to LDAP this way: `ldap://localhost:389`.
But currently, slapd listen only on IPv4. On my side I've this result with `ss`:

```
ss -ntlp | grep 389
LISTEN 0      1024       127.0.0.1:389        0.0.0.0:*    users:(("slapd",pid=1434,fd=9)) 
```

We have this issue here: https://forum.yunohost.org/t/seafile-11-0-9-cant-login-with-client/30504/2

## Solution

Configure slapd to listen also on ipv6.

## PR Status

This is ready

## How to test

Tested with netcat and now connection work with IPv4 and IPv6.

We can also test with `ss` with has this result:
```
ss -ntlp | grep 389
LISTEN 0      1024       127.0.0.1:389        0.0.0.0:*    users:(("slapd",pid=1434,fd=9))                                                                                                 
LISTEN 0      1024           [::1]:389           [::]:*    users:(("slapd",pid=1434,fd=8))                                                                                                 
```

### DIFF
--- a/conf/slapd/slapd.default
+++ b/conf/slapd/slapd.default
@@ -21,7 +21,7 @@ SLAPD_PIDFILE=
 # sockets.
 # Example usage:
 # SLAPD_SERVICES="ldap://127.0.0.1:389/ ldaps:/// ldapi:///"
-SLAPD_SERVICES="ldap://127.0.0.1:389/ ldaps:/// ldapi:///"
+SLAPD_SERVICES="ldap://localhost:389/ ldaps:/// ldapi:///"
 
 # If SLAPD_NO_START is set, the init script will not start or restart
 # slapd (but stop will still work).  Uncomment this if you are


alexAubin:
cf https://github.com/YunoHost/yunohost/commit/fbfa249ff5e1a601475da5d9968ebc47cdb190dc , listening only to ipv4 only is actually not a bug but a feature because this crashes on ipv4-only instances, which then results in disaster as in "being completely locked out of the server" because LDAP is down ... It feels much simpler to change the app's conf to bind to 127.0.0.1 rather than localhost ... But if we really do want IPv6 then we should have a similar mechanism to what's done for sshd_config where we only enable it if ipv6 seems to be available

Josue-T:
Ok, it was what I suspeceted. But well the issue is that many apps are impacted by this and also the doc don't warn about this. Note also that some apps just don't work if IPv6 is disabled cf https://github.com/YunoHost-Apps/synapse_ynh/pull/458

And finally last question, why disabling completely IPv6 🤔

Josue-T:
Well, it would be interesting to know it my change still work with IPv6 disabled, as listening on localhost which is implicitly on IPv4+IPv6 is different  than the previous version which is explicit to listen on IPv6 and which is normal that it don't work if IPv6 is disabled.

alexAubin:
Tested by disabling ipv6 from sysctl.conf and it does only listens to ipv4 indeed, so let's merge this and pray to god that no user has some stupidly specific network conf where slapd will explode because of this ...

PR #1917, opened 2024-07-21
app/permissions: Move permissions data out of LDAP, introduce a new 'app core' config panel allowing to change the app logo, label, etc.

alexAubin:
## The problem

As foretold a while ago, for various reason, we are going to want to tweak the permission datastructure (because of the auth header policy evolution, or being able to tweak the label, description, logo in the portal) and all the infos are currently in LDAP for not real reason beyond the fact that this is where we store the user<->group<->ACL relationships - but all the other infos (label, url, show_tile etc.) are only meant to be used by ssowat without relying on LDAP.

This is not practical because everytime we want to change the datastructure, we need an LDAP migration, and having everything in settings would also to just simply be able to edit a flat yaml text file which has plenty of benefits.

## Solution

- Move all permission infos out of LDAP : url, additional_urls, label, show_tile, and protected infos, but also the "allowed" list of group/users
- The old LDAP attributes are marked as OBSOLETE, couldn't be removed entirely otherwise LDAP will spit warnings because the old info still exists in DB ... and yet it cannot be removed either because LDAP supposedly doesnt know about them ...
- The perm infos are now part of the app settings, using a special `_permissions` key which is a `dict` mapping each perm to the corresponding infos
- The "system perm" infos are stored in a new configuration file, `/etc/yunohost/permissions.yml`
- The "allowed/member" relationship between user and permission is resynchronized to the LDAP (from the "truth source" being the app settings + system perm conf) using an internal function `_sync_permissions_with_ldap` which may also be called manually from CLI using `yunohot user permissions ldapsync` ... this CLI command is mainly here for debugging purpose and/or scenario in which somebody tweaks access rights manually in the app settings ... Kind of similar to `yunohost app ssowatconf`
- Drop the group<->perm relationship from LDAP because it's not needed anymore
- Remove `--short` option for `yunohost user permission list` because it was pretty much unused and confusing because the return type of the function was different with/without this option
- Remove `yunohost user permission reset <app>` because it's pretty much unused and exists mainly because of the legacy permission system
- `user_permission_create/update/delete` do not generate an operation log anymore because it felt a bit bloated that it did ... especially now that what it does is mainly to edit the app settings ... but admin-triggered operation like user_permission_update (from CLI/API) or add/remove members will generate a "flash" operation

## PR Status

Yolodraft

- Rereview the migration after the second iteration that also moves the "allowed" info
- Rereview the sync_perm=True/False because the ldap sync function doesn't call app_ssowat anymoar

## How to test

...


src/migrations/0032_rework_permission_infos.py
```diff
@@ -0,0 +1,87 @@
...
+    @Migration.ldap_migration
+    def run(self, *args):
+
+        regen_conf(["slapd"], force=True)
+
+        self.ldap_migration_started = True
+
+        permissions_per_app = self.read_legacy_permissions_per_app()
+        for app, permissions in permissions_per_app.items():
+            app_setting(app, "_permissions", permissions)
```
alexAubin, line 31: Fixme: it doesn't properly migrate the "label" info ?

zamentur:
Currently, `/etc/ssowat/conf.json` and `/etc/yunohost/portal/DOMAIN.json` contain structure like this:
```
"APP.main": {
    "users": [ LIST_OF_USERS ]
}
```

If you have hundreds of users, you have for each permissions and in each of this file, all allowed users listed, even if you use groups to define those permission.

I understand that we need some cache to avoid perf issue with ldap request. Could it be possible to factorize those file with something like:
`/var/cache/yunohost/groups.yml`
```
all_users: [ USER_LIST ]
other_group: [ "user1","user2" ]
```
And refer to it in `/etc/ssowat/conf.json` and `/etc/yunohost/portal/DOMAIN.json` with something like this instead:
```
"APP.main": {
    "allowed": [ "other_group", "user3" ]
}
```

Next we could recompile the "flat" list of allowed users in `_get_portal_settings()`.
In `/etc/ssowat/conf.json`, we could simply have a copy of groups:
```
groups:
  all_users: [ USER_LIST ]
  other_group: [ "user1","user2" ]
```
Or even directly give read write on `/var/cache/yunohost/groups.yml` to ssowat lua code...

May be an other PR. It's just it seems to me quite annoying to have tons of users listed in each apps main permissions.

PR #1936, opened 2024-08-27
enh: add option to install a lets encrypt certificate when adding a subdomain of an already added domain

Axolotle:
## The problem
When an already added domain dns records has a wildcard, we could safely install a let's encrypt certificate when adding a subdomain.
This avoid some repetitive actions/navigation in cli and web-admin and will allow to add a subdomain easily while being in the app install form (some apps requires to create a dedicated subdomain to be installed)

## Solution
- detect wildcards in `certificate_status`
- add new bool option `install_letsencrypt_cert` to `domain_add`

## PR Status
could'nt test the letsencrypt install properly (can't install certs on my setup idk why)


src/domain.py
```diff
@@ -350,10 +359,31 @@ def domain_add(
...
+            ]["has_wildcards"]
+        )
+
+        if can_install_letsencrypt:
+            try:
+                _certificate_install_letsencrypt([domain], force=True, no_checks=True)
+            except:
+                failed_letsencrypt_cert_install = True
+        else:
+            failed_letsencrypt_cert_install = True
```
alexAubin, line 378:
```suggestion
            logger.warning("Skipping Let's Encrypt certificate attempt because there's no wildcard configured on the parent domain's DNS records.")
            failed_letsencrypt_cert_install = True
```

alexAubin (approved):

PR #1937, opened 2024-08-28
Add ynh_add_swap and ynh_smart_mktemp to official helpers

selfhoster1312:
These two helpers are widely used and have been part of experimental helpers for over 5 years. Maybe it's time to include them so scripts don't have to include their own custom version?

Case in point, i found 3 different versions of `ynh_add_swap` used across apps:

- one that does `chattr +C`, which will not work on most filesystems like ext4
- one that checks for LXC (or LXD?) virtualization and skips the swap setup (because it would fail)
- one that does neither

In this PR, ynh_add_swap checks that no virtualization is used at all otherwise it just skips the setup.

### DIFF
--- a/helpers/helpers.v2.1.d/utils
+++ b/helpers/helpers.v2.1.d/utils
@@ -456,3 +456,148 @@ ynh_spawn_app_shell() {
         # Spawn the app shell
         su -s /bin/bash $app
 }
+
+# Add swap
+#
+# usage: ynh_add_swap --size=SWAP in Mb
+# | arg: -s, --size= - Amount of SWAP to add in Mb.
+ynh_add_swap () {
+  if systemd-detect-virt --container --quiet; then
+    ynh_print_warn --message="You are inside a container/VM. swap will not be added, but that can cause troubles for the app $app. Please make sure you have enough RAM available."
+    return
+  fi
+
+	# Declare an array to define the options of this helper.
+	declare -Ar args_array=( [s]=size= )
+	local size
+	# Manage arguments with getopts
+	ynh_handle_getopts_args "$@"
+
+	local swap_max_size=$(( $size * 1024 ))
+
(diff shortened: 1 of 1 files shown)


helpers/helpers.v2.1.d/utils
```diff
@@ -456,3 +456,148 @@ ynh_spawn_app_shell() {
...
+            local tmpdir=/var
+    elif is_there_enough_space /; then
+            local tmpdir=/   
+    elif is_there_enough_space /home; then
+            local tmpdir=/home
+    else
+            ynh_die "Insufficient free space to continue..."
+    fi
+
+    echo "$(sudo mktemp --directory --tmpdir="$tmpdir")"
```
selfhoster1312, line 602: Not sure why there was a sudo here in one of the variants...

alexAubin:
```suggestion
    echo "$(mktemp --directory --tmpdir="$tmpdir")"
```

Salamandar:
Hmmm this PR is a bit bigger than adding the two helpers, maybe split ? :D

helpers/helpers.v2.1.d/utils
```diff
@@ -456,3 +456,148 @@ ynh_spawn_app_shell() {
         # Spawn the app shell
         su -s /bin/bash $app
 }
+
+# Add swap
+#
+# usage: ynh_add_swap --size=SWAP in Mb
+# | arg: -s, --size= - Amount of SWAP to add in Mb.
+ynh_add_swap () {
+  if [ "$(systemd-detect-virt)" != "none" ]; then
```
Salamandar, line 465:
```suggestion
  if [ "$(systemd-detect-virt)" != "none" ]; then
  f systemd-detect-virt --container --quiet; then
```
That will allow swap in KVM and virtual machines as well

Salamandar (approved):

PR #1944, opened 2024-08-31
Add maintenance/pre-commit.sh script

selfhoster1312:
The first commit fixes shfmt.sh which was just... not working?

Second one introduces a script people can use like `maintenance/pre-commit.sh check|fix`. It also works without argument and should be able to be called from any folder. It will check both python and bash formatting.

### DIFF
--- a/maintenance/pre-commit.sh
+++ b/maintenance/pre-commit.sh
@@ -1,0 +1,116 @@
+#! /usr/bin/env bash
+
+set -Eeuo pipefail
+
+YUNODIR="$(realpath "$(dirname "$0")"/..)"
+SHFMT="$YUNODIR"/maintenance/shfmt.sh
+
+help() {
+    echo "pre-commit.sh ACTION [LANG]"
+    echo "  Run common pre-commit checks to make CI happy"
+    echo "  ACTIONS:"
+    echo "    - check: provide recommendations"
+    echo "    - fix: apply recommendations automatically"
+    echo "  LANG: python|bash, by default applies to all languages"
+}
+
+check_deps() {
+    for dep in "$@"; do
+        if ! command -v "$dep" > /dev/null 2>&1; then
+            echo "Missing dependency: $dep. Please install it first."
+            exit 1
+        fi
+    done
+}
+
+check_bash() {
+    check_deps shfmt
+    "$SHFMT" -d "$YUNODIR"/helpers/helpers \
+        "$YUNODIR"/helpers/helpers.v2.1.d/* \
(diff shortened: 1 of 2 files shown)


maintenance/shfmt.sh
```diff
@@ -1,16 +1,11 @@
...
-    --indent 4
-    --keep-padding      # keep column alignment paddings
-    --space-redirects   # redirect operators will be followed by a space
-    --binary-next-line  # binary ops like && and | may start a line
-    --case-indent       # switch cases will be indented
+    -i=4
+    -kp      # keep column alignment paddings
+    -sr   # redirect operators will be followed by a space
+    -bn  # binary ops like && and | may start a line
+    -ci       # switch cases will be indented
```
Salamandar, line 8: The long args work with the more recent version as you saw ;)

maintenance/pre-commit.sh
```diff
@@ -0,0 +1,116 @@
...
+        fi
+    done
+}
+
+check_bash() {
+    check_deps shfmt
+    "$SHFMT" -d "$YUNODIR"/helpers/helpers
+    "$SHFMT" -d "$YUNODIR"/helpers/helpers.v2.1.d/*
+    "$SHFMT" -d "$YUNODIR"/helpers/helpers.v2.d/*
+    "$SHFMT" -d "$YUNODIR"/hooks/*
```
Salamandar, line 31: I see one major flaw at calling multiple times shfmt: if you use `check`, and you have issues in multiple helpers dirs, it will fail on the first and NOT show issues from the subsequent ones.

So i suggest calling it **once** with all the files on the command line.

maintenance/pre-commit.sh
```diff
@@ -0,0 +1,116 @@
...
+
+check() {
+    case "$1" in
+        "python")
+            check_python
+            ;;
+        "bash")
+            check_bash
+            ;;
+        "both")
```
Salamandar, line 48: Better to name it `all`. The day we have more than 2 checks…

maintenance/shfmt.sh
selfhoster1312: If that's ok with you i'm happy to keep a version that works on my system ^^

Salamandar: Yeah alright, if that's more portable that's fine with me !

alexAubin:
Superseded by https://github.com/YunoHost/yunohost/commit/f14f5bbeed4541745cc480f02314ba075a11cdc5 and https://github.com/YunoHost/yunohost/commit/7020a4892a647a658060b6a2259c456a2753d4c1

PR #1949, opened 2024-09-11
Enh: add CLI TOS acknowledgement

Axolotle:
## Solution
Add TOS acknowledgement with simple Y/N question for postinstall and dyndns registration

## PR Status
ready?

## How to test
`yunohost tools postinstall`
`yunohost domain add test.noho.st`

## Preview
Post install:
image

dyndns:
image


Axolotle:
In the post install case, if you also add a dyndns domain, you'll be prompted twice for acknowledgement. Should we bypass this in the dyndns registration?

Salamandar:
In any case I think I would prefer to see this ToS in a separate file (e.g in /usr/share/licenses for example).

PR #1951, opened 2024-09-15
Enh: Make auto dns feature optional

Axolotle:
## The problem
- For the auto dns push feature, we can't remove a registrar API key and other unknown auth fields.
- Special domains infos are not correctly displayed

## Solution

- add a new boolean `use_auto_dns` in the DNS section to choose to use the auto dns feature or not (default to true for compatibility)
  - when setting false, remove all auth fields values + managed_dns_records_hashes` in settings
- make registrar auth fields optional and add `visible` prop depending on `use_auto_dns`
- fix special-use tld case
- add some helpers methods on `ConfigPanelModel` + also pass config model to `ConfigPanel._apply` 

## PR Status
OK

Fix https://github.com/YunoHost/issues/issues/2293

### DIFF
--- a/src/domain.py
+++ b/src/domain.py
@@ -773,6 +773,18 @@ def _get_DomainConfigPanel():
                     self.entity, next_settings["recovery_password"]
                 )
 
+            if "use_auto_dns" in next_settings and not next_settings["use_auto_dns"]:
+                # disable auto dns by reseting every registrar form values
+                options = [
+                    option
+                    for option in config.get_section("registrar").options
+                    if not option.readonly
+                    and option.id != "use_auto_dns"
+                    and hasattr(form, option.id)
+                ]
+                for option in options:
+                    setattr(form, option.id, option.default)
+
             custom_css = next_settings.pop("custom_css", "").strip()
             if custom_css:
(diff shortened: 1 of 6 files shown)


src/domain.py
```diff
@@ -772,6 +774,19 @@ def _apply(
                     self.entity, next_settings["recovery_password"]
                 )
 
+            remove_auto_dns_feature = "use_auto_dns" in next_settings and not next_settings["use_auto_dns"]
+            if remove_auto_dns_feature:
```
alexAubin, line 778:
```suggestion
            # NB: this is subtlely different from just checking `next_settings.get("use_auto_dns") since we want to find the exact situation where the admin *disables* the autodns`
            remove_auto_dns_feature = "use_auto_dns" in next_settings and not next_settings["use_auto_dns"]
            if remove_auto_dns_feature:
```

alexAubin: (Adding a comment - if i understood correctly the intent here)

alexAubin (approved):

src/domain.py
Axolotle: Indeed, we want to remove thing only when the value changes and is falsy

alexAubin: yeah that's what i was trying to convey with my comment :sweat_smile:

PR #1953, opened 2024-09-17
Storage managment API: disks infos

christophehenry:
## The problem

Rel: YunoHost/issues#1823.
Add API for getting infos on hard drives.

### DIFF
--- a/doc/generate_bash_completion.py
+++ b/doc/generate_bash_completion.py
@@ -42,7 +42,7 @@ BASH_COMPLETION_FILE = BASH_COMPLETION_FOLDER + "/yunohost"
 def get_dict_actions(OPTION_SUBTREE, category):
     ACTIONS = [
         action
-        for action in OPTION_SUBTREE[category]["actions"].keys()
+        for action in OPTION_SUBTREE[category].get("actions", {}).keys()
         if not action.startswith("_")
     ]
     ACTIONS_STR = "{}".format(" ".join(ACTIONS))
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -2125,7 +2125,7 @@ storage:
                         action: store_true
                 info:
                     action_help: Get hard-drive information
-                    api: GET /users/<name>
+                    api: GET /storage/disk/info/<name>
                     arguments:
                         name:
                             help: Name of the hard drive as returned by 'list' command
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -2099,3 +2099,37 @@ diagnosis:
                     help: Remove a filter (it should be an existing filter as listed with "ignore --list")
                     nargs: "*"
                     metavar: CRITERIA
+
+
+#############################
+#           Storage         #
+#############################
+storage:
+    category_help: Manage hard-drives, filesystem, pools
+    subcategories:
+        disk:
+            subcategory_help: Manage et get infos about hard-drives
+            actions:
+                # storage_disks_list
+                list:
+                    action_help: List hard-drives currently attached to this system optionnaly with infos
+                    api: GET /storage/disk/list
+                arguments:
+                    -i:
+                        full: --with-info
+                        help: Get all informations for each archive
+                        action: store_true
+                    -H:
+                        full: --human-readable
+                        help: Print sizes in human readable format
+                        action: store_true
+                info:
+                    action_help: Get hard-drive information
+                    api: GET /users/<name>
+                    arguments:
+                        name:
+                            help: Name of the hard drive as returned by 'list' command
+                    -H:
+                        full: --human-readable
+                        help: Print sizes in human readable format
+                        action: store_true
--- a/src/disks.py
+++ b/src/disks.py
@@ -1,9 +1,5 @@
-from logging import getLogger
-
 import dbus
-from _dbus_bindings import PROPERTIES_IFACE
-
-logger = getLogger("yunohost.storage")
+from yunohost.utils.system import binary_to_human
 
 
 __all__ = ["info", "list"]
@@ -13,30 +9,21 @@ UDISK_DRIVE_PATH = "/org/freedesktop/UDisks2/drives/"
 UDISK_DRIVE_IFC = "org.freedesktop.UDisks2.Drive"
 
 
-def _humaize(byte_size):
-    suffixes = "kMGTPEZYRQ"
-
-    byte_size = float(byte_size)
-
-    if byte_size < 1024:
-        return f"{byte_size}B"
-
-    for i, s in enumerate(suffixes, start=2):
-        unit = 1024**i
-        if byte_size <= unit:
-            return f"{(1024 * (byte_size / unit)):.1f} {s}B"
-
-
-def _query_udisks():
+def _query_udisks() -> list[tuple[str, dict]]:
     bus = dbus.SystemBus()
     manager = dbus.Interface(
         bus.get_object("org.freedesktop.UDisks2", "/org/freedesktop/UDisks2"),
         "org.freedesktop.DBus.ObjectManager",
     )
 
-    for name, dev in manager.GetManagedObjects().items():
-        if name.startswith(UDISK_DRIVE_PATH):
-            yield name.removeprefix(UDISK_DRIVE_PATH), dev[UDISK_DRIVE_IFC]
+    return sorted(
+        (
+            (name.removeprefix(UDISK_DRIVE_PATH), dev[UDISK_DRIVE_IFC])
+            for name, dev in manager.GetManagedObjects().items()
+            if name.startswith(UDISK_DRIVE_PATH)
+        ),
+        key=lambda item: item[1]["SortKey"],
+    )
 
 
 def _disk_infos(name: str, drive: dict, human_readable=False):
@@ -45,7 +32,7 @@ def _disk_infos(name: str, drive: dict, human_readable=False):
         "model": drive["Model"],
         "serial": drive["Serial"],
         "removable": bool(drive["MediaRemovable"]),
-        "size": _humaize(drive["Size"]) if human_readable else drive["Size"],
+        "size": binary_to_human(drive["Size"]) if human_readable else drive["Size"],
     }
 
     if connection_bus := drive["ConnectionBus"]:
@@ -55,7 +42,7 @@ def _disk_infos(name: str, drive: dict, human_readable=False):
         result.update(
             {
                 "type": "HDD",
-                "rpm": "Unknown",
+                "rpm": "Unknown" if human_readable else None,
             }
         )
     elif rotation_rate == 0:
@@ -75,15 +62,12 @@ def list(with_info=False, human_readable=False):
     if not with_info:
         return [name for name, _ in _query_udisks()]
 
-    result = {}
+    result = []
 
     for name, drive in _query_udisks():
-        result[name] = _disk_infos(name, drive, human_readable)
+        result.append(_disk_infos(name, drive, human_readable))
 
-    if human_readable and not result:
-        return "No external media found"
-
-    return result
+    return {"disks": result}
 
 
 def info(name, human_readable=False):
@@ -92,7 +76,7 @@ def info(name, human_readable=False):
         bus.get_object(
             "org.freedesktop.UDisks2", f"/org/freedesktop/UDisks2/drives/{name}"
         ),
-        PROPERTIES_IFACE,
+        dbus.PROPERTIES_IFACE,
     ).GetAll("org.freedesktop.UDisks2.Drive")
 
     return _disk_infos(name, drive, human_readable)
--- a/src/disks.py
+++ b/src/disks.py
@@ -1,0 +1,98 @@
+from logging import getLogger
+
+import dbus
+from _dbus_bindings import PROPERTIES_IFACE
+
+logger = getLogger("yunohost.storage")
+
+
+__all__ = ["info", "list"]
+
+
+UDISK_DRIVE_PATH = "/org/freedesktop/UDisks2/drives/"
+UDISK_DRIVE_IFC = "org.freedesktop.UDisks2.Drive"
+
+
+def _humaize(byte_size):
+    suffixes = "kMGTPEZYRQ"
+
+    byte_size = float(byte_size)
+
+    if byte_size < 1024:
+        return f"{byte_size}B"
+
+    for i, s in enumerate(suffixes, start=2):
+        unit = 1024**i
+        if byte_size <= unit:
+            return f"{(1024 * (byte_size / unit)):.1f} {s}B"
+
+
+def _query_udisks():
+    bus = dbus.SystemBus()
+    manager = dbus.Interface(
+        bus.get_object("org.freedesktop.UDisks2", "/org/freedesktop/UDisks2"),
+        "org.freedesktop.DBus.ObjectManager",
+    )
+
+    for name, dev in manager.GetManagedObjects().items():
+        if name.startswith(UDISK_DRIVE_PATH):
+            yield name.removeprefix(UDISK_DRIVE_PATH), dev[UDISK_DRIVE_IFC]
+
+
+def _disk_infos(name: str, drive: dict, human_readable=False):
+    result = {
+        "name": name,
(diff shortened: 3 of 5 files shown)


src/tests/test_storage.py
```diff
@@ -0,0 +1,8 @@
+def setup_function(function):
+    ...
+
+def teardown_function(function):
+    ...
+
+def test_storage_disks_infos():
+    ...
```
christophehenry, line 8: I don't really know how to test that. It would need simulating hard drives but the current implementation excludes virtual devices. Any idea?

alexAubin: Don't know much what needs testing but I suppose we should look into mocking `pyudev`

share/actionsmap.yml
```diff
@@ -2082,3 +2082,15 @@ diagnosis:
...
+
+
+#############################
+#           Storage         #
+#############################
+storage:
+  category_help: Manage hard-disks, filesystem, pools
+  actions:
+    # storage_disks_list
+    disks_infos:
```
christophehenry, line 2094: Is it possible to declare nested subcomands? Like `yunohohost storage disk infos`?

alexAubin: yes, have a look at the "cert" subcategory inside "domain"

christophehenry:
Ok I think this is a good first step. Here is an exemple of the result this produces:

```
{
    "sd": {
        "devname": "/dev/sda",
        "model": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "serial": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "size": xxxxxxxxxxxxxxxxxxxxxxxxxxxxx,
        "links": [
            "/dev/disk/by-diskseq/1",
            "/dev/disk/by-id/ata-xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
            "/dev/disk/by-id/wwn-xxxxxxxxxxxxxxxxxxxxxxxx",
            "/dev/disk/by-path/pci-0000:01:00.1-ata-1",
            "/dev/disk/by-path/pci-0000:01:00.1-ata-1.0"
        ],
        "partitions": {
            "sda1": {
                "devname": "/dev/sda",
                "filesystem": "btrfs",
                "encrypted": true,
                "mountpoint": "/home"
            }
        }
    },
    "sdb": {
        "devname": "/dev/sdb",
        "model": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "serial": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
        "size": xxxxxxxxxxxxxxxxxxxxxxxxxxxxx,
        "links": [
            "/dev/disk/by-diskseq/2",
            "/dev/disk/by-id/ata-xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
            "/dev/disk/by-id/wwn-xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
            "/dev/disk/by-path/pci-0000:0a:00.0-ata-6",
            "/dev/disk/by-path/pci-0000:0a:00.0-ata-6.0"
        ],
        "partitions": {
            "sdb1": {
                "devname": "/dev/sdb",
                "filesystem": "vfat",
                "encrypted": false,
                "mountpoint": "/boot/efi"
            },
            "sdb2": {
                "devname": "/dev/sdb",
                "filesystem": "ext4",
                "encrypted": false,
                "mountpoint": "/boot"
            },
            "sdb3": {
                "devname": "/dev/sdb",
                "filesystem": "btrfs",
                "encrypted": true,
                "mountpoint": "/"
            }
        }
    }
}
```

In a future PR, I'd like to add the device type (HDD, SSD, …) and RPM for HDDs. I don't really know how to do that. TrueNAS does it by using [its own library](https://github.com/truenas/py-sgio). In the meantime, this is ready for review.

src/disks.py
```diff
@@ -0,0 +1,101 @@
...
+        ):
+            encrypted_provider = glob(f"/sys/block/dm-*/slaves/{child_dev.sys_name}")
+            if encrypted_provider:
+                # retrive the dm-x part
+                dm = encrypted_provider[0].split("/")[3]
+                enc_dev = pyudev.Devices.from_name(device.context, "block", dm)
+                # This work for LUKS, what about other partition mecanisms?
+                partname = f"/dev/mapper/{enc_dev.properties['DM_NAME']}"
+                encrypted = True
+            else:
```
christophehenry, line 39: This method detects LUKS partitions but I'm not sure it does not detect other unecrypted partitions, like LVM. If someone with more system knowlege than me could check this.

christophehenry: I could use [`pyparted`](https://github.com/dcantrell/pyparted) here but this needs sufficient rights. If I'm not mistaken, `yunohost-api` is running as root, so this shouldn't be an issue, right?

christophehenry: Turns out this solution detects LVM (non-LUKS) partitions as encrypted. And I can't figure out how to retrieve FS with pyparted. I don't know how gparted and KDE partition manager detect the partition filesystem.

christophehenry:
A few notes here: I rewrote this PR to use the udisks2 API which provides much more details about disks and is easier to use that udev + other sources of infos.

But as a pure dbus API, the code is less self-explanatory. It's mainly about parsing dictionary.

Maybe I can make use of something like [`dbus-fast`](https://pypi.org/project/dbus-fast/), IDK.

Tell me if it's OK to add the udisk daemon to Ynh. I'm not aware of everything it entails.

alexAubin:
Naively udisk sounds fine to me :+1:

zamentur:
Yes i think udisk is a good idea, also because it could help to automount external usb or disk...

zamentur:
> In a future PR, I'd like to add the device type (HDD, SSD, …) and RPM for HDDs. I don't really know how to do that. TrueNAS does it by using [its own library](https://github.com/truenas/py-sgio). In the meantime, this is ready for review.

You can know if it's a HDD or SSD by checking directly the info inside /sys...
```
cat /sys/block/sda/queue/rotational
```
0 -> SSD
1 -> HDD

/sys/block/sda/queue/removable to know if it's an usb removable disk i guess

We could also imagine check for ssd how many written is made, in order to compute an end of life percentage.

christophehenry:
> You can know if it's a HDD or SSD by checking directly the info inside /sys...

udisks2 provides the information so I don't need a solution anymore. Thanks.

src/disks.py
```diff
@@ -0,0 +1,49 @@
...
+
+def infos():
+    result = {}
+
+    bus = dbus.SystemBus()
+    manager = bus.get_object("org.freedesktop.UDisks2", "/org/freedesktop/UDisks2")
+
+    for k, v in manager.get_dbus_method(
+        "GetManagedObjects", "org.freedesktop.DBus.ObjectManager"
+    )().items():
```
christophehenry, line 32: Does someone knows a more efficient way to interact with dbus than just list *all the objects managed by Udisks2* here? Or is dbus definitely a crap?

christophehenry:
Quick note here: this PR uses `dbus-python` which multiple sources mention is slowly being deprecated but I can't find a more modern library. [dbus wiki](https://www.freedesktop.org/wiki/Software/DBusBindings/) lists `pydbus` a "a modern, pythonic D-Bus library" but it is dead since 2016… `python-sdbus ` seems async only, unless I'm wrong. Maybe gdbus but Gnome has to be installed on the system so it's a no-go, I suppose?

src/disks.py
```diff
@@ -0,0 +1,49 @@
...
+    serial: str
+    size: int
+    type: str
+    rpm: Optional[int] = None
+
+
+def infos():
+    result = {}
+
+    bus = dbus.SystemBus()
```
christophehenry, line 27: Should it be file-scoped? I don't know how it is efficient to create a bus at each call.

christophehenry:
I don't believe it is useful to test this. We'd basically be testing the `udisks2` library which is never a good idea.

src/disks.py
zamentur: but maybe dbus is ok... (i am not a dbus expert)

src/disks.py
```diff
@@ -0,0 +1,37 @@
+import dbus
+
+from moulinette.utils.log import getActionLogger
```
christophehenry, line 3: Has this function disappeared? This line raises an error after rabase.

alexAubin: Ah yes, you can check the other file for the new syntax, as far as i remember it's just `import logging` and then `logging.getLogger(...)`

share/actionsmap.yml
```diff
@@ -2003,3 +2003,18 @@ diagnosis:
...
+#           Storage         #
+#############################
+storage:
+  category_help: Manage hard-drives, filesystem, pools
+  subcategories:
+    disk:
+      subcategory_help: Manage et get infos about hard-drives
+      actions:
+        # storage_disks_list
+        infos:
```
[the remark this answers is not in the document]
zamentur: About --human-readable we are doing this in several part of the app, but regarding the cli is often use by a human, i ask me if we should not put by default the --human-readable and configure an other option to get the size in Bytes...

zamentur (changes requested):
I suggest to split into 2 commands 
```
yunohos storage disk list
yunohos storage disk info <disk>
```

zamentur:
I tried to test into an incus container, but drives are not discovered by dbus (lsblk does). It's due to the fact that the main storage is on /dev/loopXX on incus. I tried to add an entire usb key, but it was a fail.

src/disks.py
```diff
@@ -0,0 +1,38 @@
...
+        name = name.removeprefix(UDISK_DRIVE_PATH)
+        rotation_rate = drive["RotationRate"]
+        result[name] = {
+            "name": name,
+            "model": drive["Model"],
+            "serial": drive["Serial"],
+            "size": drive["Size"],
+            "type": "HDD" if rotation_rate != 0 else "SSD",
+            "rpm": rotation_rate if rotation_rate != 0 else None,
+        }
```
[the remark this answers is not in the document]
christophehenry: When using the ` --human-readable` option you mean?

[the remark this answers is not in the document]
zamentur: With --human-readable : 

```
size: 60 GB
```

Without:
```
size: 64424509440
```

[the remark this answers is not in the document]
christophehenry: No I meant about the case where there's no disks at all. With --human-readable, a message, without, empty list. Is that OK?

src/disks.py
```diff
@@ -0,0 +1,38 @@
...
+        "org.freedesktop.DBus.ObjectManager",
+    )
+
+    for name, dev in manager.GetManagedObjects().items():
+        if not name.startswith(UDISK_DRIVE_PATH):
+            continue
+
+        drive = dev[UDISK_DRIVE_IFC]
+        name = name.removeprefix(UDISK_DRIVE_PATH)
+        rotation_rate = drive["RotationRate"]
```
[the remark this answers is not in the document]
christophehenry: [According to documentation](https://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Drive.html#gdbus-property-org-freedesktop-UDisks2-Drive.RotationRate), this value always exists. It is 0 if known to be non-rotating media and -1 if known to be rotating media but rotation rate isn't known. So this should not be necessary.

christophehenry:
@zamentur So I implemented you suggestions and tested with a USB flash drive, USB hard drive and SD card.

<details>
<summary>Here are the results</summary>

  ```
  {
      # This is the SD card
      "******************": {
          "name": "******************",
          ******************,
          ******************,
          "removable": true,
          "size": "3.6 GB",
          "connection_bus": "sdio",
          "type": "SSD"
      },
      # This is an internal SSD
      "******************": {
          "name": "******************",
          ******************,
          ******************,
          "removable": false,
          "size": "223.6 GB",
          "type": "SSD"
      },
      # This is the USB HDD
      "******************": {
          "name": "******************",
          ******************,
          ******************,
          "removable": false,
          "size": "1.8 TB",
          "connection_bus": "usb",
          "type": "HDD",
          "rpm": 5400
      },
      # This is an internal SSD
      "******************": {
          "name": "******************",
          ******************,
          ******************,
          "removable": false,
          "size": "1.8 TB",
          "type": "SSD"
      },
      # This is an USB flash drave
      "******************": {
          "name": "******************",
          ******************,
          ******************,
          "removable": true,
          "size": "7.2 GB",
          "connection_bus": "usb",
          "type": "HDD",
          "rpm": "Unknown"
      }
  }
  ```

Weirdly, the USB flash drive is detected as being rotational with unkown RPM. I don't know how this should be handled.
</details>

src/disks.py
```diff
@@ -0,0 +1,98 @@
...
+    result = {
+        "name": name,
+        "model": drive["Model"],
+        "serial": drive["Serial"],
+        "removable": bool(drive["MediaRemovable"]),
+        "size": _humaize(drive["Size"]) if human_readable else drive["Size"],
+    }
+
+    if connection_bus := drive["ConnectionBus"]:
+        result["connection_bus"] = connection_bus
```
christophehenry, line 49: Should those keys be translated when using the `human_readable` option?

src/disks.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+__all__ = ["info", "list"]
+
+
+UDISK_DRIVE_PATH = "/org/freedesktop/UDisks2/drives/"
+UDISK_DRIVE_IFC = "org.freedesktop.UDisks2.Drive"
+
+
+def _humaize(byte_size):
```
zamentur, line 16: Maybe you should use this function instead ?
https://github.com/YunoHost/yunohost/blob/4e9d94c836fa34b37805bf07b61aa8aa0d9d61e3/src/utils/system.py#L120

src/disks.py
zamentur: I think it's not necessary.

src/disks.py
```diff
@@ -0,0 +1,98 @@
...
+
+    result = {}
+
+    for name, drive in _query_udisks():
+        result[name] = _disk_infos(name, drive, human_readable)
+
+    if human_readable and not result:
+        return "No external media found"
+
+    return result
```
zamentur, line 86:
```suggestion
    return {"disks": result}
```
It's done like that in user_list backup_list and permission_list, with `disks:` no need to add a message that there is no disks at all i guess.

christophehenry: Ok, then I guess the result itself should be a list and not a dict. What do you say?

src/disks.py
```diff
@@ -75,15 +62,15 @@ def list(with_info=False, human_readable=False):
...
 
     for name, drive in _query_udisks():
-        result[name] = _disk_infos(name, drive, human_readable)
+        result.append(_disk_infos(name, drive, human_readable))
 
     if human_readable and not result:
         return "No external media found"
 
-    return result
+    return {"disks": result}
```
christophehenry, line 73: Should `result` be sorted some way?

doc/generate_bash_completion.py
```diff
@@ -42,7 +42,7 @@
 def get_dict_actions(OPTION_SUBTREE, category):
     ACTIONS = [
         action
-        for action in OPTION_SUBTREE[category]["actions"].keys()
+        for action in OPTION_SUBTREE[category].get("actions", {}).keys()
```
christophehenry, line 45: Il n'y a pas de section `action` dans `storage` pour le moment.

src/disks.py
zamentur: I guess it could be better to have sda before sdb... So alphanumeric sort ?

christophehenry: [Udisks2 provides a `SortKey` property](https://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Drive.html#gdbus-property-org-freedesktop-UDisks2-Drive.SortKey). I don't know how it is computed though. I can do a `sda`/`sdb`, etc. sorting. But do we have the garantuee that a disk is always present in `/dev`?

christophehenry: Ok, done.

zamentur (approved):

zamentur:
We probably should merge into a temporary branch to run all CI tests...

PR #1955, opened 2024-09-27
Automatically apply permissions for stuff provisioned with `ynh_setup_source`.

orhtej2:
## The problem

- `ynh_setup_source "$install_dir/subpath"` is excluded from default permissions
- adding subdirectory to `$install_dir` is exempt from default permissions, unlike adding a file there.

## Solution

Subdirectories of `$install_dir` and `$data_dir` get `$app:$app` `o+rwX` ownership & permissions by default, in line with files added there.

## PR Status

It worked once when installing YunoHost-Apps/syncserver-rs_ynh#58

## How to test

Install an app that uses helpers v2.1 (duh) and provisions sources into a subdirectory.

### DIFF
--- a/helpers/helpers.v2.1.d/utils
+++ b/helpers/helpers.v2.1.d/utils
@@ -228,9 +228,14 @@ _ynh_apply_default_permissions() {
     # App files can have files of their own
     if ynh_system_user_exists --username="$app"; then
         # If this is a file in $install_dir or $data_dir : it should be owned and read+writable by $app only
-        if [ -f "$target" ] && (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
-            chmod 600 "$target"
-            chown "$app:$app" "$target"
+        if (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
+            if [ -d "$target" ]; then
+              chmod -R o+rwX "$target"
+              chown -R "$app:$app" "$target"
+            else
+              chmod 600 "$target"
+              chown "$app:$app" "$target"
+            fi
             return
         fi
         # If this is the install dir (so far this is the only way this helper is called with a directory - along with $data_dir via ynh_restore?)
--- a/helpers/helpers.v2.1.d/utils
+++ b/helpers/helpers.v2.1.d/utils
@@ -230,7 +230,7 @@ _ynh_apply_default_permissions() {
         # If this is a file in $install_dir or $data_dir : it should be owned and read+writable by $app only
         if (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
             if [ -d "$target" ]; then
-              chmod -R o+rwX "$target"
+              chmod -R u+rwX "$target"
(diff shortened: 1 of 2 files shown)


helpers/helpers.v2.1.d/utils
```diff
@@ -228,9 +228,14 @@ _ynh_apply_default_permissions() {
     # App files can have files of their own
     if ynh_system_user_exists --username="$app"; then
         # If this is a file in $install_dir or $data_dir : it should be owned and read+writable by $app only
-        if [ -f "$target" ] && (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
-            chmod 600 "$target"
-            chown "$app:$app" "$target"
+        if (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
+            if [ -d "$target" ]; then
+              chmod -R o+rwX "$target"
```
zamentur, line 233: Seems a security hole. If we do `ynh_setup_source` or `ynh_setup_source "$install_dir"` we apply here a o+rwX permission by default, but we should not allowed other users to access this dir (by default) i guess.

In more adding a -R options onto data_dir could be looooong (if there is a lot of file), so this should be done only one time by yunohost install or upgrade...

I suggest to study all chown, chmod apps to understand what should be done by default and how give advice in the right way to packagers if needed.

zamentur (changes requested):
Remove the chmod -R o+rwX

helpers/helpers.v2.1.d/utils
orhtej2: It was supposed to be `u` not `o` x_x

orhtej2:
> Remove the chmod -R o+rwX

`chmod` is sort of heart of this change, but it was meant to be `chmod u` not `o` (fixed).

alexAubin:
So eeeeh, should we move forward with this ?

orhtej2:
> So eeeeh, should we move forward with this ?

Honestly 🤷 , I guess zamentur has a valid idea to assess if these are permissions right for my one app or general defaults for the apps in catalogue.

helpers/helpers.v2.1.d/utils
```diff
@@ -228,9 +228,14 @@ _ynh_apply_default_permissions() {
     # App files can have files of their own
     if ynh_system_user_exists --username="$app"; then
         # If this is a file in $install_dir or $data_dir : it should be owned and read+writable by $app only
-        if [ -f "$target" ] && (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
-            chmod 600 "$target"
-            chown "$app:$app" "$target"
+        if (is_in_dir "$target" "${install_dir:-}" || is_in_dir "$target" "${data_dir:-}" || is_in_dir "$target" "/etc/$app"); then
+            if [ -d "$target" ]; then
+              chmod -R u+rwX "$target"
+              chown -R "$app:$app" "$target"
```
zamentur, line 234: It should be $app:$group.

I suggest to adapt `if [ "$target" == "${install_dir:-}" ]; then` below by replacing it with: `if is_in_dir "$target" "${install_dir:-}" ; then`

Same for data_dir (and maybe /etc/$app)

zamentur:
Dear contributors,

PR #1957, opened 2024-09-29
apps/helpers2.1: Properly detect 'is_data' (data dir or log folder) when calling ynh_backup

orhtej2:
## The problem

`ynh_backup "$data_dir"` runs (is elgible) during pre-upgrade backup.

## Solution

Properly match paths?

## PR Status

we may consider extracting `is_in_dir` from `_ynh_apply_default_permissions` as it contains this logic already.

## How to test

1. Install `filebrowser_ynh` (an app confirmed to use `data_dir` and helpers v2.1, use different app if needed)
2. Upload any file
3.a `yunohost app upgrade filebrowser -F`

Created backup should not include `data_dir`.

3b. Create backup (from cli, webadmin, whatever), choose `filebrowser`

Created backup should include `data_dir`.

### DIFF
--- a/helpers/helpers.v2.1.d/backup
+++ b/helpers/helpers.v2.1.d/backup
@@ -43,7 +43,7 @@ ynh_backup() {
         else
             ynh_print_info "$target will not be saved, because 'do_not_backup_data' is set."
         fi
-        return 1
+        return 0
     fi
 
     # ==============================================================================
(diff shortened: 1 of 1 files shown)


alexAubin (approved):

orhtej2:
Confirmed described test scenario(s) pass on bookworm with patch applied.

PR #1963, opened 2024-10-09
Fix formatting error.

orhtej2:
## The problem

#1959 broke copy/tar backups, preventing apps upgrades :(

## Solution

Verify that a property exists before using it.

## PR Status

Worked once on my machine (TM)

## How to test

Backup any app from WebUI.

### DIFF
--- a/src/backup.py
+++ b/src/backup.py
@@ -766,18 +766,35 @@ class BackupManager:
         """Apply backup methods"""
 
         for method in self.methods:
-            logger.debug(
-                m18n.n(
-                    "backup_applying_method_" + method.method_name, method=method.method
+            try:
+                logger.debug(
+                    m18n.n(
+                        "backup_applying_method_" + method.method_name,
+                        method=method.method,
+                    )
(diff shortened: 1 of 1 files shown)


src/backup.py
```diff
@@ -766,18 +766,35 @@ def backup(self):
...
+                    m18n.n(
+                        "backup_applying_method_" + method.method_name,
+                        method=method.method,
+                    )
+                )
+            except AttributeError:
+                logger.debug(
+                    m18n.n(
+                        "backup_applying_method_" + method.method_name,
+                        method=method.method_name,
```
alexAubin, line 780: eeeh it's really not obvious from reading the code what's the difference between these two blocks ... can't we use `if hasattr()` instead ?

alexAubin: eg. `method=method.method if hasattr(method, "method") else method.method_name` ?

(Im not even sure to understand why method needs to be passed to the string considering the string key is already method-specific?)

PR #1966, opened 2024-10-11
Allow bypassing the terms of services at postinstall time

Salamandar:
This is required for automated install / image build.


tituspijean:
It should be `terms-of-services` 😅

Salamandar:
Ah crap was already merged.

PR #1968, opened 2024-10-12
Copy over patches when checking out x_x

orhtej2:
## The problem

Apps using helpers v2.1 fail to apply patches, even properly moved to `sources/patches/$source_id/`

## Solution

Copy over patches during checkout, look in the right directory.

## PR Status

Tested once (worked!)

## How to test

Install Wallabag from https://github.com/YunoHost-Apps/wallabag2_ynh/tree/move_patches, log in (should work with LDAP patch applied).

### DIFF
--- a/helpers/helpers.v2.1.d/sources
+++ b/helpers/helpers.v2.1.d/sources
@@ -211,8 +211,8 @@ ynh_setup_source() {
     fi
 
     # Apply patches
-    if [ -d "$YNH_APP_BASEDIR/patches/" ]; then
-        local patches_folder=$(realpath "$YNH_APP_BASEDIR/patches/$source_id")
+    if [ -d "$YNH_APP_BASEDIR/sources/patches/" ]; then
+        local patches_folder=$(realpath "$YNH_APP_BASEDIR/sources/patches/$source_id")
         pushd "$dest_dir"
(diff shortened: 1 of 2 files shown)


zamentur:
Maybe the idea was to move sources/patches/ to patches/ ?

orhtej2:
> Maybe the idea was to move sources/patches/ to patches/ ?

May as well be, but copying over `sources` was an easier thing to do ;)

JimboJoe:
Hello, if I understand well, this fix for core would solve the never-ending forum thread about failing wallabag authentication (patches not applied at install/upgrade). I'm also surprised it didn't cause problems in other apps... 🤔

orhtej2:
This band-aid fix requires further investigation + some decision on app assets handling (the contents of `extra_files` directory).

Stay tuned I guess 🤷

matjack1:
I think this PR is not necessary as the fix should be done in Wallabagger: https://github.com/YunoHost-Apps/wallabag2_ynh/pull/229

As stated in the docs here: https://doc.yunohost.org/en/packaging_apps_helpers_v2.1#ynh-setup-source the code is correct regarding the patches location:

> Patches named patches/${src_id}/*.patch will be applied to $dest_dir

PR #1969, opened 2024-10-14
Fix the SSL config for the yunohost.org domain

Salamandar:
Store yunohost CA certs in a directory named yunohost, not yunohost.org, to allow the actual yunohost.org domain to be installed

(untested)

### DIFF
--- a/bin/yunoprompt
+++ b/bin/yunoprompt
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Fetch x509 fingerprint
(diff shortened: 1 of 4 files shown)


alexAubin (changes requested):
Honestly hmpf i'd rather have us keep tweaking our own infra rather than changing this in the code, there are too many symlinks in the wild pointing to /etc/yunohost/certs/yunohost.org somehow, it's a mess to untangle, this is gonna be a user support disaster just to save us max. 1h of tweaking every time we have to redeploy yunohost.org ...

PR #1971, opened 2024-10-15
[fix] Wrong warning runtime ram missing during upgrade

zamentur:
## The problem
On the chat: 
> About RAM requirement for installation. For forgejo [manifest](https://jo.trax.im/fork/forgejo_ynh/src/branch/master/manifest.toml) I have (not my doing; this has been like this for at least months):
> 
> ```
> [integration]
> disk = "150M"
> ram.build = "50M"
> ram.runtime = "2G"
> ```
> 
> Install proceeds fine, but "app upgrade" complains "This app requires 2G RAM to install/upgrade but only 716.0M is available..."
> 
> (Using Yunohost 11.2.30)
> 
> That seems to be a wrong warning. Do I misunderstand it? Couldn't find an open issue with search for "requires RAM".

## Solution

Consider the app is already running, before upgrade operation, so no need to ask for available ram in more. 

## PR Status

Untested

## How to test

...

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -2881,12 +2881,22 @@ def _check_manifest_requirements(
     # Is "include_swap" really useful ? We should probably decide wether to always include it or not instead
     if ram_requirement.get("include_swap", False):
         ram += swap
+
+    # During upgrade build operations, the service is stopped
     can_build = ram_requirement["build"] == "?" or ram > human_to_binary(
         ram_requirement["build"]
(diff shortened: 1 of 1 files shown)


Salamandar:
That might be linked to a time when we made bad RAM measurements in the CI. This is fixed now.
https://ci-apps-bookworm.yunohost.org/ci/job/10456 reports ~220MB.

PR #1911, opened 2024-07-16
YunoHost 11.3: Bullseye -> Bookworm migration

Salamandar:
cf https://github.com/YunoHost/issues/issues/2050

#### Install a testing environment
```bash
## Install YunoHost
curl https://install.yunohost.org | bash -s -- -a
# Drink coffee
yunohost tools postinstall -d bullseye-bookworm-migration.test -u camille -F Camille -p Yunohost
# Sip more coffee

## Install your favorite apps, for example:
# Nextcloud (PHP+mysql)
yunohost app install nextcloud --args "domain=bullseye-bookworm-migration.test&path=/cloud&admin=camille&init_main_permission=visitors&user_home=true&system_addressbook_exposed=yes"
# Wekan (Nodejs + mongodb) 
yunohost app install wekan --args "domain=bullseye-bookworm-migration.test&path=/wekan&admin=camille&init_main_permission=visitors"
# Terraforming mars (nodejs+postgresql) Actually requires a few GB of RAM to npm build ? :|
# yunohost app install terraforming-mars --args "domain=bullseye-bookworm-migration.test&path=/tfm&init_main_permission=visitors"
# Coin (Python + postgresql)
yunohost domain add coin.bullseye-bookworm-migration.test
yunohost app install coin --args "domain=coin.bullseye-bookworm-migration.test&admin=camille&email=camille@example.tld&isp_name=YunoHost&isp_site=yunohost.org"
```

#### Run the actual migration
```bash
## Requirements for the migration
# Gotta make sure we get to unstable, because only unstable packages of yunohost 12.x are available now
sed -i 's/stable/stable testing unstable/g' /etc/apt/sources.list.d/yunohost.list
apt update
apt full-upgrade

# Manual apply of this pr
mv /usr/lib/python3/dist-packages/yunohost/migrations/0027_migrate_to_bookworm.py.disabled /usr/lib/python3/dist-packages/yunohost/migrations/0027_migrate_to_bookworm.py

# Validate the migration is there
yunohost tools migrations list --pending

# Run the migration from the command line (OR restart yunohost-api and run the migration from the webadmin?)
yunohost tools migrations run --accept-disclaimer --debug

# Here be dragons
```

PR #1973, opened 2024-10-19
Use the new repo.yunohost.org debian repository

Salamandar:
<untested>
I suggested to use a unified domain for "built stuff we serve", repo.yunohost.org. It could be nice to use it before releasing bookworm.


alexAubin:
Hmpf if we really want to do this then it should go target the migration branch in https://github.com/YunoHost/yunohost/pull/1911, and we also need a system to patch existing bookworm install, and it also needs to be propagated on the x86 isos and arm image build scripts

Salamandar:
Yeah, I agree, and was thinking about how to do that.
It could be handled by the regen-conf hooks (they don't do that yet).

PR #1997, opened 2024-11-09
Add possibility to allow/disallow main email, email alias and email forward form portal

Josue-T:
PR related to: https://github.com/YunoHost/yunohost-portal/pull/21

## The problem

In Yunohost portal, user is always allowed to edit the mail alias and mail forward rule.

In some instance it could be a security issue to allow to edit by example the mail alias.

## Solution

Add 3 new Yunohost settings to define if the user is allowed to edit
- the main email
- the mail alias
- the mail forward

## PR Status

Tested locally it work

## How to test

- Checkout the branch on yunohost and yunhost-portal
- Allow the 3 rules in "Tools" > "Yunohost settings" > "Portal" from admin
- Test that from the yunohost portal, the user can edit everything.
- Disallow the 3 rules in "Tools" > "Yunohost settings" > "Portal" from admin
- Test that from the yunohost portal, the user can't edit everything.


zamentur:
On instances with a lot of users, the problem is that it's possible for a user to define alias like `administrator@domain.tld`... or to create emails near other user emails `sam.dupont@domain.tld` even if  `sam.dupond@domain.tld` exists for an other user.

src/portal.py
```diff
@@ -198,14 +207,47 @@ def portal_update(
...
+                    "mail_already_exists", mail=mainemail
+                )
+
+            if domain not in domains or not user_is_allowed_on_domain(username, domain):
+                raise YunohostValidationError("mail_alias_unauthorized", domain=domain)
+            new_mails[0] = mainemail
+        else:
+            # email already exist in the list we just move it on the first place
+            new_mails.remove(mainemail)
+            new_mails.insert(0, mainemail)
```
zamentur, line 235: So if we have in the form
```
Mail: sam@example.com
Mail alias: 
```
And we replace `sam@example.com` by `camille@example.com`, we get
```
Mail: camille@example.com
Mail alias: sam@example.com
```
instead of
```
Mail: camille@example.com
Mail alias: 
```

zamentur:
```suggestion
            new_mails = [mainemail] + new_mails[1:]
```

share/config_global.toml
```diff
@@ -71,6 +71,18 @@ i18n = "global_settings_setting"
         optional = true
         default = ""
 
+    [security.portal]
+        [security.portal.portal_allow_edit_email]
+        type = "boolean"
```
zamentur, line 76:
```suggestion
        type = "boolean"
        default = 0
```
If the user change the mail left part, the autoconfig won't work anymore.

zamentur: or we should a dynamic language to generate the autoconfig...

src/settings.py
```diff
@@ -246,6 +249,17 @@ def _apply(
...
+        global_portal_settings_path = Path(f"{PORTAL_SETTINGS_DIR}/global_settings.json")
+        portal_allow_edit_email = form.get("portal_allow_edit_email", self.get("security.portal.portal_allow_edit_email"))
+        portal_allow_edit_email_alias = form.get("portal_allow_edit_email_alias", self.get("security.portal.portal_allow_edit_email_alias"))
+        portal_allow_edit_email_forward = form.get("portal_allow_edit_email_forward", self.get("security.portal.portal_allow_edit_email_forward"))
+        write_to_json(
+            str(global_portal_settings_path), {"allow_edit_email": portal_allow_edit_email,
+                                               "allow_edit_email_alias": portal_allow_edit_email_alias,
+                                               "allow_edit_email_forward":portal_allow_edit_email_forward
+                                               }, sort_keys=True, indent=4
+        )
```
zamentur, line 261: I think this should already managed by ConfigPanel. It doesn't work without ?

Josue-T: The issue currently is that we store 2 config at different places. One global setting which is not accessible by the portal API and an other one more specific to the portal api which is accessible by the portal api (and which is specific to the domain).

The issue currently is that to me this setting is expected to be global it don't really make sens to be different between 2 domain as it's also a security thing. So I added this parameters in the global setting. But this can't be read by the config panel. So it's why I added one global config for the portal API (sychromized from the global setting) so the portal api can read the config.

Josue-T:
Hello,

Any news around this ? @alexAubin maybe can we merge ?

share/config_global.toml
```diff
@@ -71,6 +71,19 @@ i18n = "global_settings_setting"
...
+        default = 0
+
+        [security.portal.portal_allow_edit_email_alias]
+        type = "boolean"
+        default = 1
+
+        [security.portal.portal_allow_edit_email_forward]
+        type = "boolean"
+        default = 1
+
```
alexAubin, line 86:
```suggestion
    [security.portal]
        [security.portal.portal_allow_edit_email]
        type = "boolean"
        default = false

        [security.portal.portal_allow_edit_email_alias]
        type = "boolean"
        default = true

        [security.portal.portal_allow_edit_email_forward]
        type = "boolean"
        default = true

```

src/portal.py
```diff
@@ -185,6 +192,7 @@ def portal_me():
 
 def portal_update(
     fullname: Union[str, None] = None,
+    mainemail: Union[str, None] = None,
```
alexAubin, line 195: From https://github.com/YunoHost/yunohost-portal/pull/21 it looks like the data is called `mail` and not `mainemail` so uuuuuh did you actually try the code that change the mainemail or am i missing something x_X

(In fact i would rather stick to calling this arg `mail` to be somewhat consistent with the admin API which also use `mail` for the main email arg in `user_update()`)

src/portal.py
```diff
@@ -242,6 +284,10 @@ def portal_update(
         new_attr_dict["mail"] = mails
```
alexAubin, line 285: At the very end of this function, we return a dict with fullname, mailalias, mailforward

We probably want to add the main email to be consistent

src/portal.py
Josue-T: Well, euh maybe at the time when I did this I didn't get that theses fields was mapped. From what I remember it was working so it's strange that it was working. Anyway I did the fix and tested, now it's working so it should be ok.

src/portal.py
Josue-T: Thanks, done

Josue-T:
Thank you @alexAubin for the review, I did the fix so it should be ok now.

PR #2005, opened 2024-12-03
Pydantic v2

Axolotle:
## Solution
Use new features of pydantic v2 to improve type coercing/validation.

Constraints:
Now use custom "validators" (constraints) that can be used directly in type definition but also as dynamic annotation.
Changed the way Option's annotation is built with those constraints

Option's `normalize` and `humanize` now properly validate passed values and serialize it depending on modes: 
- "string", for bash scripts which is the mode used in `normalize` for now but will next be called in a new method `stringify`
- "human", called in `humanize` to display human readable values
- Another mode will be implemented at some point: "normal" which will be traditional python types and will be the serializing mode of `normalized`
Those modes are also available to serialize a full form (`normal` mode will be used to serialize non app config panels settings)

None:
Harmonized "none" values for every type, `(None, "", "none", "_none")` (and uppercase) are treated as None and serialized depending on type.

New attributes: multiple
Options can now use `multiple: true` to accept a list of values, it will properly parse coma list string like `F1,F2,F3` as lists then will perform validation on each item. So you can use `multiple: true` on `type: string` with a custom pattern that will be used on all list items. All option types can be `multiple` except `password|file`.
Added some new tests with this attr.

Removed call to xfail/xpass in tests and fixed remaining parsing/validation errors

## PR Status
WIP

## TODO
- [ ] reflect changes on web admin
- [ ] implement mode `normal` (serialized as python, not "bash/string")
- [ ] test redact
- [x] check/clean/add i18n errors keys

## How to test
Yes


zamentur:
I think `multiple:` attribute is not handled correctly for now in helpers/config

PR #2010, opened 2024-12-09
fix unset variable in helper 2.1 mysql

kay0u:
## The problem

During an upgrade of wordpress:

> WARNING - /usr/share/yunohost/helpers.v2.1.d/mysql: ligne 36: character_set : variable sans liaison

## Solution

I think there is a typo, replace `character_set` by `default_character_set` should fix the issue

## PR Status

...

## How to test

...


alexAubin:
heuarg why am i only seeing this PR now @_@

PR #2013, opened 2024-12-13
Typing

Salamandar:
## The problem

CLI via API

## PR Status

untested

### DIFF
--- a/bin/yunohost-api
+++ b/bin/yunohost-api
@@ -49,6 +49,14 @@ def _parse_api_args() -> argparse.Namespace:
         type=int,
         help="Port to listen on (default: %d)" % DEFAULT_PORT,
     )
+    srv_group.add_argument(
+        "-s",
+        "--socket",
+        action="store",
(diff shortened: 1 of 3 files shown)


zamentur:
About doing CLI via API, it will be more difficult to debug in a simple way with pdb...

But about this pr, it means the socket it properly protected, maybe we should integrate a security to check that the socket is only usable by authorized users (root ?) ?

Salamandar:
> About doing CLI via API, it will be more difficult to debug in a simple way with pdb...

Agreed, although i never used pdb.
I think we need to keep a "native" cli (that would run IN PLACE of the server)

> But ab this pr, it means the socket it properly protected, maybe we should integrate a security to check that the socket is only usable by authorized users (root ?) ?

This socket file can be created by systemd / tmpfiles to ensure propre creation.

Also, not root bu a "yunohost-admins" group. I don't know (yet) how the server can know which user is accessing programmatically.

PR #2017, opened 2024-12-16
Shellcheck : Fix hooks/

Salamandar:


hooks/backup/50-conf_manually_modified_files
```diff
@@ -18,7 +18,13 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
+# Exit hook on subcommand error or unset variable
+set -Eeuo pipefail
+
+# Source YNH helpers
+# shellcheck source=../../helpers/helpers
```
zamentur, line 25: o_O

hooks/conf_regen/06-slapd
```diff
@@ -101,13 +102,14 @@ do_pre_regen() {
...
     # Define if we need to migrate from hdb to mdb
-    curr_backend=$(grep '^database' /etc/ldap/slapd.conf 2> /dev/null | awk '{print $2}')
-    if [ -e /etc/ldap/slapd.conf ] && [ -n "$curr_backend" ] \
-        && [ $curr_backend != 'mdb' ]; then
-        backup_dir="/var/backups/dc=yunohost,dc=org-${curr_backend}-$(date +%s)"
-        mkdir -p "$backup_dir"
-        slapcat -b dc=yunohost,dc=org -l "${backup_dir}/dc=yunohost-dc=org.ldif"
-        echo "$backup_dir" > "$tmp_backup_dir_file"
+    if [ -e /etc/ldap/slapd.conf ]; then
+        curr_backend=$(grep '^database' /etc/ldap/slapd.conf | awk '{print $2}')
```
zamentur, line 106: Is there a reason to remove the `2> /dev/null` ?

hooks/conf_regen/06-slapd
```diff
@@ -101,13 +102,14 @@ do_pre_regen() {
...
+        if [ "$curr_backend" != 'mdb' ]; then
```
zamentur, line 107: What if $curr_backend is empty string `""` ? Are you sure it secure to remove `[ -n "$curr_backend" ]`
```suggestion
        if [ "$curr_backend" != 'mdb' ] && [ -n "$curr_backend" ]; then
```

hooks/conf_regen/06-slapd
zamentur:
```suggestion
        curr_backend=$(grep '^database' /etc/ldap/slapd.conf 2> /dev/null | awk '{print $2}')
```

hooks/conf_regen/15-nginx
```diff
@@ -18,9 +18,12 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
+# Exit hook on subcommand error or unset variable
 set -e
```
zamentur, line 22: Test seems to fail for regenconf nginx, so may be that's why we keep only set -e ?

zamentur (changes requested):
nginx error on adding domain shoudl be fixed before merging

hooks/backup/50-conf_manually_modified_files
zamentur: I don't understand if it's really needed to add `shellcheck source=../../helpers/helpers`

Salamandar: The linter doesn't understand what `source /usr/share/yunohost/helpers` means: we usually don't develop on our yunohost machines.
We need to tell shellcheck where to find the helpers, and that is a relative path to the current file.

hooks/conf_regen/06-slapd
Salamandar: I think this can't happen: the "grep" would fail. and the script would exit. But I'm not sure.

PR #2018, opened 2024-12-17
Use samesite lax for portal cookies

Josue-T:
## The problem

https://github.com/YunoHost-Apps/synapse_ynh/issues/497

After more investigation it seem to be related to the `samesite` settings of the `yunohost.portal` cookie.

So I did 2 test one with yunohost 11 and one with yunohost 12. With this process:
- Install synapse
- Install element (on the yunohost instance)
- Login on the SSO
- Try to login on matrix from the local element instance
  - Pass on Yunohost 11
  - Pass on Yunohost 12
- Try to login on matrix from https://app.element.io (which is a completely other domain)
  - Pass on Yunohost 11
  - Fail on Yunohost 12

Now try to do this small fix on yunohost 12 and see that it solve the issue. 

Note that there was already a discussion about this some year ago here: https://github.com/YunoHost/SSOwat/pull/103

## Solution

Use the same settings than on bullseye for the SSO cookie.

## PR Status

Tested and it fixed the issue

## How to test

Try to connect to a synapse server on a yunohost instance from https://app.element.io

### DIFF
--- a/src/authenticators/ldap_ynhuser.py
+++ b/src/authenticators/ldap_ynhuser.py
@@ -252,7 +252,7 @@ class Authenticator(BaseAuthenticator):
             secure=True,
             httponly=True,
             path="/",
-            samesite="strict" if not is_dev else None,
+            samesite="lax" if not is_dev else None,
             domain=f".{request.get_header('host')}",
             max_age=SESSION_VALIDITY
             - 600,  # remove 1 minute such that cookie expires on the browser slightly sooner on browser side, just to help desimbuigate edge case near the expiration limit
@@ -305,7 +305,7 @@ class Authenticator(BaseAuthenticator):
             secure=True,
             httponly=True,
             path="/",
-            samesite="strict" if not is_dev else None,
+            samesite="lax" if not is_dev else None,
             domain=f".{request.get_header('host')}",
             max_age=SESSION_VALIDITY
             - 600,  # remove 1 minute such that cookie expires on the browser slightly sooner on browser side, just to help desimbuigate edge case near the expiration limit


Thatoo:
Would that solve also the searx issue https://github.com/YunoHost-Apps/searxng_ynh/issues/363 ?

Josue-T:
> Would that solve also the searx issue [YunoHost-Apps/searxng_ynh#363](https://github.com/YunoHost-Apps/searxng_ynh/issues/363) ?

I don't know but you can check easily by applying the patch manually and then check if it fix the issue. Don't forget to restart the portal-api service after the change 😉

Josue-T:
Just one comment for impacted user by this issue. Here are the process to apply the match manually:
- Edit `/usr/lib/python3/dist-packages/yunohost/authenticators/ldap_ynhuser.py` with your editor
- Replace `samesite="strict" if not is_dev else None,` by `samesite="lax" if not is_dev else None,` (just replace `strict` by `lax`), on line 255 and 308 (maybe the line could be a little bit different).
- Save your changes and close the editor
- Run this command: `systemctl restart yunohost-portal-api.service`

felurx:
We have a weird issue similiar to https://github.com/YunoHost/issues/issues/2493, where on Android (with Firefox, not sure about other browsers) if a link is clicked in another app or a homescreen shortcut is opened, the user would get redirected to the app list (no matter if  logged in previously or not). Copy-pasting a link into the address bar didn't trigger that behaviour.

I tried manually applying the patch and that seems to have fixed the problem. 
Thanks to @Josue-T for writing it!

zamentur:
@Josue-T could you detail the domain list in used in your tests ?

https://github.com/YunoHost/issues/issues/2474
https://github.com/YunoHost/issues/issues/2522

Josue-T:
> @Josue-T could you detail the domain list in used in your tests ?
> 
> Indeed, thos bugs exists too, and the situation depends how domain and sub domain are deployed... [YunoHost/issues#2474](https://github.com/YunoHost/issues/issues/2474) [YunoHost/issues#2522](https://github.com/YunoHost/issues/issues/2522)

Well by example I tested with https://app.element.io/ on my local matrix server on `matrix.mydomain.tld`. Mainly the issue on synapse is when the login request come from a completely different domain (the element app) and target synapse on a complelty different domain. If we have by example element on `mydomain.tld` and we try to connect on `matrix.mydomain.tld` it will probably work.

zamentur (approved):
I made some research on this topic and some tests with a fresh firefox and trying several attacks. All attacks i tried (even with yunohost app package too permissive)  was quite impossible due to browser default policies like 
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

I think we could merge this PR which will improve user experience in several scenario.

PR #2023, opened 2024-12-29
Improving go helper

yalh76:
## The problem

There was some issues and missing parts in the go helper

## Solution

Use the Ruby helper as starting point to improve the go helper

## PR Status

Ready

## How to test

Put it in package like lxd and use it / I can do some test PR to show that it works


tituspijean:
`ynh_use_go` is missing from the v2.1 file. 🤔

yalh76:
> `ynh_use_go` is missing from the v2.1 file. 🤔

Yes, it has been removed during https://github.com/YunoHost/yunohost/commit/a8cd94d3db9db86b4a7830888a993ebddcc77bd7 "replaced" by `_ynh_load_go_in_path_and_other_tweaks`

helpers/helpers.v1.d/go
```diff
@@ -90,7 +80,7 @@ ynh_use_go() {
     ynh_go_load_path="PATH=$PATH"
 
     # Sets the local application-specific Go version
-    pushd $install_dir
+    pushd ${install_dir:-$install_dir}
```
Salamandar, line 83: What's the meaning of this substitution ?

helpers/helpers.v1.d/go
```diff
@@ -130,36 +120,51 @@ ynh_install_go() {
 
     # Install or update goenv
     mkdir -p $goenv_install_dir
-    pushd "$goenv_install_dir"
-    if ! [ -x "$goenv_install_dir/bin/goenv" ]; then
-        ynh_print_info --message="Downloading goenv..."
-        git init -q
-        git remote add origin https://github.com/syndbg/goenv.git
+    goenv="$(command -v goenv $goenv_install_dir/bin/goenv | grep "$goenv_install_dir/bin/goenv" | head -1)"
```
Salamandar, line 123: This command will fail if goenv is not present. You should something like that instead:
```
if command -v …
```

helpers/helpers.v1.d/go
yalh76: Bad copy paste from: https://github.com/YunoHost/yunohost/blob/a96cb9bb2c1a74806b9571daf4009652d46b35bb/helpers/helpers.v1.d/ruby#L95

helpers/helpers.v1.d/go
yalh76: In that case

https://github.com/YunoHost/yunohost/blob/a96cb9bb2c1a74806b9571daf4009652d46b35bb/helpers/helpers.v1.d/ruby#L135
and
https://github.com/YunoHost/yunohost/blob/a96cb9bb2c1a74806b9571daf4009652d46b35bb/helpers/helpers.v2.1.d/ruby#L77

should not be ok

helpers/helpers.v1.d/go
```diff
@@ -226,29 +235,37 @@ ynh_remove_go() {
 ynh_cleanup_go() {
 
     # List required Go versions
-    local installed_apps=$(yunohost app list --output-as json --quiet | jq -r .apps[].id)
+    local installed_apps=$(yunohost app list | grep -oP 'id: \K.*$')
```
alexAubin, line 238: Uuuuh not sure why switching to the new version ... the current one with json and `jq` seems much more clean and robust ? Is it that the ruby helpers still use an hacky grep ?

helpers/helpers.v1.d/go
```diff
@@ -192,7 +200,8 @@ eval \"\$(goenv init -)\"
 #goenv" > /etc/profile.d/goenv.sh
 
     # Load the environment
-    HOME=$install_dir eval "$(goenv init -)"
+    export HOME=${HOME:-"/root/"}
+    eval "$(goenv init -)"
```
alexAubin, line 204: I would tend to keep the previous syntax such that `HOME` doesn't get change in the rest of the script which could have side effect idk ... also we probably want to make sure that HOME is the app's install dir and not /root/ ?

helpers/helpers.v1.d/go
```diff
@@ -174,12 +179,15 @@ ynh_install_go() {
     test -x /usr/bin/go_goenv && mv /usr/bin/go_goenv /usr/bin/go
 
     # Install the requested version of Go
-    local final_go_version=$("$goenv_latest_dir/bin/goenv-latest" --print "$go_version")
-    ynh_print_info --message="Installation of Go-$final_go_version"
-    goenv install --skip-existing "$final_go_version"
+    local final_go_version=$(goenv latest --print $go_version)
+    if ! [ -n "$final_go_version" ]; then
```
alexAubin, line 183:
```suggestion
    if [[ -z "$final_go_version" ]]; then
```

helpers/helpers.v1.d/go
```diff
@@ -18,41 +18,32 @@
...
-    fi
-}
-
 goenv_install_dir="/opt/goenv"
 go_version_path="$goenv_install_dir/versions"
-# goenv_ROOT is the directory of goenv, it needs to be loaded as a environment variable.
+
+# GOENV_ROOT is the directory of goenv, it needs to be loaded as a environment variable.
 export GOENV_ROOT="$goenv_install_dir"
+export goenv_root="$goenv_install_dir"
```
alexAubin, line 26: (are we sure that having that same var in lowercase is actually useful for anything x_x ... i guess it comes from copypasting the same stuff as for ruby but that sounds very meh)

helpers/helpers.v2.1.d/go
```diff
@@ -19,15 +19,17 @@
...
+# GOENV_ROOT is the directory of goenv, it needs to be loaded as a environment variable.
 export GOENV_ROOT="$GOENV_INSTALL_DIR"
+export goenv_root="$GOENV_INSTALL_DIR"
 
 _ynh_load_go_in_path_and_other_tweaks() {
 
-    # Get the absolute path of this version of go
-    go_dir="$GOENV_INSTALL_DIR/versions/$app/bin"
+    # Get the absolute path of this version of Go
+    go_dir="$GOENV_INSTALL_DIR/versions/$go_version/bin"
```
alexAubin, line 30: I'm super confused because it is `$app` for the ruby helper x_X Or is the ruby helper also wrong ? https://github.com/YunoHost/yunohost/blob/dev/helpers/helpers.v2.1.d/ruby#L30

I don't understand what defines the name of these folder ... I assume they are created when calling `rbenv install` ? X_x

helpers/helpers.v1.d/go
alexAubin: To me what's troublesome is : what is it doing in the first place x_X

It looks like this is meant to verify that `$RBENV_INSTALL_DIR/bin/rbenv` exists (and maybe is executable) ? Shouldn't we just use something like 

```
rbenv="$RBENV_INSTALL_DIR/bin/rbenv"
if [[ -x "$rbenv" ]]
then
     ...
else
     ...
fi
```

(In fact that was the original code x_x)

alexAubin:
(Now i'm just super confused about what's the best starting point between ruby and go, but it definitely looks like the ruby helper code could use some simplifications too @_@)

helpers/helpers.v1.d/go
yalh76: Yes ruby helper still using grep
https://github.com/YunoHost/yunohost/blob/f7ed895c1811d13e444e043b0c9ace6401518bbc/helpers/helpers.v2.1.d/ruby#L231
and
https://github.com/YunoHost/yunohost/blob/f7ed895c1811d13e444e043b0c9ace6401518bbc/helpers/helpers.v1.d/ruby#L285

helpers/helpers.v2.1.d/go
yalh76: Ruby helper use the alias plugins : 
https://github.com/YunoHost/yunohost/blob/f7ed895c1811d13e444e043b0c9ace6401518bbc/helpers/helpers.v2.1.d/ruby#L126
and
https://github.com/YunoHost/yunohost/blob/f7ed895c1811d13e444e043b0c9ace6401518bbc/helpers/helpers.v1.d/ruby#L184

That plugins is not available for goenv, so there is no `$app` in `$GOENV_INSTALL_DIR/versions/`

yalh76:
> (Now i'm just super confused about what's the best starting point between ruby and go, but it definitely looks like the ruby helper code could use some simplifications too @_@)

I don't know either.

The point is: Go helper was a copy/pasta of Ruby as experimental helper. 
goenv is like a fork of rbenv
So both helper should be relativly the same: that was the case when you implemented 2.1 helpers

helpers/helpers.v1.d/go
yalh76: I don't think, it's usefull, but was in ruby herlper

helpers/helpers.v1.d/go
yalh76: I think we should remove :
the part :
https://github.com/YunoHost/yunohost/blob/cd12849b94454f06486a7657292d1efb462ef4ee/helpers/helpers.v2.1.d/go#L150-L159
from each helper Go and Ruby; helper 2 and helper2.1

The first part `/etc/profile.d/****` is to set the environment variables at startup of the Bash shell. But we don't need go and ruby environment variables set to each users.

The second part `eval \"\$(goenv init -)\"` is to initialise environment for current user, but we don't need that either, as all we need is done in `_ynh_load_go_in_path_and_other_tweaks` same for ruby

Limezy:
Hi all, thanks for your work on the matter !
Do you have any visibility on what is blocking this PR to move further ?
There are a few apps left in a broken state for installation as of now (at least Headscale and Outline, but maybe a few others)
I wish I could help but I'm unfortunately very limited in my knowledge of go (and Yunohost helpers mechanisms)
Thanks !

ImaCrea:
Hello around :) A warm little message here of support <3 (I'd love to try out Outline but need to install Dex first 🙏 )

yalh76:
> Hi all, thanks for your work on the matter ! Do you have any visibility on what is blocking this PR to move further ? There are a few apps left in a broken state for installation as of now (at least Headscale and Outline, but maybe a few others) I wish I could help but I'm unfortunately very limited in my knowledge of go (and Yunohost helpers mechanisms) Thanks !

@alexAubin  is already working on #2042

Limezy:
Great piece of news !
Will also unlock Headscale package

zamentur:
Related PR https://github.com/YunoHost/yunohost/pull/2288

PR #2026, opened 2025-01-15
Change default value of show_other_domains_apps to True

Salamandar:
Not having other domains' apps is a special use case, and newbies will just want to see all apps.
It's more intuitive to set it to True

### DIFF
--- a/share/config_domain.toml
+++ b/share/config_domain.toml
@@ -27,7 +27,7 @@ i18n = "domain_config"
 
         [feature.portal.show_other_domains_apps]
         type = "boolean"
-        default = false
+        default = true
 
         [feature.portal.portal_title]
         type = "string"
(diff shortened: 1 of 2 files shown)


zamentur:
So it currently means that if we have domainA.tld and domainB.tld, when i click from domainA.tld/yunohost/sso/ on tile going on domainB.tld, i arrive on the second portal domainB.tld/yunohost/sso/ , i have to relogin (due to cookie same_site=strict cookie policy and domain limited) and after i can access a protected APP...

See https://github.com/YunoHost/yunohost/pull/2018
https://github.com/YunoHost/issues/issues/2522
https://github.com/YunoHost/issues/issues/2474

Salamandar:
> i have to relogin

Yes, indeed, that might be something nice to fix.

But… At least people can *see* the tile for the other domain's app and click on it. So even if they won't understand why they need to re-login, they can see the app is installed and won't ask "hey $packager, your app doesn't appear on the dashboard".

PR #2036, opened 2025-01-22
Allow ignoring of YNH version.

orhtej2:
## The problem

Some apps require 12.x even though they should work on 11.x.
The idea is to merge this to `dev` then backport to `bullseye`.

## Solution

Allow users to ignore required YNH version, even if it may break.

## PR Status

To be tested.

## How to test

Install Nextcloud 30.x on YNH 11.x

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -869,6 +869,10 @@ app:
                     full: --force
                     help: Do not ask confirmation if the app is not safe to use (low quality, experimental or 3rd party), or when the app displays a post-install notification
                     action: store_true
+                -i:
+                    full: --ignore-yunohost-version
+                    help: Attempt to install the app even if your YunoHost version is below the required one
+                    action: store_true
 
         ### app_remove()
         remove:
@@ -907,6 +911,10 @@ app:
                 -c:
                     full: --continue-on-failure
                     help: Continue to upgrade apps even if one or more upgrade failed
+                    action: store_true
+                -i:
+                    full: --ignore-yunohost-version
+                    help: Attempt to install the app even if your YunoHost version is below the required one
                     action: store_true
 
         ### app_change_url()
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -914,7 +914,7 @@ app:
                     action: store_true
                 -i:
                     full: --ignore-yunohost-version
-                    help: Attempt to install the app even if your YunoHost version is below the required one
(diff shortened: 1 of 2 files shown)


yalh76:
I think, it would be a better idea to decrease the YunoHost version required on the package that still works on bullseye.

In the other hand, increaing the yunohost version required, reduce the work needed to maintain bullseye version and invite user to upgrade from bullseye to bookworm

orhtej2:
This is meant to be opt-in you're-on-your-own for people who are confident about what they're doing. I'm fully supporting 'you should move to 12.x' initiative.
This relies on #2034 getting merged in the first place, otherwise the point is moot.

orhtej2:
Works:

```
> sudo yunohost app upgrade fittrackee -i
Info: Now upgrading fittrackee…
Warning: This app requires YunoHost >= 12.0 but current installed version is 11.3.0.2
```

I was able to upgrade `dendrite` to 0.14.1 and `nextcloud` to 30.0.4 no problem after applying #2034 to `bullseye` server.

share/actionsmap.yml
```diff
@@ -908,6 +912,10 @@ app:
                     full: --continue-on-failure
                     help: Continue to upgrade apps even if one or more upgrade failed
                     action: store_true
+                -i:
+                    full: --ignore-yunohost-version
+                    help: Attempt to install the app even if your YunoHost version is below the required one
```
zamentur, line 917:
```suggestion
                    help: Attempt to upgrade the app even if your YunoHost version is below the required one
```

src/app.py
```diff
@@ -1104,6 +1108,8 @@ def app_install(
                 _ask_confirmation(
                     "confirm_app_insufficient_ram", params=values, force=force
                 )
+            elif name == "required_yunohost_version" and ignore_yunohost_version:
+                logger.warning(m18n.n(err **values))
```
zamentur, line 1112:
```suggestion
                logger.warning(m18n.n(err, **values))
```

src/app.py
```diff
@@ -667,6 +668,8 @@ def app_upgrade(
                     _ask_confirmation(
                         "confirm_app_insufficient_ram", params=values, force=force
                     )
+                elif name == "required_yunohost_version" and ignore_yunohost_version:
+                    logger.warning(m18n.n(err **values))
```
zamentur, line 672:
```suggestion
                    logger.warning(m18n.n(err, **values))
```

zamentur (approved):

PR #2038, opened 2025-02-05
Fix firewall

Salamandar:
## The problem

https://forum.yunohost.org/t/python-became-crazy/35186

### DIFF
--- a/src/firewall.py
+++ b/src/firewall.py
@@ -239,8 +239,8 @@ class YunoUPnP:
         for protocol, port in firewall.upnp_to_close:
             status = status and self.close_port(protocol, port)
 
-        for protocol, ports in firewall.config.items():
-            for port, info in ports.items():
+        for protocol in ["tcp", "udp"]:
+            for port, info in firewall.config[protocol].items():
                 if self.enabled():
                     status = status and self.open_port(protocol, port, info["comment"])
                 else:


src/firewall.py
```diff
@@ -237,14 +237,14 @@ def refresh(self, firewall: "YunoFirewall") -> bool:
 
         status = True
         for protocol, port in firewall.upnp_to_close:
-            status = status or self.close_port(protocol, port)
+            status = status and self.close_port(protocol, port)
 
-        for protocol, ports in firewall.config.items():
-            for port, info in ports.items():
+        for protocol in ["tcp", "udp"]:
+            for port, info in firewall.config[protocol].items():
```
alexAubin, line 243: Hmmm that kind of stuff makes me paranoid about wether or not the expected keys does exist in the file ... We've seen in the past situation where yaml file (app settings) became empty because at some point yunohost tries to update the file, but there's no space left on the server (eg because a backup went crazy and filled the disk) and then somehow the file still exists but is empty ...

I suppose one day or another, we'll also have users randomly editing the file with typos so I would also make sure that somehow the expected keys (tcp, udp, ..?) do exist in the file

Can we iterate on the `def read(...)` method for YunoFirewall adding failsafes such as returning a dict with the appropriate structure ? :grimacing:

alexAubin: Hmyeah I'd say return an error (or a warning and then some failsafe values) if for some reason we're unable to read the file or it doesn't have the expected structure

PR #2041, opened 2025-02-07
[enh] Allow to login with a mail

zamentur:
## The problem

Some user tried to log with a mail, it seems previously it was possible ???
https://github.com/YunoHost/issues/issues/2475

## Solution

Search the username binded to a mail

## PR Status

Tested manually

## How to test

Try to log with username or with email

### DIFF
--- a/src/authenticators/ldap_ynhuser.py
+++ b/src/authenticators/ldap_ynhuser.py
@@ -27,6 +27,7 @@ from pathlib import Path
 
 import jwt
 import ldap
+import ldap.filter
 import ldap.sasl
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import padding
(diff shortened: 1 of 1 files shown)


src/authenticators/ldap_ynhuser.py
```diff
@@ -201,6 +202,13 @@ def _authenticate_credentials(self, credentials=None):
         except ValueError:
             raise YunohostError("invalid_credentials")
 
+        username = ldap.filter.escape_filter_chars(username)
+        # Search username, if user give a mail instead
+        if "@" in username:
+            user = _get_ldap_interface().search("ou=users", f"mail={username}", ["uid"])
```
zamentur, line 208: @alexAubin  do you think there is a security risk to expose the _get_ldap_interface during authentication ?

alexAubin: Hmmm idk, why do you think it could be a security risk ? :thinking: I don't quite see what kind of injection could happen there

zamentur: Don't know how this API should be isolated or not from the other part of the code that was written with hypothesis that the API user is an admin...

I don't remember well how is made _get_ldap_interface and if we could imagine injection through username  (or other strange thing like attack on memory).

alexAubin: Nah it's fine, the portal API runs as non-root and our LDAP ACL are decent

`_get_ldap_interface()` will only open an anonymous connection in the context of the portal, cf : https://github.com/YunoHost/yunohost/blob/dev/src/utils/ldap.py#L90

Worst case scenario, the attacker could find some way to inject stuff via the provided username (but `escape_filter_chars` seems to prevent that i suppose ?) and maybe enumerate / leak the whole list of existing username maybe but that's about it

zamentur: Ok so i mark this as solved

PR #2042, opened 2025-02-09
Helpers / app resources: simplify helpers and implement new resources for nodejs/ruby/go/composer

alexAubin:
## The problem

Redundant code in app packages, nodejs / ruby / go / composer should be managed declaratively like apt deps etc

## Solution

Implement them as resource

## PR Status

Kinda yolo, not complete

The code is kind of a mess because it needs to call bash commands with proper env vars x_x

### DIFF
--- a/helpers/helpers.v2.1.d/go
+++ b/helpers/helpers.v2.1.d/go
@@ -61,8 +61,6 @@ fi
 #
 # - `$path_with_go` (the value of the modified `$PATH`, but you dont really need it?)
 # - `$go_dir` (the directory containing the specific go version)
-#
-# This helper also creates a /etc/profile.d/goenv.sh that configures PATH environment for goenv
 ynh_go_install() {
 
     [[ -n "${go_version:-}" ]] || ynh_die "\$go_version should be defined prior to calling ynh_go_install"
--- a/helpers/helpers.v2.1.d/go
+++ b/helpers/helpers.v2.1.d/go
@@ -44,6 +44,12 @@ _ynh_load_go_in_path_and_other_tweaks() {
     popd
 }
 
+# Auto-load Nodejs path tweaks if this app uses the nodejs resource in the manifest
+if [ -n "${go_version:-}" ] && (cat "$YNH_APP_BASEDIR/manifest.toml" | toml_to_json | jq -q ".resources.go" >/dev/null)
+then
+    _ynh_load_go_in_path_and_other_tweaks
+fi
+
 # Install a specific version of Go using goenv
 #
 # The installed version is defined by `$go_version` which should be defined as global prior to calling this helper
--- a/helpers/helpers.v2.1.d/go
+++ b/helpers/helpers.v2.1.d/go
@@ -78,7 +78,7 @@ ynh_go_install() {
     go_version=$final_go_version
     ynh_app_setting_set --app="$app" --key="go_version" --value="$go_version"
 
-    ynh_print_info "Installation Go $go_version"
+    ynh_print_info "Installing Go $go_version"
     $GOENV_ROOT/bin/goenv install --quiet --skip-existing "$go_version" 2>&1
 
     # Cleanup Go versions
@@ -116,7 +116,7 @@ _ynh_go_cleanup() {
     for installed_go_version in $installed_go_versions; do
         if ! grep -qE "^go_version: '?$installed_go_version'?" /etc/yunohost/apps/*/settings.yml
         then
-            ynh_print_info "Removing Go-$installed_go_version"
+            # ynh_print_info "Removing Go-$installed_go_version"
             $GOENV_ROOT/bin/goenv uninstall --force "$installed_go_version"
         fi
     done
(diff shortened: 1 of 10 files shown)


yalh76:
maybe the good time to switch nodejs from https://github.com/tj/n to https://github.com/nodenv/nodenv
as nodeenv, works like rbenv or goenv

There is also other *env available, can be seen as https://github.com/momo-lab/xxenv-latest

helpers/helpers.v2.1.d/go
```diff
@@ -40,11 +39,17 @@ _ynh_load_go_in_path_and_other_tweaks() {
...
 
     # Sets the local application-specific go version
-    pushd "$install_dir"
-    $GOENV_INSTALL_DIR/bin/goenv local "$go_version"
+    pushd "${install_dir}"
+        "$GOENV_ROOT/bin/goenv" local "$go_version"
     popd
 }
 
+# Auto-load Nodejs path tweaks if this app uses the nodejs resource in the manifest
```
kay0u, line 47:
```suggestion
# Auto-load Go path tweaks if this app uses the Go resource in the manifest
```

helpers/helpers.v2.1.d/utils
```diff
@@ -607,3 +607,32 @@ ynh_smart_mktemp() {
...
+#
+function _ynh_git_clone()
+{
+    local url="$1"
+    local dest_dir="$2"
+    # FIXME: hmmm we might want some way to specify the branch ? idk
+    #local url="$(cut -f1 -d: <<< "$url_and_branch")"
+    #local branch="$(cut -f2 -d: <<< "$url_and_branch")"
+    local branch
+    branch=${branch:-master}
```
kay0u, line 624:
```suggestion
    local branch=${3:-master}
```
Something like that?

src/utils/resources.py
```diff
@@ -1608,4 +1608,395 @@ def deprovision(self, context: Dict = {}):
...
+            env={"RBENV_ROOT": self.RBENV_ROOT},
+        )
+        self.set_setting("ruby_version", ruby_version)
+        logger.info(f"Building Ruby {ruby_version}, this may take some time...")
+        self._run_script(
+            "provision_or_update",
+            f"""
+            #export RBENV_ROOT='{self.RBENV_ROOT}'
+            export RUBY_CONFIGURE_OPTS='--disable-install-doc --with-jemalloc'
+            export MAKE_OPTS='-j2'
```
kay0u, line 1798: Does it takes time to compile? Should we use number of cores instead?

kay0u (approved):
LGTM

helpers/helpers.v2.1.d/go
```diff
@@ -171,39 +110,22 @@ ynh_go_remove() {
...
     done
 
-    # If none Go version is required
-    if ((${#required_go_versions[@]} == 0)); then
+    # If no app uses Go anymore
+    if ! grep -q "^go_version:" /etc/yunohost/apps/*/settings.yml; then
         # Remove goenv environment configuration
-        ynh_print_info "Removing of goenv"
-        ynh_safe_rm "$GOENV_INSTALL_DIR"
+        ynh_print_info "Removing goenv"
```
ericgaspar, line 127:
```suggestion
```
I find GO helper unnecessarily talkative

src/utils/resources.py
alexAubin: Yeah I was wondering too about where this `2` comes from, it's purely a copypasta of the options in the current helpers ... I dunno what we want to do exactly, maybe it is only 2 to not overload the server ... but maybe using half of the available core could be smarter ?

Anyway it seem to take time to compile indeed, on my test machine (thinkpad) it did took like 10-20 minutes I think ? And it does take significant time on the CI too because it does compile Ruby :/ ... (I don't understand why there can't be pre-compiled versions like Debian ships ...)

helpers/helpers.v2.1.d/go
alexAubin: Ugh yeah indeed especially since at the end of the day I think it's only downloading a 50~100 MB file and not compiling Go locally ?

src/utils/resources.py
alexAubin:
```suggestion
            # Use half of available CPUs for compiling
            # The trick with 1 + ...  -1 is to prevent having '0' when there's only one CPU available.
            NB_CPU_TO_USE="$(( 1 + ($(grep -c ^processor /proc/cpuinfo) - 1) / 2 ))"
            export MAKE_OPTS="-j$(NB_CPU_TO_USE)"
```

tituspijean:
FYI in v12.1.3 while updating apps:

```
Warning: /usr/share/yunohost/helpers.v2.1.d/nodejs: line 49: toml_to_json: command not found
Warning: jq: Unknown option -q
Warning: Use jq --help for help with command-line options,
Warning: or see the jq manpage, or online docs  at https://stedolan.github.io/jq
```

PR #2046, opened 2025-02-21
Add SMART status to disk infos

christophehenry:
Rel: YunoHost/issues#1823

So this one may be a bit controversial since it rewrite the D-Bus part using [`sdbus-python`](https://python-sdbus.readthedocs.io/en/latest/index.html) which is not yet bundled in Debian. I think, however using `sdbus-python` will have a positive effect on code since it D-Bus proxies are plain Python classes which improves code expressiveness and API discovery.


src/utils/udisks2_interfaces.py
```diff
@@ -0,0 +1,818 @@
...
+    def __init__(self, bus: SdBus):
+        super().__init__(UDISKS2_SERVICE_NAME, f"{UDISKS2_BASE_PATH}/Manager", bus)
+
+    @dbus_method(
+        input_signature="s",
+        result_signature="(bs)",
+        flags=DbusUnprivilegedFlag,
+    )
+    def can_format(self, type: str) -> Tuple[bool, str]:
+        raise NotImplementedError
```
zamentur, line 85: Have you tried to delete all NotImplemented methods ?

zamentur:
Could you explain advantages to use python-sdbus ? Maybe we could try to make pro and cons for the 2 solutions ?

src/utils/udisks2_interfaces.py
christophehenry: Those interfaces a automatically generated by the CLI as mentioned at the beginning of the file. The implementation is provided by `@dbus_method`. The `NotImplementerError` is just here to have a method content. It could be `pass` but that's what the CLI generates so I kept it.

christophehenry:
With sdbus, properties and methods are defined by the generated interfaces. This makes the infos as easy to retrieve as accessing a class property whereas with the dbus-python library you have to unmarshall several layers of dictionary. It makes the code way cleaner and I think it's necessary as I plan to add more features using Udisks2 like formatting. Now the methods are already defined.

PR #2048, opened 2025-02-22
Storage managment API: make MyPy happy :)

christophehenry:


christophehenry:
Tant pis :sweat_smile:

PR #2063, opened 2025-03-06
Add check_status on UPnP device, rework how to detect if there is an UPnP device

kay0u:
## The problem

...

## Solution

...

## PR Status

Tested on my prod lul

## How to test

...

### DIFF
--- a/src/firewall.py
+++ b/src/firewall.py
@@ -197,7 +197,7 @@ class YunoUPnP:
     def enabled(self, new_status: bool | None = None) -> bool:
         if new_status is not None:
             self.firewall.config["router_forwarding_upnp"] = new_status
-        self.firewall.write()
+            self.firewall.write()
         return self.firewall.config.get("router_forwarding_upnp", False)
 
     def ensure_listen_port(self) -> None:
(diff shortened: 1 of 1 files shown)


src/firewall.py
```diff
@@ -189,7 +189,10 @@ class YunoUPnP:
     def __init__(self, firewall: "YunoFirewall") -> None:
         self.firewall = firewall
         self.description = "Yunohost firewall"
-        self.upnpc: miniupnpc.UPnP | None = None
+        self.upnpc = miniupnpc.UPnP()
+        self.upnpc.localport = self.UPNP_PORT
```
kay0u, line 193: I'm not sure why we define a port here, it works with the default one.

src/firewall.py
```diff
@@ -200,33 +203,52 @@ def enabled(self, new_status: bool | None = None) -> bool:
     def ensure_listen_port(self) -> None:
         self.firewall.open_port("udp", self.UPNP_PORT, self.UPNP_PORT_COMMENT)
```
kay0u, line 204: This function is never called, I think we can remove it?

alexAubin (approved):
Quick review, LGTM

src/firewall.py
Salamandar: I think i kept this from the original code.

kay0u:
Ready to merge IMO

PR #2066, opened 2025-03-12
[fix] FDN dns resolver are not available for all AS

zamentur:
Only FFDN subnet are authorized now...

## The problem

https://www.fdn.fr/attaque-serveurs-dns-2025/

## Solution

Remove FDN resolver

## PR Status

Ready

## How to test

...

### DIFF
--- a/conf/dnsmasq/plain/resolv.dnsmasq.conf
+++ b/conf/dnsmasq/plain/resolv.dnsmasq.conf
@@ -6,6 +6,7 @@
 
 # List taken from
 # http://diyisp.org/dokuwiki/doku.php?id=technical:dnsresolver
+# and some others from https://sebsauvage.net/wiki/doku.php?id=dns-alternatifs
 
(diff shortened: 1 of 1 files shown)


zamentur:
This PR should be merged in bullseye to... cause it could break the upgrade.

We need to process quickly and made an announcement.

alexAubin:
>This PR should be merged in bullseye to... cause it could break the upgrade.

Not sure to see how it would break the upgrade, we have several resolvers in the list precisely to avoid everything breaking up when some goes down ...?

zamentur:
> Not sure to see how it would break the upgrade, we have several resolvers in the list precisely to avoid everything breaking up when some goes down ...?

Yes but if a resolver accept the packet and doesn't respond it could trigger timeout (30s). I remember there was a bug in dnsmasq about this (and about the order of the list, reason why we finally shuffle it).

PR #2075, opened 2025-03-23
[fix] Send mail to external mailbox of a managed domain

zamentur:
## The problem

https://forum.yunohost.org/t/how-to-keep-your-mailbox-outside-yunohost/4860
https://github.com/YunoHost/issues/issues/2275

## Solution

List domains for which we managed mailboxes into a file /etc/postfix/virtual-mailbox-domains

## PR Status
Regenconf hooks untested (just tested by editing file in /etc/postfix)

## How to test

...


conf/postfix/main.cf
```diff
@@ -99,7 +99,7 @@ inet_interfaces = all
 message_size_limit = 35914708
 
 # Virtual Domains Control
-virtual_mailbox_domains = ldap:/etc/postfix/ldap-domains.cf
```
alexAubin, line 102: So shall we get rid of that file from the regenconf ? https://github.com/YunoHost/yunohost/blob/dev/conf/postfix/plain/ldap-domains.cf

zamentur: Done by emptying the file

conf/dnsmasq/domain.tpl
```diff
@@ -5,5 +5,9 @@ interface-name={{ domain }},{{ interface }}
 {% if ipv6 %}
 host-record={{ domain }},{{ ipv6 }}
 {% endif %}
+{% if mail_out == "1" %}
 txt-record={{ domain }},"v=spf1 mx a -all"
```
zamentur, line 9: What if another server is used to send mails to for a domain ?
For example we have a serverA, a yunohost with mailbox toto@domain.tld

And a serverB, with a website that send emails noreply@domain.tld to toto@domain.tld.

The official zone includes the IP of the mailserver of serverB in SPF, but here we have our dnsmasq lying, so i guess the postfix of ynh serverA could refuse the mail !

zamentur: In fact, i think this line is only here for the usecase where DNSmasq is configured as the authority DNS server...

zamentur: I think i will creates an other PR for this...

Josue-T:
Great feature. 

I don't know if it could also fit into this work but an other great feature would be to be able to set a user email which is out of the Yunohost managed domain. So by example the user email could be `hello@gmail.com` even if Yunohost only manage the domain `example.com`. In case I would like to provide a Yunohost account to some user but I don't would like to provide the email to the user, currently we are quite limited.

Ideally we would have something like this:

- If the user have the `mail.main` permission` he must have a Yunohost domain email.
- If the user don't have the `mail.main` permission, in this case we can set an external email (ideally we must but for retro compatibility it could be more complicated to force to have an external email when `mail.main` is not allowed). In this case we shouldn't be able to set the mail alias and mail forward field because it don't really make sense in this case.

What do you think ?

zamentur:
@Josue-T I agree with you, it could be very nice to have this (for example to avoid sending nextcloud notif on an internal mail when users have external mails...).

However, it's completely an other PR i think. I have created an issues https://github.com/YunoHost/issues/issues/2565

PR #2081, opened 2025-03-29
tools_rootpw: Replace spwd with direct call to passwd.

Salamandar:
## The problem

spwd was deprecated in python 3.11, removed in 3.13 (trixie).

## Solution

Call directly passwd, that uses the PAM config.

## PR Status

ready to test

## How to test

...

### DIFF
--- a/src/tools.py
+++ b/src/tools.py
@@ -52,10 +52,7 @@ def tools_versions():
 
 
 def tools_rootpw(new_password, check_strength=True):
-    import spwd
-
     from yunohost.utils.password import (
-        _hash_user_password,
         assert_password_is_compatible,
         assert_password_is_strong_enough,
     )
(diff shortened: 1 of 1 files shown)


src/tools.py
```diff
@@ -52,10 +52,7 @@ def tools_versions():
 
 
 def tools_rootpw(new_password, check_strength=True):
-    import spwd
-
     from yunohost.utils.password import (
-        _hash_user_password,
```
alexAubin, line 58: Hmmmm but there are other places where `_hash_user_password` is used, shouldn't we fix that function rather than just `tools_rootpw` ?

src/tools.py
```diff
@@ -64,29 +61,21 @@ def tools_rootpw(new_password, check_strength=True):
...
-        with open("/etc/shadow", "r") as before_file:
-            before = before_file.read()
+    proc = subprocess.Popen(
+        ["passwd"],
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+    out, err = proc.communicate(f"{new_password}\n{new_password}\n".encode("utf-8"))
+    result = proc.wait()
```
alexAubin, line 71: These days apparently it's more practical to use `subprocess.run()` I think ? Maybe something like : 

```
out, err = subprocess.run(["passwd"], input=f"{new_password}\n{new_password}\n".encode("utf-8"), capture_output=True)
```

src/tools.py
alexAubin: (Ah I just realized the new code calls passwd, it doesn't hash the password per se)

src/tools.py
Salamandar: Ah didn't know you could pass `input` like that ! yeah feels about right.

PR #2082, opened 2025-03-31
fix ynh_apt_install_dependencies_from_extra_repository

kay0u:
## The problem

I had an error because `apt-mark auto` was called without any argument.

## Solution

We want a non-empty array here

## PR Status

...

## How to test

...


Thovi98:
I confirm it resolves one issue I asked in the apps chat recently, thank you!

Salamandar:
🥵

PR #2084, opened 2025-04-09
Refactorize conf-regen for apt

Salamandar:
* Use conf files instead of bash strings
* Replace list files with deb822 source files
* Manage yunohost's sources file and gpg key


conf/apt/sources.list.d/yunohost.sources
```diff
@@ -0,0 +1,5 @@
+Types: deb
+URIs: http://forge.yunohost.org/debian/
+Suites: bookworm
+Components: stable
+Signed-By: /etc/apt/trusted.gpg.d/yunohost-bookworm.gpg
```
alexAubin, line 5: Eeeeeh that sounds cool but that's going to break stuff like https://install.yunohost.org/switchtoTesting and https://github.com/YunoHost/yunohost/blob/dev/src/diagnosers/00-basesystem.py#L156-L168

And some old setups are likely to have the yunohost repo inside `/etc/apt/sources.list` ooooooor maybe even all installs done via the debian ISO ? I'm not sure ... cf https://github.com/YunoHost/custom-debian-iso/blob/main/files_to_inject/preseeds/default.preseed#L58-L61 (there's nothing suggesting it creates an independent file ? idk)

(btw isn't it `repo` instead of `forge` now ?)

Salamandar: > (btw isn't it repo instead of forge now ?)

Yes agreed, I wanted to change that too.

I'll handle the rest of the scripts changes.

Salamandar:
I tested it and the legacy (unmanaged) yunohost.list file is left as "pending" when running regen-conf. Any idea ?

PR #2085, opened 2025-04-12
Implement git in ynh_setup_source

yalh76:
## The problem
- Some packages require a git repository when building the app
- Some packages can't be handle simply with ynh_setup_source because the is some sub-git repository for example https://github.com/misskey-dev/misskey
- Some package upgrade would be improved if using git instead of `--keep="a directory with a lot of stuff"` for example: https://github.com/YunoHost-Apps/mastodon_ynh/issues/424
- Bored to see package linter warning `! Using 'git clone' is not recommended ... most forge do provide the ability to download a proper archive of the code for a specific commit. Please use the 'sources' resource in the manifest.toml in combination with ynh_setup_source. `

## Solution

Implement git in ynh_setup_source

## PR Status

Ready and working

## How to test

A proof of concept has been done on https://github.com/YunoHost-Apps/peertube-search-index_ynh/pull/47

### DIFF
--- a/helpers/helpers.v2.1.d/sources
+++ b/helpers/helpers.v2.1.d/sources
@@ -207,14 +207,7 @@ ynh_setup_source() {
     elif [[ "$src_format" == "docker" ]]; then
         "$YNH_HELPERS_DIR/vendor/docker-image-extract/docker-image-extract" -p "$src_platform" -o "$dest_dir" "$src_url" 2>&1
     elif [[ "$src_format" == "git" ]]; then
-        if [ -d "$dest_dir/.git" ]; then
-            chown -R root:root "$dest_dir"
-            git -C "$dest_dir" fetch --quiet
-        else
-            git clone "$src_url" "$dest_dir" --quiet
-        fi
-        git -C "$dest_dir" checkout "$src_sum" --quiet
-        git -C "$dest_dir" submodule update --init --recursive
+        _ynh_git_clone "$src_url" "$dest_dir" "master" "$src_sum"
     elif [[ "$src_format" == "zip" ]]; then
         # Zip format
         # Using of a temp directory, because unzip doesn't manage --strip-components
--- a/helpers/helpers.v2.1.d/sources
+++ b/helpers/helpers.v2.1.d/sources
@@ -207,6 +207,7 @@ ynh_setup_source() {
     elif [[ "$src_format" == "docker" ]]; then
         "$YNH_HELPERS_DIR/vendor/docker-image-extract/docker-image-extract" -p "$src_platform" -o "$dest_dir" "$src_url" 2>&1
     elif [[ "$src_format" == "git" ]]; then
+        chown -R root:root "$dest_dir"
         _ynh_git_clone "$src_url" "$dest_dir" "master" "$src_sum"
     elif [[ "$src_format" == "zip" ]]; then
         # Zip format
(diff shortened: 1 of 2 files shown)


alexAubin (commented):
Soooooooo agreed on principle :+1: The situation with Mastodon is pretty uuugh "interesting" and using git would be a nice way to adress it i suppose (assuming that's not what's already done?)

However digging into the specific I have a couple question / remarks :grimacing: : 
1. it the goal to use commit ids as checksum ? Or do you foresee cases where using a branch is absolutely needed (ideally commit ids are better, both because the point to a very precise state so we can properly say "this is version X.Y.Z" *and* because in terms of security, the upstream project cannot simply ship malicious code by updating the branch/tag)
2. I suppose sooner or later somebody is gonna complain and/or be confused that we write `sha256 = 'some commit id'` in the manifest because commit ids are in fact sha1 ... I don't know if we should really care about this, changing the keyword for this case may trigger some explosion in other tools x_x
3. ideally it would be nice to use the new `_ynh_git_clone` introduced for the go/ruby helpers : https://github.com/YunoHost/yunohost/blob/dev/helpers/helpers.v2.1.d/0-utils#L615 (many tweaks probably needed to adapt its behavior to also fit in ynh_setup_source ? at least the submodule stuff)
in particular it uses this trick of *not* git-cloning because `git fetch` can be used to fetch a single branch or even commit instead of downloading the entire history, which may improve performances (idk if we have use case where we care about that tho)
4. the `chown -R root:root $dest_dir` is spooky, though I assume you added this to work around the infamous git error of `dubious ownership` etc ? Ideally I think I would rather aim to run all the git commands as `$app` instead ?

Salamandar:
I would love to support the """standard""" pkgbuild git source syntax:

https://wiki.archlinux.org/title/PKGBUILD#Sources

the "schema" would be :
```
{rename}::{protocol}://{uri}#{refspec}
```
where protocol can be `http(s)`, `git+ssh`, `git+https(s)` and refspec would be `branch=XXX` or `commit=XXX` or `tag=XXX`.

yalh76:
1. Yes, the goal is to use the commit id
2. The change from `sha256` to sha1... not sure, I think there are two options :
   a. maybe changing `sha256 = 'some commit id'` to `checksum = 'some commit id'` in [Packaging v3](https://github.com/YunoHost/yunohost/pull/2070)
   b. stay like it is actually, as there are some plan to change commit id from sha1 to sha256 (https://git-scm.com/docs/hash-function-transition, https://gist.github.com/ChristopherA/65da131fea445f58b9075d4e8c85ed80 )
3. just switched to `_ynh_git_clone` on https://github.com/YunoHost-Apps/peertube-search-index_ynh/pull/47, I let you know when it's working
4. yes it's cause of `dubious ownership`, I will look what is better: keeping the `chown` swichting to `ynh_exec_as_app`

yalh76:
@alexAubin seems quite ok now

Tested on https://github.com/YunoHost-Apps/peertube-search-index_ynh/pull/47

helpers/helpers.v2.1.d/sources
```diff
@@ -206,6 +206,9 @@ ynh_setup_source() {
         fi
     elif [[ "$src_format" == "docker" ]]; then
         "$YNH_HELPERS_DIR/vendor/docker-image-extract/docker-image-extract" -p "$src_platform" -o "$dest_dir" "$src_url" 2>&1
+    elif [[ "$src_format" == "git" ]]; then
+        chown -R root:root "$dest_dir"
+        _ynh_git_clone "$src_url" "$dest_dir" "master" "$src_sum"
```
Thovi98, line 211: AFAIU, this will only work for repo that have a master branch.

I wanted to try this pr for Discourse, and I had this issue because Discourse has a `main` branch, not a master one. For the test, I've modified from `master` to `main` and it works great
https://github.com/YunoHost-Apps/discourse_ynh/pull/407

zamentur:
Dear contributors,

PR #2088, opened 2025-04-14
Fix tls_passthrough_list regex pattern, the regex should match the whole list and not individual elts?

Salamandar:
## The problem

```
List of forwarding: repo.yunohost.org;repo.local;443,release-creator.yunohost.org;repo.local;443,install.yunohost.org;repo.local;443,build.yunohost.org;repo.local;443,forge.yunohost.org;repo.local;443
Error: You should specify a list of items formatted as DOMAIN;DESTINATION;DESTPORT, such as yolo.test;192.168.1.42;443
```

## Solution

pattern matches "no comma at the beginning" and then yeah, comma separated list

### DIFF
--- a/share/config_global.toml
+++ b/share/config_global.toml
@@ -172,7 +172,7 @@ i18n = "global_settings_setting"
         [misc.tls_passthrough.tls_passthrough_list]
         type = "tags"
         # Regex is just <domain regex>;<destination>;<port>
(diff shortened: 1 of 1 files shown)


zamentur:
For me as it's a tags question the comma should be implicit (seems it's not the case, it's quite strange) !

zamentur:
An issue says that validation error of regex is not triggered ! https://github.com/YunoHost/issues/issues/2777

What's the truth ?

PR #2097, opened 2025-04-24
Add support for multi db support

Josue-T:
## The problem

https://github.com/YunoHost/issues/issues/2245

## Solution

Implement the feature

## PR Status

Tested with a specific seafile branch and it seem working

## How to test

Tested with a specific branch of seafile: https://github.com/YunoHost-Apps/seafile_ynh/pull/142

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -1474,15 +1474,21 @@ class DatabaseAppResource(AppResource):
     """
     Initialize a database, either using MySQL or Postgresql. Relevant DB infos are stored in settings `$db_name`, `$db_user` and `$db_pwd`.
 
-    NB: only one DB can be handled in such a way (is there really an app that would need two completely different DB ?...)
-
     NB2: no automagic migration will happen in an suddenly change `type` from `mysql` to `postgresql` or viceversa in its life
 
-    ### Example
+    ### Example for single DB
 
     ```toml
     [resources.database]
     type = "mysql"   # or : "postgresql". Only these two values are supported
+    ```
+
+    ### Example for 2 DB
+
+    ```toml
+    [resources.database]
+    main.type = "mysql"   # first DB
+    second.type = "postgresql" # Second DB
     ```
 
     ### Properties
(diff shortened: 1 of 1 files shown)


src/utils/resources.py
```diff
@@ -1572,40 +1585,48 @@ def provision_or_update(self, context: Dict = {}):
...
-                    f"ynh_mysql_create_db '{db_name}' '{db_user}' '{db_pwd}'",
-                )
-            elif self.dbtype == "postgresql":
-                self._run_script(
-                    "provision",
-                    f"ynh_psql_create_user '{db_user}' '{db_pwd}'; ynh_psql_create_db '{db_name}' '{db_user}'",
-                )
+        for name in self.dbtypes.keys():
+            db_suffix = f"_{name}" if name != "main" else ""
+            db_name = self.get_setting(f"db_name{db_suffix}") or (db_user + db_suffix)
```
Josue-T, line 1590: Note that for now I did a simple solution. With the current implementation we will end with something like this:
- `myapp`: main db name
- `myapp_admin`: additional db for `myapp`

But the risk is the name collision. By example if we have an other app named `myapp_admin` (we have something like this by example for `synapse-admin` / `synapse`) we might have a db name collision because the app `myapp_admin` will also have the db name `myapp_admin`.

The solution that I thought could be to ensure to add always a suffix to all db name. So with the last example the db name will be:
- `myapp_main`: main db name
- `myapp_admin`: additional db for `myapp`

In this case the db name for the app `myapp_admin` will be `myapp_admin_main`.

The big advantage is that we will never have any name collision this way, but there are some issue, mainly:
- Retro compatibility: currently it's not the case, changing this might break some apps
- Migration: we have 2 choices:
  - Migrate the current db name but it's really risky....
  - Keep the current db as it is: we can live with it but in this case all old instances will have the db name which will have a different name strategy than the new instances which is not ideal.

Josue-T: Yes would be an interesting solution.

alexAubin:
Closing this during PR-Lanta tonight, to be supported in packaging v3 sorryyyyyyy :sob:

PR #2098, opened 2025-04-25
Increase DKIM key length

Josue-T:
## The problem

Using in 2025 a RSA 1024 key for DKIM is not really strong

Related issues: not found

## Solution

Use a 2048 instead. It seem that we have some way to pass around the nsupdate limitation cf:
- https://lists.isc.org/pipermail/bind-users/2020-June/103238.html
- https://gitlab.isc.org/isc-projects/bind9/-/issues/1907

So the workaround  following the message is:

> Break up the record in chunks of less than 255 byte, enclose each of these parts with "" and feed nsupdate all of these chunks seperated with a space on one line.

Seeing the answer they will probably never provide a better solution in nsupdate to handle the 2048 key, maybe we need to integrate the workaround. I don't know exactly where we use nsupdate so I didn't provide for now the workaround on the impacted code.

## PR Status

Tested
### Code TODOs
 - [ ] May be add a comment /help in dns suggest to suggest to remove " " if the registrar record forms needs to remove that
 - [x] Discuss about this change with others contributors (on chat, contributors meeting, forum, issues or pr)
 - [x] Check if yunohost-admin should be updated
 - [x] Check if dynette should be updated
 - [x] Write migration to upgrade 1024 dkim keys
   - [x] ~~Add some missing translations keys~~ n/a
   - [x] ~~Update/write unit tests~~
 - [x] ~~Update/write documentation: ~~

### Tests TODOs
   - [x] Test with `yunohost domain add`
   - [x] Test with `yunohost dyndns update`
   - [x] Test with `yunohost diagnosis run dnsrecords --force` `yunohost diagnosis show dnsrecords`
   - [x] Test with `yunohost domain dns push`

## How to test

...


src/dns.py
```diff
@@ -307,56 +307,33 @@ def _get_DKIM(domain):
...
             dkim.group("host"),
-            '"v={v}; h={h}; k={k}; p={p}"'.format(
-                v=dkim.group("v"),
-                h=dkim.group("h"),
-                k=dkim.group("k"),
-                p=dkim.group("p"),
-            ),
+            re.sub(r'"[\s\n]+"', '" "',
+                   dkim.group("c").strip(),
+                   re.M | re.S),
```
zamentur, line 334:
```suggestion
            re.sub(r'"[\s\n]+"', '',
                   dkim.group("c").strip(),
                   re.M | re.S),
```

zamentur, line 334: DeprecationWarning: 'count' is passed as positional argument

zamentur:
The splitted key need to be tested on dyndns and push API.

TODO:
 - [x] Test with yunohost dyndns update
 - [x] Test with yunohost domain dns push
 - [ ] ~~May be add a comment /help in dns suggest to suggest to remove `" "` if the registrar record forms needs to remove that~~

zamentur:
Note that some emails are refused due to this short key length, reason why i tag it as bugfix

Josue-T:
Note: feel free to work on this PR because I not sure I will have time to work on this one for theses few next months.

orhtej2:
I've applied the code on my instance. While I was able to run the migration and push the updated records to OVH mxtoolbox claims the generated DKIM TXT is invalid (invalid base64 key) 🤷 

The mail seems to be working just fine nevertheless

zamentur:
Summary of the discussion during PRlanta:

Due to DNS TTL (and resolver which doesn't respect TTL), it could be possible an instance send mail with 1024bits key during the migration, or with the 2048bits during the cache TTL of DNS resolver. That means some mails could be considered as using false DKIM keys.

We have studied a way to do the migration with a second DKIM selector. However this method needs several zone changes from admin and the add of a new entry (and not just update the entry), even for people who have already done the change manually... That's means it could be a problem for several type of public:
 * admins who are not comfortable with changing this kind of DNS records (and are not using dyndns)
 * profesionnal adminsys who coturn zone delegation using a CNAME (and could be in difficulties to coordinate with customers this kind of action) rodinux or ljf are concerned
 * admins who doesn't care of mails for their server but want a green diagnostic...

In more those change need a lot of rewrite code, and the gain is not easy. It could create a new confusion by introducing a new DNS record mail2._domainkey...

 - [x] So after discussing with @Josue-T we have considered that adding a check that the mailq is empty and some advice to avoid sending emails during the migration could be a reasonable way to proceed.

zamentur (approved):
LGTM (if someone can improve wording, it could be nice)

Josue-T:
LGTM, I think we can merge it.

Note ideally in long term it would be great to have full DKIM rotation management. I created https://github.com/YunoHost/issues/issues/2810 for this.

zamentur:
- [ ] ~~.key and .txt are in "manually modified" state~~ Not managed by regenconf.yml cause key are not conf...
- [x] update domains with no mail_in features but with an existing dkim keys already generated (all?)

PR #2099, opened 2025-04-26
WIP: App domain resource

Josue-T:
## The problem

https://github.com/YunoHost/issues/issues/2563

## Solution

Implement the feature

## PR Status

In draft

## How to test

A test app was made with [cryptpad](https://github.com/YunoHost-Apps/cryptpad_ynh/pull/243) and with the [MAS branch of synapse](https://github.com/YunoHost-Apps/synapse_ynh/pull/536) with postgresql

### DIFF
--- a/helpers/helpers.v1.d/nginx
+++ b/helpers/helpers.v1.d/nginx
@@ -35,23 +35,30 @@
 #
 # Requires YunoHost version 4.1.0 or higher.
 ynh_add_nginx_config() {
+    local finalnginxconf
+    local domain_var_name
+    local extra_domains_list
+    extra_domains_list="$(ynh_app_setting_get --app="$app" --key=_extra_domains |
+        jq -r 'to_entries | .[] | {key: ("_" + .key), value: .value} | join(",")')"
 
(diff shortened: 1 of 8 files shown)


zamentur:
Should it be a domain question ? in order to customize the second extra domains ???

Some apps are not web app, so may be we should declare the main domain as a ressources too ? (don't remeber how are declared the non webapp)

Josue-T:
> Should it be a domain question ? in order to customise the second extra domains ???

Initially I didn't have the idea to provide the possibility to customise the extra domain. Maybe it could be useful in some case but I also think that maybe in same case we might need to have the possibility to have a static value, so the user can't customise. By example for XMPP I'm not sure that it's possible to customise the extra domains.

> Some apps are not web app, so may be we should declare the main domain as a ressources too ?

Yes it's also a possibility to have also the main domain as resource but currently it seem that a resource is defined only by the packager and can't be modified, while the domain is something that can selected at the install time.

>  (don't remeber how are declared the non webapp)

I think the app like this just don't ask for a domain at install time.

PR #2102, opened 2025-05-02
add komgo to the exemption list of basic_auth

Thovi98:
## The problem

https://forum.yunohost.org/t/komga-premiere-install-et-sso-first-install-and-sso/36520/3?u=tho

## Solution

Add komga to the list

## PR Status

ready

## How to test

Adapt /usr/lib/python3/dist-packages/yunohost/app.py and install the app.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1818,6 +1818,7 @@ def app_ssowatconf() -> None:
             "baikal",
             "ihatemoney",
             "keeweb",
+            "komga",
             "monica",
             "my_webdav",
             "nextcloud",


alexAubin:
Ideally it's better if the app scripts themselves set the `protect_against_basic_auth_spoofing` setting to true

That stuff in the code exists mostly because it was easier than doing a PR on each app *and* asking everybody to upgrade the apps after upgrading to Bookworm

Thovi98:
Understood! Thanks for the explanation.

PR #2105, opened 2025-05-05
Use YunoHost behind a reverse proxy using proxy_protocol

Tagadda:
This is yolocommit :) Not really tested yet
This PR should allow YunoHost to accept proxy protocol requests on port 444 from allowed IPs, while accepting regular HTTPS requests on port 443.
FIXME:
- Locales
- Check nftables syntax


conf/nginx/server.tpl.conf
```diff
@@ -15,16 +15,9 @@ server {
     location ^~ '/.well-known/ynh-diagnosis/' {
         alias /var/www/.well-known/ynh-diagnosis/;
     }
-    {%- if mail_enabled == "True" %}
-
-    location ^~ '/.well-known/autoconfig/mail/' {
-        alias /var/www/.well-known/{{ domain }}/autoconfig/mail/;
-    }
-    {%- endif %}
```
alexAubin, line 23: \<whisper\>not sure why this got deleted :eyes: \</whisper\>

Tagadda: heh I guess it shouldn't be accessible by unencrypted plain text http and redirect to https instead :shrug:

alexAubin: Ah yes I didnt realize it's for HTTP hmmmm I wonder why it's still there, I sort of remember reworking and pondering many time what we leave in the plain HTTP part ... there might be a boring story of "actually mail client may test the HTTP first and then 'fallback'(?)(sigh) to HTTPS ... I hope since then mail client do check HTTPS first and anyway yeah it should be ok to throw a 302 :white_check_mark:

conf/nginx/server.tpl.conf
```diff
@@ -41,33 +34,24 @@ server {
...
 server {
-    {%- if tls_passthrough_enabled != "True" %}
-    listen 443 ssl http2;
-    listen [::]:443 ssl http2;
-    {%- else %}
-    listen 127.0.0.1:444 ssl proxy_protocol;
+    listen 444 ssl http2 proxy_protocol;
     port_in_redirect off;
-    set_real_ip_from 127.0.0.1/32;
+    {% if trusted_proxies_list is defined and trusted_proxies_list|length > 0 -%}
```
alexAubin:
```suggestion
    {% if trusted_proxies_list is defined and trusted_proxies_list.strip()|length > 0 -%} 
```

conf/nginx/tls_passthrough.conf
```diff
@@ -2,18 +2,17 @@
 stream {
 
     map $ssl_preread_server_name $name {
-        {% for domain_ip in domain_ip_map %}
+	hostnames;
```
alexAubin:
```suggestion
        hostnames;
```

hooks/conf_regen/15-nginx
```diff
@@ -89,25 +90,21 @@ do_pre_regen() {
...
-            ynh_render_template "tls_passthrough_server.conf" "${nginx_conf_dir}/${tls_passthrough_domain}.forward80.conf"
-        done
-    else
-        touch "${tls_passthrough_module}"
-    fi
+    ynh_render_template "tls_passthrough.conf" "${tls_passthrough_module}"
+    for tls_passthrough_domain_and_ip in ${tls_passthrough_list//,/ }; do
+        export tls_passthrough_domain=$(echo "$tls_passthrough_domain_and_ip" | awk -F';' '{print $1}')
+        export tls_passthrough_ip=$(echo "$tls_passthrough_domain_and_ip" | awk -F';' '{print $2}')
+        export tls_passthrough_port=$(echo "$tls_passthrough_domain_and_ip" | awk -F';' '{print $3}')
```
alexAubin:
```suggestion
        export tls_passthrough_domain=$(awk -F';' '{print $1}' <<< "$tls_passthrough_domain_and_ip")
        export tls_passthrough_ip=$(awk -F';' '{print $2}' <<< "$tls_passthrough_domain_and_ip")
        export tls_passthrough_port=$(awk -F';' '{print $3}' <<< "$tls_passthrough_domain_and_ip")
```

Nitpicking here though I may be the one that wrote the original code, but using `<<<` instead of `echo` prevents the content from actually ending up in the logs which in turn should prevent it to leak when sharing logs (though we have other ways to anonymize but maybe only the domain names "actually registered" are redacted)

locales/en.json
```diff
@@ -499,6 +499,9 @@
     "global_settings_setting_postfix_compatibility": "Postfix Compatibility",
     "global_settings_setting_postfix_compatibility_help": "Compatibility vs. security tradeoff for the Postfix server. Affects the ciphers (and other security-related aspects)",
     "global_settings_setting_postfix_name": "Postfix (SMTP email server)",
+    "global_settings_setting_trusted_proxies_list": "Trusted proxies list",
+    "global_settings_setting_trusted_proxies_list_help": "",
```
alexAubin, line 503: Ideally it's better to add "trusted_proxies_list" to the "settings_without_help_key" here : https://github.com/YunoHost/yunohost/blob/dev/maintenance/missing_i18n_keys.py#L217 (it's a bit hackish but that's what we have for now)

On the other hand, maybe we *do* need an explication here ? If I understand correctly, this is the list of the "front" servers when being the back / reverse-proxied server ?

src/settings.py
```diff
@@ -311,7 +311,6 @@ def regen_ssowatconf(setting_name, old_value, new_value):
         app_ssowatconf()
 
 
-@post_change_hook("tls_passthrough_enabled")
```
alexAubin, line 314: Maybe we should have a `post_change_hook` to auto-regen the firewall conf when `trusted_proxies_list` changes ?

locales/en.json
```diff
@@ -526,9 +529,7 @@
...
-    "global_settings_setting_tls_passthrough_explain": "This feature is ADVANCED and EXPERIMENTAL and will trigger major changes in the nginx configuration of this server. Please DO NOT use it if you don't know what you are doing! In particular, you must be aware that fail2ban cannot be implemented on the proxied server (nftables cannot ban malicious traffic as all IP packets appear as coming from the front server). In addition, for now the proxied server's nginx configuration needs to manually tweaked to accept the `proxy_protocol`.",
+    "global_settings_setting_tls_passthrough_explain": "This feature is ADVANCED and EXPERIMENTAL. Please DO NOT use it if you don't know what you are doing! In particular, you must be aware that fail2ban cannot be implemented on the proxied server (nftables cannot ban malicious traffic as all IP packets appear as coming from the front server). In addition, for now the proxied server's nginx configuration needs to manually tweaked to accept the `proxy_protocol`.",
```
alexAubin, line 532: It sounds important to keep an explanation somewhere of what this feature is about ?

```suggestion
    "global_settings_setting_tls_passthrough_explain": "This is an advanced feature to reverse-proxy an entire domain to another machine *without* decrypting the traffic. Useful when you want to expose several machines behind the same IP but still allow each machine to handle the SSL termination. This feature is ADVANCED and EXPERIMENTAL. Please DO NOT use it if you don't know what you are doing! In particular, you must be aware that fail2ban cannot be implemented on the proxied server (nftables cannot ban malicious traffic as all IP packets appear as coming from the front server). In addition, for now the proxied server's nginx configuration needs to manually tweaked to accept the `proxy_protocol`.",
```

Also maybe we want to update the last sentence to refer to the "trusted proxies list" ?

conf/nftables/nftables.d/yunohost-firewall.tpl.conf
```diff
@@ -4,6 +4,7 @@ define tcp_ports = { {{ tcp_ports.strip().split(' ') | join(', ') }} }
 {% if udp_ports.strip() %}
 define udp_ports = { {{ udp_ports.strip().split(' ') | join(', ') }} }
 {% endif %}
+define trusted_proxies_list = { 127.0.0.1,{{ trusted_proxies_list }} }
```
alexAubin, line 7: Is that bit ok if `trusted_proxies_list` is empty ?

Tagadda: yup!

locales/en.json
Tagadda: Yup, the IP the proxy server is using to communicate with your YunoHost server. That IP would be different from the public IP the proxy is using to be reached.
IP addresses allowed to act as a proxy for your server.

src/settings.py
Tagadda: https://github.com/YunoHost/yunohost/pull/2105/files#diff-3ac2227be4f47aa1118e44e00a63059ddc4518fe5a95186bb7fc841109278adcR324-R327

conf/nginx/server.tpl.conf
```diff
@@ -41,33 +34,24 @@ server {
 }
 
 server {
-    {%- if tls_passthrough_enabled != "True" %}
-    listen 443 ssl http2;
-    listen [::]:443 ssl http2;
-    {%- else %}
-    listen 127.0.0.1:444 ssl proxy_protocol;
+    listen 444 ssl http2 proxy_protocol;
```
zamentur, line 37: SO if i understand well, server.tpl.conf use 444 by default the proxy_protocol and http2 are enabled on it.

## Warning on address IP spoofing
Spoofing address ip is prevent because only some ips are allowed to by nftable. However if nftable is down or accept too much traffic, an attacker could spoof the IP. In more, i don't know if it's secure to use proxy_protocol outside of a local network...

Note: that 127.0.0.1 is always allowed so someone with a non root ssh account might use this account to spoof an ip by doing a request with proxy protocol. (i don't know if curl is able to use proxy_protocol).

## Regenerate nginx conf and old customized change
I think by default the regen conf don't remove old customize conf. So we could be in a situation where tls_passthrough.conf is added but the old server.tpl.conf is maintained with 443 port

hooks/conf_regen/15-nginx
```diff
@@ -89,25 +90,21 @@ do_pre_regen() {
...
-        for tls_passthrough_domain_and_ip in ${tls_passthrough_list//,/ }; do
-            export tls_passthrough_domain=$(echo "$tls_passthrough_domain_and_ip" | awk -F';' '{print $1}')
-            export tls_passthrough_ip=$(echo "$tls_passthrough_domain_and_ip" | awk -F';' '{print $2}')
-            export tls_passthrough_port=$(echo "$tls_passthrough_domain_and_ip" | awk -F';' '{print $3}')
-            ynh_render_template "tls_passthrough_server.conf" "${nginx_conf_dir}/${tls_passthrough_domain}.forward80.conf"
-        done
-    else
-        touch "${tls_passthrough_module}"
-    fi
+    ynh_render_template "tls_passthrough.conf" "${tls_passthrough_module}"
```
zamentur, line 101: Have we an idea on performance impact of making this by default ?

share/config_global.toml
```diff
@@ -175,22 +175,21 @@ i18n = "global_settings_setting"
...
 
-    [misc.tls_passthrough]
-
-        [misc.tls_passthrough.tls_passthrough_enabled]
-        type = "boolean"
-        default = false
+    [misc.trusted_proxies]
+        [misc.trusted_proxies.trusted_proxies_list]
+        type = "tags"
+        default = ""
```
zamentur, line 181: Maybe we should add a pattern to check element are in IP format ?

zamentur: Maybe local ip only...

zamentur:
Dear contributors,

PR #2114, opened 2025-05-21
Resolve logger warnings

emmanuel-ferdman:
## The problem
The `logger.warn` is deprecated so it throws annoying deprecation warnings:
```python
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```
## Solution
Migrates to `logger.warning`. Also fixes a few typos along the way.


alexAubin:
Ah cheers !

PR #2115, opened 2025-05-28
apps: refactor the upgrade flows, check requirements beforehand, allow to have a different upgrade source (eg testing branch), group system package by categories to improve UI/UX

alexAubin:
## The problem

Some people may launch an app upgrade only for yunohost to report an error, complaining that the new version of the app needs a higher yunohost version (for example 12.x when still being on 11.x)

## Solution

Add the proper info to the API such that the webadmin may display the upgrade button as disabled with the info about the failed requirements (+ same stuff into the app info page which now includes the upgrade button)

## PR Status

Kinda yolocommited / roughly tested

## How to test

Zblerg idk query the API on the GET update route / or run `yunohost app list --upgradable`


src/app.py
```diff
@@ -191,39 +183,51 @@ def app_info(app: str, full: bool = False, upgradable: bool = False) -> AppInfo:
...
+            tmp_notifications = {}
+
+        if tmp_notifications.get("PRE_UPGRADE"):
+            ret["upgrade"]["notifications"] = _filter_and_hydrate_notifications(
+                tmp_notifications["PRE_UPGRADE"],
+                ret["version"],
+                ret["settings"],
+            )
+
+        ret["upgrade"]["notifications"] = {"main": "gni", "other": "other"}
```
alexAubin, line 212: oopsies

PR #2116, opened 2025-05-31
Make basic-space-cleanup delete YunoHost logs too

tituspijean:
Addresses https://github.com/YunoHost/issues/issues/2329

I'm unsure what are all the `.YYYY-MM-DD...` symlinks for.

## PR Status

Ready?

## How to test

...

### DIFF
--- a/src/tools.py
+++ b/src/tools.py
@@ -603,7 +603,7 @@ def tools_shell(command=None):
         shell.interact()
 
 
-def tools_basic_space_cleanup():
+def tools_basic_space_cleanup(days=90):
     """
     Basic space cleanup.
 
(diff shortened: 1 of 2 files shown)


alexAubin:
(Same comment as ljf, if you don't read the log carefully, you'd think the `--days` arg applies to all files)

alexAubin (commented):
LGTM apart from the discussion on the design of --days

src/tools.py
```diff
@@ -603,21 +603,23 @@ def tools_shell(command=None):
...
     archived logs removal
+    yunohost logs removal (older than 90 days by default)
     """
     subprocess.run("apt autoremove && apt autoclean", shell=True)
     subprocess.run("journalctl --vacuum-size=50M", shell=True)
     subprocess.run("rm /var/log/*.gz", shell=True)
     subprocess.run("rm /var/log/*/*.gz", shell=True)
     subprocess.run("rm /var/log/*.?", shell=True)
     subprocess.run("rm /var/log/*/*.?", shell=True)
+    subprocess.run(f"find /var/log/yunohost/operations/ -type f,l -mtime +{days} -execdir rm {} +", shell=True)
```
alexAubin: Sounds like Python is not happy about the empty `{}` because it's a f-string ?

`SyntaxError: f-string: valid expression required before '}'`

PR #2122, opened 2025-06-14
Implement RedisAppResource

fflorent:
## The problem

Motivated by this issue: https://github.com/YunoHost/issues/issues/2583#issuecomment-2949275846

TL;DR: I saw that some packages don't look for free redis DB, like some of mine or even [PeerTube](https://github.com/YunoHost-Apps/peertube_ynh/issues/482). 

## Solution

In the above mentioned issues, instead of making a linter, @zamentur rightfully suggested to support Redis as an AppResource in the v3 manifest. Therefore it would be the role of the system to provision / deprovision redis database correctly.

Also it would make redis helpers obsolete.

## PR Status

Need help [on these questions](https://github.com/YunoHost/yunohost/pull/2122/files#r2160415723)

## How to test

```
./ynh-dev test yunohost:app_resources:resource_redis
```

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -1670,8 +1670,12 @@ class RedisDatabaseAppResource(AppResource):
             )
         super().__init__({"redis": properties}, *args, **kwargs)
 
+    def _ensure_array(self, val: str|list[str]):
+        return [val] if isinstance(val, str) else val
+
     def _get_all_alt_names(self, properties: Dict[str, Any]):
-        return [alt_name for subprops in properties.values() for alt_name in subprops.get('previous_names', [])]
+        # FIXME: Maybe this is a bit too condensed
+        return [alt_name for subprops in properties.values() for alt_name in self._ensure_array(subprops.get('previous_names', []))]
 
     def _free_db(self, count=1):
         info = check_output(
@@ -1680,12 +1684,13 @@ class RedisDatabaseAppResource(AppResource):
         databases_num_str = re.findall(r'^db(\d+):', info, flags=re.MULTILINE)
         databases_num = [int(num) for num in databases_num_str]
         redis_conf_path='/etc/redis/redis.conf'
-        max_dbs = searchf(r'^(?<=database\s+)(\d+)', redis_conf_path, flags=re.MULTILINE)
-        if max_dbs is None:
+        max_dbs_match = searchf(r'^databases\s+(\d+)', redis_conf_path, count=1, flags=re.MULTILINE)
+        if max_dbs_match is None:
             raise YunohostError(
                 f"Can't read the maximal number of Redis database allowed in {redis_conf_path}",
                 raw_msg=True
             )
+        max_dbs = int(max_dbs_match)
         free_dbs = set(range(0, max_dbs)) - set(databases_num)
         if len(free_dbs) < count:
             raise YunohostError(
@@ -1704,7 +1709,8 @@ class RedisDatabaseAppResource(AppResource):
             if self.get_setting(db_name) is not None:
                 dbs_to_provision.remove(db_name)
                 continue
-            for alt_name in self.redis.get(db_name, {}).get('previous_names', []):
+            previous_names = self.redis.get(db_name, {}).get('previous_names', [])
+            for alt_name in self._ensure_array(previous_names):
                 db_num = self.get_setting(alt_name)
                 if db_num is not None:
                     self.set_setting(db_name, db_num)
@@ -1715,10 +1721,13 @@ class RedisDatabaseAppResource(AppResource):
         new_db_nums = self._free_db(len(dbs_to_provision))
         for (db_name, db_num) in zip(dbs_to_provision, new_db_nums):
             self.set_setting(db_name, db_num)
+            # Set some unrelevant key so it is reserved for this app
+            # even if no key is put
+            os.system(f'redis-cli -n "{db_num}" SET __YNH_PROVISIONNED true')
 
     def deprovision(self, context: Dict = {}):
         for env in self.redis:
-            os.system(f"redis-cli -n \"{self.get_setting(env)}\" flushdb")
+            os.system(f'redis-cli -n "{self.get_setting(env)}" flushdb')
             self.delete_setting(env)
 
 
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -1735,7 +1735,7 @@ class RedisDatabaseAppResource(AppResource):
             # Set some unrelevant key so it is reserved for this app
             # even if no key is put
             assert (
-                os.system(f'redis-cli -n "{db_num}" SET __YNH_PROVISIONNED true > /dev/null') == 0
+                os.system(f'redis-cli -n "{db_num}" SET __YNH_PROVISIONNED {self.app} > /dev/null') == 0
             ), "Uhoh, cannot provision a Redis DB"
 
         self.set_setting('provisionned_redis_db_names', ",".join(db_names))
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -1665,7 +1665,7 @@ class RedisDatabaseAppResource(AppResource):
         colliding_names = set(db_names) & set(alt_names)
         if colliding_names:
             raise YunohostError(
-                f"Cannot define properties whose names are also specified as previous_names: {", ".join(colliding_names)}",
+                f'Cannot define properties whose names are also specified as previous_names: {", ".join(colliding_names)}',
                 raw_msg=True
             )
         super().__init__({"redis": properties}, *args, **kwargs)
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -21,6 +21,7 @@
 import copy
 import os
 import random
+import re
 import shutil
 import subprocess
 import tempfile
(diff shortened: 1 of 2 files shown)


src/utils/resources.py
```diff
@@ -1608,6 +1609,119 @@ def deprovision(self, context: Dict = {}):
...
+        free_dbs = set(range(0, max_dbs)) - set(databases_num)
+        if len(free_dbs) < count:
+            raise YunohostError(
+                f'Uhoh, you seem to have reached the maximum of Redis databases allowed. You may increase the value of "database" in {redis_conf_path} to increase it.',
+                raw_msg=True
+            )
+        return list(free_dbs)[:count]
+
+    def provision_or_update(self, context: Dict = {}):
+        # I would like to fetch the previous properties so I can deprovision redis databases missing after an upgrade
```
fflorent: I need help for that part. 

The idea is that if the packager removes a Redis database (say `[resources.redis.celery_redis_db]` above), it deprovisions only this database.

I would be tempted to make the AppResourceManager pass the old properties in the context. Would it make sense?

src/utils/resources.py
```diff
@@ -1608,6 +1609,142 @@ def deprovision(self, context: Dict = {}):
...
+            previous_names = self.redis.get(db_name, {}).get('previous_names', [])
+            for alt_name in self._ensure_array(previous_names):
+                db_num = self.get_setting(alt_name)
+                if db_num is not None:
+                    self.set_setting(db_name, db_num)
+                    self.delete_setting(alt_name)
+                    dbs_to_provision.remove(db_name)
+                    break
+
+        previously_provisioned_db = set(str(self.get_setting('provisionned_redis_db_names') or '').split(","))
```
fflorent, line 1726: I am not very fond of `provisionned_redis_db_names`, I feel that it is a bit awkward. But it sounds like a viable hack.

I have proposed an alternate implementation in #2124 .

Please let me know what approach sounds the best to you, and I'll keep only one PR.

src/utils/resources.py
```diff
@@ -1608,6 +1609,142 @@ def deprovision(self, context: Dict = {}):
...
+            if self.get_setting(db_to_deprovision) is not None:
+                self._deprovision_db(db_to_deprovision)
+
+        new_db_nums = self._free_db(len(dbs_to_provision))
+        for (db_name, db_num) in zip(dbs_to_provision, new_db_nums):
+            self.set_setting(db_name, db_num)
+            # Set some unrelevant key so it is reserved for this app
+            # even if no key is put
+            assert (
+                os.system(f'redis-cli -n "{db_num}" SET __YNH_PROVISIONNED true > /dev/null') == 0
```
fflorent: Probably more interesting for debugging to set the value to the app name.

```suggestion
                app_name = self.app.replace("-", "_").replace(".", "_")
                os.system(f'redis-cli -n "{db_num}" SET __YNH_PROVISIONNED_BY {app_name} > /dev/null') == 0
```

src/utils/resources.py
```diff
@@ -1608,6 +1609,142 @@ def deprovision(self, context: Dict = {}):
...
+                raw_msg=True
+            )
+        max_dbs = int(max_dbs_match)
+        free_dbs = set(range(0, max_dbs)) - set(databases_num)
+        if len(free_dbs) < count:
+            raise YunohostError(
+                f'Uhoh, you seem to have reached the maximum of Redis databases allowed. You may increase the value of "database" in {redis_conf_path} to increase it.',
+                raw_msg=True
+            )
+        return list(free_dbs)[:count]
```
fflorent, line 1703: A big issue I have just discovered when working on a package: not only we have to look for a free database on installation / upgrade, but also on restoration.

I have then 2 questions in mind:
 - are these AppResource's function invoked when restoring an app? (after a quick look, I would say no)
 - how to for the app regenerate its configurations that references the redis database? (for example like [grist_ynh does here](https://github.com/YunoHost-Apps/grist_ynh/blob/f8302e3a7594de75932fefe26aca9816f8bf0117/scripts/restore#L42-L44))

Not very sure about what's the best idea. 
My guess is that we *might* introduce a linter for the restore script that:
 -  would ensure that we lookup for free database again and store these values in the settings manually (or make the AppResourceManager and the AppResource classes do that?) ;
 - would check the conf template files who read the redis databases variables and check whether they are regenerated;

Help appreciated :pray:

src/utils/resources.py
```diff
@@ -1608,6 +1609,142 @@ def deprovision(self, context: Dict = {}):
...
+        new_db_nums = self._free_db(len(dbs_to_provision))
+        for (db_name, db_num) in zip(dbs_to_provision, new_db_nums):
+            self.set_setting(db_name, db_num)
+            # Set some unrelevant key so it is reserved for this app
+            # even if no key is put
+            assert (
+                os.system(f'redis-cli -n "{db_num}" SET __YNH_PROVISIONNED {self.app} > /dev/null') == 0
+            ), "Uhoh, cannot provision a Redis DB"
+
+        self.set_setting('provisionned_redis_db_names', ",".join(db_names))
```
alexAubin, line 1741: So if I understand correctly, that trick with `provisionned_redis_db_names` is to be able to keep track of what redis DB were provisioned in the past such that we can free the ones that are not in the manifest anymore ?

The way I dealt with this kind of stuff in other resource, eg `ports`, is to assume that every instance was either named `port` (the main one) or `port_{suffix}` and...

aaaaaaaand actually i'm looking at the current code and nope ... during `provision_or_update` i'm not trying to check if there's an old `port_<somthing>` value that would no exist anymore in the manifest ... merp

But that could work in theory and in this case we could assume redis DB are in setting either `redis_db` (the main one) or `redis_db_suffix` ?  ¯\\\_(ツ)\_/¯

fflorent: > So if I understand correctly, that trick with provisionned_redis_db_names is to be able to keep track of what redis DB were provisioned in the past such that we can free the ones that are not in the manifest anymore ?

Exactly! That's not something I am very fond of, that's the reason why I proposed an alternative implementation in #2124 where I compare the resource of the previous manifest and the new one to know what has changed (and maybe that's debatable).

> But that could work in theory and in this case we could assume redis DB are in setting either redis_db (the main one) or redis_db_suffix ? ¯\_(ツ)_/¯

You mean enforcing redis db to start with `redis_db_*`? We may do that (especially if we can use `previous_names`)
But I feel like there could exist edge cases causing confusion because a packager has created a setting named `redis_db_*`.
But at least the name would follow a nomenclature and it would be symmetric to what have been made for ports.

src/utils/resources.py
```diff
@@ -1608,6 +1609,142 @@ def deprovision(self, context: Dict = {}):
...
+    ### Example
+
+    ```toml
+    [resources.redis]
+    # This should be fine empty for most cases. If you want to customize the env variable or to provision other redis databases,
+
+    [resources.redis.redis_db] # Will provision a database available as $redis_db (or __REDIS_DB__ in conf files)
+
+    [resources.redis.celery_redis_db] # Will provision a database available as $celery_redis_db (or __CELERY_REDIS_DB__ in conf files)
+    previous_names = [ "celery_db" ] # During an upgrade, if `$celery_db` exists, it will be renamed to `$celery_redis_db`.
```
fflorent, line 1630: Maybe do something a bit more symmetrical to the ports resource? (Aleks' idea)

```toml
[resource.redis]
main.setting_name = "redis_db"
celery.setting_name = "celery_redis_db"
celery.previous_names = ["celery_db"]
```

fflorent: Suggestion: 
```toml
[resources.redis]
[resources.redis.main] # Setting name would be redis_db. Required only if there exist more than one redis db.
[resources.redis.celery] # setting name would be celery_redis_db
previous_names = "celery_db"
```

src/utils/resources.py
```diff
@@ -1608,6 +1609,142 @@ def deprovision(self, context: Dict = {}):
...
+        free_dbs = set(range(0, max_dbs)) - set(databases_num)
+        if len(free_dbs) < count:
+            raise YunohostError(
+                f'Uhoh, you seem to have reached the maximum of Redis databases allowed. You may increase the value of "database" in {redis_conf_path} to increase it.',
+                raw_msg=True
+            )
+        return list(free_dbs)[:count]
+
+    def _deprovision_db(self, db_name):
+        os.system(f'redis-cli -n "{self.get_setting(db_name)}" flushdb >/dev/null 2>/dev/null')
```
fflorent, line 1706: Rather use similar logic as for the ports:
https://github.com/YunoHost/yunohost/blob/7f763be3138f482a9512ad2e4d334c79b0a44812/src/utils/resources.py#L1398-L1404

Josue-T:
Seem that redis DB ID is not always a good idea as explained here: https://stackoverflow.com/questions/10644158/multiple-redis-instances#10644197

Also some package like seafile just don't support this settings cf https://github.com/haiwen/seafile-server/pull/805

So maybe we might need to rethink about how to handle this.

fflorent:
After discussing during the PR Lanta:
1. Indeed Seafile is not willing to configure the DB id
2. https://stackoverflow.com/questions/10644158/multiple-redis-instances#10644197 -> suggests that a better practice to spawn as many redis-server instances as there exists apps
3. We can create systemd service using the at symbol: https://superuser.com/questions/393423/what-is-the-significance-of-the-symbol-in-systemctl-scripts-and-specifically
4. We can use `include` in redis, so a per-app config file will just add few settings (the port and the dumpfile location), and include the standard config file
  Side note: we should take care of restoring the config file by overwriting it thanks to the settings stored
5. It would be better to optionally provide a user and a password with the resource, or better: use sockets when possible with no username or password (example: `unix:///var/run/redis/APP.pid?db=0`)
6. Needs to backup the dump file (under `/var/lib/redis`).
7. Also great news: we won't have to deal with db ids, the package or the app can choose their db id!

About the migration:
1. Remove the use of the redis helpers
2. Ensure the app use the redis port in their URL (or the socket)
3. Raise a warning in the CI when the package use the old helper

PR #2124, opened 2025-06-15
Redis app resource (implementation alternative)

fflorent:
## The problem

See #2122 

This PR introduces an alternate implementation (see the latest commit).

## Solution

The idea is to let the AppResource store the properties of the previous manifest (the `self.old` property). Therefore they could handle the difference, in our case by deprovisionning the redis databases that are removed in the new version.

You can check the difference with #2122 here:
https://github.com/fflorent/yunohost/compare/redis-app-resource...fflorent:yunohost:redis-app-resource-alt

## PR Status

Need help [on these questions](https://github.com/YunoHost/yunohost/pull/2122/files#r2160415723)

## How to test

```
./ynh-dev test yunohost:app_resources:resource_redis
```

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -26,7 +26,7 @@ import shutil
 import subprocess
 import tempfile
 from logging import getLogger
-from typing import Any, Callable, Dict, List, Union
+from typing import Any, Callable, Dict, List, Union, Self
 
 from moulinette import m18n
 from moulinette.utils.filesystem import chmod, chown, mkdir, rm, write_to_file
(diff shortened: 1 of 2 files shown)


src/utils/resources.py
```diff
@@ -1608,6 +1611,150 @@ def deprovision(self, context: Dict = {}):
...
+                raw_msg=True
+            )
+        max_dbs = int(max_dbs_match)
+        free_dbs = set(range(0, max_dbs)) - set(databases_num)
+        if len(free_dbs) < count:
+            raise YunohostError(
+                f'Uhoh, you seem to have reached the maximum of Redis databases allowed. You may increase the value of "database" in {redis_conf_path} to increase it.',
+                raw_msg=True
+            )
+        return list(free_dbs)[:count]
```
fflorent, line 1708: [Same questions as here](https://github.com/YunoHost/yunohost/pull/2122/files#r2160415723)

Josue-T:
Seem that redis DB ID is not always a good idea as explained here: https://stackoverflow.com/questions/10644158/multiple-redis-instances#10644197

Also some package like seafile just don't support this settings cf https://github.com/haiwen/seafile-server/pull/805

So maybe we might need to rethink about how to handle this.

fflorent:
See this comment: https://github.com/YunoHost/yunohost/pull/2122#issuecomment-4502279075

PR #2127, opened 2025-06-26
document supported CPU archs

oleole39:


Salamandar:
Actually we support every architecture that might be printed by `dpkg --print-architecture`.

You can see the list by running: `dpkg-architecture --list-known` (yeah it's a bit stupidly long).

oleole39:
> The schema lists the archs we might have tested for,

Hmm I guess the doc should stick to what the schema allows - should the schema be updated?

Salamandar:
The schema can be updated to allow more archs. The schema is not used by the code, only by stuff like IDE linters and the CI :)

PR #2144, opened 2025-08-04
fix/tools_update: return upgradable apps, not all apps

kay0u:
## The problem

`apps` contains all the apps installed, not the upgradable ones

## Solution

use `upgradable_apps ` instead

## PR Status

...

## How to test

...

### DIFF
--- a/src/tools.py
+++ b/src/tools.py
@@ -466,7 +466,7 @@ def tools_update(
 
     return {
         "system": upgradable_system_packages_per_categories,
-        "apps": apps,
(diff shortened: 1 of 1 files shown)


alexAubin:
Eeeh but that's actually a feature and not a bug (i think ?) because the point was to also list apps that are in special cases such as `url_required`, possibly because they were in the catalog in the past but are not anymore (otherwise you may never notice it)

Can you elaborate what made you notice this and want to patch it ?

PR #2147, opened 2025-08-06
helpers2.1: skip apt update when cache was already updated less than 30 minutes ago

alexAubin:
## The problem

It's annoying to wait like half a minute everytime you run app installs on low connectivity because of apt update

## Solution

Skip apt update when cache is not older than 30 minutes

## PR Status

Tested

## How to test

Run something like idk, my_webapp install two times in a row

### DIFF
--- a/helpers/helpers.v2.1.d/apt
+++ b/helpers/helpers.v2.1.d/apt
@@ -351,7 +351,7 @@ _ynh_apt() {
     # Optimization when just calling apt update : check if the cache was
     # already refreshed in the last 30 min, which should be enough and prevent
     # unecessary traffic and annoying wait time during app dependency installs etc
-    if [[ "$@" == "update" ]]
+    if [[ "$*" == "update" ]]
     then
         # trick from https://stackoverflow.com/a/205710
         local aptcache="/var/cache/apt/pkgcache.bin"


helpers/helpers.v2.1.d/apt
```diff
@@ -347,6 +347,24 @@ _ynh_apt_package_version() {
...
+    # Optimization when just calling apt update : check if the cache was
+    # already refreshed in the last 30 min, which should be enough and prevent
+    # unecessary traffic and annoying wait time during app dependency installs etc
+    if [[ "$@" == "update" ]]
+    then
+        # trick from https://stackoverflow.com/a/205710
+        local aptcache="/var/cache/apt/pkgcache.bin"
+        touch -d '-30 minutes' /tmp/.30_minutes_ago
+        sleep 1
+        if [[ -e $aptcache ]] && [[ $aptcache -nt /tmp/.30_minutes_ago ]] && [[ -z "$(find /etc/apt/sources.list* -type f -newer $aptcache)" ]]; then
```
fflorent: Wouldn't it work?

```
if [[ -z "$(find /var/cache/apt/pkgcache.bin -mmin -30 2>/dev/null)" ]] && [[ -z "$(find /etc/apt/sources.list* -type f -mmin -30 ]];
```

Source: https://unix.stackexchange.com/a/271058

helpers/helpers.v2.1.d/apt
```diff
@@ -347,6 +347,21 @@ _ynh_apt_package_version() {
 
 # APT wrapper for non-interactive operation
 _ynh_apt() {
+
+    # Optimization when just calling apt update : check if the cache was
+    # already refreshed in the last 30 min, which should be enough and prevent
+    # unecessary traffic and annoying wait time during app dependency installs etc
+    if [[ "$@" == "update" ]]
```
alexAubin:
```suggestion
    if [[ "$*" == "update" ]]
```

PR #2150, opened 2025-08-11
Add jinja filter to load and dump json, yaml and toml

Josue-T:
## The problem

Iterating over complex data in template is a pain, because ne need delimiter and it's not strong solution

By example, this part of code could be simplified with jinja filters: https://github.com/YunoHost-Apps/forgejo_ynh/blob/75750de715f852989780ddb95f86d43410cc8253/scripts/login_source.sql#L34-L37

## Solution

Use json, toml or yaml serialization and decode in template with filters

## PR Status

Tested locally and it seem working

## How to test

Just made a file `test.j2` with: 
```
from json to yaml
{{ hellovar | from_json | to_yaml }}

from json to json

{{ hellovar | from_json | to_json }}

from json to toml

{{ hellovar | from_json | to_toml }}

from yaml to json

{{ hellovaryaml | from_yaml | to_json }}

from toml to json

{{ hellovartoml | from_toml | to_json }}
```
And then run `j2 test.j2 --filters /usr/lib/python3/dist-packages/yunohost/utils/jinja_filters.py`

### DIFF
--- a/helpers/helpers.v2.1.d/templating
+++ b/helpers/helpers.v2.1.d/templating
@@ -50,6 +50,25 @@
 # For a full documentation of jinja's syntax you can refer to [the official Jinja documentation](https://jinja.palletsprojects.com/en/3.1.x/templates/).
 #
 # Note that in YunoHost context, all variables are from shell variables and therefore are strings
+# To help handling complex data Jinja engine are executed with theses additional filters:
+# - from_json: load a string as Json and return an object
+# - from_yaml: load a string as Yaml and return an object
+# - from_toml: load a string as Toml and return an object
+# - to_json: serialize to string an object to Json
+# - to_yaml: serialize to string an object to Yaml
+# - to_toml: serialize to string an object to Toml
+#
+# So by example, if you want to convert a json string `$my_json` to Toml, you can to this way:
+# {{ my_json | from_json | to_toml }}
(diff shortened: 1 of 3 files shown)


helpers/helpers.v1.d/templating
```diff
@@ -61,6 +61,25 @@
...
+# To help handling complex data Jinja engine are executed with theses additional filters:
+# - from_json: load a string as Json and return an object
+# - from_yaml: load a string as Yaml and return an object
+# - from_toml: load a string as Toml and return an object
+# - to_json: serialize to string an object to Json
+# - to_yaml: serialize to string an object to Yaml
+# - to_toml: serialize to string an object to Toml
+#
+# So by example, if you want to convert a json string `$my_json` to Toml, you can to this way:
+# {{ $my_json | from_json | to_toml }}
```
alexAubin:
```suggestion
# {{ my_json | from_json | to_toml }}
```

guessing there's no $ here ?

Josue-T: Ah yes 👍

Salamandar (commented):
Although this looks quite OK in general, the hard-coded path `/usr/lib/python3/dist-packages/yunohost/utils/` doesn't feel quite right to me. We might try to find another way to pass the "installation path" of yunohost to helpers.

PR #2152, opened 2025-08-12
Upgrade helper version for hook in version 2.1

Josue-T:
## The problem

We still are using helper v2.0 for hooks

## Solution

Use helpers v2.1.

Note I specified the helper version in each file to avoid to upgrade the hook version in case some apps still use helper v2.0 for hooks.

## PR Status

Can be tested, seem working

## How to test

Run the regen-conf, backup, restore


helpers/helpers.v2.1.d/backup
```diff
@@ -88,7 +95,12 @@ ynh_backup() {
     fi
 
     # Add the relative current working directory to the destination path
-    local rel_dir="${YNH_CWD#"$YNH_BACKUP_DIR"}"
+    local rel_dir
+    if [ "${YNH_CORE_SCRIPT:-}" != 1 ]; then
+        rel_dir="${YNH_CWD#"$YNH_BACKUP_DIR"}"
+    else
+        rel_dir="${YNH_CORE_BACKUP_DIR#"$YNH_BACKUP_DIR"}"
```
zamentur: Use YNH_CWD instead of YNH_CORE_BACKUP_DIR like in https://doc.yunohost.org/admin/backups/include_exclude_files

debian/postinst
```diff
@@ -32,6 +32,9 @@ do_configure() {
         if [ -d /etc/yunohost/apps/ ] && ls /etc/yunohost/apps/* >/dev/null 2>&1; then
             echo "Sounds like /etc/yunohost/installed mysteriously disappeared ... You should probably contact the Yunohost support ..."
         else
+            export YNH_CONTEXT=regenconf
+            export YNH_HELPERS_VERSION=2
+            export YNH_CORE_SCRIPT=1
```
zamentur: Remove YNH_CORE_SCRIPT and use another context var

Remove YNH_HELPERS_VERSION=2

Salamandar:
This will break restoration of old backups.

Josue-T:
should be mostly good except one thing, how to handle the backup dir because currently we use YNH_CWD to define the destination and as the core backup is called all in one by as a hook we can't set YNH_CWD for each hook.

Josue-T:
> This will break restoration of old backups.

Well, normally no because the change try to stay retro compatible. And the original path is saved in the CSV, so at the end in most of case we don't really take care of where we backup. But fee free to bring some improvement in this PR if you have a better idea of how to handle this.

helpers/helpers.v2.1.d/backup
```diff
@@ -45,7 +45,14 @@ ynh_backup() {
...
 
     # If the path starts with /var/log/$app or $data_dir
-    if [[ (-n "${app:-}" && "$target" =~ ^/var/log/$app/?\*?$) || (-n "${data_dir:-}" && "$target" =~ ^$data_dir)     ]]; then
+    if [[ \
+        # Apps backup
+        (-n "${app:-}" && "$target" =~ ^/var/log/$app/?\*?$) \
+        || (-n "${data_dir:-}" && "$target" =~ ^$data_dir) \
+        # for core backups
+        || (-z "${app:-}" && "$target" == '/home/'*) \
+        || (-z "${app:-}" && "$target" == '/var/mail') \
```
alexAubin, line 54: It's unclear why this was changed, this may change some existing behavior such as people calling something along the lines of:

```shell
BACKUP_CORE_ONLY=1 yunohost backup create
```

hoping to backup "everything" (apps and core) but without the app's core data ... with this change it would also exclude the /home/ and /var/mail dirs

(dunno how much of a big deal this is...)

alexAubin: Update: supposedly that's because the core called `ynh_backup foo 1` with the "1" corresponding to the legacy "is_big" uuueargha

hooks/backup/18-data_multimedia
```diff
@@ -22,15 +22,13 @@
...
-# Backup destination
-backup_dir="${1}/data/multimedia"
-
 if [ ! -e "/home/yunohost.multimedia" ] || [ -e "/home/yunohost.multimedia/.nobackup" ]; then
     exit 0
 fi
 
 # Backup multimedia directory
-ynh_backup --src_path="/home/yunohost.multimedia" --dest_path="${backup_dir}" --is_big --not_mandatory
+ynh_backup "/home/yunohost.multimedia"
```
alexAubin:
```suggestion
[ ! -e "/home/yunohost.multimedia" ] || ynh_backup "/home/yunohost.multimedia"
```

alexAubin (approved):
(LGTM modulo how happy the tests are)

Josue-T:
@alexAubin do you think we could merge it maybe in Trixie ?

alexAubin (approved):
I'm gonna say it looks okay to me to merge in current `dev` but we may get some surprises even though we pretty much carefully reviewed it several times

Any opinion @zamentur @Salamandar ?

zamentur:
About `set -eu` i found this commit https://github.com/YunoHost/yunohost/commit/1dece38a36bd58b3260902c6737498306bcdb219 . If i understand well `set` policy is now managed by 2.1 helpers, so may be we should discuss together to search if some things can break due to the new policy

zamentur:
Unit tests failed due to missing helpers.

I remove this from YunoHost 13 project cause it's not a requirements for trixie, in more i am not sure it's reasonable to change regen-conf on this big release.

PR #2156, opened 2025-08-18
Add tool to remove old, unused kernels

gabjy:
## The problem

We need a tool for the user to remove old kernels

## Solution

Adding the kernel removal process to the basic cleanup tool

## PR Status

Couldn't test with LXC; works fine on host system

## How to test

Running the added code in a Python console


src/tools.py
```diff
@@ -713,6 +713,14 @@ def tools_basic_space_cleanup() -> None:
...
     subprocess.run("rm /var/log/*.?", shell=True)
     subprocess.run("rm /var/log/*/*.?", shell=True)
+    # Removing kernel except last 2
+    output = subprocess.run("dpkg -l | grep linux-image | awk '{print$2}'", capture_output=True, text=True, shell=True)
+    kernels = output.stdout.split("\n")
+    kernels = kernels[:-4]
+    for kernel in kernels:
+        subprocess.run("apt remove -y --purge " + kernel, shell=True)
+
+    subprocess.run("update-grub2", shell=True)
```
tituspijean: I do not have that command on my server, so I do not think this is ubiquitous.

Josue-T: This could really depends if it's bar metal or lxd/lxc or an arm board.

alexAubin: Supposedly i don't think it's even needed ? As far as I understand some proper calls are inside the .deb of the kernel to auto-regen grub and whatever is needed when installing/uninstalling kernels ? (I'm only 90% sure but never had to advise people to run an explicit grub update after helping them to uninstall kernel on the forum)

tituspijean: @Josue-T ah right, I'm in LXC (but then the command would fail, which is not nice UX).

gabjy: Thanks for the inputs. I committed some changes based on them

src/tools.py
```diff
@@ -713,6 +713,22 @@ def tools_basic_space_cleanup() -> None:
     subprocess.run("rm /var/log/*/*.gz", shell=True)
     subprocess.run("rm /var/log/*.?", shell=True)
     subprocess.run("rm /var/log/*/*.?", shell=True)
+    # Removing kernel except last 2
+    uname_output = subprocess.run("uname -r", capture_output=True, text=True, shell=True)
+    running_kernel = uname_output.stdout.split("\n")[:-1]
+    dpkg_output = subprocess.run("dpkg -l | grep linux-image | awk '{print$2}'", capture_output=True, text=True, shell=True)
```
alexAubin:
```suggestion
def tools_clean_old_kernels() -> None:
    # Removing kernel except last 2
    uname_output = subprocess.run("uname -r", capture_output=True, text=True, shell=True)
    running_kernel = uname_output.stdout.split("\n")[:-1]
    dpkg_output = subprocess.run("dpkg -l | grep linux-image | awk '{print$2}'", capture_output=True, text=True, shell=True)
```

In terms of design : I think this should be a different command line action than `basic-space-cleanup` which should remain "basic", and is typically needed when `/` ran out of space. Cleaning up the old kernels is little bit more touchy, and typically is needed when `/boot/` is full.

If you agree with this, I propose this suggestion (dunno how familiar you are with github/gitlab suggestions but they can be commited from Github/Gitlab web interface, though you can also copypasta or manually apply the change yourself locally) to have it in a `tools_clean_old_kernels()` function. If you accept, it should also be added to the `actionmap.yml` in `share/` with a block like `clean-old-kernels:` in the `tools` section.

src/tools.py
```diff
@@ -713,6 +713,22 @@ def tools_basic_space_cleanup() -> None:
...
+    if dpkg_output.stdout == '':
+        print("Couldn't find installed kernels. Are you running Yunohost in LXC/LXD ?")
```
alexAubin:
```suggestion
    if not dpkg_output.stdout.strip():
        logger.info("Couldn't find installed kernels. Maybe you are running Yunohost in an LXC container ?")
        return
```

Typically we want to use `logger.info()` / `.warning()` etc. for such message :+1:. Also we can directly `return` to avoid needing an `else:` indent level

src/tools.py
alexAubin:
```suggestion
    dpkg_output = subprocess.run("dpkg -l | grep '^ii\s*linux-image' | awk '{print $2}'", capture_output=True, text=True, shell=True)
```

Syntax nitpicking, I'm more comfortable with a space between `print` and `$2` even though it works ;P

I also tweaked the grep pattern to be more accurate because `dpkg --list` actually list some package it knows about, but which are not actually installed ... the one actually installed are the lines starting with `ii`

src/tools.py
```diff
@@ -713,6 +713,22 @@ def tools_basic_space_cleanup() -> None:
...
+    else:
+        kernels = dpkg_output.stdout.split("\n")[:-4]
```
alexAubin: Hmmmm I'm not sure to understand the `[:-4]` here :thinking:  :sweat_smile: 

Also typically we probably want to use `.strip()` here before `.split()`, because we will typically get an output like : 

```python
"foo\nbar\n"
```

(note the trailing `\n`) Which will `.split("\n")` into : 

```python
["foo", "bar", ""]
```

(note the last element being empty string

src/tools.py
```diff
@@ -713,6 +713,22 @@ def tools_basic_space_cleanup() -> None:
...
+        if kernels == []:
+            print("No kernel to remove")
```
alexAubin:
```suggestion
        if kernels == []:
            logger.info("No kernel to remove")
            return
```

Similar stuff here : we can use a `return` statement to not have to use an `else` statement and save an indentation level

src/tools.py
```diff
@@ -713,6 +713,22 @@ def tools_basic_space_cleanup() -> None:
     subprocess.run("rm /var/log/*/*.gz", shell=True)
     subprocess.run("rm /var/log/*.?", shell=True)
     subprocess.run("rm /var/log/*/*.?", shell=True)
+    # Removing kernel except last 2
+    uname_output = subprocess.run("uname -r", capture_output=True, text=True, shell=True)
+    running_kernel = uname_output.stdout.split("\n")[:-1]
```
alexAubin:
```suggestion
    # running_kernel is expected to be something like : 6.1.0-28-amd64
    running_kernel = uname_output.stdout.strip()
```

Similar as another comment below : let's use `.strip()` to get rid of the trailing `\n`. Also `uname -r` is expected to be a single line so hmmm no need to split it ?

src/tools.py
```diff
@@ -713,6 +713,22 @@ def tools_basic_space_cleanup() -> None:
...
+    else:
+        kernels = dpkg_output.stdout.split("\n")[:-4]
+        if kernels == []:
+            print("No kernel to remove")
+        else:    
+            for kernel in kernels:
+                if running_kernel[0] not in kernel:
+                    subprocess.run("apt remove -y --purge " + kernel, shell=True)
+                else:
+                    print("This is the running kernel, it won't be removed")
```
alexAubin: Imho we probably want to have a single call to apt using something like below. (I hope list comprehension is not too magic but that's a nice context to learn about it ;P)

Note that I'm not comparing `running_kernel` to `kernel` because `running_kernel` is only the **version** (as returne by `uname -r`) and `kernel` is the apt/dpkg package, something like `linux-image-1.2.3-4-amd64`

To run a single apt command, i'm using `' '.join(kernels_to_remove)` inside an f-string (also not sure if you're familiar with those, I'm a big f-an of them but maybe that's too much magic :sweat_smile: )

```suggestion
            kernels_to_remove = [kernel for kernel if running_kernel not in kernel]
            cmd = f"apt remove -y --purge {' '.join(kernels_to_remove)}"
            subprocess.run(cmd, shell=True)
```

To further improve upon this I also suggest : 
- i would ask the admin for confirmation before actually running the apt, because removing kernels feel a bit touchy ... Something along the lines of "The system is currently running kernel "1.2.3.4". This will remove the following kernels : linux-image-0.1.2, linux-image-0.1.3... Proceed ?" You can use `_ask_confirmation(...)` for this for which you'll find example usage in the same `tools.py`
- we should check the `returncode` of the subprocess because apt may fail for some reason, in which case we probably want to have some `logger.error()` message

But this can be iterated upon in a second phase, I hope this ain't too overwheling :sweat_smile:

alexAubin (commented):
General principle looks fine to me but I have a bunch of comments :sweat_smile:  I hope it' not too overwhelming, I tried to provide full explanation for the changes I'm proposing :sweat_smile:

src/tools.py
gabjy: I remember we talked about it yep. In the end I chose to put it in the already existing command to avoid creating yet another one but I agree that's a little involving to remove old kernels

src/tools.py
gabjy: I take note of the `logger.` for next time :)

src/tools.py
gabjy: Noted for the `strip()` function

Regarding the `[:-4]` it's because there are :  a trailling empty string (which should be removed with the split function), a "linux-image-amd64" (which I guess is the the name/type of the kernel), plus we should remove the last 2 kernels from the list of kernels we want to remove from the system, just in case.
Note that keeping the last 2 only is an arbitrary choice, it could be the last 3, last 4, etc

src/tools.py
gabjy: Oh I get what you did here ! Guess I'm not used to Python trick and oneliner yet haha. I'll look into it

src/tools.py
```diff
@@ -713,6 +713,20 @@ def tools_basic_space_cleanup() -> None:
     subprocess.run("rm /var/log/*/*.gz", shell=True)
     subprocess.run("rm /var/log/*.?", shell=True)
     subprocess.run("rm /var/log/*/*.?", shell=True)
+def tools_clean_old_kernels() -> None:
+    # Removing kernel except last 2
+    uname_output = subprocess.run("uname -r", capture_output=True, text=True, shell=True)
+    # running_kernel is expected to be something like : 6.1.0-28-amd64
+    running_kernel = uname_output.stdout.strip()
+    dpkg_output = subprocess.run("dpkg -l | grep '^ii\s*linux-image' | awk '{print $2}'", capture_output=True, text=True, shell=True)
```
alexAubin:
```suggestion
    dpkg_output = subprocess.run("dpkg -l | grep '^ii\s*linux-image-[0-9]' | awk '{print $2}'", capture_output=True, text=True, shell=True)
```

Let's further refine the grep regex to only catch the "actual version" and not the `linux-image-amd64` (on other arch there may be other stuff like i have `linux-image-rpi-<stuff>` on my RPi

src/tools.py
```diff
@@ -713,6 +713,20 @@ def tools_basic_space_cleanup() -> None:
...
+    kernels = dpkg_output.stdout.strip().split("\n")[:-3]
+    if not kernels:
+        logger.info("No kernel to remove")
+        return
+    else:    
+        kernels_to_remove = [kernel for kernel in kernels if running_kernel not in kernel]
+        cmd = f"apt remove -y --purge {' '.join(kernels_to_remove)}"
+        subprocess.run(cmd, shell=True)
```
alexAubin:
```suggestion
    kernels = dpkg_output.stdout.strip().split("\n")[:-3]
    kernels_to_remove = [kernel for kernel in kernels if running_kernel not in kernel]
    if not kernels_to_remove:
        logger.info("No kernel to remove")
        return
    else:
        cmd = f"apt remove -y --purge {' '.join(kernels_to_remove)}"
        subprocess.run(cmd, shell=True)
```

Let's test on `if not kernels_to_remove`, otherwise using `if kernels`, kernels will typically never be empty because it contains the current kernel

src/tools.py
gabjy: I added the error message and the confirmation as suggested.
Note that I used directly the strings and not the translation framework you're using. Should I create an entry in the locals files or will it be managed automatically by Weblate ?

src/tools.py
```diff
@@ -713,6 +713,24 @@ def tools_basic_space_cleanup() -> None:
...
     subprocess.run("rm /var/log/*.?", shell=True)
     subprocess.run("rm /var/log/*/*.?", shell=True)
+def tools_clean_old_kernels() -> None:
+    # Removing kernel except last one
+    uname_output = subprocess.run("uname -r", capture_output=True, text=True, shell=True)
+    # running_kernel is expected to be something like : 6.1.0-28-amd64
+    running_kernel = uname_output.stdout.strip()
+    dpkg_output = subprocess.run("dpkg -l | grep '^ii\s*linux-image-[0-9]' | awk '{print $2}'", capture_output=True, text=True, shell=True)
+    kernels = dpkg_output.stdout.strip().split("\n")[:-1]
+    kernels_to_remove = [kernel for kernel in kernels if running_kernel not in kernel]
```
alexAubin:
```suggestion
    installed_kernels = dpkg_output.stdout.strip().split("\n")
    # Consider for removal all the installed kernels except the running one
    # + the most recent one (which may be more recent than the running one if system didnt reboot since last kernel upgrade)
    kernels_to_remove = [kernel for kernel in sorted(installed_kernels)[:-1] if running_kernel not in kernel]
```

src/tools.py
```diff
@@ -713,6 +713,24 @@ def tools_basic_space_cleanup() -> None:
...
+    if not kernels_to_remove:
+        logger.info("No kernel to remove")
+        return
+    else:
+        from .utils.app_utils import _ask_confirmation
+        _ask_confirmation(f"The system is currently running kernel {running_kernel}. This will remove the following kernels : {', '.join(kernels_to_remove)}. Proceed ?")
+        cmd = f"apt remove -y --purge {' '.join(kernels_to_remove)}"
+        kernel_removal = subprocess.run(cmd, shell=True)
```
alexAubin: In fact we can avoid having to use `shell=True` because the command is somewhat simple in this case (and `shell=True` is less safe, though in this case we are not using use inputs or whatever)

```suggestion
        cmd = ["apt", "remove", "-y", "--purge"] + kernels_to_remove
        kernel_removal = subprocess.run(cmd)
```

alexAubin (approved):

alexAubin (approved):
LGTM after the last remaining comments

src/tools.py
```diff
@@ -713,6 +713,26 @@ def tools_basic_space_cleanup() -> None:
...
+    installed_kernels = dpkg_output.stdout.strip().split("\n")
+    # Consider for removal all the installed kernels except the running one
+    # + the most recent one (which may be more recent than the running one if system didnt reboot since last kernel upgrade)
+    kernels_to_remove = [kernel for kernel in sorted(installed_kernels)[:-1] if running_kernel not in kernel]
+    if not kernels_to_remove:
+        logger.info("No kernel to remove")
+        return
+    else:
+        from .utils.app_utils import _ask_confirmation
+        _ask_confirmation(f"The system is currently running kernel {running_kernel}. This will remove the following kernels : {', '.join(kernels_to_remove)}. Proceed ?")
```
alexAubin, line 745: cf discussion irl : `_ask_confirmation` in fact expects an i18n key. We can also provide `params={ ... }` with a dict of values for the i18n string

So let's have something like : 

```suggestion
        _ask_confirmation(f"tools_clean_old_kernels_confirm", params={"running_kernel": running_kernel, "kernels_to_remove"=', '.join(kernels_to_remove)})
```

and then you need to edit `locales/en.json` to add a new entry `"tools_clean_old_kernels_confirm"` with value :

```
"The system is currently running kernel {running_kernel}. This will remove the following kernels : {kernels_to_remove}. Proceed ?"
```

src/tools.py
```diff
@@ -713,6 +713,26 @@ def tools_basic_space_cleanup() -> None:
...
+        cmd = ["apt", "remove", "-y", "--purge"] + kernels_to_remove
+        kernel_removal = subprocess.run(cmd)
+        if kernel_removal.returncode != 0:
+            logger.error(f"There was an error during the kernels removal. The return code of the 'apt remove' command is : {kernel_removal.returncode}")
```
alexAubin, line 749: Similar i18n consideration here though the syntax is a bit different (oneday™ we will make everything consistent)

```suggestion
            logger.error(m18n.n(""tools_clean_old_kernels_error", returncode=kernel_removal.returncode))
```

and then gotta add another key to en.json `tools_clean_old_kernels_error` with something like

```
"There was an error during the kernels removal. The return code of the 'apt remove' command is : {returncode}"
```

alexAubin:
Ah :sweat_smile: 

Hmmm not sure how it works exactly, on my end it I have version 22, 23, and 28. System is running version 28, but `apt autoremove` only wants to remove version 22 and keep version 23 ? I guess it wants to keep one previous kernel ?

Still I'm thinking in some case, the `/boot/` may be so small it can only contain maybe 2 or 3 kernel, which and you may run out of space just because `apt` wants to install a more recent kernel

For scenario for servers with a lot of uptime, you may not even be running the latest version already installed because that would involve a reboot, so you may have (version N-1), (version N = the running one), (version N+1 = need to reboot to actually run it) AND apt wants to install "version N+2", so that's 4 versions

src/tools.py
```diff
@@ -727,6 +727,27 @@ def tools_basic_space_cleanup() -> None:
         shell=True,
     )
 
+def tools_clean_old_kernels() -> None:
+    # Removing kernel except last one
+    uname_output = subprocess.run("uname -r", capture_output=True, text=True, shell=True)
+    # running_kernel is expected to be something like : 6.1.0-28-amd64
+    running_kernel = uname_output.stdout.strip()
+    dpkg_output = subprocess.run("dpkg -l | grep '^ii\s*linux-image-[0-9]' | awk '{print $2}'", capture_output=True, text=True, shell=True)
```
alexAubin:
```suggestion
    dpkg_output = subprocess.run("dpkg -l | grep '^ii\\s*linux-image-[0-9]' | awk '{print $2}'", capture_output=True, text=True, shell=True)
```

Linter is complaining about `Invalid escape sequence \s`

(Déso j'ai recréer un commentaire parce que j'utilise ca comme demo pour ma formation git ;P)

gabjy:
Hi, I wanted to add the key for the string inside the _ask_confirmation call inside the en.json file but it can't seem to find it. Should I run a specific command for it to be taken into account ?

alexAubin:
Eeeeh nope supposedly it should use the entry from `locales/en.json`, no need to regen anything :thinking:

PR #2161, opened 2025-08-25
Update registrars list

Thovi98:
## The problem

Lexicon version is way to old

## Solution

Adapt the registrar_list for the latest version of lexicon (based on https://dns-lexicon.github.io/dns-lexicon/configuration_reference.html)

## PR Status

Requires https://github.com/YunoHost/lexicon/pull/3

## How to test

...

### DIFF
--- a/share/registrar_list.toml
+++ b/share/registrar_list.toml
@@ -7,6 +7,11 @@
     type = "string"
     redact = true
 
+[arvancloud]
+    [arvancloud.auth_token]
+    type = "string"
+    redact = true
+
 [aurora]
     [aurora.auth_api_key]
     type = "string"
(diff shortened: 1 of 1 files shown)


share/registrar_list.toml
```diff
@@ -626,24 +760,51 @@
...
+    type = "string"
+    redact = true
     
 [vultr]
     [vultr.auth_token]
     type = "string"
     redact = true
 
-[webgo]
+[wedos]
```
orhtej2, line 779: fairly sure you have to change the `webgo` references below, otherwise the TOML is likely invalid

Thovi98: Good catch, fixed, thanks!

alexAubin:
Cheers ! Tonight we updated the lexicon version shipped via our repo, so update to lexicon 3.21.1 is already available

PR #2167, opened 2025-08-31
fix: portal domain conf not properly updated with the portal_allow_mail_edit settings

tituspijean:
## The problem

Email settings seem not to be applied.

## Solution

...

## PR Status

Please check, my brain is sunday-fried.

## How to test

...

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -2128,14 +2128,17 @@ def app_ssowatconf() -> None:
     }
     for domain, apps in portal_domains_apps.items():
         portal_settings = {}
-        portal_settings.update(portal_email_settings)
 
+        # If possible, load the existing file
         portal_settings_path = Path(PORTAL_SETTINGS_DIR) / f"{domain}.json"
         if portal_settings_path.exists():
             this_domain_portal_settings: dict[str, Any] = read_json(
                 str(portal_settings_path)
             )  # type: ignore[assignment]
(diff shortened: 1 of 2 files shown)


tituspijean:
It looks like the portal edit page gets updated only if I run `yunohost app ssowatconf` one more time manually. 🤔

src/app.py
```diff
@@ -2128,15 +2128,18 @@ def app_ssowatconf() -> None:
...
+        # If possible, load the existing file
         portal_settings_path = Path(PORTAL_SETTINGS_DIR) / f"{domain}.json"
         if portal_settings_path.exists():
             this_domain_portal_settings: dict[str, Any] = read_json(
                 str(portal_settings_path)
             )  # type: ignore[assignment]
             portal_settings.update(this_domain_portal_settings)
 
+        # Update with the new settings
+        portal_settings.update(portal_email_settings)
```
alexAubin, line 2141: Eeeh not sure to understand why adding these before or after the other domain settings would matter ?

alexAubin: (Cherry-picked the other commit to dev in the meantime)

tituspijean: The initial code would prepare the new `portal_settings` (line 2131), then override it with old values on line 2138. We want it the other way around.

PR #2182, opened 2025-09-17
Core: Use helpers v2.1

Salamandar:


Josue-T:
Seem a bit a duplicate of https://github.com/YunoHost/yunohost/pull/2152

Anyway feel free to work on it as I've a lot of other stuff to handle...

Salamandar:
Yes I saw afterwards that you already did the work :(

Josue-T:
Note I just need to push something and then it should be mostly good except one thing, how to handle the backup dir because currently we use YNH_CWD to define the destination and as the core backup is called all in one by as a hook we can't set YNH_CWD for each hook.

Josue-T:
> I just need to push something

Done by c427bdc8f572001e5dcb18bf49955afe6b7d2a39

zamentur:
Seems this PR is less advanced than the other so i close. I make that often (starting to work from start on a PR that already exists... quite frustrating :/ )

PR #2184, opened 2025-09-20
Make app restore more easy to debug with no-remove-on-failure option

Josue-T:
## The problem

We have a possibility to avoid to remove the app if the app install fail. It's really useful for debuging. For app restore we don't have it and it could be not easy to understand why the restore fail when the app are removed.

## Solution

Add the same feature than for app install.

## PR Status

Ready but not tested at all.

## How to test

Try app restore of app which fail on restore.

### DIFF
--- a/src/backup.py
+++ b/src/backup.py
@@ -865,7 +865,7 @@ class RestoreManager:
         return restore_manager.result
     """
 
-    def __init__(self, name, method="tar"):
+    def __init__(self, name, method="tar", no_remove_on_failure=False):
         """
         RestoreManager constructor
 
@@ -892,6 +892,7 @@ class RestoreManager:
         self.name = name
         self.method = BackupMethod.create(method, self)
         self.targets = BackupRestoreTargetsManager()
+        self.no_remove_on_failure = no_remove_on_failure
 
     #
     # Misc helpers                                                          #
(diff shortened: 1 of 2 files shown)


zamentur (commented):
LGTM
I think we can merge this after a small test

src/backup.py
```diff
@@ -1469,7 +1470,8 @@ def copytree(src, dst, symlinks=False, ignore=None):
                 # Cleaning temporary scripts directory
                 shutil.rmtree(tmp_workdir_for_app, ignore_errors=True)
 
-                app_remove(app_instance_name, force_workdir=app_workdir)
+                if not self.no_remove_on_failure:
+                    app_remove(app_instance_name, force_workdir=app_workdir)
```
alexAubin, line 1474:
```suggestion
                    app_remove(app_instance_name, force_workdir=app_workdir)
                else:
                    logger.error(f"The restore of {app_instance_name} failed, but was not cleaned up as requested by --no-remove-on-failure.")
```

Josue-T:
Tested by with an app which fail and success on restore and with and without option `--no-remove-on-failure`. Sound good.

Salamandar (approved):

Salamandar:
Can you just squash your two commits ? Then let's merge this.

PR #2186, opened 2025-09-22
[fix] Boolean are transform into string by ynh_write_var_in_file

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/2671

## Solution

- Avoid to cast boolean and integer as string in js/json/php/py file.
- Deduce the format of the boolean from the previous value.
- Document that this helpers should not be use for keys which store values of different types.
- Update Unit test to check the type in the file and not only the value

## PR Status

Tested manually

## How to test

...


zamentur:
I think the failed test on test_process is not related to this PR https://gitlab.com/YunoHost/yunohost/-/jobs/11450278752

alexAubin:
Superseded by https://github.com/YunoHost/yunohost/pull/2209

PR #2190, opened 2025-09-30
certificate: Allow IPv6-only domains to run ACME

Salamandar:
## The problem

The test is buggy and doesn't allow ipv4=MISSING, ipv6=OK.

### DIFF
--- a/src/certificate.py
+++ b/src/certificate.py
@@ -846,10 +846,8 @@ def _check_domain_is_ready_for_ACME(domain):
     # - 'MISSING' for IPv6 ain't critical for ACME
     # - IPv4 can be None assuming there's at least an IPv6, and viveversa
     #    - (the case where both are None is checked before)
-    if not (
-        A_record_status in [None, "OK"]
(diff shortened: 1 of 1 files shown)


OniriCorpe (approved):

src/certificate.py
```diff
@@ -846,10 +846,8 @@ def _check_domain_is_ready_for_ACME(domain):
     # - 'MISSING' for IPv6 ain't critical for ACME
     # - IPv4 can be None assuming there's at least an IPv6, and viveversa
     #    - (the case where both are None is checked before)
-    if not (
-        A_record_status in [None, "OK"]
-        and AAAA_record_status in [None, "OK", "MISSING"]
-    ):
+    statuses = [A_record_status, AAAA_record_status]
+    if "WRONG" in statuses or "OK" not in statuses:
```
alexAubin, line 850: eeeh ok but what about the `None` case though

Salamandar: What about it ?

PR #2191, opened 2025-10-03
[HELPER] Set ynh default php version to default debian php version

ewilly:
## The problem

current default php version for ynh is 7.4 and does not rely to default debian php version

## Solution

retrieve the debian default php version and use it
(not sure that the curl command is the best way but it's a working idea)

## PR Status

...

## How to test

I use it here : https://github.com/YunoHost-Apps/zwiicms_ynh/blob/1c96e467db03d66f1c17b90483fe5ae5ee2c4cdf/manifest.toml#L75

### DIFF
--- a/helpers/helpers.v2.1.d/php
+++ b/helpers/helpers.v2.1.d/php
@@ -19,7 +19,11 @@
 #
 
 # (this is used in the apt helpers, big meh ...)
-readonly YNH_DEFAULT_PHP_VERSION=7.4
+DEBIAN_DEFAULT_PHP_VERSION=$(curl -fsSL "https://packages.debian.org/$YNH_DEBIAN_VERSION/php" \
+    | grep -Eo "/$YNH_DEBIAN_VERSION/php[0-9].[0-9]" \
+    | tail -n1 | sed "s@^/$YNH_DEBIAN_VERSION/php@@")
+    
+readonly YNH_DEFAULT_PHP_VERSION=$( [[ ! -z ${DEBIAN_DEFAULT_PHP_VERSION} ]] && echo ${DEBIAN_DEFAULT_PHP_VERSION} || echo "8.2" )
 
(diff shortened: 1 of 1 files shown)


ewilly:
`DEBIAN_DEFAULT_PHP_VERSION=$( apt-cache depends php | grep -Eo "php[0-9].[0-9]" | sed "s@^php@@" )` is perhaps a proper way ?

helpers/helpers.v2.1.d/php
```diff
@@ -19,7 +19,9 @@
 #
 
 # (this is used in the apt helpers, big meh ...)
-readonly YNH_DEFAULT_PHP_VERSION=8.2
+DEBIAN_DEFAULT_PHP_VERSION=$(apt-cache depends php | grep -Eo "php[0-9].[0-9]" | sed "s@^php@@")
```
alexAubin, line 22: Uuuuuuh wokay but `apt-cache` is not free (it takes 1.5s to run this command on my desktop computer so I'd expect even more on a low-end ARM board) yet sourcing the helpers happens quite frequently and it's just way faster to hardcode the value considering it only changes between major debian versions ? x_x

ewilly: hum right, that something I didn't consider...
The main idea was to avoid something like :
``
if [[ $YNH_DEBIAN_VERSION == "bookworm" ]]; then
    readonly YNH_DEFAULT_PHP_VERSION=8.2
elif [[ $YNH_DEBIAN_VERSION == "trixie" ]]; then
    readonly YNH_DEFAULT_PHP_VERSION=8.4
fi
``

Salamandar: Considering the few places this variable is used in, we could replace this variable with a function, to be called only when necessary.

zamentur:
Dear contributors,

PR #2192, opened 2025-10-03
[HELPER] Fix swap helpers

ewilly:
## The problem
Related to https://github.com/YunoHost/issues/issues/2670

## Solution

* ynh_add_swap helper: check if main partition is a btrfs file system before applying chattr
* ynh_del_swap : check that the file is mounted as swap before trying swappoff

## PR Status

...

## How to test

...


helpers/helpers.v2.1.d/0-utils
```diff
@@ -523,7 +523,9 @@ ynh_add_swap() {
         truncate -s 0 "/swap_$app"
 
         # try to set the No_COW attribute on the swapfile with chattr (depending of the filesystem type)
-        chattr +C "/swap_$app" 2> /dev/null
+        if grep -qs ' / .*btrfs' /proc/mounts; then
```
alexAubin: Eeeh shouldn't it be the *negation* of this grep ? i.e; we want to run the chattr only if the filesystem is *not* btrfs?

helpers/helpers.v2.1.d/0-utils
```diff
@@ -523,7 +523,9 @@ ynh_add_swap() {
...
+			chattr +C "/swap_$app"
+		fi
```
alexAubin:
```suggestion
        local root_fs_type=$(cat /proc/mounts | awk '{ if ($2 == '/') print $3 }')
        if [[ "$root_fs_type" == "btrfs" ]]; then
            echo "Skipping chattr +C for swap because the root filesystem is btrfs  which doesn't support this attribute"
        else
			chattr +C "/swap_$app"
		fi
```

helpers/helpers.v2.1.d/0-utils
OniriCorpe: no, the chattr command must only run on btrfs fs

helpers/helpers.v2.1.d/0-utils
ewilly: * I correct that one by inverting content of then and else as chattr should be applied only on btrfs fs.
* To keep the same writing logic bettween ynh_add_swap and ynh_del_swap if use a variable for awk: `local root_fs_type=$(cat /proc/mounts | awk '{ if ($2 == var) print $3 }' var="/")`

kay0u:
I'm not sure it's available on debian, but we can use [this](https://wiki.archlinux.org/title/Btrfs#Swap_file) for swapfile on btrfs:

```
btrfs subvolume create /ynh_swap/
btrfs filesystem mkswapfile --size 4g --uuid clear /ynh_swap/swap_$app
```

Because from what I saw, if there is a swapfile on a subvolume (/ for example), the snapshot fails

helpers/helpers.v2.1.d/0-utils
```diff
@@ -523,7 +523,12 @@ ynh_add_swap() {
         truncate -s 0 "/swap_$app"
 
         # try to set the No_COW attribute on the swapfile with chattr (depending of the filesystem type)
-        chattr +C "/swap_$app" 2> /dev/null
+        local root_fs_type=$(cat /proc/mounts | awk '{ if ($2 == var) print $3 }' var="/")
+        if [[ "$root_fs_type" == "btrfs" ]]; then
+			chattr +C "/swap_$app"
+        else
+            echo "Skipping chattr +C for swap because the root filesystem is btrfs  which doesn't support this attribute"
```
zamentur, line 530:
```suggestion
            echo "Skipping chattr +C for swap because the root filesystem is '$root_fs_type' which doesn't support this "no copy-on-write" attribute"
```

helpers/helpers.v2.1.d/0-utils
```diff
@@ -523,7 +523,12 @@ ynh_add_swap() {
         truncate -s 0 "/swap_$app"
 
         # try to set the No_COW attribute on the swapfile with chattr (depending of the filesystem type)
-        chattr +C "/swap_$app" 2> /dev/null
+        local root_fs_type=$(cat /proc/mounts | awk '{ if ($2 == var) print $3 }' var="/")
+        if [[ "$root_fs_type" == "btrfs" ]]; then
```
zamentur, line 527: I suggest to add f2fs https://elixir.bootlin.com/linux/v6.18.6/A/ident/FS_NOCOW_FL
```suggestion
        if [[ "$root_fs_type" == "btrfs" || "$root_fs_type" == "f2fs" ]]; then
```

zamentur:
> I'm not sure it's available on debian, but we can use [this](https://wiki.archlinux.org/title/Btrfs#Swap_file) for swapfile on btrfs:
> 
> ```
> btrfs subvolume create /ynh_swap/
> btrfs filesystem mkswapfile --size 4g --uuid clear /ynh_swap/swap_$app
> ```
> 
> Because from what I saw, if there is a swapfile on a subvolume (/ for example), the snapshot fails

In my ynh-dev container (using btrfs) it's available

@kay0u could you say if the snapshot fail with chattr +C or without ?

helpers/helpers.v2.1.d/0-utils
```diff
@@ -523,7 +523,12 @@ ynh_add_swap() {
         truncate -s 0 "/swap_$app"
 
         # try to set the No_COW attribute on the swapfile with chattr (depending of the filesystem type)
-        chattr +C "/swap_$app" 2> /dev/null
+        local root_fs_type=$(cat /proc/mounts | awk '{ if ($2 == var) print $3 }' var="/")
+        if [[ "$root_fs_type" == "btrfs" || "$root_fs_type" == "f2fs" ]]; then
+			chattr +C "/swap_$app"
+        else
+            echo "Skipping chattr +C for swap because the root filesystem is '$root_fs_type' which doesn't support this "no copy-on-write" attribute"
```
Salamandar, line 562: I'd remove this echo and put a debug print on the `if` part instead.

Salamandar (commented):
Could you please fix indentation with spaces instead of tabs please ? :)

ewilly:
> Could you please fix indentation with spaces instead of tabs please ? :)

done

helpers/helpers.v2.1.d/0-utils
ewilly: Not sure to understand the point and what you suggest

zamentur:
Dear contributors,

PR #2193, opened 2025-10-03
[enh] Include spam and trash in quota

zamentur:
## The problem
https://forum.yunohost.org/t/quota-configuration-of-dovecot/40228

## Solution
I just remove the ignore rules

## PR Status

Completely untested, i just open that in order to open the discussion

## How to test

...

### DIFF
--- a/conf/dovecot/dovecot.conf
+++ b/conf/dovecot/dovecot.conf
@@ -143,8 +143,6 @@ plugin {
 
 plugin {
   quota = maildir:User quota
-  quota_rule2 = SPAM:ignore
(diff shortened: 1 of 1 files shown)


alexAubin (approved):

Josue-T (approved):

zamentur:
I finally made test on that. On my side i have no SPAM dir, instead it was Junk dir. SO the ignore was not so useful anyway.

I had to run `doveadm -f flow quota recalc -u USER` for testing. FOr me this change is ok

PR #2208, opened 2025-10-21
Helper doc:  ynh_replace does not accept wildcards in its `file` parameter

oleole39:
This works: `sed 's/old/new/g' files-*.php`

This doesn't work: `ynh_replace --match=old --replace=new --files=files-*.php`, because of the double quotes around `$file` and bash expansion rules or so: https://github.com/YunoHost/yunohost/blob/b91ab52deb47b942572790cc4b4b36b4b15ec365/helpers/helpers.v2.1.d/string#L67

This PR mentions this limitation and suggests a workaround.


oleole39:
An untested alternative could be to add the workaround directly in the helper: 
```bash
for f in "$file"
do
    sed --in-place "s${delimit}${match}${delimit}${replace}${delimit}g" $f
done 
```

With that I would expect `ynh_replace --match=old --replace=new --files=files-*.php` to work with the wildcards. But I am not sure to what extent it would be better than removing the double quotes (which might be there for a good reason)

zamentur:
Dear contributors,

PR #2209, opened 2025-10-22
Fix ynh write var in file boolean2

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/2671

## Solution

 * Avoid to cast boolean and integer as string in js/json/php/py file.
 * Deduce the format of the boolean from the previous value.
 * Document that this helpers should not be use for keys which store values of different types.
 * Update Unit test to check the type in the file and not only the value

## PR Status

Ready

## How to test

...

### DIFF
--- a/helpers/helpers.v2.1.d/templating
+++ b/helpers/helpers.v2.1.d/templating
@@ -296,6 +296,11 @@ ynh_read_var_in_file() {
 # | arg: --key=     - the key to set
 # | arg: --value=   - the value to set
 # | arg: --after=   - the line just before the key (in case of multiple lines with the name of the key in the file)
+#
+# This helpers replaces several var affectation use case in several languages
+# We don't use jq or equivalent to keep comments and blank space in files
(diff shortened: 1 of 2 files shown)


helpers/helpers.v2.1.d/templating
```diff
@@ -373,9 +379,42 @@ ynh_write_var_in_file() {
         value="$(echo "$value" | sed "s/'/\\\\\\\\'/g")"
         sed -ri "${range}s$delimiter(^${var_part}')([^']|\\')*('"'[\s,;]*)(\s*'"$comments"'.*)?$'$delimiter'\1'"${value}'${endline}${delimiter}i" "$file"
     else
-        if [[ "$value" == *"'"* ]] || [[ "$value" == *'"'* ]] || [[ "$ext" =~ ^php|py|json|js$ ]]; then
+        formated="false"
```
tituspijean, line 382: FYI it's `formatted`

PR #2210, opened 2025-10-25
Pydantic firewall

Salamandar:
## The problem

TypedDict


Salamandar:
Might fix https://github.com/YunoHost/issues/issues/2696

PR #2211, opened 2025-10-26
Replace calls to m18n.n with calls to i18n._ as a first step to gettext

Salamandar:
## The problem

...

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/maintenance/missing_i18n_keys.py
+++ b/maintenance/missing_i18n_keys.py
@@ -39,7 +39,7 @@ REFERENCE_FILE = LOCALE_FOLDER + "en.json"
 
 def find_expected_string_keys():
     # Try to find :
-    #    m18n.n(   "foo"
+    #    _(   "foo"
     #    YunohostError("foo"
     #    YunohostValidationError("foo"
     #    # i18n: foo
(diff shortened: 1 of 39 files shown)


maintenance/missing_i18n_keys.py
```diff
@@ -18,111 +18,114 @@
...
-    #    m18n.n(   "foo"
+    #    _(   "foo"
     #    YunohostError("foo"
     #    YunohostValidationError("foo"
     #    # i18n: foo
-    regex_m18n = re.compile(r"m18n\.n\(\n*\s*[\"\'](\w+)[\"\']")
-    regex_ynherr = re.compile(r"YunohostError\(\n*\s*[\'\"](\w+)[\'\"]")
-    regex_ynhxerr = re.compile(
-        r"Yunohost(?:Validation|Authentication)Error\(\n*\s*[\'\"](\w+)[\'\"]"
+    p1 = re.compile(r"m18n\.n\(\n*\s*[\"\'](\w+)[\"\']")
```
zamentur, line 46:
```suggestion
    p1 = re.compile(r"_\(\n*\s*[\"\'](\w+)[\"\']")
```
But it probably doesn't work cause some function could end with `_`

PR #2220, opened 2025-11-01
Fix variable expansion in dovecot config

orhtej2:
## The problem

yunohost/issues#2692

## Solution

Despite [what plugins conf man page claims](https://doc.dovecot.org/2.3/configuration_manual/plugins/mail_event_logging/) [the variable expansion syntax is now ](https://doc.dovecot.org/2.4.2/core/settings/syntax.html#variable-expansion) `$SET:key_name`
 
## PR Status

Untested

## How to test

...

### DIFF
--- a/conf/dovecot/dovecot.conf
+++ b/conf/dovecot/dovecot.conf
@@ -18,7 +18,7 @@ mail_gid = 8
 
 protocols = imap sieve
 
-mail_plugins = $mail_plugins quota notify push_notification
+mail_plugins = $SET:mail_plugins quota notify push_notification
 
 ###############################################################################
 # generated 2025-10-31, Mozilla Guideline v5.7, Dovecot 2.4.1, OpenSSL 3.5.1, intermediate config


olberger:
I'm not sure this is valid. After setting this in `/etc/dovecot/dovecot.conf`:
```
mail_plugins = $SET:mail_plugins quota notify push_notification
```
I `systemctl restart dovecot`, and get:
```
doveconf: Fatal: Error in configuration file /etc/dovecot/dovecot.conf line 21: Failed to expand $SET:mail_plugins: Setting type can't be expanded to string
```

olberger:
> I'm not sure this is valid. After setting this in `/etc/dovecot/dovecot.conf`:
> 
> ```
> mail_plugins = $SET:mail_plugins quota notify push_notification
> ```
> 
> I `systemctl restart dovecot`, and get:
> 
> ```
> doveconf: Fatal: Error in configuration file /etc/dovecot/dovecot.conf line 21: Failed to expand $SET:mail_plugins: Setting type can't be expanded to string
> ```

I guess the $SET syntax only applies to dovecot 3 and not dovecot 2.4, currently in trixie ?

Salamandar:
I can see the config doc here: https://doc.dovecot.org/main/core/settings/types.html#boolean-list

Salamandar:
Superseeded by 0b7d69d8667bddfb9f7a1af949d76060bab4685e

PR #2225, opened 2025-11-16
Make app list sorted case-insensitive

orhtej2:
## The problem

...

## Solution

...

## PR Status

...

## How to test

...


alexAubin (approved):

orhtej2:
Actually it gets resorted UI-side

PR #2235, opened 2025-11-21
Fix regex to redact passwords in the logs

hidrarga:
## The problem

Some passwords are not redacted in the logs. For instance passwords containing whitespaces or passwords in bash conditions because backslashes are added before each character.

## Solution

I'm improving the regex that searches for passwords to redact to include whitespaces. Then I replace both the password and the same string with backslashes.

## PR Status

I could test this on my server, but I'm not sure why the regex was limited to non-whitespace characters in the first place.

### DIFF
--- a/src/log.py
+++ b/src/log.py
@@ -573,19 +573,22 @@ class RedactingFormatter(Formatter):
         # Wrapping this in a try/except because we don't want this to
         # break everything in case it fails miserably for some reason :s
         try:
-            # This matches stuff like db_pwd=the_secret or admin_password=other_secret
-            # (the secret part being at least 3 chars to avoid catching some lines like just "db_pwd=")
+            # This matches stuff like db_pwd=the_secret, admin_password=other_secret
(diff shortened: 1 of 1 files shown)


src/log.py
```diff
@@ -573,19 +573,22 @@ def identify_data_to_redact(self, record):
...
                 match
-                and match.group(2) not in self.data_to_redact
+                and match.group(3) not in self.data_to_redact
                 and match.group(1) not in ["key", "manifest_key"]
             ):
-                self.data_to_redact.append(match.group(2))
+                self.data_to_redact.append(match.group(3))
+                # In conditions, the string might be split with backslashes, so we need
+                # to replace these as well.
+                self.data_to_redact.append("".join("\\" + c for c in match.group(3)))
```
alexAubin, line 591: Soooo I partially merged only this bit via 06f5d9195 ... i'm closing the remaining of the PR because the whole regex thingy is to be reworked in #2236

PR #2237, opened 2025-11-24
trixie: add psql and venv migrations

kay0u:
## The problem

https://github.com/YunoHost-Apps/immich_ynh/issues/322

## Solution

...

## PR Status

Not tested, I accidentally dropped all my psql db lul

## TODO Postgresql
- [x] Add rebuild index
- [ ] Test migration

## TODO Python
- [ ] Use `python -m venv --upgrade VENV_DIR` instead of rebuilding the venv
- [ ] Test on not yet supported python apps

## How to test

...

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -679,6 +679,7 @@
     "migration_not_enough_space": "Make sufficient space available in {path} to run the migration.",
     "migration_postgresql_previous_not_installed": "PostgreSQL was not installed on your system. Nothing to do.",
     "migration_postgresql_target_not_installed": "PostgreSQL 13 is installed, but not PostgreSQL 15!? Something weird might have happened on your system :(…",
+    "migration_postgresql_reindexing_databases": "Reindexing databases and refreshing collations after migrations, please do not interrupt.",
     "migration_python_venv_rebuild_broken_app": "Skipping {app} because virtualenv can't easily be rebuilt for this app. Instead, you should fix the situation by forcing the upgrade of this app using `yunohost app upgrade --force {app}`.",
     "migration_python_venv_rebuild_disclaimer_base": "Following the upgrade to Debian Bookworm, some Python applications needs to be partially rebuilt to get converted to the new Python version shipped in Debian (in technical terms: what's called the 'virtualenv' needs to be recreated). In the meantime, those Python applications may not work. YunoHost can attempt to rebuild the virtualenv for some of those, as detailed below. For other apps, or if the rebuild attempt fails, you will need to manually force an upgrade for those apps.",
     "migration_python_venv_rebuild_disclaimer_ignored": "Virtualenvs can't be rebuilt automatically for those apps. You need to force an upgrade for those, which can be done from the command line with: `yunohost app upgrade --force APP`: {ignored_apps}",
--- a/locales/en.json
+++ b/locales/en.json
@@ -716,6 +716,8 @@
     "migration_description_0035_fix_apps_nodejs_version": "Fix nodejs versions in app systemd configurations",
     "migration_description_0036_migrate_to_trixie": "Upgrade the system to Debian Trixie and YunoHost 13",
     "migration_description_0037_upgrade_dkim_keys": "Upgrade DKIM keys to improve mail deliverability",
+    "migration_description_0037_postgresql_15_to_17": "Migrate databases from PostgreSQL 15 to 17",
+    "migration_description_0038_rebuild_python_venv_in_trixie": "Repair Python app after trixie migration",
     "migration_ldap_backup_before_migration": "Creating a backup of LDAP database and apps settings prior to the actual migration.",
     "migration_ldap_can_not_backup_before_migration": "The backup of the system could not be completed before the migration failed. Error: {error}",
     "migration_ldap_migration_failed_trying_to_rollback": "Could not migrate… trying to roll back the system.",
--- a/locales/en.json
+++ b/locales/en.json
@@ -708,6 +708,8 @@
     "migration_description_0034_fix_missing_admins_aliases": "Fix missing mail aliases for the admins group",
     "migration_description_0035_fix_apps_nodejs_version": "Fix nodejs versions in app systemd configurations",
     "migration_description_0036_migrate_to_trixie": "Upgrade the system to Debian Trixie and YunoHost 13",
+    "migration_description_0037_postgresql_15_to_17": "Migrate databases from PostgreSQL 15 to 17",
+    "migration_description_0038_rebuild_python_venv_in_trixie": "Repair Python app after trixie migration",
     "migration_ldap_backup_before_migration": "Creating a backup of LDAP database and apps settings prior to the actual migration.",
     "migration_ldap_can_not_backup_before_migration": "The backup of the system could not be completed before the migration failed. Error: {error}",
     "migration_ldap_migration_failed_trying_to_rollback": "Could not migrate… trying to roll back the system.",
--- a/src/migrations/m0037_postgresql_15_to_17.py
+++ b/src/migrations/m0037_postgresql_15_to_17.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2026 YunoHost Contributors
+#
+# This file is part of YunoHost (see https://yunohost.org)
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+from .postgresql import PostgreSQLMigration
+
+
+class MyMigration(PostgreSQLMigration):
+    "Migrate DBs from Postgresql 15 to 17 after migrating to Trixie"
+
+    previous_version = "15"
+    target_version = "17"
+
+    dependencies = ["migrate_to_trixie"]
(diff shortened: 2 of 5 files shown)


src/migrations/0038_rebuild_python_venv_in_trixie.py
```diff
@@ -0,0 +1,207 @@
...
+        "microblogpub",  # uses poetry ? x_x
+        "mopidy",  # applies a custom patch?
+        "motioneye",  # install stuff from a .tar.gz
+        "pgadmin",  # bunch of manual patches
+        "searxng",  # uses --system-site-packages ?
+        "synapse",  # specific stuff for ARM to prevent local compiling etc
+        "matrix-synapse",  # synapse is actually installed in /opt/yunohost/matrix-synapse because ... yeah ...
+        "tracim",  # pip install -e .
+        "weblate",  # weblate settings are .. inside the venv T_T
+    ]
```
alexAubin, line 56: Zmlerf theoretically we should re-review all the app catalog by grepping apps which seem to use a virtualenv and check which apps are exception that can't be properly rebuilt because they use non-standard stuff

kay0u: After a quick review of all apps with a venv: 
- [searxngs_ynh](https://github.com/YunoHost-Apps/searxng_ynh/blob/master/scripts/_common.sh#L31) seems ok now
- [paperless-ngx_ynh](https://github.com/YunoHost-Apps/paperless-ngx_ynh/blob/master/scripts/install#L42) I'm not sure, could work
- [mopidy](https://github.com/YunoHost-Apps/mopidy_ynh/blob/master/scripts/_common.sh#L13) seems ok
- matrix-synapse: I can't find this app
---
- [diacamma_ynh](https://github.com/YunoHost-Apps/diacamma_ynh/blob/master/scripts/_common.sh#L42) Does an ugly sed
- [django-for-runners_ynh](https://github.com/YunoHost-Apps/django-for-runners_ynh/blob/master/scripts/_common.sh#L94) pip-sync is used, I'm not sure if it's a problem
- [django-fritzconnection_ynh](https://github.com/YunoHost-Apps/django-fritzconnection_ynh/blob/master/scripts/_common.sh#L94) same, pip-sync
- [funkwhale_ynh](https://github.com/YunoHost-Apps/funkwhale_ynh/blob/master/scripts/upgrade#L64) install from a folder ./api?
- [homeassistant_ynh](https://github.com/YunoHost-Apps/homeassistant_ynh/blob/master/scripts/_common.sh#L47) uv is used
- [immich_ynh](https://github.com/YunoHost-Apps/immich_ynh/blob/master/scripts/_common.sh#L198) uv is used
- [indico_ynh](https://github.com/YunoHost-Apps/indico_ynh/blob/master/scripts/install#L76) symlink between venv and static web pages
- [kresus](https://github.com/YunoHost-Apps/kresus_ynh/blob/master/scripts/install#L46) uses virtualenv
- [motioneye](https://github.com/YunoHost-Apps/motioneye_ynh/blob/master/scripts/install#L46) install from .tar.gz
- [lasuite-docs_ynh](https://github.com/YunoHost-Apps/lasuite-docs_ynh/blob/main/scripts/install#L94) moving stuff into the venv
- [librephotos_ynh](https://github.com/YunoHost-Apps/librephotos_ynh/blob/master/scripts/_common.sh#L101) setup.py install
- [mautrix_telegram_ynh](https://github.com/YunoHost-Apps/mautrix_telegram_ynh/blob/master/scripts/install#L113) install from .tar.gz
- [microblogpub_ynh](https://github.com/YunoHost-Apps/microblogpub_ynh/blob/master/scripts/install#L47) uses poetry
- [pgadmin](https://github.com/YunoHost-Apps/pgadmin_ynh/blob/master/scripts/_common.sh#L56) bunch of manual patches
- [pretalx_ynh](https://github.com/YunoHost-Apps/pretalx_ynh/blob/master/scripts/install#L73) ynh_replace into the venv
- [synapse_ynh](https://github.com/YunoHost-Apps/synapse_ynh/blob/master/scripts/_common.sh#L34) ynh_setup_source into the venv
- [syncserver-rs_ynh](https://github.com/YunoHost-Apps/syncserver-rs_ynh/blob/master/scripts/install#L80) uses poetry
- [tracim](https://github.com/YunoHost-Apps/tracim_ynh/blob/master/scripts/install#L87) pip install .
- [weblate_ynh](https://github.com/YunoHost-Apps/weblate_ynh/blob/master/scripts/install#L118) settings are .. inside the venv T_T
- [yunohost_appgenerator_ynh](https://github.com/YunoHost-Apps/yunohost_appgenerator_ynh/blob/master/scripts/install#L27) uses pdm

src/migrations/0037_postgresql_15_to_17.py
```diff
@@ -0,0 +1,115 @@
...
+            return
+
+        if not self.package_is_installed(f"postgresql-{PREVIOUS_VERSION}"):
+            logger.warning(m18n.n(f"migration_{MIGRATION_NUMBER:04d}_postgresql_{PREVIOUS_VERSION}_not_installed"))
+            return
+
+        if not self.package_is_installed(f"postgresql-{TARGET_VERSION}"):
+            raise YunohostValidationError(f"migration_{MIGRATION_NUMBER:04d}_postgresql_{TARGET_VERSION}_not_installed")
+
+        # Make sure there's a 15 cluster
```
kay0u, line 60:
```suggestion
        # Make sure there's a "PREVIOUS_VERSION" cluster
```

src/migrations/0037_postgresql_15_to_17.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2024 YunoHost Contributors
```
Thovi98, line 3:
```suggestion
# Copyright (c) 2025 YunoHost Contributors
```
I don’t know maybe those dates are automatically updated on push?

Josue-T:
Probably related to https://github.com/YunoHost/issues/issues/2737

ewilly:
@kay0u some comments about the postgresql migration:
* I had to modify the file `/etc/postgresql-common/user_clusters` from `* * 15 main postgres` to `* * 17 main postgres`
* I had to fix the collation mismatch with that bash command:
```
databases=$(LC_ALL=C sudo --login --user=postgres PGUSER=postgres PGPASSWORD="$(cat /etc/yunohost/psql)" psql --tuples-only --no-align --dbname="postgres" --command="SELECT datname FROM pg_database WHERE datistemplate = false OR datname = 'template1';")
for db in $databases
do
  LC_ALL=C sudo --login --user=postgres PGUSER=postgres PGPASSWORD="$(cat /etc/yunohost/psql)" psql --dbname="$db" --command="REINDEX DATABASE $db;"
  LC_ALL=C sudo --login --user=postgres PGUSER=postgres PGPASSWORD="$(cat /etc/yunohost/psql)" psql --dbname="$db" --command="ALTER DATABASE $db REFRESH COLLATION VERSION;"
done
```
sorry I'm not able to translate that in python language

Salamandar:
I guess let's first split the actual code from the bookworm migrations and reuse it in the trixie migrations?

Salamandar:
See #2274

kay0u:
> @kay0u some comments about the postgresql migration:
> 
>     * I had to modify the file `/etc/postgresql-common/user_clusters` from `* * 15 main postgres` to `* * 17 main postgres`
> 
>     * I had to fix the collation mismatch with that bash command:
> 
> 
> ```
> databases=$(LC_ALL=C sudo --login --user=postgres PGUSER=postgres PGPASSWORD="$(cat /etc/yunohost/psql)" psql --tuples-only --no-align --dbname="postgres" --command="SELECT datname FROM pg_database WHERE datistemplate = false OR datname = 'template1';")
> for db in $databases
> do
>   LC_ALL=C sudo --login --user=postgres PGUSER=postgres PGPASSWORD="$(cat /etc/yunohost/psql)" psql --dbname="$db" --command="REINDEX DATABASE $db;"
>   LC_ALL=C sudo --login --user=postgres PGUSER=postgres PGPASSWORD="$(cat /etc/yunohost/psql)" psql --dbname="$db" --command="ALTER DATABASE $db REFRESH COLLATION VERSION;"
> done
> ```
> 
> sorry I'm not able to translate that in python language

@Salamandar do you think we should add this to the migration script?

What about the `ignored_python_apps`?

Salamandar:
> @Salamandar do you think we should add this to the migration script?

I guess so although i don't understand why the /etc/postgresql-common/user_clusters file had to be edited.

> What about the ignored_python_apps?

hmmmm idk. Also I would like to re-try just upgrading the venv without recreating it completely.

ewilly:
> I guess so although i don't understand why the /etc/postgresql-common/user_clusters file had to be edited.

@Salamandar : humm you are right, that's due to immich package. On bookworm I use postgreql17 coming from postgresql repo. To ensure that the default cluster stick to debian one (15 in bookwrom) I fill that file (see https://github.com/YunoHost-Apps/immich_ynh/blob/0b045730eb1e999cff149d7c2089d073ee217476/scripts/_common.sh#L461). With trixie we could go back to because debian provide postgreql17 as default. But for now I am not sure what's the best way how to proceed.

Salamandar:
> > I guess so although i don't understand why the /etc/postgresql-common/user_clusters file had to be edited.
> 
> @Salamandar : humm you are right, that's due to immich package. On bookworm I use postgreql17 coming from postgresql repo. To ensure that the default cluster stick to debian one (15 in bookwrom) I fill that file (see [YunoHost-Apps/immich_ynh@`0b04573`/scripts/_common.sh#L461](https://github.com/YunoHost-Apps/immich_ynh/blob/0b045730eb1e999cff149d7c2089d073ee217476/scripts/_common.sh#L461)). With trixie we could go back to because debian provide postgreql17 as default. But for now I am not sure what's the best way how to proceed.

I think the best is to just remove the line `* * 15 main postgres`. Other lines shall stay I think.

src/migrations/0038_rebuild_python_venv_in_trixie.py
tituspijean: As of 2026-08-05, the only change to this list is `syncserver-rs_ynh` which does not use `poetry` anymore, though it uses `uv` but the resulting `venv` seems cleanly used.

tituspijean (approved):
Merging now, when linter is happy, test to be done in the next beta release.

PR #2241, opened 2025-12-07
Ensure file endline

Salamandar:
We should prefer files ending with an empty line: most editors do it automatically, thus adding a useless diff when manually editing / rolling back manual modifications.

### DIFF
--- a/src/utils/file_utils.py
+++ b/src/utils/file_utils.py
@@ -138,6 +138,16 @@ def _ensure_write_file_valid(file_path: str | Path) -> Path:
     return file_path
 
 
+def ensure_file_endline_terminated(file: str | Path) -> None:
+    """Ensure the file is terminated by an empty line"""
+    with Path(file).open("rb+") as stream:
+        stream.seek(-1, os.SEEK_END)
+        last_char = stream.read(1)
+        if last_char != b"\n":
+            stream.seek(0, os.SEEK_END)
+            stream.write(b"\n")
+
+
 def write_to_file(
     file_path: str | Path, data: str | bytes | list, file_mode: str = "w"
 ) -> None:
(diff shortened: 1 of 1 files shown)


alexAubin:
>The problem
>
>...

... can we have an explanation of what problem this does address x_x ...

Salamandar:
> ... can we have an explanation of what problem this does address x_x ...

Yeah, I usually put this in commit messages, but i'll edit the PR descr.

zamentur:
Could we have in the description of the PR info on which kind of tests has been made on this code ? (it's ok (for me) if there are no test and you write "untested", just doing it, help us during PR review to understand the risk to merge or not this kind of code.

`write_to_file` function is used everywhere in ynh, for a lot of files, and on my side i am not totally confident that all software strip the endline ...

For example 
with `PIDs_to_keep=[158,162]`
`write_to_file(MOULINETTE_LOCK, "\n".join(PIDs_to_keep))`

and next 
```
PIDs = read_file(MOULINETTE_LOCK).split("\n")
PIDs_to_keep = [PID for PID in PIDs if int(PID) != PID_to_remove]
```
=> `ValueError: invalid literal for int() with base 10: ''` cause `PIDs=['158', '162', '']`

zamentur:
And on my side, i think, this is dangerous, when i write `write_to_file(XXX, some_string)` i expect my string to be saved without change like this.

So i vote for closing this PR.

alexAubin:
>For example
>[...]
>=> `ValueError: invalid literal for int() with base 10: ''` cause `PIDs=['158', '162', '']`

Urgh that's terrifying and i'm realizing there may be similar issue when, for example, writing that file and then appending the file which may result in an empty line and fucking things up

Not really feeling like reviewing every place in the code where we write a file. I don't even understand the initial motivation of "some text editor do it automatically and this may create unecessary diff" like what concrete examples are we talking about ... afaik even app conf files that admins may tweak manually and may likely end up being diffed are not written using python ... at least that's clearly too vague to justify such a subtle yet "dangerous" change ...

PR #2244, opened 2025-12-27
Replace cron jobs with systemd timers and services

Salamandar:
* untested
* maybe check the systemd capabilities?

### DIFF
--- a/conf/yunohost/yunohost-diagnosis.timer
+++ b/conf/yunohost/yunohost-diagnosis.timer
@@ -1,0 +1,12 @@
+[Unit]
+Description=YunoHost Automatic Diagnosis
+
+[Timer]
+OnCalendar=*-*-* 0/12:00:00
+# TODO forky: Use RandomizedOffsetSec instead and remove FixedRandomDelay
+RandomizedDelaySec=12h
+FixedRandomDelay=1
+Persistent=true
+
+[Install]
+WantedBy=timers.target
--- a/conf/yunohost/yunohost-dyndns.service
+++ b/conf/yunohost/yunohost-dyndns.service
@@ -1,0 +1,52 @@
+[Unit]
+Description=YunoHost DynDNS update service
+After=network.target
+
+[Service]
+Type=simple
+User=root
+Group=root
+WorkingDirectory=/root
+# Ensure IPv4 is available
+ExecCondition=ping -q -W5 -c1 ipv4.yunohost.org
+# Ensure an operation is not already running
+ExecCondition=test ! -e /var/run/moulinette_yunohost.lock
+ExecStart=yunohost dyndns update --email
+
+# Sandboxing options to harden security
+# Depending on specificities of your service/app, you may need to tweak these
+# .. but this should be a good baseline
+# Details for these options: https://www.freedesktop.org/software/systemd/man/systemd.exec.html
+NoNewPrivileges=yes
+PrivateTmp=yes
+PrivateDevices=yes
+RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6 AF_NETLINK
+RestrictNamespaces=yes
+RestrictRealtime=yes
+DevicePolicy=closed
+ProtectClock=yes
+ProtectHostname=yes
+ProtectProc=invisible
+ProtectSystem=full
+ProtectControlGroups=yes
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+LockPersonality=yes
+SystemCallArchitectures=native
+SystemCallFilter=~@clock @debug @module @mount @obsolete @reboot @setuid @swap @cpu-emulation
+# @privileged # (not sure why this need to be removed...)
+
+# Denying access to capabilities that should not be relevant for webapps
+# Doc: https://man7.org/linux/man-pages/man7/capabilities.7.html
+CapabilityBoundingSet=~CAP_RAWIO CAP_MKNOD
+CapabilityBoundingSet=~CAP_AUDIT_CONTROL CAP_AUDIT_READ CAP_AUDIT_WRITE
+CapabilityBoundingSet=~CAP_SYS_BOOT CAP_SYS_TIME CAP_SYS_MODULE CAP_SYS_PACCT
+CapabilityBoundingSet=~CAP_LEASE CAP_LINUX_IMMUTABLE CAP_IPC_LOCK
+CapabilityBoundingSet=~CAP_BLOCK_SUSPEND CAP_WAKE_ALARM
+CapabilityBoundingSet=~CAP_SYS_TTY_CONFIG
+CapabilityBoundingSet=~CAP_MAC_ADMIN CAP_MAC_OVERRIDE
+CapabilityBoundingSet=~CAP_NET_ADMIN CAP_NET_BROADCAST CAP_NET_RAW
+CapabilityBoundingSet=~CAP_SYS_ADMIN CAP_SYS_PTRACE CAP_SYSLOG
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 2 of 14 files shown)


conf/yunohost/yunohost-dyndns.service
```diff
@@ -0,0 +1,49 @@
+[Unit]
+Description=YunoHost DynDNS update service
+After=network.target
+
+[Service]
+Type=simple
+User=root
+Group=root
+WorkingDirectory=/root
+ExecStart=bash -c '! ping -q -W5 -c1 ipv4.yunohost.org >/dev/null 2>&1 || test -e /var/run/moulinette_yunohost.lock || yunohost dyndns update'
```
zamentur: What if it failed ? Before i think an email was send by the cron, but systemd-timer don't send email by default.

Should we want send email for that ?

conf/yunohost/yunohost-renew-certs.service
```diff
@@ -0,0 +1,48 @@
+[Unit]
+Description=Renew the HTTPS SSL certificates
+After=network.target
+
+[Service]
+Type=simple
+User=root
+Group=root
+WorkingDirectory=/root
+ExecStart=yunohost domain cert renew --email
```
zamentur, line 10: When the diagnosis return a port is down (sometimes it's not true, it's just the remote diagnosis that timeout before), the cert renew fail, and people need to add --force

conf/yunohost/yunohost-update-catalog.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Fetch the YunoHost apps catalog daily
+
+[Timer]
+OnCalendar=*-*-* 00:00:00
+RandomizedOffsetSec=24h
```
zamentur: Note that before it was between 00:00 and 01:00. I think it's maybe better like you suggest it.

hooks/conf_regen/01-yunohost
```diff
@@ -230,80 +230,58 @@ do_pre_regen() {
...
-HandleLidSwitchExternalPower=ignore
-EOF
+    cp logind-ynh-override.conf "$pending_dir/etc/systemd/logind.conf.d/ynh-override.conf"
 
     cp yunohost-api.service "$pending_dir/etc/systemd/system/yunohost-api.service"
     cp yunohost-portal-api.service "$pending_dir/etc/systemd/system/yunohost-portal-api.service"
     cp yunoprompt.service "$pending_dir/etc/systemd/system/yunoprompt.service"
     cp proc-hidepid.service "$pending_dir/etc/systemd/system/proc-hidepid.service"
-    # Delete legacy yunohost-firewall service
-    touch "$pending_dir/etc/systemd/system/yunohost-firewall.service"
```
zamentur, line 300: Note: if the machine has not been updated since march 2025, it could be a problem

src/utils/jinja_filters.py
```diff
@@ -71,3 +76,25 @@ def to_toml(value) -> str:
     Serialize to string an object to Toml
     """
     return toml.dumps(value)
+
+
+T = TypeVar("T")
+
+
+def stable_shuffle(values: list[T]) -> list[T]:
```
zamentur: I don't see where this filter is used, i guess it should be used into resolv.dnsmasq.conf.j2

hooks/conf_regen/43-dnsmasq
```diff
@@ -37,7 +34,19 @@ do_init_regen() {
         cat /etc/machine-id || true
         date +%m%Y
```
zamentur: I see that the order change every month, there are pros and cons for that.

conf/yunohost/yunohost-diagnosis.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=YunoHost Automatic Diagnosis
+
+[Timer]
+OnCalendar=*-*-* 00,12:00:00
+RandomizedOffsetSec=12h
```
zamentur: why not it was 7h and 19h before, i guess it was the idea that 7h is the beginning of the day, so we want a diagnosis to know if something is broken.

However your solution could be a good things to avoid remote diagnosis overload.

zamentur (changes requested):
Some change to make in comments

hooks/conf_regen/43-dnsmasq
```diff
@@ -69,17 +78,13 @@ do_pre_regen() {
...
+    # The resolver list is randomized with a seed based on the machine + month number
     # This way, the shuffle is random but should be stable accross the same month
     # and make sure the regenconf is idempotent (at least during the same month)
     # i.e. that it doesn't re-shuffle the file everytime we run the regenconf
-    SEED=$( (
-        cat /etc/machine-id || true
-        date +%m%Y
-    ) | md5sum)
-    cat plain/resolv.dnsmasq.conf | grep "^nameserver" | shuf --random-source=<(echo "$SEED") > "${pending_dir}/etc/resolv.dnsmasq.conf"
+    export custom_resolvers=""
```
zamentur: Is it related to https://github.com/YunoHost/yunohost/pull/2246

conf/yunohost/yunohost-update-catalog.timer
Salamandar: Yes, the change was voluntary :) There's no reason to have everything during only 1 hour.

src/utils/jinja_filters.py
Salamandar: Yes, sorry, I started working on another branch but forgot to change branches. I removed the changes related to dnsmasq.

hooks/conf_regen/43-dnsmasq
Salamandar: Yes, I removed the changes related to dnsmasq from this PR.

conf/yunohost/yunohost-diagnosis.timer
Salamandar: Having all instances run their diagnosis on a 20 minutes window might be a bit too much on our infrastructure.
I thought having just 2 diagnosis at random times per day would be enough. Email is asynchronous so we don't really care when the diagnosis email is sent, right ?

conf/yunohost/yunohost-renew-certs.service
Salamandar: Hmmm right, but people can still edit the .service file to add --force, right ? 

(actually it might be --no-checks, not --force, but the point still is valid)

conf/yunohost/yunohost-dyndns.service
Salamandar: Hmmmm agreed, lets extend dyndns update then.

zamentur: https://github.com/YunoHost/yunohost/pull/2265

conf/yunohost/yunohost-renew-certs.service
zamentur: Indeed it's `--no-checks`. Maybe it's an ipv4 / ipv6 issue. I guess let's encrypt acme check and yuno acme checks is not totally the same, cause i had this issue on several servers...

Anyway, this comments should not block this PR, if we don't find solution before merging it, we could create an issue to document the problem

conf/yunohost/yunohost-diagnosis.timer
zamentur: I think we can close this comment review.

conf/yunohost/yunohost-dyndns.service
Salamandar: Ok the email is tested, working fine (on incus).

Salamandar: Added --email to dyndns.

PR #2249, opened 2026-01-05
nginx: Rename fastcgi and proxy params files with file extension

Salamandar:
Also use includes to reduce code duplication

## The problem

file extensions please !

## Solution

...

## PR Status

...

## How to test

...


alexAubin:
They were probably named with no extension because the `fastcgi_params` and `proxy_params` file shipped in nginx default conf have no extension either ... Changing this now will mean changing every autopatch from last week too ...

zamentur:
I think i have seen some app including those files with no ext into their nginx conf. Do we really need this, i think it's a risky change for small gain.

I close this PR, feel free to reopen it if you are sure it doesn't need more huge works.

PR #2260, opened 2026-01-19
fix(form): allow empty domain when optional

kay0u:
## The problem

I have an optional domain parameter (so it's empty by default), but during the initialization his value is set do the default domain: https://ci-apps-dev.yunohost.org/ci/job/17058

## Solution

If optional is True, return empty string

## PR Status

Done

## How to test

...

### DIFF
--- a/src/utils/form.py
+++ b/src/utils/form.py
@@ -1701,10 +1701,9 @@ class DomainOption(BaseChoicesOption):
 
     @validator("default", pre=True, always=True)
     def inject_default(cls, value: str | None, values: Values) -> str | None:
-        # If the option is optional and no default was explicitly set,
-        # don't force the main domain as default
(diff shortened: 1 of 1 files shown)


src/utils/form.py
```diff
@@ -1701,6 +1701,11 @@ def inject_domains_choices(
 
     @validator("default", pre=True, always=True)
     def inject_default(cls, value: str | None, values: Values) -> str | None:
+        # If the option is optional and no default was explicitly set,
+        # don't force the main domain as default
+        if values.get("optional") and value in (None, ""):
```
kay0u:
```suggestion
        # If the option is optional, don't force the main domain as default
        if values.get("optional"):
```

kay0u:
Well, that may create more problems than it solves.

PR #2268, opened 2026-02-04
ssh: Rework/update ciphers and algorithms (Mozilla recommendations are too outdated)

zamentur:
## The problem

Mozilla configurations are intended for older versions of OpenSSH server (Modern: 6.7, Intermediate: 5.3). It does not make sense to use them with newer versions of openssh.

## Solution

@Josue-T and I have analyzed different openssh versions (latest, trixie, bookworm) and mozilla recommendation and defined a method to select algorithms when new version of openssh or yunohost are released.

- We replace the `intermediate` conf, by the default conf of the current debian.
- We have improved 'modern' conf from mozilla by taking into account latest openssh default conf.
- We introduce a no downgrade and postquantic `paranoid` mode (to satisfy paranoid people and avoid regular PR about ciphers)
- Like mozilla guidelines, we prioritizes largest key and security than performance (so hmac before umac)

See comments for better explanations.

## PR Status

Untested

## How to test

...

## TODO
 - [ ] Add mlkem768x25519-sha256 in YunoHost trixie version (based on openssh 10)


locales/en.json
```diff
@@ -528,7 +528,7 @@
     "global_settings_setting_smtp_relay_port": "SMTP relay port",
     "global_settings_setting_smtp_relay_user": "SMTP relay user",
     "global_settings_setting_ssh_compatibility": "SSH Compatibility",
-    "global_settings_setting_ssh_compatibility_help": "Compatibility vs. security tradeoff for the SSH server. Affects the ciphers (and other security-related aspects). See https://infosec.mozilla.org/guidelines/openssh for more info.",
+    "global_settings_setting_ssh_compatibility_help": "Compatibility vs. security tradeoff for the SSH server. Affects the ciphers (and other security-related aspects). See comments into /etc/ssh/sshd_config for more info.",
```
Josue-T: Maybe we can point the user to the template so the user can see the full description (not the only one that he selected).
```suggestion
    "global_settings_setting_ssh_compatibility_help": "Compatibility vs. security tradeoff for the SSH server. Affects the ciphers (and other security-related aspects). See comments into /usr/share/yunohost/conf/ssh/sshd_config for more info.",
```

Josue-T (approved):

Salamandar (approved):

SveDec:
Hi, is there a reason to use `/etc/ssh/sshd_config` for SSH settings set by Yunohost ?
If not, I'd like to propose a PR to restore the vanilla Debian conf file, and use a dedicated file for Yunohost inside the `/etc/ssh/sshd_config.d/` directory. This would allow advanced users to put their own settings in a separate file in the same directory, and avoid conflicts during major Debian upgrades, when the package upgrade tries to update the file.
Thanks !

alexAubin:
Mostly for historical reasons I suppose, I think the include / `sshd_config.d` mechanism is somewhat recent i think ? Maybe only available in bookworm ?

I'm not sure to understand what benefits there would have in moving all yunohost tweaks to a file in `sshd_config.d`, keep in mind that many people will just expect to find the bulk of the config in the regular ssh config file, rather than being "hidden" elsewhere. Also we do modify the original configuration because we want to have an exhaustive guarantee of the value of each parameters. If we keep the default conf, this may introduce unexpected variation, because VPS providers may enable/tweak an unexpected option X or Y which will introduce unexpected behaviors or security issues compared to the default yunohost config.

However it could indeed be cool to have an include statement for people to override the config without actually editing the file

SveDec:
> Mostly for historical reasons I suppose, I think the include / `sshd_config.d` mechanism is somewhat recent i think ? Maybe only available in bookworm ?

It appeared with Debian 11 Bullseye

> I'm not sure to understand what benefits there would have in moving all yunohost tweaks to a file in `sshd_config.d`, keep in mind that many people will just expect to find the bulk of the config in the regular ssh config file, rather than being "hidden" elsewhere.

Mmh, that may be a bubble effect, but I would have claimed the opposite : I see most people around me use the dedicated directory and expect the base file to stay untouched.

> Also we do modify the original configuration because we want to have an exhaustive guarantee of the value of each parameters. If we keep the default conf, this may introduce unexpected variation, because VPS providers may enable/tweak an unexpected option X or Y which will introduce unexpected behaviors or security issues compared to the default yunohost config.

Unless VPS providers remove the conf dir include, conf files inside this dir take precedence, cf `sshd_config` [man page](https://manpages.debian.org/trixie/openssh-server/sshd_config.5.en.html) :
_/etc/ssh/sshd_config.d/*.conf files are included at the start of the configuration file, so options set there will override those in /etc/ssh/sshd_config._

(this is because OpenSSH uses the first value set for most directives).

Indeed, you can't be sure the base conf file won't be totally messed up and ignore the conf dir if you don't control it, but if we want to harden ourselves, we could just ensure the vanilla file is installed.

The only drawback with this solution is that we have to be cautious on the files' name, as they are read in alphanumeric order (check for example [this page](https://utcc.utoronto.ca/~cks/space/blog/sysadmin/OpenSSHConfigOrderMatters) for a feedback on this) ; concretely, we will probably prefer `00-yunohost.conf` over `yunohost.conf` ^^

> However it could indeed be cool to have an include statement for people to override the config without actually editing the file

If there is a consensus on this solution, I can also submit a (much simpler, I admit) PR to add the line, probably on `dev` rather than `trixie` though :)

Salamandar:
> Unless VPS providers remove the conf dir include, conf files inside this dir take precedence, cf sshd_config [man page](https://manpages.debian.org/trixie/openssh-server/sshd_config.5.en.html) :

Yes, but that means we need to ensure our config file is as exhaustive as possible to override the possibly overriden values that we want to be kept to the default value… That's more of a hassle than replacing the original file.

SveDec:
> Yes, but that means we need to ensure our config file is as exhaustive as possible to override the possibly overriden values that we want to be kept to the default value… That's more of a hassle than replacing the original file.

Mmh, you're right ... but we're gonna have the same problem if Yunohost replaces the main conf file while including all the files in the conf directory : providers could add their own files in this directory (as they _should_ do since Bullseye).

To cover all cases we have to either be exhaustive, or not include all `*.conf` files (I guess we can include for example `ynh-*.conf` files or something similar).

zamentur:
My suggestion make tests and merge this PR, and create an other discussion / PR about /etc/sshd_config.d

oleole39:
@zamentur successful test on YNH 12.1.40.1

1. Replaced old files with the PR ones:
    - replace file in `/usr/share/yunohost/config_global.toml`
    - replace file in `/usr/share/yunohost/conf/ssh/sshd_config`

2. Tested all 3 modes via CLI:
    - `yunohost settings set security.ssh.ssh_compatibility -v intermediate` successful and a new SSH session can be opened
    - `yunohost settings set security.ssh.ssh_compatibility -v modern` successful and a new SSH session can be opened
    - `yunohost settings set security.ssh.ssh_compatibility -v paranoid` successful and a new SSH session can be opened

3. Tested all 3 modes via WebUI, after restarting  `yunohost-api` service.
4. Rebooted the instance in paranoid mode and was able to log in via SSH again without issue.

PR #2270, opened 2026-02-05
trixie: nsswitch -> nss regen-conf should clean the /etc/yunohost/regenconf.yml file (fix #2754 & # 2748)

ewilly:
## The problem

see https://github.com/YunoHost/issues/issues/2745 (or its duplicate it seems https://github.com/YunoHost/issues/issues/2748)

## Solution

update regenconf.py to remove nsswitch data

## PR Status

- [x] test on my system
- [x] ready for review

## How to test

...

### DIFF
--- a/src/regenconf.py
+++ b/src/regenconf.py
@@ -478,6 +478,8 @@ def _get_regenconf_infos():
             del data["metronome"]
         if "rspamd" in data:
             del data["rspamd"]
+        if "nsswitch" in data:
+            del data["nsswitch"]
         return data
     except Exception:
         return {}


ewilly:
It only removes the nsswitch category when there is a pending configuration files and run without category name argument (so basically on `yunohost tools regen-conf --force`). Seems to be correct. 
The only point I see it that the category is no removed if there is no pending, so will not solve situation for those already on trixie and if the upgrade of yunohost package does not provide any change in conf_regen hooks.

ewilly:
@Salamandar that one is ready for your review

Salamandar:
> The only point I see it that the category is no removed if there is no pending, so will not solve situation for those already on trixie and if the upgrade of yunohost package does not provide any change in conf_regen hooks.

I guess that's fine. I think we'll release a disclaimer saying stuff was borked on alpha/beta and an upgrade to stable might not be a good idea.

PR #2271, opened 2026-02-07
Force upgrade shall actually upgrade broken apps

orhtej2:
## The problem

`yunohost app upgrade <some_app> -F` does not upgrade apps marked as broken, even if potentially they're fixed and just wait for a catalog rebuild.

## Solution

Ignore broken state of single app requested for force upgrade.

## PR Status

YOLO

## How to test

1. (Force) Install a broken app
2. Try to upgrade it

PR #2272, opened 2026-02-07
Normalize casing of `--force` shorthand in `app` cli

orhtej2:
## The problem

`yunohost app upgrade` and `yunohost app install` use different shorthand for `--force` (`-F` vs `-f` respectively)

## Solution

Bite the bullet and use `-f` in both cases.

## PR Status

YOLO

## How to test

Invoke from cli 🤷

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -1004,10 +1004,10 @@ app:
                 -u:
                     full: --url
                     help: Git url to fetch for upgrade
-                -f:
+                -F:
                     full: --file
                     help: Folder or tarball for upgrade
-                -F:
+                -f:
                     full: --force
                     help: Force the update, even though the app is up to date
                     action: store_true


alexAubin (commented):
Maybe trixie can be a good time for this kind of breaking change ?

zamentur:
Note that it seems not to be the same behaviour between --force (install) and --force (upgrade)

Install
> Do not ask confirmation if the app is not safe to use (low quality, experimental or 3rd party), or when the app displays a post-install notification

Upgrade
> Force the update, even though the app is up to date

EDIT: related to https://github.com/YunoHost/yunohost/pull/2271

fflorent:
Not sure what is the status of this PR. Despite @zamentur's comment, both options are named `--force`, so I think that would bring consistency anyway. Does it need further discussions?

Despite not being *that* important, Alex suggested that because it is a breaking change, it should land along with the Trixie release. Feel like it is a good timing :).

orhtej2:
No outstanding tasks from my end, this PR is ready for merging once it gets an actual approval.

debian/changelog
```diff
@@ -33,6 +33,41 @@ yunohost (13.0.6) testing; urgency=low
 
  -- Félix Piédallu <felix@piedallu.me>  Thu, 06 Aug 2026 14:13:12 +0200
 
+yunohost (13.0.5) testing; urgency=low
```
orhtej2: what is this

share/actionsmap.yml
```diff
@@ -1004,10 +1004,10 @@ app:
                 -u:
                     full: --url
                     help: Git url to fetch for upgrade
-                -f:
+                -F:
```
orhtej2, line 1007: nope, mismerged

PR #2275, opened 2026-02-08
ynh_systemctl: add a --mount option similar to --service

Gredin67:
## The problem

...

## Solution

...

## PR Status

...

## How to test

...

### DIFF
--- a/helpers/helpers.v2.1.d/systemd
+++ b/helpers/helpers.v2.1.d/systemd
@@ -75,6 +75,7 @@ ynh_config_remove_systemd() {
 #
 # usage: ynh_systemctl [--service=service] [--action=action] [ [--wait_until="line to match"] [--log_path=log_path] [--timeout=300] [--length=20] ]
 # | arg: --service=      - Name of the service to start. Default : `$app`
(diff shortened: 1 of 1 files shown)


helpers/helpers.v2.1.d/systemd
```diff
@@ -75,28 +75,35 @@ ynh_config_remove_systemd() {
...
     ynh_handle_getopts_args "$@"
     service="${service:-$app}"
+    mount="${mount:-}"
     action=${action:-start}
     wait_until=${wait_until:-}
     length=${length:-20}
     log_path="${log_path:-/var/log/$service/$service.log}"
     # ===========================================
 
+    if [ -e "$mount" ]; then
```
alexAubin: Do you really mean `-e` ? `-e` is to check that the file exists, did you meant that you want to check if the variable is not empty string ? (aka `-n` ) ?

Gredin67: yep, I got confused between those 2

Gredin67:
can we merge ? @alexAubin

PR #2280, opened 2026-02-22
[fix] trixie-security instead of trixie #2719

zamentur:
## The problem

https://github.com/YunoHost/issues/issues/2719

## PR Status

Untested

## How to test

...

### DIFF
--- a/src/migrations/m0036_migrate_to_trixie.py.disabled
+++ b/src/migrations/m0036_migrate_to_trixie.py.disabled
@@ -550,7 +550,7 @@ sideway (typically by SSHing from the local network, or through rescue access on
             new_file_pars = []
             paragraphs = Deb822.iter_paragraphs(file.read_text())
             for paragraph in paragraphs:
-                paragraph["Suites"] = "trixie"
+                paragraph["Suites"] = paragraph["Suites"].replace("bookworm", "trixie")
 
                 if re.match(
                     r"^https?://(repo|forge)\.yunohost\.org.*", paragraph["URIs"]


eduardomozart:
Are you sure it fixes the issue at https://github.com/YunoHost/issues/issues/2719 ? Because if it's `bookworm`, it will replace by `trixie` instead of `trixie-security`, as far as I'm concerned from this PR contents.

Salamandar (approved):

Salamandar:
> Are you sure it fixes the issue at [YunoHost/issues#2719](https://github.com/YunoHost/issues/issues/2719) ? Because if it's `bookworm`, it will replace by `trixie` instead of `trixie-security`, as far as I'm concerned from this PR contents.

It just replaces `bookworm` with `trixie` in the string. So `Suites: bookworm bookworm-security` will be `Suites: trixie trixie-security`.

eduardomozart:
> > Are you sure it fixes the issue at [YunoHost/issues#2719](https://github.com/YunoHost/issues/issues/2719) ? Because if it's `bookworm`, it will replace by `trixie` instead of `trixie-security`, as far as I'm concerned from this PR contents.
> 
> It just replaces `bookworm` with `trixie` in the string. So `Suites: bookworm bookworm-security` will be `Suites: trixie trixie-security`.

But I'm not sure if this replacement is enough. Because on the APT file of https://github.com/YunoHost/issues/issues/2719 the script already replaced `bookworm` by `trixie`, but now it is `trixie-security` not simply `trixie`.

Salamandar:
> the script already replaced bookworm by trixie

No, the script was wrongly replacing the *whole line* with `trixie`. Now we're only replace occurrences of `bookworm` with `trixie`.

EDIT: I checked out the content of the `debian.sources` of Debian Cloud 12, here it is for reference:

```deb822
Types: deb deb-src
URIs: mirror+file:///etc/apt/mirrors/debian.list
Suites: bookworm bookworm-updates bookworm-backports
Components: main
Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg

Types: deb deb-src
URIs: mirror+file:///etc/apt/mirrors/debian-security.list
Suites: bookworm-security
Components: main
Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
```

so the replacement should be what's needed here.

PR #2286, opened 2026-03-17
security: restrict permissions on /var/log/nginx/*.log

alexAubin:
## The problem

Turns out that `/var/log/nginx/domain.tld-*.log`, on a fresh server, are world readable

```
# namei -l /var/log/nginx/yolo.test-access.log
f: /var/log/nginx/yolo.test-access.log
drwxr-xr-x root root /
drwxr-xr-x root root var
drwxr-xr-x root root log
drwxr-xr-x root adm  nginx
-rw-r--r-- root root yolo.test-access.log
```

In the general case, badly conceived apps (packages or upstream) may perform HTTP request that leak secrets in their URL or query args ... which in turn could be read by a compromise account to escalate privileges etc.

## Solution

In that, in turns out that on an "old" server. the -access.log et al are owned by www-data:adm with rw-r----- ... because logrotate actually comes with a statement `create 0640 www-data adm` ... so every log rotated gets these new permissions. Which sort of automatically adresses the problem for old server. But it also demonstrates that it should be safe to apply these permissions even for new servers (I mean we could imagine that some apps like monitoring, or maybe matomo needs access to these but apparently not ?)

Anyway, this PR adds the corresponding chown/chmod in the regenconf

## PR Status

Yolocomitted

## How to test

Eeh idk run the regen conf on a new server


hooks/conf_regen/15-nginx
```diff
@@ -177,6 +177,10 @@ do_post_regen() {
         chmod 644 /etc/nginx/conf.d/*.d/*.conf
     fi
 
+    # We don't want nginx logs to be world-readable which could leak secrets from GET query args in some cases/apps...
+    find /var/log/nginx -type f | xargs -r chmod 0640
+    find /var/log/nginx -type f | xargs -r chown www-data:adm
```
Josue-T, line 182: Just thinking why we can't just protect the whole `/var/log/nginx` directory instead of each files in the `/var/log/nginx` dir ?

alexAubin: I'm not sure, maybe www-data needs read permission but the owner/group is apparently root/adm ... So would need to sit and think about it ... Fixing the permission as logrotate implicitly does is just more straightfoward and has more guarantee that we don't forget some sort of edge case of subtleties in permissions

Josue-T: Yeah, well the issue that I see here is that doing this in the regen-conf feel like not really strong because the regen-conf is not executed frequently. Let's take this scenario:
- Nginx create the file `/var/log/nginx/access.log`
- We run the regen-conf, this will protect `/var/log/nginx/access.log`
- then logrotate move `/var/log/nginx/access.log` to `/var/log/nginx/access.log.1`
- Nginx create again `/var/log/nginx/access.log`
- I would suppose `/var/log/nginx/access.log` will be again with `-rw-r--r--` and  this could be for a long time.

In my opinion we would need to ensure that nginx create the file with the correct rights, or ensure that the parent directoy `/var/log/nginx` are protecting the content.

Thatoo:
Is there any link to this issue https://github.com/nginx/nginx/issues/319 ?
In that case, the issue is that permissions for log files, such as access.log and error.log, are set to a hardcoded constant value NGX_FILE_DEFAULT_ACCESS in this file https://github.com/nginx/nginx/blob/18ccebb1a889eb6989c64754f4f9b2512d58a491/src/os/unix/ngx_files.h#L105

zamentur:
Tested by thatoo: yes it works

Thatoo:
!testme

Thatoo (approved):

hooks/conf_regen/15-nginx
alexAubin: >then logrotate move /var/log/nginx/access.log to /var/log/nginx/access.log.1

But logrotate *does* auto-fix the situation, the files created by logrotate (both access.log.1 and access.log, if I'm not mistaking, since when you look at an existing machine it also has the proper permission)

The initial issue is that the file permissions are not secure **until** logrotate rotates the file for the first time

PR #2288, opened 2026-03-27
[TO DISCUSS?] Use mise to ease helpers

Thovi98:
## Why?

More and more apps require a different python version than the current one. Several apps already use `uv` or other tools to circumvent this issue, but IMHO this situation has to be handled by an official helper. I searched for a solution that could do the job and found [mise](https://mise.jdx.dev/) (IIRC some others mentioned it in the apps room also).

The fact with mise is that - on the contrary of the versions managers we already have - it can handle a lot of tools.

Based on the current helpers, I’ve written an helper to install `mise`, and one to install/remove python the same way we did before integrating node/ruby/go into the resources. As mise handle these tools too, it was pretty easy to adapt all those helpers to use mise too.

This has been my reasoning:
- one mise to rule them all
- if we want to add a new tool ([julia](https://github.com/YunoHost/apps/pull/3401)?), we already have the structure and it’s a matter of changing the name of the helper (with some tweaks here and there).
- mise support prebuilt packages for a lot of tools, including ruby ([which we compile for now, which takes a lot of resources](https://github.com/YunoHost/issues/issues/2690))

Mise is now installed in `/opt/mise/bin/mise` and the tools installed are located in `/opt/mise/binaries/` (several apps using the same nodejs version for example use the same binary). I wanted first to install the binaries in every app’s directory (which was convenient for backup/restore ; and to run it with `ynh_exec_as_app`), but it increased the installation/backup size up to 300/500Mb and duplicated each dependency when needed by several apps, which was not a great thing…

## PR Status

- mise installation and the python helper seems to work well. I’ve managed to test it several times on my test server and on the ci for apps that require at least python 3.13. e.g. https://ci-apps-dev.yunohost.org/ci/job/19449 (on my test server I removed the official node helper to prevent conflict). 
- other helpers were tested also but I don’t know how to convert them into the python resource (as I’m really not used about python yet).
- mise has new versions every day. Should we find a way to update it 1 time/month via workflows or we increase it whenever we want manually (but we should not forget it)?
- I can split this pr to have mise and python helpers separate of the rest that already exist
- I’m open to feedback/discuss. I enjoyed working on it and better understanding the helpers. I’m quiet proud on the result but I understand if it’s not what we want. 

Examples of application:
Python:
https://ci-apps-dev.yunohost.org/ci/job/19449

Go:
https://ci-apps-dev.yunohost.org/ci/job/19446

Ruby:
https://ci-apps-dev.yunohost.org/ci/job/19448

Nodejs:
all of the above
 
## How to test

Backup the old helpers in another directory and install some of the mentioned apps.


Thovi98:
Latest changes make sure that the used tool is properly removed when no apps use it anymore.
The mise install dir (`/opt/mise`) is also removed when no apps use it anymore. 

So this is ready for review, when times and energy permit it.

zamentur:
Dear contributors,

PR #2289, opened 2026-03-31
port resource: take secondary ports into account

hbrunn:
## The problem

port conflicts when an app uses multiple ports and isn't running at install time of second instance

## Solution

adjust the regex searching for ports in use to also search for secondary ports

## PR Status

...

## How to test

install an app that uses multiple ports, ie libererp_ynh. stop the app, install another instance of the app and observe that without the patch, the chat port of the second instance will be 8072, while it should be 8073


zamentur:
Dear contributors,

PR #2290, opened 2026-04-10
Fix SSE during upgrade

kay0u:
## The problem

Sometimes we only see the backup logs but not the upgrade ones

## Solution

Pre create the operation_logger for the upgrade, so the backup operation is seen as a child and we don't have to close/recreate a new stream between the backup and the upgrade

PS: **Use "hide whitespace" during the review**, otherwise it's unreadable for nothing

## PR Status

Tested directly in production, as usual

## How to test

...

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -907,223 +907,234 @@ def app_upgrade(
                 else:
                     upgrade_type = "UPGRADE_APP"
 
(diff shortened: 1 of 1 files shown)


kay0u:
closing if favor or #2291

PR #2291, opened 2026-04-10
Fix SSE during upgrade

kay0u:
## The problem

See #2290

## Solution

Only have one socket for all the operations, start it when the broker is started and reuse it every time

## PR Status

Tested in prod

## How to test

...


src/utils/sse.py
```diff
@@ -120,6 +123,20 @@ class SSEEventHeartbeat(SSEEventBase):
 )
 
 
+_shared_pub_socket: Any = None
```
Salamandar: Let's use stuff like functools.cache ?

zamentur:
Dear contributors,

PR #2292, opened 2026-04-20
友 Use po files and gettext instead of json files

Salamandar:


zamentur:
Could you add a `Status` section (with todo list?) ?

PR #2294, opened 2026-04-25
make upstream config usable to make loadbalancing possible on reverse_proxy

rosbeef:
## The problem

Can not make a reverse_proxy with load balance using the redirect app 

## Solution

Include  a upstream.config file located in the domain.d directory

## PR Status

Not tested :/ please help me!

## How to test

Create a redirect app which should be modified too

### DIFF
--- a/conf/nginx/server.tpl.conf
+++ b/conf/nginx/server.tpl.conf
@@ -2,10 +2,11 @@ map $http_upgrade $connection_upgrade {
     default upgrade;
     ''      close;
 }
-
+{%- if balancer_enable == "True" %}
 upstream {{ domain }} {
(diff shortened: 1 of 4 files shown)


rosbeef:
Need the file created ans not empty

rosbeef:
with a upstream.config file filled with  "server 127.0.0.1; " nginx is not complaining.
while this file is not used it seems to not make any change for yunohost

rosbeef:
Change the way of thinking,
Is reverse proxy and redirect should be part of an external app or a domain option?

rosbeef:
After thinking with yunohost member on the matrix channel, this way is not the prefered to modify a domain.conf file for now, as it should be discussed in the core team because it's impact more stuf.

For now i will edit the domain.conf from the redirect app.

PR #2002, opened 2024-11-21
Support Debian 13 / Trixie

Salamandar:
Fix https://github.com/YunoHost/issues/issues/2469

This is a draft PR as Trixie is not released yet.

For Debian Trixie freeze dates, see https://release.debian.org/trixie/freeze_policy.html 

Blockers:

- [x] PHP repository Sury [are waiting for transition freeze](https://github.com/oerdnj/deb.sury.org/issues/2176#issuecomment-2480532004) to start working on a Trixie repository. Most PHP apps are broken until then
- [x] Missing package `python3-lexicon` on Trixie: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1003200
- [x] [Pydantic v2](https://github.com/YunoHost/yunohost/pull/2001) (v1 is not provided by Trixie anymore)
- [x] dovecot configuration
- [ ] postfix service down but isn't ?
- [x] App tests failing
- [x] Pydantic errors in CI
- [x] Migration from Bookworm
  - [ ] `networking.service` failed on VPS after migration?
  - [x] `sssd-*.socket` failed after migration: misconfiguration?
  - [ ] Legacy apt files still present ?

TODO: apt go to stable only

PR #2295, opened 2026-04-26
trixie: configure sssd_ldap to use ldap:// on localhost without any TLS

artlog:
## The problem

sssd_ldap use a ldaps:// on localhost what adds no security ( since localhost ) and does consume cpu and create delays.

## Solution

remove anything about tls, switch from ldaps:// to ldap://

## Additionnal informations

to get it working correctly in tls if later on an external ldap server would be in use :

a real certificate with a trusted ca should be used ldap_tls_reqcert = demand should be used with ldaps://<domain.tld>:636 url

still some extra info about removed fields that were not relevant anyway :

remove ldap_tls_cacert
/etc/yunohost/certs/yunohost.org/crt.pem is a self signed certificate it is for yunohost.org that won't match
it is not a CA ( CA:FALSE )
found information indicating self-signed certificates can't be verified with ssdd anyway see ( https://docs.pagure.org/sssd.sssd/users/faq.html )

remove ldap_tls_key
this would not be required even with tls
it is for client authentication and we don't provide ldap_tls_cert afaik ynh ldap server is not configured to accept client certificates

## PR Status

It did work in my yunohost dev environment ...

## How to test

restart sssd 

create an new user
login to system with it
getent passwd <new_user>

### DIFF
--- a/conf/sssd/sssd.conf
+++ b/conf/sssd/sssd.conf
@@ -17,11 +17,9 @@ auth_provider = ldap
 chpass_provider = ldap
 sudo_provider = ldap
 
-ldap_uri = ldaps://localhost:636
+ldap_uri = ldap://localhost:389
 ldap_search_base = dc=yunohost,dc=org
 
-ldap_tls_reqcert = allow
-ldap_tls_cacert = /etc/yunohost/certs/yunohost.org/crt.pem
-ldap_tls_key = /etc/yunohost/certs/yunohost.org/key.pem
+ldap_tls_reqcert = never
 
 cache_credentials = True


artlog:
@alexAubin this pr is dedicated to you since it answers one of your request.
@Salamandar , if you can review it ;-) thanks.

Salamandar:
I prefer to keep TLS, but that's only my opinion. If the project decides to disable it (because, yeah, localhost), that's fine with me and in this case this PR is lgtm.

alexAubin (approved):
Personally I don't understand what the argument for having TLS would be, TLS is for authenticity and confidentiality, and that's irrelevant when localhost is talking to localhost ...

artlog:
@Salamandar In fact it is not only becasue it is localhost.
Currently we are doing TLS but providing no additional security since ldap_tls_reqcert = allow will just allow any kind of certificate without authentication of remote.

if we want TLS so we have do enforce more things :

ldap_tls_cacert : should be a CA , then certificate should be a signed by CA one.
ldap_tls_reqcert = hard : we enforce the certificate to be good, so certificate domain should match.
ldap_tls_key : useless , we are not authenticating ourselves with a key.

It would perhaps be possible to use ldapi// with trust on unix user, but i am not sure sss_ldap supports it, if that's possible it allows to not even use a bind password, that's not ciphered by this is to the kernel level handling.

artlog:
At least we could do a !testme to check it does not break anything ...

alexAubin:
!testme is an app-packaging thing, the core is tested via a gitlab pipeline, but since this PR is from a repo fork, it's not clear to me how to trigger a test without pulling the changes into a branch that's on the base repo

artlog:
i will try to give some ground on my suggestion because this should be discussed and not pushed by force.

- in ynh 12 we are using nslcd and its configuration is ldap://localhost without any TLS. so not having TLS here is not a regression.
- sssd is really meant to be run on the very same host, if that has to change then TLS would be required.

For a **power consumption reduction** and **latency**. Using TLS does involve TLS handshakes that are done on the same system both on client library side and on ldap server side. TLS handshake are using asymetric cryptography that is expensive cpuwise.

Then of course we need security

- ldap or ldaps are both using clear ldap password, this mean people who can see content of traffic can get passwords. or another trust method ( ldapi::// sasl ) should be used.
- TLS is used to hide content a communication channel for eyesdropping
- If TLS is used but not enforced by CA it is trival to forge one. So first we need to make TLS working with trust. And we are currently not meeting that requirement.

So threat to not use TLS would be loopback sniffing by a process running on host.

This might be more complex to set as Man in the middle on loopback and we are back to the fact that it is looback that protect us. Here to attack this it is a matter or opening a tun or tap interface on loopback and this really requires privileges, it really does not seems easy.

Having TLS on loopback would add a layer of protection, But this might be unusual as having a certificate with localhost as DNSname in any case it can't be a letsencrypt one. Still this is relevant only if we can't find a better solution for IPC than non unix sockets.

Of course if we could have a unix domain socket, and enforce user level privilege it would really be nice. Here we could (perahps ?) use ldapi: so using unix domain socket and then user privilege checking, that is not TLS at all but we protect against the fact that loopback connection is accessible by any process whatever its privilege is. At that level we could even wonder if sssd should not access ldap database internally through its lmdb database ie by linking directly ldap backend, but ... no i'am going too far, let's stay focus ;-)

But in current yunohost system any ldap authenticated App is using ldap:// so this traffic already exists, protection sssd won't protect other traffics. So this loopback potential sniffing is already there.

alexAubin:
I don't know if we really need to overthink the whole thing....

>threat to not use TLS would be loopback sniffing by a process running on host.

> this really requires privileges

The threat of "the attacker has enough privileges to sniff on localhost traffic BUT not enough privileges to do something much more obvious like being able to read critical secrets on the filesystem directly" seems super imaginary to me ...

If we do think this is an actual threat, then the whole YunoHost stack is totally fucked, because a) the attacker could also sniff the traffic between nginx and reverse-proxied apps, b) the attacker could also sniff the traffic between apps and mysql/postgresql, c) the attacker could probably start its own daemon on some relevant port, with its own self-signed certificate, and therefore using TLS or not is irrelevant

The TLS option exist for a reason : it is when you want to interact with stuff on another machine, in which case you want authenticity and confidentiality. It is not relevant in localhost<->localhost communication, period ... It's only technical overhead, complexity overhead, and another moving part in a critical piece of the stack, ultimately leading to broken system for broken reason like "the system clock was reset therefore the self-signed cert is invalid (sigh)"

artlog:
@alexAubin i agree.

Here the background of this Pull Request :
When i read https://github.com/YunoHost/yunohost/pull/2002 ( around https://github.com/YunoHost/yunohost/pull/2002/changes/60de75e49bd4155dfb75b94234472549cf97da6b#r2485016014 ) that configuration to remove TLS was not found. so i proposed one that is working on my dev system.

But given @Salamandar remark i see now that consensus around TLS removal was not reached.
TLS was introduced very recentlty together with sssd and only in trixie.

I have no right over this project, so i can't do more than propose.

This then lead to a general question : how such security decisions are taken in yunohost project ?

Salamandar:
> But given @Salamandar remark i see now that consensus around TLS removal was not reached.

That's fine with me to remove TLS :)

artlog:
Given consensus, Waving whoever is maintainer to consider integration, please ;-)

zamentur (approved):
LGTM, i agree with Aleks analyzes, user need sudo permission (or to be root) in order to make sniff on lo interface. So if a malicious user has this kind of right, there are others way to attack the system.

zamentur:
Waiting for automatic tests to run (i just clicked on the checks workflow permission)

zamentur:
I create an internal branch with the code to run the CI...

PR #2297, opened 2026-05-09
Implement `yunohost app db app` to spawn a client db prompt

fflorent:
## The problem

It can be painful to remember how to open a db prompt for an app:
1. You have to identify whether the app use mysql or postgresql in the manifest
2. You have to remember the command to run (especially for postgresql with which you have to login first as `postgres`)

## Solution

Offer the `yunohost app db APP` for that.

This command also takes a different approach than the helpers: I force to connect as the app db user (rather than the admin account) for arguable reasons:
1. So the user has exactly the same rights as the app (can avoid confusions)
2. If the user does something dangerous, it is only scoped to the app.

## PR Status

Tested on both bookworm and trixie ✅

## How to test

- You may install piwigo to check it works with mysql/mariadb
- You may install lufi to check it works with postgresql
- You may run `yunohost app db lufi` or `yunohost app db piwigo` and see that you can operate on their respective DB
- You may run `yunohost app db piwigo <<< "select * from users;"` to see that we can run queries this way
- You may run `echo "select * from files limit 1;" > /tmp/lufi.sql; yunohost app db lufi < /tmp/lufi.sql` to see that we can execute sql commands this way too.

### DIFF
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -1085,6 +1085,8 @@ app:
             arguments:
                 app:
                     help: App ID
+                    extra:
+                        autocomplete: *apps_list
 
         ### app_register_url()
         register-url:
--- a/share/actionsmap.yml
+++ b/share/actionsmap.yml
@@ -1078,6 +1078,14 @@ app:
                     extra:
                         autocomplete: *apps_list
 
+        ### app_db()
+        db:
+            action_help: Open an interactive database client prompt for the app
+            api: GET /apps/<app>/db
+            arguments:
+                app:
+                    help: App ID
+
         ### app_register_url()
         register-url:
             hide_in_help: True
--- a/src/app.py
+++ b/src/app.py
@@ -1892,6 +1892,62 @@ def app_shell(app: str) -> None:
         env=env,
     )
 
+def app_db(app):
+    """
+    Open an interactive DB pompt for the app
+
+    Keyword argument:
+        app -- App ID
+
+    """
+    import subprocess
+    import tempfile
+
+    _assert_is_installed(app)
+    local_manifest = _get_manifest_of_app(app)
+    settings = _get_app_settings(app)
+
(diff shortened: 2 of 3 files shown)


share/actionsmap.yml
```diff
@@ -1078,6 +1078,14 @@ app:
                     extra:
                         autocomplete: *apps_list
 
+        ### app_db()
+        db:
+            action_help: Open an interactive database client prompt for the app
```
fflorent, line 1083: I would ideally document too how that it's possible to run queries this way:
```bash
yunohost app db APP <<< "select * from MY_APP_TABLE"
yunohost app db APP <<< /path/to/my_commands.sql
```

But it looks like newlines are ignored here.
Maybe we would have to instead offer arguments for that?

src/app.py
```diff
@@ -1893,6 +1893,64 @@ def app_shell(app: str) -> None:
...
+    else:
+        raise YunohostValidationError("app_db_prompt_type_not_supported", type=type)
+
+    # NOTE: Postgresql (and probably mariadb too) requires the password file to
+    # be a regular one. We can't work with pipes.
+    # Let's create a temp file that is destroyed when quitting
+    with tempfile.NamedTemporaryFile() as fp:
+        fp.write(password_file_content.encode())
+        # FIXME: should close + call NamedTemporaryFile with `delete_on_close=False`
+        # when we can (requires 3.13 so when we deprecate bookworm)
```
fflorent:
```suggestion
```

Josue-T (approved):

share/actionsmap.yml
```diff
@@ -1078,6 +1078,14 @@ app:
...
+            api: GET /apps/<app>/db
+            arguments:
+                app:
+                    help: App ID
```
fflorent, line 1087:
```suggestion
                    help: App ID
                    extra:
                        autocomplete: *apps_list
```

PR #2298, opened 2026-05-11
[enh] Add dash in username and force first char to be alphanum

zamentur:
## The problem

In /etc/adduser.conf a default REGEX for user and group name exists:

```
# Non-system user- and groupnames are checked against this regular
# expression.
# Default: NAME_REGEX="^[a-z][-a-z0-9_]*\$?$"
#NAME_REGEX="^[a-z][-a-z0-9_]*\$?$"
```

We currently don't accept dash chars, but we probably should as we accept dot (that is not in default REGEX_NAME...).
## Solution

Use the default regex, with dot in more and without the $ at the end. We also tolerate usernames starting with a number in order to avoid some user to be banned from their server :/
```python3
r"^[a-z0-9][-a-z0-9_.]*$"
```

Note: some yunohost app could not support this kind of names (with dot, dash or even underscore) however, i think the high risk is for the dot char... [And we already decided to accept it](https://github.com/YunoHost/issues/issues/2287)
Note2: if we want to ease the way to autoconfigure email, we need username to be the first part of the email, so allowing dash is necessary cause lot of people want this kind of address ex: `jean-michel.smith@domain.tld`

## PR Status
Tested manually: user create, user import, user list
Waiting for auto test 

## How to test

...

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -753,7 +753,7 @@
     "pattern_password": "Must be at least 3 characters long",
     "pattern_password_app": "Sorry, passwords can not contain the following characters: {forbidden_chars}",
     "pattern_port_or_range": "Must be a valid port number (i.e. 0-65535) or range of ports (e.g. 100:200)",
-    "pattern_username": "Must be lower-case alphanumeric and underscore characters only",
+    "pattern_username": "Must be lower-case alphanumeric, dot, dash and underscore characters only",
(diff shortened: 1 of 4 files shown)


Thatoo:
It will make it easy to manage ynh users from Dolibarr as members, Dolibarr accepting [-a-z0-9_.]

Josue-T (approved):
LGTM, we just need to ensure that the tests are working.

PR #2300, opened 2026-05-19
security: handle the kernel special case when checking for security issues via the security.toml index

alexAubin:
## The problem

cf the discussion in https://github.com/YunoHost/apps/pull/3454/ about copy.fail and other security issues in the kernel, the kernel is not a regular package and it's not easy to properly list vulnerabilities in here. We need a proper `uname -r`

## Solution

Add a special case for a package named `kernel` in the security.toml check thingy in the diagnosis

## PR Status

Yolocommited

## How to test

...


alexAubin:
Tested on dev env during today's meeting, works as expected, yolomerging

PR #2304, opened 2026-05-23
enh(sshd_config)!: Refactor sshd_config template for different auth methods

louis-vgn:
## Problem
https://github.com/YunoHost/issues/issues/2804

## Solution
This commit slightly refactors the template code to allow for more explicit authentication methods notably because the `AuthenticationMethods` option in sshd_config requires that each method be explicitly listed; it’s cleaner to remove the previous `ssh_password_authentication` yunohost option.

Also, this allows for future potentially more secure auth methods as we can easily add a new "profile" like:

```
choices.advanced = "Public key and OTP authentication required"
```

Issue: https://github.com/YunoHost/issues/issues/2804
BREAKING CHANGE: The setting `security.ssh.ssh_password_authentication` is replaced with `security.ssh.ssh_authentication_methods`.

## STATUS
Work in progress

### TODO
 - [ ] Discuss design
   - [ ] Which default policy (do we allowed password auth by default) ?
   - [x] Which settings value to referrer to different policy ? insecure/basic/intermediate/advanced or pass_or_pubkey, pubkey, pass_and_pubkey, pubkey_and_otp ?
 - [x] Missing translation key 
   - [x] Config panel question `ssh_authentication_methods` need a translation key inside en.json (see https://github.com/YunoHost/yunohost/blob/18ac16a5c71ef6af1d1c274a70bbaee7b42cecfe/locales/en.json#L530 for example)
   - [x] Rename or delete `global_settings_setting_ssh_password_authentication_help` `global_settings_setting_ssh_password_authentication`
 - [x] Edit the hook `hooks/conf_regen/03-ssh` to give ssh_authentication_methods to the template
 - [x] Rebuild sshd_config after editing the value : `post_change_hook` decorator like here https://github.com/YunoHost/yunohost/blob/18ac16a5c71ef6af1d1c274a70bbaee7b42cecfe/src/settings.py#L342
 - [x] Add a warning in diagnosis to report that password auth only is insecure.
 - [ ] Write a migration to avoid breaks ssh config of people
   - [ ] The fact is ynh instances have already a defined ssh auth policy (and sometimes custom change onto sshd_config), so we need a migration to convert `ssh_password_authentication` into `ssh_authentication_methods`. You can find example migrations here https://github.com/YunoHost/yunohost/tree/dev/src/migrations
 - [x] Documentation
 - [ ] Final tests (and CI checks)
## How to test


share/config_global.toml
```diff
@@ -37,9 +37,13 @@ i18n = "global_settings_setting"
         type = "number"
         default = 22
 
-        [security.ssh.ssh_password_authentication]
-        type = "boolean"
-        default = true
+        [security.ssh.ssh_authentication_methods]
```
zamentur: This label needs translation

louis-vgn:
@zamentur Thanks for the triage of this PR. I see the label `work needed`. Would you mind pointing me in the right direction so I can work on it? I’m sure there is more to do, but looking at the source code and the yunohost docs, I couldn’t figure it out by myself.

Notably, I set up the dev environment with incus. The CLI does accept the new option `security.ssh.ssh_authentication_methods`, but it doesn’t apply the changes in `sshd_config`. In the logs I see the following line `Nothing to do after changing setting ssh_authentication_methods`. I’m clearly missing something, but I don’t know what.

<details><summary>Details</summary>
<p>

```bash
$ sudo yunohost settings set security.ssh.ssh_authentication_methods -v intermediate --debug
32   DEBUG   acquiring lock...
41   DEBUG   lock has been acquired
367  DEBUG   loading python module yunohost.settings took 0.326s
367  DEBUG   processing action 'yunohost.settings.set'
368  DEBUG   Import and parse pre-answered options
396  DEBUG   initializing root ldap interface
397  DEBUG   Ask unanswered question and prevalidate data

========================================
>>>> Security
========================================

# SSH
1467 INFO    Saving the new configuration...
1469 DEBUG   Nothing to do after changing setting ssh_authentication_methods
1474 SUCCESS Config updated as expected
1475 DEBUG   To view the log of the operation 'Apply settings', use the command 'yunohost log show 20260524-111926-settings_set'
1487 DEBUG   action executed in 1.120s
1487 DEBUG   lock has been released
```

</p>
</details>

zamentur:
Hi @louis-vgn , sorry i didn't pay attention that it was your first contrib on this repo. 

**Welcome ! 🎉**

I just edited short todo in your first post to give more explanation on what should be done in my point of view.

The doc PR seems ok, but we should discuss the idea to put arbitrary security level like basic or intermediate on this. In future we could want pubkey_and_otp to be intermediate for example... So i suggest others values that directly describe the conf choice and manage security level in label. (see my comment in the code).

I ping @Josue-T to let him take this point in mind (cause he works on OIDC/authelia integration, not totally related, but OTP may be done with authelia...)

zamentur:
> Notably, I set up the dev environment with incus. The CLI does accept the new option `security.ssh.ssh_authentication_methods`, but it doesn’t apply the changes in `sshd_config`. In the logs I see the following line `Nothing to do after changing setting ssh_authentication_methods`. I’m clearly missing something, but I don’t know what.

Good catch :)

You have to setup a `post_change_hook` decorator like here https://github.com/YunoHost/yunohost/blob/18ac16a5c71ef6af1d1c274a70bbaee7b42cecfe/src/settings.py#L342

louis-vgn:
> Hi @louis-vgn , sorry i didn't pay attention that it was your first contrib on this repo.

No worries. I should’ve mentioned it.

> **Welcome ! 🎉**
>
> I just edited short todo in your first post to give more explanation on what should be done in my point of view.

Thanks! I’ll get to work on those items.
 
> The doc PR seems ok, but we should discuss the idea to put arbitrary security level like basic or intermediate on this. In future we could want pubkey_and_otp to be intermediate for example... So i suggest others values that directly describe the conf choice and manage security level in label. (see my comment in the code).

Right. This seems less judgmental and more descriptive, although I think somewhere else than the documentation, password authentication should be stated to be insecure and highly prioritize pubkey auth (at least).

zamentur:
> Right. This seems less judgmental and more descriptive, although I think somewhere else than the documentation, password authentication should be stated to be insecure and highly prioritize pubkey auth (at least).

Maybe we should add a warning in diagnosis for that

Josue-T:
> I ping @Josue-T to let him take this point in mind (cause he works on OIDC/authelia integration, not totally related, but OTP may be done with authelia...)

Yeah for OTP with ssh, we would need a shared OTP backend which is compatible with ssh and authelia. This could be challenging to have something shared from a really quick research. Maybe the other solution could be just to have 2 independent OTP system for the SSO and for ssh but this could be confusing for end users.

louis-vgn:
> Maybe we should add a warning in diagnosis for that

For the warning in the diagnosis, should I edit the `70-regencon.py`, or the `00-basesystem.py`? I only see ssh related warnings in `70-*`, but it seems targeted to issues related to regeneration of the config files, while I feel it would make more sense to print a warning in the base system related to security. What do you think?

louis-vgn:
Regarding the default policy, I think it should be left to password only as that’s the one being enforced as of now. When installing yunohost, it all starts with a root account, hence ssh in the root account and in the admin account created by the postinstall script allows for ssh with password only. Setting a new default policy to passkey would require more work in the postinstall script and, more importantly, a change in the user onboarding:

One would have to have an ssh-key up and ready, run postinstall script, and at some point the script would need to ask the user to launch a new terminal, `ssh-copy-id` to their server while keeping the other shell (the password authenticated --probably still root-- shell) opened, then try if key based ssh authentication works, and if it works, we’re good and the script can keep going and change the default policy. 

All of that needs to happen outside of the postinstall script, because the user must take actions. Hence, I believe it is best to leave the default policy to password only for now, but print a serious warning in the diagnosis (which this commit does https://github.com/YunoHost/yunohost/pull/2304/commits/7a9f186c550b290697dc201c638273b8b50f68a3).

Implementing the workflow I described above would nevertheless make a much better default and seriously augment the default security posture[^1] of a yunohost instance. I could make another PR for this.

[^1]: That is not *necessarily* true since it will highly depend on how the user handles their private ssh key on disk. It still pushes for a best practice yunohost wise, but the ssh authentication security of the instance really all depends on how the user handles it. It’s out of scope of yunohost, of course, but it’s still important to keep that in mind to avoid a security theater.

louis-vgn:
#@zamentur Regarding the migration srcipt, I’ve looked at the examples, but I’m not sure what needs to be migrated.

Do you mean do some kind of search and replace in the `sshd_config` to look for new strings added by the PR (for instance the comments and the AuthenticationMethods key)? I think I get it for the coments, but for the actual config, wouldn’t the script handling the system setting handle it? Meaning that if I run `yunohost settings set security.ssh.ssh_authentication_methods -v  <policy>`, it will change what’s needed accordingly if the setup is not too customized. If it is and there were manual modifications to the files, then this would catch and print a warning no?
https://github.com/YunoHost/yunohost/blob/18ac16a5c71ef6af1d1c274a70bbaee7b42cecfe/src/diagnosers/70-regenconf.py#L30-L55

zamentur:
Dear contributors,

PR #2312, opened 2026-06-20
fix : Ipv6 pref sending if available

### LINKED ISSUE #2465: Configuration of IPv6 for email sending
# Context

When sending email from my YunoHost server, it happens that the email doesn't reach the recipient's inbox. For example, Gmail often puts my email in the spam folder. And Hotmail just ignores my email and the recipient never sees it (not even in the spam directory).

A friend suggested to configure IPv6 in Postfix and SPF to improve ma server's reputation.  I see two problems with emails sent through IPv4:

- Your IP address may have been used by a spammer in the past before it is attributed to you. When that happens, your address has a bad reputation and appears in a lot of block lists, before you even start to use it… This usually doesn’t happen with IPv6 (because there are so many of them)
- People self-hosting there SMTP server are really sparse. So when an address reserved to an [ISP](https://en.wikipedia.org/wiki/Internet_service_provider) starts sending emails, it is usually a machine infected by a malware and transformed into a zombie machine. This is why a lot of platform block email coming from IPv4 address reserved to an ISP.

And this is why I think sending emails though IPv6 is a better option.

After 3 weeks of testings, I am pretty happy. People using hotmail adresses finally receive my messages.

# Add the IPv6 in the SPF entry

The admin panel of Yunohost provides a suggestion for the configuration of the DNS record. (It can also configure it itself, but I never tried that feature). In the suggestions, we find entries related to the email (SPF, DKIM and DMARC). The suggested SPF entry looks like this:

    “v=spf1 a mx -all”

I added the ip address of the server in the entry (especially the ipv6):

    “v=spf1 a mx ip4:XXX.XXX.XXX.XXX ip6:YYYY:YYYY:YYYY:YYYY:YYYY:YYYY -all”

# Activate IPv6 in Postfix

By default, Postfix uses IPv4 to send the email. Some major email provider have drastic rules for emails received via IPv4, but are more permissive for email received via IPv6. So it is intresting to configure Postfix so it sends emails via IPv6 when possible. To do so, I edited the file /etc/postfix/main.cf and add two lines after inet_interfaces = all (the position is not important, but I put it where it seems coherent):

    inet_protocols = all
    smtp_address_preference = ipv6

(For information, since I did this change, Yunohost’s diagnosis system warns me that the file was manually modified.)

frju365:
AW: I just added the solution provided by @jershon 

## The problem

For SPF record : 
As I understood, it could be problem if main domain (exemple for a website) has different ipv4 than actual mail server.

For ipv6 : 
Some email provider communicate only in ipv6. Apparently following issue this can be a solution of not receiving mail on some email hosting providers.

**Related issues:** : fixes Yunohost/Issues#2465
**Related PRs:**

## Solution

- add preference to ipv6 if avilable
- Correct SPF record. Allow domain to be handle by server IP even if main site is no handle by server.

**AI transparency:** Understand difference between record (I didn't fully remember what was the good way to handle such case).

## PR Status
*Tested partially*

### Code TODOs

 - [x] Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)
 - [x] Write data or settings migration
   - [x] Add some missing translations keys
   - [x] Update/write unit tests
 - [x] Update/write documentation

### Tests TODOs

## How to test
...

### DIFF
--- a/src/dns.py
+++ b/src/dns.py
@@ -212,10 +212,10 @@ def _build_dns_conf(
             spf4 = ""
             spf6 = ""
             if ipv4:
-                spf4 = ' ip4:' + ipv6
+                spf4 = ' ip4:' + ipv4
             if ipv6: 
                 spf6 = ' ip6:' + ipv6
-            mail.append((basename, ttl, "TXT", '"v=spf1 a mx',spf4, spf6, ' -all"'))
+            mail.append((basename, ttl, "TXT", f'"v=spf1 a mx{spf4}{spf6} -all"'))
 
             # DKIM/DMARC record
             dkim_host, dkim_publickey = _get_DKIM(domain)
--- a/src/dns.py
+++ b/src/dns.py
@@ -208,12 +208,13 @@ def _build_dns_conf(
             mail.append((basename, ttl, "MX", f"10 {domain}."))
 
         if settings["mail_out"]:
-            # Tentative de mitigation en cas où le serveur du domaine d'envoi du mail est différent du serveur du domaine (site externalisé) Issue: #2465
+            # Tentative de mitigation en cas où le serveur du domaine d'envoi du mail est différent
+            # du serveur du domaine (site externalisé) Issue: #2465
             spf4 = ""
             spf6 = ""
             if ipv4:
                 spf4 = ' ip4:' + ipv4
-            if ipv6: 
+            if ipv6:
                 spf6 = ' ip6:' + ipv6
             mail.append((basename, ttl, "TXT", f'"v=spf1 a mx{spf4}{spf6} -all"'))
 
--- a/src/dns.py
+++ b/src/dns.py
@@ -209,11 +209,13 @@ def _build_dns_conf(
 
         if settings["mail_out"]:
             # Tentative de mitigation en cas où le serveur du domaine d'envoi du mail est différent du serveur du domaine (site externalisé) Issue: #2465
-            if ip4:
-                spf4 = 'ip4:' + ipv6
-            if ip6: 
-                spf6 = 'ip6:' + ipv6
-            mail.append((basename, ttl, "TXT", '"v=spf1 a mx ',spf4, ' ', spf6, ' -all"'))
+            spf4 = ""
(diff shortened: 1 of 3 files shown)


src/dns.py
```diff
@@ -208,7 +208,15 @@ def _build_dns_conf(
             mail.append((basename, ttl, "MX", f"10 {domain}."))
 
         if settings["mail_out"]:
-            mail.append((basename, ttl, "TXT", '"v=spf1 a mx -all"'))
+            # Tentative de mitigation en cas où le serveur du domaine d'envoi du mail est différent
+            # du serveur du domaine (site externalisé) Issue: #2465
```
zamentur, line 212: The issue 2465 seems not about distinct mail and web server for a same domain.

src/dns.py
```diff
@@ -208,7 +208,15 @@ def _build_dns_conf(
...
+            spf4 = ""
+            spf6 = ""
+            if ipv4:
+                spf4 = ' ip4:' + ipv4
+            if ipv6:
+                spf6 = ' ip6:' + ipv6
+            mail.append((basename, ttl, "TXT", f'"v=spf1 a mx{spf4}{spf6} -all"'))
```
zamentur, line 219: `a` already means `AAAA` if the mail is sent with ipv6.

SO for me the only reasons to add this, is the case where no MX or A records exists for the IP of the server. This happens if `mail_in` is disabled and if a website is hosted on the domain on another host.

https://github.com/Nerteron/yunohost/blob/12c2594b1024a604d5bd3d96007b79905d4de87d/src/dns.py#L207

I think there could be a lot of situations for which SPF should be customized, the important thing is to check with diagnosis that current IPs are allowed.

For me this part should be in another PR , maybe with a settings to include a custom part to SPF

zamentur, line 219:
```suggestion
            mail.append((basename, ttl, "TXT", '"v=spf1 a mx -all"'))
```

zamentur (changes requested):
I suggest to remove the SPF part

zamentur:
Dear contributors,

PR #2313, opened 2026-06-20
enh(mdns): replace 'device-info' service type by 'workstation'

tituspijean:
## The problem

After exploring my own network with `avahi-browse`, I discovered that the proper way to signal computers and servers is with the `_workstation._tcp.local.` service type, and not `_device-info._tcp.local.` (which is expected to be accompanied with a TXT record containing... device information).

## Solution

Change the device type broadcast by yunomdns, add a small factorization, and add the dummy port 9 in the service information.

## PR Status
Ready

### Code TODOs

 - [x] Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.): N/A
 - [x] Write data or settings migration: N/A
   - [x] Add some missing translations keys: N/A, yunomdns has no user interface
   - [x] Update/write unit tests: N/A
 - [x] Update/write documentation: no documentation about the service type used by yunomdns

### Tests TODOs

## How to test
- patch your local service
- empty your mdns cache
- check that `avahi-browse -a`, or `ping yunohost.local` still work.


bin/yunomdns
```diff
@@ -169,11 +170,11 @@ def main() -> bool:
...
                 ServiceInfo(
-                    type_="_device-info._tcp.local.",
-                    name=f"{interface}: {d_domain}._device-info._tcp.local.",
-                    parsed_addresses=ips,
-                    port=80,
+                    type_=servicetype,
+                    name=f"{d_domain}.{servicetype}",
                     server=f"{d}.",
+                    parsed_addresses=ips,
+                    port=9,
```
zamentur, line 177: Should this port be open on firewall ?

tituspijean: No, it's used for WoL and the Discard Protocol, clients are not expected to actually use it. I'll check if the function accepts removing this parameter, but I don't think so.

tituspijean:
@zamentur OK to merge now?

PR #2314, opened 2026-06-20
fix : exclude second port used by app

### LINKED ISSUE #2404: Edge case in port availability check
### Describe the bug
There is an edge case issue in the port-availability check, performed when updating an application, that can prevent user from updating apps on their environments.

The `PortsResource`'s function `provision_or_update()` performs a check to see if the ports required by an application are being used by any other service/app in the system. This check is performed  via the `port_is_used()`[method](https://github.com/YunoHost/yunohost/blob/6aa9d0537284406c91687b63403789a339741776/src/utils/resources.py#L1273), which considers two conditions:
- the port is listed in the `ss --numeric --listening --tcp --udp
` output
- the port is listed inside at least one settings.yml file in any directory under the /etc/yunohost/apps/ folder. (`f"grep --quiet --extended-regexp \"port: '?{port}'?\" /etc/yunohost/apps/*/settings.yml"`)

This means that the latter check will always fail when users are upgrading an app whose new version requires to open a port that was already listed in the app's own `settings.yml`, even if they correctly remove the port from the firewall's configuration.

That's the case, for instance, for AdGuard:
```
$ python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> port = 853
>>> os.system(f"grep --extended-regexp \"port: '?{port}'?\" /etc/yunohost/apps/*/settings.yml")
/etc/yunohost/apps/adguardhome/settings.yml:adguard_DoT_port: '853'
0
```

A possible workaround would be to exclude the app's directory from the second check listed above, using a command like this:
`os.system(f"grep --extended-regexp \"port: '?{port}'?\" /etc/yunohost/apps/*/settings.yml | grep -v --extended-regexp '^/etc/yunohost/apps/{appname}/'")`

```
>>> appname = 'adguardhome'
>>> os.system(f"grep --extended-regexp \"port: '?{port}'?\" /etc/yunohost/apps/*/settings.yml | grep -v --extended-regexp '^/etc/yunohost/apps/{appname}/'")
256
```

I found out this by trying to upgrade AdGuard Home on my server, and the same issue was experienced by [other users as well](https://forum.yunohost.org/t/adguard-home-update-failed/29758).

### Context

- Hardware: *Fujitsu Esprimo Q920*
- YunoHost version:  11.2.12
- I have access to my server: *Through SSH | through the webadmin *
- Are you in a special context or did you perform some particular tweaking on your YunoHost instance?: *no*

### Logs
[https://paste.yunohost.org/raw/esokoyilez](https://paste.yunohost.org/raw/esokoyilez)

frju365:
## The problem

If an app have a second port, when it upgrades the apps, it will get the second port used by the app. So, it won't allow the user to upgrade application and will consider the port is "used by anpther app".

**Related issues:** : 
* fixes Yunohost/issues#2404 
* https://github.com/YunoHost-Apps/adguardhome_ynh/issues/180

## Solution

Exclude port "greped" from setting of the app itself.

**AI transparency:** *If AI has been used, explain how. See https://doc.yunohost.org/en/dev/?persistLocale=true#%EF%B8%8F-develop*

## PR Status
*Tested partially*
Tested with AdguardHome from 0.107.48~ynh2 to now.

### Code TODOs

 - [x] Discuss about this change with others contributors (on chat, contributors meeting, forum, issues or pr)
 - [x] Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)
 - [x] Write data or settings migration
   - [x] Add some missing translations keys
   - [x] Update/write unit tests
 - [x] Update/write documentation

### Tests TODOs

## How to test
Modify file : https://github.com/YunoHost-Apps/adguardhome_ynh/issues/180
Install adguardhome version 0.107.48~ynh2
Upgrade it.

### DIFF
--- a/src/utils/resources.py
+++ b/src/utils/resources.py
@@ -1399,7 +1399,7 @@ class PortsResource(AppResource):
         # This second command is mean to cover (most) case where an app is using a port yet ain't currently using it for some reason (typically service ain't up)
         used_by_app = (
             os.system(
-                f"grep --quiet --extended-regexp \"port: '?{port}'?\" /etc/yunohost/apps/*/settings.yml"
+                f"grep --quiet --extended-regexp \"port: '?{port}'?\" --exclude=/etc/yunohost/apps/{self.app}/settings.yml /etc/yunohost/apps/*/settings.yml"
             )
             == 0
         )


tituspijean (approved):

zamentur:
Thanks for your fix

PR #2316, opened 2026-06-30
Handle decoding errors in subprocess output

MayeulC:
I have encountered decoding errors making restore operations crash. This change replaces invalid characters with the utf-8 replacement character "�" instead of crashing.

## The problem
Here is a failed restore operation:
https://paste.yunohost.org/raw/uponuherog

Edit: since the problem is in the logging machinery, it does not appear in the logs. This is what I had in the console:
```python-traceback
Info : Skipping /home/yunohost.app/nextcloud__2 which doesn't exists in the archive, probably because restoring from a safety-backup-before-upgrade
Info : [#####++.............] > Restoring the MySQL database...
Exception in thread Thread-8:
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
Attention : --------------
Attention : INSERT INTO `memories_planet_geometry` VALUES
    self.run()
  File "/usr/lib/python3/dist-packages/yunohost/utils/process.py", line 157, in run
    self.queue.put((self.log_callback, line.decode("utf-8").strip("\n")))
                                       ^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd5 in position 66: invalid continuation byte

^CErreur : Impossible de restaurer nextcloud__2 : L'opération a-t-elle été interrompue manuellement ?
Info : L'opération 'Restaurer 'nextcloud__2' depuis une sauvegarde' a échoué ! Pour obtenir de l'aide, merci de partager le journal de l'opération en utilisant la commande 'yunohost log share 20260630-121810-backup_restore_app-nextcloud__2'
Info : Suppression de nextcloud__2…
```

I interrupted it after multiple hours (visible ^C) as there was no visible progress (I imagine that a pipe was left open).

I would like to note that this isn't the first time I hit this error.

## Solution
I do not think that it is critical to leave the output intact, as 

**AI transparency:** No AI was used; I just let copilot suggest a commit title.

## PR Status
Fully tested: Restore operation was a success after I applied this change locally.

I did this PR via the web UI, but I can edit a local clone if you need some changes, such as some tests. They might come in a few days though.

### Code TODOs

 - [X] (N/A) Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)
 - [X] (N/A) Write data or settings migration
   - [X] (N/A) Add some missing translations keys
   - [ ] Update/write unit tests -> I guess I can do that

### Tests TODOs
 - [X] Test change on configuration on an instance

## How to test

I suppose that the python function can be easily tested by running a command that produces invalid utf-8 output, such as `echo -e 'invalid\x89utf-8'`


MayeulC:
I would also like confirmation that this decoding only happens before printing/saving the logs: I hope I didn't mess up the values written to the database?

zamentur (approved):
LGTM (made some manual test with python and bash utf-8 string + search for a way to use that in a malicious way, i found nothing)

PR #2319, opened 2026-07-10
[fix] Broken OptionType comparisons in config panel (is -> ==)

### LINKED ISSUE #2753: Trixie: config panel option type is a string instead of enum type
### Describe the bug

While testing the vpnclient for trixie, I noticed there was a bug with the config panel.

When I run `yunohost app config get vpnclient`, I get the following error:
```
Error: Config panel question 'config_file' should be initialized with a value during install or upgrade.
```

Looking at the code, I noticed that this was coming from this line: https://github.com/YunoHost/yunohost/blob/trixie/src/utils/configpanel.py#L813

So I check the value of `option.type` and `OptionType.file`, and the issue is that `option.type` is a string while `OptionType` is an enum. Thus, this condition fails when it shouldn't : https://github.com/YunoHost/yunohost/blob/trixie/src/utils/configpanel.py#L809

I double checked by replacing `option.type is OptionType.file` with `option.type == OptionType.file` and the error is gone.

Now, I don't know why in bookworm `option.type` is an enum and in trixie it became a string…

### Context

- Hardware: Incus container
- YunoHost version: 13.0.4
- I have access to my server: Through SSH
- Are you in a special context or did you perform some particular tweaking on your YunoHost instance?: no
- Using, or trying to install package version/branch: vpnclient on branch trixie : https://github.com/YunoHost-Apps/vpnclient_ynh/pull/158

ajoe:
## The problem

Fixes YunoHost/issues#2753

Since `BaseOption` uses `use_enum_values = True`, pydantic stores `option.type` as the plain string value, so the three identity comparisons against `OptionType` enum members in `src/utils/configpanel.py` are always `False` on YunoHost 13:

- `option.type is OptionType.file` in `_get_partial_raw_settings_and_mutate_config()`: the intended exemption for `file` questions never triggers, so **any app config panel containing a `file` question whose bound file does not exist yet fails to load** with `Config panel question 'X' should be initialized with a value during install or upgrade.` — as reported in the issue for `vpnclient`. I hit the exact same thing with `wireguard_client` on 13.0.4: adding an interface via the panel works once, and afterwards the panel can no longer be opened because the interface's `.conf` file hasn't been uploaded yet.
- `option.type is OptionType.password` in `_get_config()`: password values are **no longer masked** in `config get` output.
- `option.type is OptionType.button` in `ConfigPanelModel.get_options_with_trigger()`: button options are no longer yielded for their `action` trigger.

## Solution

Use `==` instead of `is`. Since `OptionType` is a `str`-based Enum, equality works whether the stored value is a plain `str` (current behaviour with `use_enum_values`) or an actual enum member.

Verified on a YunoHost 13.0.4 instance: with this change the `wireguard_client` config panel loads again.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/utils/configpanel.py
zamentur:
```suggestion
                    if option.type != OptionType.button or option.id == action_id
```

zamentur (approved):
LGTM

PR #2320, opened 2026-07-11
fix(helpers): app_shell parsing Environment options

### LINKED ISSUE #2818: env_var combines unit file variables in `app shell`
When jumping into an app shell with `yunohost app shell <app>`, because of the way variables are exported from the unit file, if there are multiple lines they get merged into a single variable which impairs the shell functionality of some apps (eg: https://github.com/YunoHost-Apps/mollysocket_ynh/issues/13)

### To reproduce
- `yunohost app shell <app>`
- `printenv <ENV1>`
- shows contents of all envs

### Expected behavior
- `yunohost app shell <app>`
- `printenv <ENV1>`
- shows contents of only env1

could

`[ -n "${env_var:-}" ] && export "${env_var?}"` [(helpers.v2.1.d/0-utils, line 455)](https://github.com/YunoHost/yunohost/blob/3a5f8bac8301c450897b96cbd43a4c7d3ba750fb/helpers/helpers.v2.1.d/0-utils#L455)

be changed to

`[ -n "${env_var:-}" ] && eval export ${env_var?}`

(assuming there aren't any apps that employ spaces in a single env var in the unit file)

tituspijean:
## The problem
The app shell helper assumes only one Environment option is present in a systemd unit file. Additionally, characters like spaces and quotes might break the current way the variables are escaped.

**Related issues:**
Fixes https://github.com/YunoHost/issues/issues/2818

**Related PRs:**

## Solution

Thanks to @DeMiro5001, who pointed to the `shlex` lib in Python to properly handle the `systemctl show` output format.

**AI transparency:** *AI has been used to get a generic example of how to use shlex, and to discover how to properly escape variables with `!r` in Python.*

## PR Status
*Fully tested & Ready*

### Code TODOs

 - [ ] ~~Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)~~
 - [ ] ~~Write data or settings migration~~
   - [ ] Update/write unit tests
 - [ ] ~~Update/write documentation~~

### Tests TODOs

## How to test
Install umami or mollysocket, who use multiple Environment options in their systemd file, then:
```
yunohost app shell <app>
printenv
# output should not show variables leaking out on the next line or other barbarisms
```
...


tituspijean:
@zamentur, you seemed wary about using `eval` in the original issue. Is that implementation OK?

DeMiro5001:
The docs of shlex say
> Changed in version 3.12: Passing None for s argument now raises an exception, rather than reading sys.stdin.

Did you test apps that do not have environment variables? I mean when "$env_var" is empty

tituspijean:
> Did you test apps that do not have environment variables? I mean when "$env_var" is empty

Looks ok, no warnings displayed:

```
 $ systemctl show "adguardhome.service" -p "Environment" --value

 $ yunohost app shell adguardhome                               
adguardhome@server:~$ 
```

zamentur:
> @zamentur, you seemed wary about using `eval` in the original issue. Is that implementation OK?

I think eval should always be avoided, it's too risky to have an injection in an eval that can execute kind of code, even with a lot of care... In more the Environment var is dependent of 1000+ repo we can not check manually.

If a solution is possible without eval, it should be implemented...

DeMiro5001:
We can use https://www.freedesktop.org/software/systemd/man/latest/systemd-run.html
(from what I understood, it gets env vars from systemd)
For scanning repos : 
- https://github.com/semgrep/semgrep
- https://github.com/gitleaks/gitleaks

I did have a fast look, these might be helpful.

tituspijean:
`systemd-run` seemed overly complex to bend to our use case (one main issue was that it would create the app shell based on the current shell, we would have needed to reimport all variables).

I kept the current design and reshaped the loop to avoid using `eval`.

zamentur (approved):
LGTM

PR #2323, opened 2026-07-16
Fix typo in permissions helpers documentation

oleole39:
Only docs of `ynh_permission_url` and `ynh_permission_update` are affected among permissions helpers.

I have not actually tested the commands, but it looks quite obvious given the doc of other permissions helpers both in v1 and v2.1.

Also the "wrong" format is used in `utils` helpers v1 (but already fixed in v2.x). But apparently it also works as such as per https://github.com/YunoHost/yunohost/pull/2323#issuecomment-4993095831). 

it might be that the `=` is actually optional in bash, but maybe harmonizing things could avoid confusion or future bugs. Not such a big deal though eventually.


Salamandar (approved):

Salamandar:
I wouldn't change the calls though, it works as is. But OK for the doc.

oleole39:
- `ynh_permission_url --permission "whatever"` format seems to be used in 10 apps repos, 2 of which are not archived: https://github.com/search?q=org%3AYunoHost-Apps%20%22--permission%20%5C%22%22&type=code
- `ynh_permission_url --permission="whatever"` format seems to be used in a lot of apps repos: https://github.com/search?q=org%3AYunoHost-Apps+%22--permission%3D%5C%22%22&type=code

oleole39:
> I wouldn't change the calls though, it works as is. But OK for the doc.

Thank you, I was wondering whether it was actually working as such.
 
As you wish, somehow there isn't much to adapt in order to harmonize and avoid confusion: only `utils` in v1 helpers (done in this PR) and 2 apps (yet to be done, cf. https://github.com/YunoHost/yunohost/pull/2323#issuecomment-4993148435), but I don't mind too much.

zamentur:
Thanks for your help

PR #2330, opened 2026-07-31
[enh] Antibot with proof of work

zamentur:
## The problem
In the user self registration PR, the antibot math and time mechanism seems too simple for 2026 (and 2006 too).

**Related issues:** 
**Related PRs:** https://github.com/YunoHost/yunohost-portal/pull/44

## The solution
Here i suggest to integrate a basic proof of work mechanism without adding another extra dependencies (i have firstly search for self-hosted captcha and if there was a standard API...).

So, for an attacker it means:
 - automate the attack with a browser automation (be slow and pay for energy)
 - or succeed to craft a countermeasure with an other language (like go)

This new mechanism run when people are filling the form, and if needed finish the computation before to send the registration request.

It asks for webclient to bruteforce 10 sha256 hashes specially crafted. Currently, each hash is made from a range of 100 000 combination (salted with the challenge token), so client has to generate about 500 000 hash to find the 10 hashes...

On the portal, it uses 5 workers to speed the computation. On my core i5 it takes ~12s to solve the 10 hash (and 36s with a mono thread approach), but it should be tested on smartphone or older cpu. Some self hosted captcha solution use a wasm to speed the operation (it could be a future improvement). I guess we can optimize by varying those params (number of hashes, hash complexity, number of workers).

If the user finish to fill and save before the end of the computation, the save button display the load indicator until the computation is finished and the form sent.

An improvement could be to increase difficulty if there are a lot of requests (on a per IP basis ?), a lot pending requests, or a lot of challenge pending.

About energy, i consider registration form of yunohost should not be use so often and for me the energy cost should not be so high in normal context. Of course, if an attacker decide to try to spam and compute the hash, it consume energy on attacker side.

**AI transparency:** no use but i took some idea from https://github.com/a-ve/wicketkeeper/blob/develop/client/src/solvers/fast.js

## PR Status
*Tested partially*

### Code TODOs
 - [ ] use `navigator.hardwareConcurrency || 1` to deduce the good number of worker
 - [ ] Add some comments
 - [ ] Increase hash complexity based on some stress indicators
 - [x] Update related repo yunohost-portal
 - ~~[ ] Write data or settings migration~~
   - [ ] Write unit tests
 - ~~[ ] Update/write documentation~~

### Tests TODOs
 - [x] Run the code from a yunohost protal run with docker compose (it's important cause the URL in http should be localhost
 - [ ] Check that hash compute ir rerun with new challenge if the first sumition failed due to challenge. Indeed the old challenge has been invalidating.

## How to test
Run the code with a yunohost portal from docker and localhost url. In http you have to run it with `localhost` or `127.0.0.1`

### DIFF
--- a/src/portal.py
+++ b/src/portal.py
@@ -463,7 +463,7 @@ def portal_invitation_consume(token, username, fullname, password, external_emai
 
 # FIXME : this is probably not a proper global state shared between all the gevent/bottle threads
 # Though for now it looks like we have a single thread anyway so it may be OK ?
-CHALLENGES: dict[str, tuple[int, int, list[str]]] = dict()
+CHALLENGES: dict[str, tuple[int, int, str]] = dict()
 
 
 def _generate_antibot_challenge() -> tuple[str, str, str]:
--- a/src/portal.py
+++ b/src/portal.py
@@ -463,13 +463,15 @@ def portal_invitation_consume(token, username, fullname, password, external_emai
 
 # FIXME : this is probably not a proper global state shared between all the gevent/bottle threads
 # Though for now it looks like we have a single thread anyway so it may be OK ?
-CHALLENGES: dict[str, tuple[int, int]] = dict()
+CHALLENGES: dict[str, tuple[int, int, list[str]]] = dict()
 
 
-def _generate_antibot_challenge() -> tuple[str, str]:
+def _generate_antibot_challenge() -> tuple[str, str, str]:
 
-    from random import randint, choice
+    from secrets import randbelow, choice
     from .utils.misc import random_ascii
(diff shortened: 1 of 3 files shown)


src/portal.py
```diff
@@ -488,13 +490,23 @@ def _generate_antibot_challenge() -> tuple[str, str]:
 
     answer = CALCULATIONS[operator](x, y)
 
+    # Create a Proof of Work challenge
+    # Force bots to bruteforce 10 hashes
```
zamentur, line 494:
```suggestion
    # Force bots to compute about 500 000 hashes (bruteforce)
```

PR #2332, opened 2026-08-01
[fix] Bad regex + fullname support

zamentur:
## The problem

-  regex use to check optional values in user_import are very very bad:  `^|(something)$` instead of `(something)?$`
-  firstname and lastname are still used... and not properly strip

**Related issues:**
**Related PRs:**

## Solution

 - Fix regex
 - Use fullname and strip it

**AI transparency:** no ai

## PR Status
Tested on cli

### Tests TODOs
 - [ ] Tests on webadmin ?

## How to test
...

### DIFF
--- a/tests/test_user-group.py
+++ b/tests/test_user-group.py
@@ -143,8 +143,7 @@ def test_import_user():
 
     fieldnames = [
         "username",
-        "firstname",
-        "lastname",
+        "fullname",
         "password",
         "mailbox-quota",
         "mail",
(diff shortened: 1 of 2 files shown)


tests/test_user-group.py
```diff
@@ -158,8 +157,7 @@ def test_import_user():
         writer.writerow(
             {
                 "username": "morgan-claude.good_7",
-                "firstname": "Morgan-Claude",
-                "lastname": "Good",
+                "firstname": "Morgan-Claude Good",
```
tituspijean, line 160:
```suggestion
                "fullname": "Morgan-Claude Good",
```

Just having a glance at the whole thing, why not "fullname" here and in the next modified lines?

PR #2339, opened 2026-08-05
Move migration 0037_upgrade_dkim_keys to 0036_dkim_keys_upgrade

Salamandar:
This is to ensure a prettier order in the migration webadmin


tituspijean:
Discussed during PRLanta, and tested in production. Having conflicting migration numbers is not a problem, and it enables us to show it at the same level as the future Trixie migration.

tituspijean (approved):

tituspijean:
The re-numbering needs to be done on the `trixie` branch as well.

PR #2340, opened 2026-08-05
enh(postfix): allow auth with email

tituspijean:
## The problem

Fixes https://github.com/YunoHost/issues/issues/2722

## Solution

Include `uid` and `mail` in the search query, and let output them both in the `result_attribute`s.
I think it only works on the main email address.

## PR Status
Ready

### Code TODOs

*Here are frequently forgotten tasks:*
 - [x] Discuss about this change with others contributors (on chat, contributors meeting, forum, issues or pr): PRLanta 2026-08-05
 - [ ] ~~Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)~~
 - [ ] ~~Write data or settings migration~~
 - [ ] ~~Pass Continuous Integration checks~~
   - [ ] ~~Add some missing translations keys~~
   - [ ] Update/write unit tests: ?
 - [ ] Update/write documentation: https://doc.yunohost.org/en/admin/email/clients

### Tests TODOs

Tested on YNH v13.0.5
Untested on YNH 12

## How to test

Apply patch, restart postfix, try logging in with either username or email address.

PR #2341, opened 2026-08-06
Mail trixie

### LINKED ISSUE #2722: Postfix does not accept authentication with username@domain.tld
### Describe the bug

Hello,
I know that the page https://doc.yunohost.org/en/admin/email/clients states that the ``username`` should be used without the ``domain.tld`` part, but it indeed does support ``username@domain.tld`` on YunoHost 12, but not on YunoHost 13. Not sure if it's a regression or an expected feature.

Works on YunoHost 13:

Image

Don't work on YunoHost 13 (but does work on YunoHost 12):

Image

### Context

- Hardware: VPS bought online
- YunoHost version: 13.0.1
- I have access to my server: Through SSH | through the webadmin | direct access via keyboard / screen
- Are you in a special context or did you perform some particular tweaking on your YunoHost instance?: no
  - If yes, please explain:
- Using, or trying to install package version/branch:
- If upgrading, current YunoHost version: 

### Logs

```
$ tail -f /var/log/mail.log
2025-11-22T16:27:26.396196+00:00 nethouse postfix/submission/smtpd[1226]: NOQUEUE: reject: RCPT from 189-30-206-91.user3p.v-tal.net.br[189.30.206.91]: 553 5.7.1 <eduardomozart@domain.tld>: Sender address rejected: not owned by user eduardomozart@domain.tld; from=<eduardomozart@domain.tld> to=<eduardomozart@domain.tld> proto=ESMTP helo=<EDUARDOMOZAFB9E> sasl_method=LOGIN sasl_username=eduardomozart@domain.tld
```

### LINKED ISSUE #2878: Trixie: pop3 not supported
### Describe the bug

Issue found by @Thatoo while qualifying Trixie:

Original notes in French (Engilsh version below)

> - [Thatoo][🚫 KO] Se connecter via POP3 (ports 110 et 995)
>     - Activer POP3 n'ouvre pas les ports automatiquement dans le pare-feu
>     - Même en ouvrant les ports manuellement, POP3 ne fonctionne pas (telnet répond "can't connect to remote host: connection refused")
>     - `sudo ss -tunlp` ne montre aucun service écoutant sur les ports 110 ni 995

Translation in English by myself:
> - [Thatoo][🚫 KO] Connect through POP3 (ports 110 and 995)
>     - Enabling POP3 does not open the ports automatically in the firewall
>     - Even when opening the ports manually, POP3 does not work (telnet responds "can't connect to remote host: connection refused")
>     - `sudo ss -tunlp` does not show any service listening on port 110 nor on 995

NOTE: I have not tried to reproduce it myself, I open this issue so it is not forgotten.

### Context

ynh-dev

### LINKED ISSUE #2879: Trixie: mail quotas does not work
### Describe the bug

@Thatoo reported that email quotas do not work.

And indeed, I could reproduce the issue on my side too.

### Context

Both on a regular server and on ynh-dev

### To reproduce

- Create a user
- Set them a mail quota to 1MiB
- Send them an attachment of 2MiB
- Send them again the same attachment, it succeeds (surprisingly!)
- check the user info and check the email, you've got a percentage of use set to `0%` instead of >100% : 
```
fullname: ...
loginShell: /bin/bash
mail: ...
mail-aliases: 
mail-forward: 
mailbox-quota: 
  limit: 1M
  use: 5.3M (0%)
username: ...
```
- you may check the command run by  `doveadm -f flow quota get -u <USERNAME>`, it returns 0%:
```
Quota name=User quota Type=STORAGE Value=5574 Limit=- %=0
Quota name=User quota Type=MESSAGE Value=16 Limit=- %=0
```

### Expected behavior

1. The percentage of use / quota should be strictly positive in this case
2. (I think) an error should occur to the sender when the quota is exceeded.

Josue-T:
## The problem

**Related issues:**
Closes https://github.com/YunoHost/issues/issues/2722
 
**Related PRs:**
#2299
Supersedes #2340

## Solution

Cherry-pick and adapt the changes form #2299 and test it.

I was quite confused because in dovecot the config was consistent, we where only checking for the username and not the email, but on postfix side it was the inverse, the postfix ldap config is only searching for email.

After doing more analysis I discovered that in postfix config there are some line of configuration which handle the authentication thought dovecot. That's why postfix was also only allowing username.

**AI transparency:**  

Not used at all

## PR Status

*Fully tested / Ready*

#### Setup

yunohost accounts which was created:

```yaml
test-mail-sender:
    fullname: Testmail Sender
    mail: test-mail-sender@domain.tld
    mail-aliases: 
    mail-forward: 
    username: test-mail-sender
test-mail-usern-main:
    fullname: Testmail Username-main
    mail: test-mail-usern-main@domain.tld
    mail-aliases: 
    - alias@c.domain.tld
    - alias@domain.tld
    mail-forward: 
    username: test-mail-usern-main
test-mail-usern-odom:
    fullname: Testmail Username-otherdom
    mail: test-mail-usern-odom@a.domain.tld
    mail-aliases: 
    mail-forward: 
    username: test-mail-usern-odom
test-mail-other-other:
    fullname: Testmail Otherother
    mail: mainmail@b.domain.tld
    mail-aliases: 
    mail-forward: 
    username: test-mail-other-other
test-mail-fw-main:
    fullname: Testmail Forward-main
    mail: test-mail-fw-main@domain.tld
    mail-aliases: 
    mail-forward: dst-fw@c.domain.tld
    username: test-mail-fw-main
test-mail-fw-alias:
    fullname: Testmail Fworard-alias
    mail: fw-alias@a.domain.tld
    mail-aliases: proxy@b.domain.tld
    mail-forward: dst-fw@c.domain.tld
    username: test-mail-fw-alias
test-mail-dst-fw:
    fullname: Testmail Dest-for
    mail: dst-fw@c.domain.tld
    mail-aliases: 
    mail-forward: 
    username: test-mail-dst-fw
test-mail-fw-to-group
    mail: fw-to-group@a.domain.tld
    mail-forward: mailgroupalias@domain.tld
    
mailgroupa:
    mail-aliases: 
    - groupa@b.domain.tld
    - mailgroupa@domain.tld
    - mailgroupa@b.domain.tld
    members: 
    - test-mail-usern-main
    - test-mail-usern-odom
mailgroup-alias:
    mail-aliases: mailgroupalias@domain.tld
    members: test-mail-fw-main
mailgroup-fw:
    mail-aliases: mailgroupfw@b.domain.tld
    members: test-mail-fw-to-group
``` 

##### Login SMTP with username:

Tested with raw text exchange:

```bash
openssl s_client -starttls smtp -connect domain.tld:587

EHLO myhost.example.com

# with username
AUTH PLAIN $(echo -en '\0testmailauth\0asdf7z2af2' | base64)

# with mail
AUTH PLAIN $(echo -en '\0testauth@domain.tld\0asdf7z2af2' | base64)
```

Tested for users:
- [x] test-mail-usern-main 
- [x] test-mail-usern-main@domain.tld
- [x] test-mail-usern-odom 
- [x] test-mail-usern-odom@a.domain.tld
- [x] test-mail-other-other
- [x] mainmail@b.domain.tld

##### Login IMAP and POP with username:

Tested with raw text exchange:

```bash
openssl s_client -crlf -connect domain.tld:993

a LOGIN testmailauth asdf7z2af2

a LOGIN testauth@domain.tld asdf7z2af2
```

```bash
openssl s_client -crlf -connect domain.tld:995

# with username
USER testmailauth 
PASS asdf7z2af2

# with mail
USER testauth@domain.tld 
PASS asdf7z2af2
```

Tested for users:

- [x] test-mail-usern-main 
- [x] test-mail-usern-main@domain.tld
- [x] test-mail-usern-odom 
- [x] test-mail-usern-odom@a.domain.tld
- [x] test-mail-other-other
- [x] mainmail@b.domain.tld

From local instance (to local instance):

- [x] ➡️ main mail:
    - [x] username @ main domain      (test-mail-usern-main / test-mail-usern-main@domain.tld)
    - [x] username @ other domain     (test-mail-usern-odom / test-mail-usern-odom@a.domain.tld)
    - [x] != username @ other domain  (test-mail-other-other / mainmail@b.domain.tld)
- [x] ➡️ mail alias
    - alias@c.domain.tld
    - alias@domain.tld
- [x] ➡️ other account (main mail) with mail forward -> main mail       (test-mail-fw-main@domain.tld)
- [x] ➡️ other account (mail alias) with mail forward -> mail alias     (proxy@b.domain.tld)
- [x] ➡️ group
    - [x] groupname @ main domain      (... / mailgroupa@domain.tld)
        - should arrive to the mailbox of:
            - test-mail-usern-main
            - test-mail-usern-odom
    - [x] groupname @ other domain     (... / mailgroupa@b.domain.tld)
        - should arrive to the mailbox of:
            - test-mail-usern-main
            - test-mail-usern-odom
    - [x] != groupname @ other domain  (... / groupa@b.domain.tld)
        - should arrive to the mailbox of:
            - test-mail-usern-main
            - test-mail-usern-odom
- [x] ➡️ group -> other account with mail forward -> main mail       (mailgroupalias@domain.tld -> test-mail-dst-fw)
- [x] ➡️ group -> other account with mail forward -> other group -> final user (mailgroupfw@b.domain.tld -> test-mail-dst-fw)

From external mail server to Yunohost instance: 

- [x] ➡️ main mail (match username)
    - [x] match username + main domain
    - [x] match username other domain
    - [x] different than username + other domain
- [x] ➡️ mail alias
- [x] ➡️ other account (main mail) with mail forward -> main mail
- [x] ➡️ other account (mail alias) with mail forward -> mail alias

- [x] From internal instance to external instance


tituspijean (approved):

tituspijean (changes requested):
There might actually be a bug.
Setup:
- user1 on domain1.tld
- user2 on domain2.tld

If user1 sends an email to user2, I get an "Undelivered Mail Returned to Sender" email with:

> <user2@domain1.tld> (expanded from <user2@domain2.tld>: user unknown

Josue-T:
Note, I'm working to find a fix for this.

Josue-T:
So well the issue here is that, currently we was mixing email and username. And the main case that we are mixing uid is in the ldap `maildrop` field. The first entry is the username. And as soon as we try to separate things we have some issue.

To explain, let's take this example:
* `user-a` can have a main email `info@domain-a.tld`
* `user-b` have the email `info@domain-b.tld`

Let's consider that the main domain is `domain-a.tld`.

We we try to send email to  `info@domain-b.tld` what happen is:
- Postfix receive a mail.
- Postfix will try to query LDAP to know where to send this email. This config will be used
```
query_filter = (&(objectClass=mailAccount)(mail=%s)(permission=cn=mail.main,ou=permission,dc=yunohost,dc=org))
result_attribute = maildrop
```
- So the resulting of this will be `user-b` because the maildrop of `user-b` is his username.
- Now postfix when he is receiving a answer like this, it will add the main domain, so it will be `domain-a.tld`.
- So postfix will try to send to dovecot the mail to a mailbox of an email `user-b@domain-a.tld` which don't exist at all.

So the fix to me is to fix the ldap maildrop field. We should put the user mail instead of the username. I tested locally and it seem fixing the issue.

Unfortunately for this we will need to create a migration to fix this...

DeMiro5001:
You probably already know this but I share it with you. 

https://web.mit.edu/netbsd/src/gnu/dist/postfix/html/LDAP_README.html

Sorry, not enough knowledge about it for now

Josue-T:
Pushed a fix and a migration. Tested really quickly, seem working better. I'll try to test it more deeply ASAP. Any feedback would be welcome.

Josue-T:
There is some security implication because of [this](https://matrix.to/#/!GgYiGWJbdhBZSAOnDq:matrix.org/$jKjw9RFgzej-3WTMDahG3PuxdqSny9GVxj-MJhwSrSk?via=matrix.org&via=tille.ch&via=matrix.skeptikon.fr).

Josue-T:
Tested deeply (see the description), discovered one minor bug that was fixed by a8b05ec. Now I think it should be working as expected.

src/migrations/m0040_migrate_ldap_maildrop_filed.py
tituspijean, line 1: :warning: Migration number conflicts with the one in https://github.com/YunoHost/yunohost/pull/2348

Josue-T: Fixed by da739f3de

tituspijean (approved):
My reported bug above is now resolved.

conf/dovecot/dovecot.conf.j2
```diff
@@ -60,11 +60,10 @@ passdb_default_password_scheme = SSHA
...
 passdb ldap {
   # bind = yes
-  filter = (&(objectClass=inetOrgPerson)(uid=%{user | username })(permission=cn=mail.main,ou=permission,dc=yunohost,dc=org))
-  #fields {
-  #  user = %{ldap:uid}
-  #  password = %{ldap:userPassword}
-  #}
+  filter = (&(objectClass=inetOrgPerson)(|(uid=%{user})(mail=%{user}))(permission=cn=mail.main,ou=permission,dc=yunohost,dc=org))
+  fields {
+    user = %{ldap:uid}
```
zamentur, line 65:
```suggestion
    user = %{ldap:uid}
    password = %{ldap:userPassword}
```
Not sure of this suggestion but dovecot doc says : 

> This maps the LDAP "uid" attribute to Dovecot's "user" field and LDAP's "userPassword" attribute to Dovecot's "password" field. These two fields should always be returned, but it's also possible to return other special [passdb: Extra Fields](https://doc.dovecot.org/main/core/config/auth/passdb.html#extra-fields).

https://doc.dovecot.org/main/core/config/auth/databases/ldap.html#dovecot-configuration

zamentur: I just understand that it's under a section "Password Lookups" an alternative to "Authentication binds". This Password lookup method is faster, but it seems it needs an admin access to the ldap...

SO i guess it's ok, i close

src/migrations/m0040_migrate_ldap_maildrop_filed.py
```diff
@@ -0,0 +1,56 @@
...
+                "uid",
+                "mail",
+                "maildrop"
+            ],
+        )
+
+        for infos in users_infos:
+            username = infos['uid'][0]
+            logger.debug("Migrating maildrop field for user %s", username)
+            new_maildrop = [infos['mail'][0]] + infos["maildrop"][1:]
```
zamentur: It might conflict with nextcloud config panel option `mail_forward` https://github.com/YunoHost-Apps/nextcloud_ynh/blob/bcba5c3e41f1641c982d0240a72fede6b763bb7a/config_panel.toml#L50

I don't remember if the ldap mechanims in nextcloud take the last maildrop or the first. Anyway i guess it doesn't take value without `@`...

zamentur: Maybe this line could raise an error, if someone as already fill the same mail in the maildrop ?

src/portal.py
```diff
@@ -242,6 +242,7 @@ def portal_update(
             new_mails = [mail] + new_mails[1:]
 
         new_attr_dict["mail"] = new_mails
+        new_attr_dict["maildrop"] = [mail] + current_user["maildrop"][1:]
```
zamentur: same here

src/user.py
```diff
@@ -527,6 +527,7 @@ def user_update(
 
         user["mail"] = [mail] + user["mail"][1:]
         new_attr_dict["mail"] = user["mail"]
+        new_attr_dict["maildrop"] = [mail] + user["maildrop"][1:]
```
zamentur: same here

src/user.py
zamentur, line 620: This change might break some post_user_update, we should run a `grep YNH_USER_MAILFORWARDS */hooks/post_user_update` to study this

zamentur:
> There is some security implication because of [this](https://matrix.to/#/!GgYiGWJbdhBZSAOnDq:matrix.org/$jKjw9RFgzej-3WTMDahG3PuxdqSny9GVxj-MJhwSrSk?via=matrix.org&via=tille.ch&via=matrix.skeptikon.fr).

Not sure of me, but maybe the `yunohost user group update add-mailalias` feature have the same issue https://github.com/YunoHost/yunohost/blob/da739f3deb8e23fe209d4b0858a61fca219e5f60/conf/postfix/plain/ldap-groups.cf#L6

hooks/conf_regen/25-dovecot
```diff
@@ -34,7 +34,6 @@ do_pre_regen() {
     mkdir -p "${dovecot_dir}/global_script"
 
     # copy simple conf files
-    cp dovecot-ldap.conf "${dovecot_dir}/dovecot-ldap.conf"
```
zamentur, line 37: If i remember well, we need to empty the file to make it deleted by regen-conf...

zamentur:
TODO check implication for the mail-forward user self registration feature

src/migrations/m0040_migrate_ldap_maildrop_filed.py
Josue-T: > Maybe this line could raise an error, if someone as already fill the same mail in the maildrop ?

I'm not sure that this could happen because it means that the user would add a mail forward to his mailbox which would means to end with a loop.

Josue-T: > It might conflict with nextcloud config panel option `mail_forward` https://github.com/YunoHost-Apps/nextcloud_ynh/blob/bcba5c3e41f1641c982d0240a72fede6b763bb7a/config_panel.toml#L50
> 
> I don't remember if the ldap mechanims in nextcloud take the last maildrop or the first. Anyway i guess it doesn't take value without `@`...

Reading the official nextcloud doc:

> LDAP attribute containing the email address (takes first if multiple are stored)

So yes we will have some issue with this.

But my opinion over this is that I think we have quite a wrong design(but yes it's an historical stuff). Having a local address which is redirected to external address as the "standard" to handles user with external address look a bit wrong and more a workaround to a missing feature than a good design. My plan was then to propose a fix for this and to provide the possibly to use external email as the main email, but well it means to create a new PR in addition to the others....

But I think as soon as the main email could be an external one, I think it would fix the issue with nextcloud. Currently this setting is mainly here because the ldap mail field can't be an external email and so we need a workaround to work for theses cases.

Josue-T:
> > There is some security implication because of [this](https://matrix.to/#/!GgYiGWJbdhBZSAOnDq:matrix.org/$jKjw9RFgzej-3WTMDahG3PuxdqSny9GVxj-MJhwSrSk?via=matrix.org&via=tille.ch&via=matrix.skeptikon.fr).
> 
> Not sure of me, but maybe the `yunohost user group update add-mailalias` feature have the same issue
> 
> https://github.com/YunoHost/yunohost/blob/da739f3deb8e23fe209d4b0858a61fca219e5f60/conf/postfix/plain/ldap-groups.cf#L6

It's true, Yes, I need to review also the group alias features.

src/user.py
Josue-T: did a `grep -r YNH_USER_MAILFORWARDS` on all app repos. It retourned nothing.

Josue-T:
> > > There is some security implication because of [this](https://matrix.to/#/!GgYiGWJbdhBZSAOnDq:matrix.org/$jKjw9RFgzej-3WTMDahG3PuxdqSny9GVxj-MJhwSrSk?via=matrix.org&via=tille.ch&via=matrix.skeptikon.fr).
> > 
> > 
> > Not sure of me, but maybe the `yunohost user group update add-mailalias` feature have the same issue
> > https://github.com/YunoHost/yunohost/blob/da739f3deb8e23fe209d4b0858a61fca219e5f60/conf/postfix/plain/ldap-groups.cf#L6
> 
> It's true, Yes, I need to review also the group alias features.

Fix now by f7e2345

src/migrations/m0040_migrate_ldap_maildrop_filed.py
Josue-T: So the conclusion is that we agreed that in mid-term we would like to improve the mail handling to handle external email so a user don't need a have internal mailbox. But even if the postfix/dovecot config is probably mostly done there are still some work to do to make it ready for this feature implementation like:
- Handling the case in the CLI
- Managing internal and external email in the webadmin and improving the UI
- Managing the external email in the portal (with the external email user should not be able to edit the mail forward and the mail alias).

So as this require some work and it's more a feature than a bug fix the idea for now is to implement a workaround done by 0f353b5.

src/migrations/m0040_migrate_ldap_maildrop_filed.py
```diff
@@ -0,0 +1,61 @@
...
+                "uid",
+                "mail",
+                "maildrop"
+            ],
+        )
+
+        for infos in users_infos:
+            username = infos['uid'][0]
+            logger.debug("Migrating maildrop field for user %s", username)
+            # Big fat WARNING: currently we put the user email in the last maildrop entry
```
zamentur:
```suggestion
            # FIXME Big fat WARNING: currently we put the user email in the last maildrop entry
```

src/portal.py
```diff
@@ -171,7 +171,12 @@ def portal_me():
         "fullname": user["cn"][0],
         "mail": user["mail"][0],
         "mailalias": user["mail"][1:],
-        "mailforward": user["maildrop"][1:],
+        # Big fat WARNING: currently we put the user email in the last maildrop entry
```
zamentur:
```suggestion
        # FIXME Big fat WARNING: currently we put the user email in the last maildrop entry
```

src/portal.py
```diff
@@ -242,6 +247,13 @@ def portal_update(
             new_mails = [mail] + new_mails[1:]
 
         new_attr_dict["mail"] = new_mails
+        # Big fat WARNING: currently we put the user email in the last maildrop entry
```
zamentur:
```suggestion
        # FIXME Big fat WARNING: currently we put the user email in the last maildrop entry
```

src/portal.py
```diff
@@ -285,10 +297,15 @@ def portal_update(
         if not is_allowed_to_edit_mail_forward:
             raise YunohostValidationError("mail_edit_operation_unauthorized")
 
-        new_attr_dict["maildrop"] = [current_user["maildrop"][0]] + [
+        # Big fat WARNING: currently we put the user email in the last maildrop entry
```
zamentur:
```suggestion
        # FIXME Big fat WARNING: currently we put the user email in the last maildrop entry
```

src/portal.py
```diff
@@ -337,7 +354,12 @@ def portal_update(
             "fullname": new_attr_dict["cn"],
             "mail": new_attr_dict["mail"][0],
             "mailalias": new_attr_dict["mail"][1:],
-            "mailforward": new_attr_dict["maildrop"][1:],
+            # Big fat WARNING: currently we put the user email in the last maildrop entry
```
zamentur:
```suggestion
            # FIXME Big fat WARNING: currently we put the user email in the last maildrop entry
```

src/user.py
```diff
@@ -527,6 +527,13 @@ def user_update(
 
         user["mail"] = [mail] + user["mail"][1:]
         new_attr_dict["mail"] = user["mail"]
+        # Big fat WARNING: currently we put the user email in the last maildrop entry
```
zamentur:
```suggestion
        # FIXME Big fat WARNING: currently we put the user email in the last maildrop entry
```

Salamandar:
mostly LGTM for me, but the git history / messages is not okay. If that's fine with you i'll rebase the branch to reword / merge commits for a proper history.

conf/postfix/plain/ldap-domains.cf
tituspijean, line 1: From @Salamandar : should we handle the removal and touch for `ldap-domains.cf` in Postfix like we did for `dovecot-ldap.conf`?

src/migrations/m0040_migrate_ldap_maildrop_field.py
tituspijean, line 1: Should it be renamed `src/migrations/m0040_migrate_ldap_maildrop_field.py` ? (instead of `filed`?)

Josue-T: ah yes we should

tituspijean (approved):
Tested on my frankenserver, still working. :+1:

conf/dovecot/dovecot.conf.j2
```diff
@@ -138,6 +141,7 @@ service auth {
   }
 }
 
+# FIXME This is probably broken because /usr/bin/quota-warning.sh seem not existing ony any Yunohost instance
```
tituspijean, line 144: [Dovecot's documentation](https://doc.dovecot.org/main/core/plugins/quota.html#example-configuration) suggests the following script. We might need to tweak it to use the proper `domain.com` and explain better the consequences of exceeding the quota.

```
#!/bin/sh
PERCENT=$1
USER=$2
cat << EOF | /usr/local/libexec/dovecot/dovecot-lda -d $USER -o quota_enforce=no
From: postmaster@domain.com
Subject: quota warning

Your mailbox is now $PERCENT% full.
EOF
```

Josue-T: Thanks for the suggestions, I'll note this in my todo, but I'm not sure about when I'll have time to handle this and also it's not a regression so I'm more in favor to firstly merge this and then I'll provide a fix for this in second step.

zamentur: I created an issue https://github.com/YunoHost/issues/issues/2891 for that, i agree with Josue to merge before

src/user.py
zamentur: ok so it 's solved

src/migrations/m0040_migrate_ldap_maildrop_field.py
```diff
@@ -0,0 +1,61 @@
...
+            [
+                "uid",
+                "mail",
+                "maildrop"
+            ],
+        )
+
+        for infos in users_infos:
+            username = infos['uid'][0]
+            logger.debug("Migrating maildrop field for user %s", username)
```
zamentur, line 50: Just to be sure in case the user force a migration rerun to not remove the first true mail forward...
```suggestion
            if '@' in infos["maildrop"][0]:
                continue
            logger.debug("Migrating maildrop field for user %s", username)
            
```

debian/control
```diff
@@ -28,7 +28,7 @@ Depends: python3-all (>= 3.13),
  , sudo, sssd, sssd-tools, libsss-sudo, sssd-ldap, slapd, ldap-utils, libnss-sss, libpam-sss
  , dnsmasq, resolvconf, libnss-myhostname
  , postfix, postfix-ldap, postfix-policyd-spf-perl, postfix-pcre, libsasl2-modules
- , dovecot-core, dovecot-imapd, dovecot-ldap, dovecot-lmtpd, dovecot-managesieved
+ , dovecot-core, dovecot-imapd, dovecot-pop3d, dovecot-ldap, dovecot-lmtpd, dovecot-managesieved
```
zamentur:
```suggestion
 , dovecot-core, dovecot-imapd, dovecot-ldap, dovecot-lmtpd, dovecot-managesieved
```
`dovecot-pop3d` is installed by the `@post_change_hook("pop3_enabled")` 

And in more it is removed when setting False in it (with `--no-remove` :) )
https://github.com/YunoHost/yunohost/blob/1928e7b31e0d6bea7de4d23cb8dcc559e1f7dd7c/src/settings.py#L374

zamentur (approved):
LGTM
tested on web admin and portal, mail sent with roundcube (and other tests made by titus and josue)

PR #2346, opened 2026-08-19
Fix migration postgresql

fflorent:
## The problem

1. When the postgresql postmigration fails (but the migration to version 17 succeeds), there is no way to run again the steps (i.e. the reindexation)
2. The postmigration to version 17 fails with the error detailed in https://github.com/YunoHost/issues/issues/2851

Fixes https://github.com/YunoHost/issues/issues/2851

## Solution

1. Run the postmigrations only when the cluster in version 17 is installed (and not the one of version 15)
2. Decode the output lines as utf8 strings

## PR Status
*Tested partially*

### Tests TODOs

### DIFF
--- a/src/migrations/postgresql.py
+++ b/src/migrations/postgresql.py
@@ -58,37 +58,42 @@ class PostgreSQLMigration(Migration):
             )
 
         # Make sure there's a 15 cluster
-        try:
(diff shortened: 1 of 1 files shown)


zamentur (approved):
LGTM seems tested

PR #2348, opened 2026-08-19
migrations: Add apt modernize-sources migration

Salamandar:
## The problem

apt.lists -> deb822 apt.sources

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -691,6 +691,7 @@
     "migration_description_0036_migrate_to_trixie": "Upgrade the system to Debian Trixie and YunoHost 13",
     "migration_description_0037_postgresql_15_to_17": "Migrate databases from PostgreSQL 15 to 17",
     "migration_description_0038_rebuild_python_venv_in_trixie": "Repair Python app after trixie migration",
+    "migration_description_0039_apt_modernize_sources": "Migrate Apt sources lists to deb822 format",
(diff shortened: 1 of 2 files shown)


tituspijean:
:warning: Migration number conflicts with the one in https://github.com/YunoHost/yunohost/pull/2341

Tagadda:
Would it make sense to handle this as part of the trixie migration instead?

Thovi98:
> Would it make sense to handle this as part of the trixie migration instead?

If we do so, would this still be available for the people who've already migrated?

zamentur (changes requested):
TODO:
 - [ ] Use deb822 in `ynh_apt_install_dependencies_from_extra_repository` (ideally without having to update all apps...)
 - [ ] Fix `backports_in_sources_list` and testing detection and `bad_sury_packages` in diagnosis
 - [ ] check if `AptDependenciesAppResource` is compatible with the new format
 - [ ] integrate in trixie PR ?
 - [ ] Manage the restore case `def run_after_system_restore(self): self.run()` (when we restore the old format apt conf)
 - [ ] Update SwitchToXXXX https://github.com/YunoHost/install_script/blob/main/switchtoTesting
 - [x] Update trixie install script https://github.com/YunoHost/install_script/blob/e381405f1311a6b84d94381704d820676d8efda7/trixie#L605

PR #2350, opened 2026-08-19
Migrations: postgresql: Rename target cluster instead of deleting it

Salamandar:
## Issue

Immich already creates a cluster at the target version let's not delete it during the migration

## Solution

Rename the existing cluster, to avoid conflicts.

## TODO

Maybe, list databases in the existing cluster, if no databases, actually delete it.

Also, the renamed clusters seems to need collation fixes too, like the main cluster.

## Tests

Tested on my server. It works:

```
salamandar:lib:# pg_lsclusters
Ver Cluster Port Status Owner    Data directory                 Log file
16  main    5433 online postgres /var/lib/postgresql/16/main    /var/log/postgresql/postgresql-16-main.log
17  main    5432 online postgres /var/lib/postgresql/17/main    /var/log/postgresql/postgresql-17-main.log
17  renamed 5434 online postgres /var/lib/postgresql/17/renamed /var/log/postgresql/postgresql-17-renamed.log
```

### DIFF
--- a/src/migrations/m0037_postgresql_15_to_17.py
+++ b/src/migrations/m0037_postgresql_15_to_17.py
@@ -24,7 +24,7 @@ from .postgresql import PostgreSQLMigration
 class MyMigration(PostgreSQLMigration):
     "Migrate DBs from Postgresql 15 to 17 after migrating to Trixie"
 
-    previous_version = "15"
-    target_version = "17"
+    previous_version = 15
+    target_version = 17
 
     dependencies = ["migrate_to_trixie"]
--- a/src/migrations/postgresql.py
+++ b/src/migrations/postgresql.py
@@ -18,6 +18,7 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
+import json
 import os
 import subprocess
 import time
@@ -36,12 +37,12 @@ class PostgreSQLMigration(Migration):
     "Migrate DBs between Postgresql versions after migrating to a new Debian version"
 
     # Provided by calling class
-    previous_version: str
-    target_version: str
+    previous_version: int
+    target_version: int
 
     def run(self):
-        ynh_deps_cmd = 'grep -A10 "ynh-deps" /var/lib/dpkg/status | grep -E "Package:|Depends:" | grep -B1 postgresql'
-        if os.system(ynh_deps_cmd) != 0:
+        ynh_deps_cmd = "grep -A10 'ynh-deps' /var/lib/dpkg/status | grep -E 'Package:|Depends:' | grep -B1 postgresql"
+        if subprocess.run(ynh_deps_cmd, shell=True, stdout=subprocess.DEVNULL, check=False).returncode != 0:
             logger.info("No YunoHost app seem to require postgresql... Skipping!")
             return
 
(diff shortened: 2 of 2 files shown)


zamentur:
If we rename it, i guess immich can't access its database

ewilly:
Which conflicts did you try to avoid by deleting the already existing right cluster?

fflorent:
I agree with @zamentur.

Also I feel like we should instead check whether postgresql is clearly in version 15 and no newer version is installed before proceeding to the upgrade, at least for the first run.

Thovi98:
I don't know how it is feasible, but after having renamed the cluster, couldn't we transfer the useful databases to /main (using pg_dump)?

So something like this:
`check 15/main -> rename 17/main to 17/renamed -> upgrade cluters from 15/main to 17/main -> transfer immich (and maybe other) db from 17/renamed to 17/main`

oleole39:
Linking this PR to https://github.com/YunoHost/issues/issues/2837 and https://github.com/YunoHost-Apps/immich_ynh/issues/432

Naive thought - is there a need to rename cluster? Can't the DBs be dumped and imported to the new cluster?

Thovi98:
> Naive thought - is there a need to rename cluster? Can't the DBs be dumped and imported to the new cluster?

From my understanding, we use `pg_upgradecluster` that upgrades the whole cluster and not each db separately

oleole39:
Given Immich cluster is already at version 17, do the DB it contains still need to be upgraded? Or could they just be dumped and imported into the debian cluster migrated from v15 to v17 ?

ewilly:
> Given Immich cluster is already at version 17, do the DB it contains still need to be upgraded?

not needed

> Or could they just be dumped and imported into the debian cluster migrated from v15 to v17 ?

yes thats my proposal here https://github.com/YunoHost-Apps/immich_ynh/issues/432#issuecomment-5388365914

ewilly:
immich is now updated to work with that PR **but will only work if immich version is beforehand -ge than `3.2.2~ynh1`** 
For more details see https://github.com/YunoHost-Apps/immich_ynh/issues/432#issuecomment-5730042766

Salamandar:
> immich is now updated to work with that PR **but will only work if immich version is beforehand -ge than `3.2.2~ynh1`** For more details see [YunoHost-Apps/immich_ynh#432 (comment)](https://github.com/YunoHost-Apps/immich_ynh/issues/432#issuecomment-5730042766)

Uhm, reading the changes on immich_ynh, it works fine if you upgrade immich AFTER the migration, right ? At least on my server, i didn't have to change the config of Immich during the trixie migration.

ewilly:
> Uhm, reading the changes on immich_ynh, it works fine if you upgrade immich AFTER the migration, right ? At least on my server, i didn't have to change the config of Immich during the trixie migration.

I update my comment on immich issue
immich is working fine after the migration even with db on  `17/renamed` cluster as the port nmber is still the same.
But the only issue i had was due to safety backup done during an upgrade after that migration.
In the backup script i make a dump of the db on the cluster name stored is the app settings (`cat /etc/yunohost/apps/immich/settings.yml | grep db_cluster`) but that one in not change by the migrations script so it still `17/main` when the db is on `17/renamed`, so safety backup fail, so upgarde fail and restoration fail too because there is no immich db provisionned on  `17/main`.
To solve that I saw 2 possibilities:
1. during the migration so at core level change the db_cluster value to  `17/renamed` in `/etc/yunohost/apps/immich/settings.yml`
2. adjust the backup script at immich level to handle that case but that need that the right backup script should be here `/etc/yunohost/apps/immich/scripts/backup`. That's why beeing on a immich version greater than or equal to `3.2.2~ynh1` before migration.
At the moment i choose 2 but of course we can discuss that :)

PR #2351, opened 2026-08-20
[fix] No SSH with public key after trixie migration

### LINKED ISSUE #2839: Trixie migration: ssh public key not recognized after upgrade
## The bug

I used to work exclusively using only public keys to authenticate.

However, after the upgrade, I got with 2 different computers this error:
```
Permission denied (publickey).
```

## The diagnosis

When taking a look at the systemd logs, I could see:
```
sshd-session[xxxx]: Authentication refused: bad ownership or modes for file /home/USER/.ssh/authorized_keys
```

And when the rights fixed (see below), I could also get:
```
sshd-session[xxxx]: Authentication refused: bad ownership or modes for directory /home/USER/.ssh
```

And same for the home (⚠️ wait, is this expected? ⚠️):
```
sshd-session[7404]: Authentication refused: bad ownership or modes for directory /home/USER
```

## The work around

As the user you want to log in with, run the following command to make the `authorized_keys` only accessible to the user:
```
chmod 0600 $HOME/.ssh/authorized_keys
```

And same but for the folder `.ssh` (but it also needs the execution right):
```
chmod 0700 $HOME/.ssh
```

And same for the home (be sure that's OK for you):
```
chmod 0600 $HOME
```

EDIT: as an alternative for the $HOME, this might work:
```
chmod og-w $HOME
```

Quoting [the sshd manpage](https://linux.die.net/man/8/sshd) (thanks to @orhtej2 for the finding):
```
If this file, the ~/.ssh directory, or the user's home directory are writable by other users, then the file could be modified or replaced by unauthorized users. In this case, sshd will not allow it to be used unless the StrictModes option has been set to ''no''. 
```

## Suggested fixes

- Either add a precondition to the migration, to check that the `$HOME`, `$HOME/.ssh` and `$HOME/.ssh/authorized_keys` are owned by the user and have rights positioned to `0700` for the directories and `0600` for the files
- Either add a step (or a migration script) to set the access rights correctly so that it is transparent for the user, but it may interfere with the `user_home` option of nextcloud.

zamentur:
## The problem

After migration ssh return `Connection reset by peer ` and new openssh forbid to log in with unsafe authorized_keys context.

**Related issues:** Fixes YunoHost/issues#2839
**Related PRs:**

## Solution

 * Reload sshd at the end of the migration
 * Cancel the migration if the server contains unsecured authorized_keys

**AI transparency:** no ai

## PR Status
Work in progress

### Code TODOs

  - [ ] Check ACL
  - [ ] Check app user (git forge? SFTP feature my_webapp ?)

### Tests TODOs

## How to test
...

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -692,6 +692,7 @@
     "migration_0036_patch_yunohost_dpkg": "Applying patch on dpkg database to workaround conflict issues…",
     "migration_0036_patching_sources_list": "Patching the sources.lists file…",
     "migration_0036_problematic_apps_warning": "Please note that the following possibly problematic installed apps were detected. It looks like those were not installed from the YunoHost app catalog, or are not flagged as 'working'. Consequently, it cannot be guaranteed that they will still work after the upgrade:",
+    "migration_0036_ssh_user_home_writable_by_group": "The user '{user}' uses ssh with a public key and have its home directory writable by a group or others. New version of openssh in YunoHost 13 forbid to login with public key inside a home directory writable by another user than the owner. You probably could fix it with <code>chmod og-w /home/{user}</code>. However, you might have some apps, process or customization needing this. ",
     "migration_0036_start": "Starting migration to Trixie…",
     "migration_0036_still_on_bookworm_after_main_upgrade": "Something went wrong during the main upgrade, the system appears to still be on Debian Bookworm.",
     "migration_0036_system_not_fully_up_to_date": "Your system is not fully up-to-date. Please perform a regular upgrade before running the migration to Trixie.",
--- a/src/migrations/0036_migrate_to_trixie.py.disabled
+++ b/src/migrations/0036_migrate_to_trixie.py.disabled
@@ -42,6 +42,7 @@ incus exec migration-to-trixie -- yunohost tools migrations run 0036 --accept-di
 import jinja2
 import logging
 import re
+import stat
 import subprocess
 import textwrap
 from datetime import date
@@ -353,6 +354,9 @@ class MyMigration(Migration):
             cmd = 'at -M now >/dev/null 2>&1 <<< "sleep 10; systemctl restart nginx yunohost-api"'
             # For some reason subprocess doesn't like the redirections so we have to use bash -c explicity...
             subprocess.check_call(["bash", "-c", cmd])
+        
+        # Reload sshd due to -R unknown option
+        subprocess.check_call(["systemctl", "reload", "sshd"])
 
         if self.yunohost_major_version() != N_CURRENT_YUNOHOST + 1:
             raise YunohostError(
@@ -449,6 +453,18 @@ class MyMigration(Migration):
 
             if upgradable_system_packages - lime2_hold_packages:
                 raise YunohostError("migration_0036_system_not_fully_up_to_date")
+
+        # Check ssh users won't be unable to login with public key after migration
+        # OpenSSH_10.0p2 forbid to login with public key that are in a /home/USER
+        # with write permissions on groups cause it's possible to move .ssh even 
+        # if groups have no write permissions on thi s.ssh directory.
+        for keys_path in Path('/home').glob('.ssh/authorized_keys'):
+            ssh_user = keys_path.parts[2]
+            user_home = Path('/home') / ssh_user
+            if bool(stat.S_IWGRP & user_home.stat().st_mode) \
+               or bool(stat.S_IWOTH & user_home.stat().st_mode):
+                raise YunohostError("migration_0036_ssh_user_home_writable_by_group", user=ssh_user)
+
 
     @property
     def disclaimer(self):


src/migrations/0036_migrate_to_trixie.py.disabled
```diff
@@ -451,6 +455,18 @@ class MyMigration(Migration):
...
+        # Check ssh users won't be unable to login with public key after migration
+        # OpenSSH_10.0p2 forbid to login with public key that are in a /home/USER
+        # with write permissions on groups cause it's possible to move .ssh even 
+        # if groups have no write permissions on thi s.ssh directory.
+        for keys_path in Path('/home').glob('.ssh/authorized_keys'):
+            ssh_user = keys_path.parts[2]
+            user_home = Path('/home') / ssh_user
+            if bool(stat.S_IWGRP & user_home.stat().st_mode) \
+               or bool(stat.S_IWOTH & user_home.stat().st_mode):
+                raise YunohostError("migration_0036_ssh_user_home_writable_by_group", user=ssh_user)
```
fflorent, line 467: It's not only the user home directory that has to be `g-w,o-w`, but also the `.ssh/` directory and the `.ssh/authorized_keys` file that both have to be `g-rwx,o-rwx` IRRC.

EDIT: actually, they are only required to be unwritable to group and others, quoting [the sshd manpage](https://linux.die.net/man/8/sshd) that @orhtej2 pointed me:
> If this file, the ~/.ssh directory, or the user's home directory are writable by other users, then the file could be modified or replaced by unauthorized users. In this case, sshd will not allow it to be used unless the StrictModes option has been set to ''no''. 



```suggestion
                raise YunohostError("migration_0036_ssh_user_home_writable_by_group", user=ssh_user)

            if bool(stat.S_IWGRP & (keys_path / '..').stat().st_mode) \
               or bool(stat.S_IWOTH & (keys_path / '..').stat().st_mode) \
               or bool(stat.S_IWGRP & keys_path.stat().st_mode) \
               or bool(stat.S_IWOTH & keys_path.stat().st_mode):
                raise YunohostError("migration_0036_ssh_dir_writable_by_group", user=ssh_user)
```

And add `migration_0036_ssh_dir_writable_by_group` translation.

BTW, wouldn't `bool((stat.S_IWGRP | stat.S_IWOTH) & user_home.stat().st_mode)` work? (but maybe less eloquent?)

Salamandar, line 467: IMHO this *needs* to be done, not just a warning: Admins might be locked out of their machine, if they disabled password login and this happens.

fflorent: (it's an exception, so it already interrupts the migration, right?)

Actually they can reenable temporarily the password login to fix the issue and reenable after.

That's how I have managed the issue on my side. But that requires to know the trick.

zamentur: If i remember well, in bookworm, `.ssh` and `.ssh/authorized_keys` was already checked by sshd. SO the issue with trixie is specifically on `/home/USER` which is a new behaviour of the new sshd version.

fflorent: Actually, I have just checked right now thanks to incus:
- on bookworm, setting write permission for `other` to either `/home/USER`, `.ssh` or `.ssh/authorized_keys` makes the authentication switch to password.
- on trixie, not only the permissions for `other` are checked but also for the `group`.

| Rights | works on bookworm? | works on trixie? |
| --- | --- | --- |
| `g-w,o-w` for `$HOME` and `$HOME/.ssh` and $HOME/.ssh/authorized_keys` | ✅ | ✅ |
| `g+w,o-w` for any of `$HOME`, `$HOME/.ssh` or `$HOME/.ssh/authorized_keys` | ✅ | ❌ |
| `g+w,o+w` for any of `$HOME`, `$HOME/.ssh` and `$HOME/.ssh/authorized_keys` | ❌ | ❌ |
| (obviously) `g-w,o+w` for any of `$HOME`, `$HOME/.ssh` and `$HOME/.ssh/authorized_keys` | ❌ | ❌ |

src/migrations/0036_migrate_to_trixie.py.disabled
```diff
@@ -404,6 +455,18 @@ class MyMigration(Migration):
...
+        # Check ssh users won't be unable to login with public key after migration
+        # OpenSSH_10.0p2 forbid to login with public key that are in a /home/USER
+        # with write permissions on groups cause it's possible to move .ssh even 
+        # if groups have no write permissions on thi s.ssh directory.
+        for keys_path in Path('/home').glob('*/.ssh/authorized_keys'):
+            ssh_user = keys_path.parts[2]
+            test_list = [ Path('/home') / ssh_user, keys_path.parent, keys_path ]
+            if any([ bool(stat.S_IWGRP & t.stat().st_mode) for t in test_list ]):
+                raise YunohostError("migration_0036_ssh_user_home_writable_by_group", user=ssh_user)
+        1/0
```
tituspijean:
```suggestion
```

? ^^

zamentur: it was a small hack to run the checks without running the migration...

locales/en.json
```diff
@@ -693,6 +693,8 @@
     "migration_0036_patching_sources_list": "Patching the sources.lists file…",
     "migration_0036_post_migrations_skipped": "Post {distrib} migrations not run automatically as requested, run them when you want :)",
     "migration_0036_problematic_apps_warning": "Please note that the following possibly problematic installed apps were detected. It looks like those were not installed from the YunoHost app catalog, or are not flagged as 'working'. Consequently, it cannot be guaranteed that they will still work after the upgrade:",
+    "migration_0036_ssh_app_need_upgrade": "The user '{user}' uses ssh with a public key but the app '{app}' can move the .ssh/authorized_keys file. Please upgrade the app or report it if this message continues to be displayed. Note: if you bypass this check, some ssh users could loose their ssh access.",
```
tituspijean, line 696:
```suggestion
    "migration_0036_ssh_app_need_upgrade": "Account '{user}' uses ssh with a public key, but the '{app}' app can move the .ssh/authorized_keys file. Please upgrade the app or report it if this message continues to be displayed. Note: if you bypass this check, some ssh users could lose their ssh access.",
```

locales/en.json
```diff
@@ -693,6 +693,8 @@
...
+    "migration_0036_ssh_user_home_writable_by_group": "The user '{user}' uses ssh with a public key but some unix group can move the .ssh/authorized_keys file due to this permission '{permission}'. New version of openssh in YunoHost 13 forbids logging in with a public key in directory tree writable by any group. You probably can fix it with <code>chmod g-w {home} {home}/.ssh {home}/.ssh/authorized_keys</code> or by equivalent setfacl command. However, be wary, you might have some apps, process, or customization needing this. Do not hesitate to ask for help on the support channels.",
```
tituspijean, line 697:
```suggestion
    "migration_0036_ssh_user_home_writable_by_group": "Account '{user}' uses ssh with a public key but some unix group can move the .ssh/authorized_keys file due to the  '{permission}' permission. New version of openssh in YunoHost 13 forbids logging in with a public key in directory tree writable by any group. You probably can fix it with <code>chmod g-w {home} {home}/.ssh {home}/.ssh/authorized_keys</code> or by equivalent setfacl command. However, be wary, you might have some apps, process, or customization needing this. Do not hesitate to ask for help on the support channels.",
```

locales/en.json
zamentur: `user` could be gitlab, so i don't know if Account is a good idea here.

PR #2352, opened 2026-08-20
migrations: trixie: Fix path to the python requirements file

Salamandar:
## The problem

...

**Related issues:**
**Related PRs:**

## Solution

...

**AI transparency:** *If AI has been used, explain how. See https://doc.yunohost.org/en/dev/?persistLocale=true#%EF%B8%8F-develop*

## PR Status
*Work in progress / Untested / Tested partially / Fully tested / Wait for other PR to be merged / Ready*

### Code TODOs

 - [ ] Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)
 - [ ] Write data or settings migration
   - [ ] Update/write unit tests

### Tests TODOs

## How to test
...

### DIFF
--- a/src/migrations/0036_migrate_to_trixie.py.disabled
+++ b/src/migrations/0036_migrate_to_trixie.py.disabled
@@ -116,7 +116,8 @@ def _backup_pip_freeze_for_python_app_venvs():
             continue
         pip_freeze = subprocess.check_output([pip, "freeze"]).decode(encoding="utf-8")
         pip_freeze = re.sub(r"^pkg(-|_)resources==.*$", "", pip_freeze)
-        (venv / VENV_REQUIREMENTS_SUFFIX).write_text(pip_freeze)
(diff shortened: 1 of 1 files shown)


fflorent (approved):
LGTM 

A bit tricky [on the `python.py` module side](https://github.com/YunoHost/yunohost/blob/6fa3c601eff5d27ec1d2edbe6a1b0da488f82468/src/migrations/python.py#L90-L100) which uses the `os.path.join()` method that produces a path that may include or not a trailing `/`. As the `listdir()` method returns file or directory names without trailing `/`, the `os.path.join` returns a path string without trailing `/` and the whole loop works.

Another side note: the PR https://github.com/YunoHost/yunohost/pull/2237 planned to test using `python -m venv --upgrade VENV_DIR`. It would be much simpler and maybe more robust.

Still this is an improvement for in the meantime, so I approve and merge it.

Thank you!

PR #2353, opened 2026-08-20
Improve the venv migration

### LINKED ISSUE #2840: Trixie migration: lack of debug logs for the pip freeze backups
I managed to migrate my server to trixie, with some failures. The operation has been logged in that file: https://paste.yunohost.org/raw/pulebokuro

After the reboot:
- an app does not work ([lasuite-meet_ynh](https://github.com/YunoHost-Apps/lasuite-meet_ynh)), because the python venv has not been migrated
- whereas another app, [grist_ynh](https://github.com/YunoHost-Apps/grist_ynh), works like a charm with its sandbox for python formulas (so the venv migration has worked successfully for it).

The problem here is that the logs don't report much things related to the `pip freeze backup` (we only see skipped venv folders):
https://github.com/YunoHost/yunohost/blob/622991f4a559caee626f2d14567883ef52fc6364/src/migrations/0036_migrate_to_trixie.py.disabled#L105-L119

I suggest to print the apps which have successfully been `pip freeze` backed up, so investigation can later be easier.

### LINKED ISSUE #2841: Trixie migration: venv not regenerated for `borg` and `borgserver`
### Describe the bug

After having migrated to trixie, neither borg nor borgserver work.
Taking a look at the migration process, this is because none of them have a `bin/activate` file.
This is because they install the virtual environment using `python3 -m venv .... --upgrade`:

https://github.com/YunoHost-Apps/borg_ynh/blob/09dedab20d29d9cf46aa65db23f8e40f7751327a/scripts/_common.sh#L12

https://github.com/YunoHost-Apps/borgserver_ynh/blob/56456675adb18ebec2c7d7979e2b0ffc73acd71e/scripts/_common.sh#L8

And I expect also compteur-du-gase and scoovie to have the same issue (and maybe other packages which pass `--upgrade-deps` but less sure):
https://github.com/search?q=org%3AYunoHost-Apps+%2Fpython.*venv.*--upgrade%2F&type=code

There are two different locations in which we look for `bin/activate`:
- in the migrate_to_trixie.py script: https://github.com/YunoHost/yunohost/blob/6a908ef946a40e26f8e32f277a85b1883cf9c9ee/src/migrations/m0036_migrate_to_trixie.py#L94
- and in the python.py migration class: https://github.com/YunoHost/yunohost/blob/6a908ef946a40e26f8e32f277a85b1883cf9c9ee/src/migrations/python.py#L93

### Suggested solution

Two possibilities:
- either we consider this as a bug of the apps (a suggestion for a fix would be appreciated)
- either we may instead look for the `pyvenv.cfg` file, which seems to be also present for the borg and borgserver venv folders (so probably a more solid approach).
  - I tend to think this check for `PYTHONHOME` and `VIRTUAL_ENV` would be unnecessary: https://github.com/YunoHost/yunohost/blob/6a908ef946a40e26f8e32f277a85b1883cf9c9ee/src/migrations/m0036_migrate_to_trixie.py#L97

### Workaround

I think that just running this command will work (replace `APP` by the desired app):
```bash
sudo yunohost app upgrade APP --force
```

### Context

Same as for #2834 

### Expected behavior

No extra steps are necessary to make the borg and borgserver app working after the trixie migration.

### LINKED ISSUE #2861: Trixie: improve venv migration

fflorent:
## The problem

1. Design issue (complexity): The python migrations proceeds to a backup of pip freeze output, which makes the migration rerun impossible if the backup failed
2. Borg (client) and Borg Server aren't migrated (because they don't have the `bin/activate` file, check https://github.com/YunoHost/issues/issues/2841)
3. Searxng is installed [using an editable](https://github.com/YunoHost-Apps/searxng_ynh/blob/904e3ad096a2f17a5e0befcefeb60f299648e5f4/scripts/_common.sh#L40), which makes the migration fails
4. The pip version required by the app might be downgraded during the migration ([for example for searxng](https://github.com/YunoHost-Apps/searxng_ynh/blob/904e3ad096a2f17a5e0befcefeb60f299648e5f4/scripts/_common.sh#L37)), by default pip-freeze does not output pip in the dependencies list

Fixes https://github.com/YunoHost/issues/issues/2841
Fixes https://github.com/YunoHost/issues/issues/2861
Fixes https://github.com/YunoHost/issues/issues/2840 (not relevant anymore)

## Solution

1. Instead of backuping the `pip freeze` command output, run the command with the `--path` option, as suggested in this Stack Overflow answer: https://stackoverflow.com/a/79894112
2. Look for `pyvenv.cfg` file instead of `bin/activate`, which are generated in any case (whenever `python -m venv` has been run once without `--upgrade` or never like for borg and borgserver)
3. Make a two step installation: one for the dependencies exclusively and one for the editables for which we pass `--no-build-isolation`)
4. Search for the pip dependency in the requirements produced by `pip freeze`, compare with the pip version of the system and keep it if it is newer (otherwise the version is removed)

**AI transparency:** Local AI (gemma3:4b) used once to see how to transform a shell command into python in a general manner, possible solutions to use a pipe (without success, I chose to use `NamedTemporaryFile`). The AI has not been given the context of the code and the produced result has not been copy-pasted, used more like searching for solutions on StackOverflow.

## PR Status
*Fully tested*: With borg, grist, lasuite-meet, searxng and I hate money.

## How to test

Use ynh-dev with bookworm, then:
- install :
   - borg (choose a path for the repo to make the backups local, much simpler)
   - grist
   - searxng 
   - and ihatemoney (see https://github.com/YunoHost/yunohost/commit/f630e32da16033a17bee36916b3a1f3a3daae85e)
- start a borg backup
- from the host and in the `yunohost` folder, switch to the PR's branch
- from the container, copy the migrator to trixie's version of this PR:  `cp /ynh-dev/yunohost/src/migrations/m0036_migrate_to_trixie.py /usr/lib/python3/dist-packages/yunohost/migrations/0036_migrate_to_trixie.py`
- comment these lines to disable the automatic run of migrations following the one for trixie: https://github.com/YunoHost/yunohost/blob/83df92b76061a332aef8b05554251d5c2ee9fec8/src/migrations/0036_migrate_to_trixie.py.disabled#L343-L346
- run the migration to trixie: `yunost tools migrations run migrate_to_trixie --accept-disclaimer`
- snapshot the container if you want
- inside the container, run `./ynh-dev use-git yunohost`
- run manually the other migrations: `yunohost --debug tools migrations run --accept-disclaimer`
- check the migrations log:
  - verify that the `pip` package is upgraded for searxng
- check the services status for the the apps, that they are functional
- check the apps themselves 
  - all apps should be functional (except ihatemoney in the current state of the PR, I need help)

### DIFF
--- a/src/migrations/python.py
+++ b/src/migrations/python.py
@@ -24,7 +24,6 @@ from logging import getLogger
 from moulinette import m18n
 
 from ..tools import Migration, tools_migrations_state
-from ..utils.file_utils import rm
 from ..utils.process import call_async_output
 from ..utils.system import debian_version
 
@@ -63,9 +62,6 @@ class PythonMigration(Migration):
     migration_id: str
     state = None
 
-    def venv_requirements_suffix(self) -> str:
-        return f".requirements_backup_for_{debian_version()}_upgrade.txt"
-
     def extract_app_from_venv_path(self, venv_path: str) -> str:
         venv_path = venv_path.replace("/var/www/", "")
         venv_path = venv_path.replace("/opt/yunohost/", "")
@@ -90,10 +86,8 @@ class PythonMigration(Migration):
         for file in os.listdir(dir):
             path = os.path.join(dir, file)
             if os.path.isdir(path):
-                activatepath = os.path.join(path, "bin", "activate")
-                if os.path.isfile(activatepath) and os.path.isfile(
-                    path + self.venv_requirements_suffix()
-                ):
+                pyvenv_path = os.path.join(path, "pyvenv.cfg")
+                if os.path.isfile(pyvenv_path):
                     result.append(path)
                     continue
                 if level < maxlevel:
@@ -133,9 +127,6 @@ class PythonMigration(Migration):
 
         venvs = self._get_all_venvs("/opt/") + self._get_all_venvs("/var/www/")
         for venv in venvs:
-            if not os.path.isfile(venv + self.venv_requirements_suffix()):
-                continue
-
             app_corresponding_to_venv = self.extract_app_from_venv_path(venv)
 
             # Search for ignore apps
@@ -177,7 +168,6 @@ class PythonMigration(Migration):
                 app_corresponding_to_venv.startswith(app)
                 for app in self.ignored_python_apps
             ):
-                rm(venv + self.venv_requirements_suffix())
                 logger.info(
                     m18n.n(
                         "migration_python_venv_rebuild_broken_app",
@@ -201,21 +191,11 @@ class PythonMigration(Migration):
                 venv_cmd.append("--system-site-packages")
 
             # Recreate the venv
-            rm(venv, recursive=True)
             callbacks = (
                 lambda l: logger.debug("+ " + l.rstrip() + "\r"),
                 lambda l: logger.warning(l.rstrip()),
             )
-            call_async_output(venv_cmd, callbacks)
-            status = call_async_output(
-                [
-                    f"{venv}/bin/pip",
-                    "install",
-                    "-r",
-                    venv + self.venv_requirements_suffix(),
-                ],
-                callbacks,
-            )
+            status = call_async_output(["python", "-m", "venv", "--upgrade", venv], callbacks)
             if status != 0:
                 logger.error(
                     m18n.n(
@@ -223,5 +203,3 @@ class PythonMigration(Migration):
                         app=app_corresponding_to_venv,
                     )
                 )
-            else:
-                rm(venv + self.venv_requirements_suffix())
--- a/src/migrations/python.py
+++ b/src/migrations/python.py
@@ -283,7 +283,8 @@ class PythonMigration(Migration):
                         extra_args=["--no-build-isolation"],
                     )
 
-                self._cleanup_old_python_assets(venv_path, old_python_version_major_minor)
+                if not venv_cfg.include_system_site_packages:
+                    self._cleanup_old_python_assets(venv_path, old_python_version_major_minor)
             except YunohostError as e:
                 logger.warn(e)
                 continue
--- a/src/migrations/python.py
+++ b/src/migrations/python.py
@@ -77,7 +77,7 @@ class PythonMigration(Migration):
 
     migration_id: str
     state = None
-    pyenv_cfg_version_extractor = re.compile(
+    pyvenv_cfg_version_extractor = re.compile(
         r"version\s*=\s*(?P<version>\d+\.\d+)"  # Omit the patch number of the version
     )
 
@@ -112,8 +112,8 @@ class PythonMigration(Migration):
         for file in os.listdir(dir):
             path = os.path.join(dir, file)
             if os.path.isdir(path):
-                pyvenv_path = os.path.join(path, "pyvenv.cfg")
-                if os.path.isfile(pyvenv_path):
+                pyvenv_cfg_path = os.path.join(path, "pyvenv.cfg")
+                if os.path.isfile(pyvenv_cfg_path):
                     result.append(path)
                     continue
                 if level < maxlevel:
@@ -280,7 +280,7 @@ class PythonMigration(Migration):
                 raise e
 
     def _extract_venv_python_version(self, venv_path: Path) -> None | str:
-        py_version_match = self.pyenv_cfg_version_extractor.search(
+        py_version_match = self.pyvenv_cfg_version_extractor.search(
             read_file(venv_path / "pyvenv.cfg")
         )
         return py_version_match.group("version") if py_version_match else None
--- a/src/migrations/python.py
+++ b/src/migrations/python.py
@@ -32,18 +32,47 @@ import re
 import tempfile
 
 from moulinette import m18n
+from configparser import ConfigParser, UNNAMED_SECTION
 
 from ..tools import Migration, tools_migrations_state
 from ..utils.error import YunohostError
-from ..utils.file_utils import read_file, rm
+from ..utils.file_utils import rm
 from ..utils.process import call_async_output, check_output
 from ..utils.system import debian_version, dpkg_compare_version
 
-type PipExecutionCallbacks = Tuple[Callable[[str], None], Callable[[str], None]]
+type ExecutionCallback = Tuple[Callable[[str], None], Callable[[str], None]]
 
 logger = getLogger("yunohost.migration")
 
-MAJOR_MINOR_PATCH_RE = r'\d+\.\d+\.\d+'
+MAJOR_MINOR_PATCH_RE = r'(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)'
+
+
+class PyVenvConfig():
+    def __init__(self, venv_path: Path):
+        self.path = venv_path / "pyvenv.cfg"
+        parser = ConfigParser(allow_unnamed_section=True)
+        with open(self.path) as file:
+            parser.read_file(file)
+        self.data = parser[UNNAMED_SECTION]
+
+    @property
+    def include_system_site_packages(self) -> bool:
+        return self.data.get('include-system-site-packages', None) == 'true'
+
+    @property
+    def version(self) -> str:
+        return self.data['version']
+
+    @cached_property
+    def version_parsed(self) -> Tuple[int, int, int]:
+        match = re.match(MAJOR_MINOR_PATCH_RE, self.version)
+        if not match:
+            raise YunohostError(f"Cannot parse version stored in {self.path}: {self.version}")
+        return (
+            int(match.group("major")),
+            int(match.group("minor")),
+            int(match.group("patch"))
+        )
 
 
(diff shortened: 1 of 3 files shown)


Salamandar:
image
:eyes:

Salamandar:
Doesn't work for me. Indeed, it updates the venv, but leaves the packages in the old `lib/python3.11/site-packages` .

fflorent:
@Salamandar Yeah, too bad. I don't have a clue if there could exist a command to migrate the libraries to the new sites package (and recompile the `.pyc` files).

src/migrations/python.py
```diff
@@ -194,27 +184,15 @@ def run(self):
...
-            status = call_async_output(
-                [
-                    f"{venv}/bin/pip",
-                    "install",
-                    "-r",
-                    venv + self.venv_requirements_suffix(),
-                ],
-                callbacks,
-            )
+            status = call_async_output(["python", "-m", "venv", "--upgrade", venv], callbacks)
```
fflorent: This command seems to produce a better result:
```bash
./bin/pip install -r <(pip freeze --path lib/python3.11/site-packages)
```

Salamandar: Yeah, I actually used `--local` instead of `--path` but it's even better.

src/migrations/python.py
```diff
@@ -107,11 +127,20 @@ def is_pending(self):
...
+    @cached_property
+    def python_version(self):
+        """ Returns the python version without the patch """
+        return f"{sys.version_info.major}.{sys.version_info.minor}"
+
     @property
     def mode(self):
-        if not self.is_pending():
-            return "auto"
-
```
fflorent, line 114: I remove this condition, I feel like it is legit to be able to rerun this migration.

PS: I am a bit disturbed by the meaning of `self.mode` VS how the value is computed. If I am right, it was meant to detect whether something had to be migrated and otherwise run the migration to automatically skip it. But this migration is automatically run at the end of the trixie migration, so it does not make sense anymore to me. Probably worth to let the parent class handle the `mode` property and make sure that `--accept-disclaimer` is not required. Can't we rename it `self.is_needed` or `self.needs_run`?

src/migrations/python.py
```diff
@@ -193,35 +250,123 @@ def run(self):
...
+            if line.startswith("-e"):
+                editables.append(line)
+            else:
+                new_requirements.append(line)
+        return (str.join("\n", new_requirements), str.join("\n", editables))
+
+    def _upgrade_venv(self, venv: str, include_system_site_packages: bool, callbacks: ExecutionCallback):
+        venv_cmd = ["python", "-m", "venv", "--upgrade", venv]
+        if include_system_site_packages:
+            venv_cmd.append("--system-site-packages")
```
fflorent, line 346: This is for ihatemoney specifically.

Please note that it has some issue and the logs show errors:
https://paste.yunohost.org/raw/pebewiduza

src/migrations/python.py
Salamandar: The whole is_pending stuff needs to be removed indeed, it looks like a legacy artifact.

src/migrations/python.py
```diff
@@ -193,35 +250,123 @@ def run(self):
...
+                old_python_version = f"{cfg_py_major}.{cfg_py_minor}"
+            except Exception as e:
+                logger.warn(e)
+                continue
 
-            # Get venv info
-            pyvenv_cfg = open(os.path.join(venv, "pyvenv.cfg")).read()
-            if "include-system-site-packages = true" in pyvenv_cfg:
-                venv_cmd.append("--system-site-packages")
+            if dpkg_compare_version(old_python_version, self.python_version) > -1:
```
Salamandar: I'd prefer `from packaging.version import Version ; Version(xxx) > Version(yyy)` tbh.

src/migrations/python.py
```diff
@@ -193,35 +250,123 @@ def run(self):
...
+
+            self._upgrade_venv(venv, venv_cfg.include_system_site_packages, callbacks)
+
+            requirements = check_output(
                 [
-                    f"{venv}/bin/pip",
+                    pip_path,
+                    "freeze",
+                    "--all",
+                    "--path=" + str(venv_path / f"lib/python{old_python_version}/site-packages"),
```
Salamandar:
```suggestion
                    "--path",
                    str(venv_path / "lib" / f"python{old_python_version}" / "site-packages"),
```

Salamandar (approved):

src/migrations/python.py
fflorent: Right, I first believed it was not required by `yunohost`, but it looks like it is:
https://github.com/YunoHost/yunohost/blob/83df92b76061a332aef8b05554251d5c2ee9fec8/pyproject.toml#L25

Much more convenient (also for `version_parsed`)

Salamandar: TBH my position is : if a dependency is not huge and helps us, let's add it.

fflorent:
Oops, I haz borken it.

Fix incoming this evening

src/migrations/python.py
fflorent: Quite surprisingly, whereas the migration fails here, when the migration to trixie is terminated and before the psql and venv migrations, the following steps seems to fix ihatemoney:
```bash
./ynh-dev use-git yunohost
yunohost tools upgrade system
yunohost tools migrations run 0037_postgresql_15_to_17

/var/www/ihatemoney/venv/bin/python3 -m venv --system-site-packages --upgrade /var/www/ihatemoney/venv/
/var/www/ihatemoney/venv/bin/pip install --use-pep517 -r <(/var/www/ihatemoney/venv/bin/pip freeze --path=/var/www/ihatemoney/venv/lib/python3.11/site-packages/ --all --exclude-editable)
```

Not sure about the differences, except maybe the position of the option `--system-site-packages`.

fflorent: I think I got it: when including the system site packages, invoking `_cleanup` breaks the app.

Not sure what happens in details nor whether this is relevant. Any idea @Salamandar?

src/migrations/python.py
```diff
@@ -193,35 +234,126 @@ def run(self):
...
+        venv_cmd = ["python", "-m", "venv", "--upgrade", venv]
+        if include_system_site_packages:
+            venv_cmd.append("--system-site-packages")
+        return self._call_async_output(venv_cmd, callbacks)
+
+    def _install_requirements(self, pip_path: Path, requirements: str, app: str, callbacks: ExecutionCallback, extra_args: list[str] = []):
+        with tempfile.NamedTemporaryFile() as fp:
+            fp.write(requirements.encode("utf8"))
+            fp.flush()
+            status = self._call_async_output(
```
fflorent, line 357: The venv upgrade and the installation of the requirements are done as root. I am not sure whether that's right.

I tend to think this could be run as root and then sets the ownership of the venv to the owner of `pyvenv.cfg` before the migration. What do you think?

Josue-T: Yes seem a good idea I think.

fflorent: Done in 98eeafa3d50d7da3877d416cc8ffba89a9458172

fflorent: @Josue-T Ahah! At the same time

src/migrations/python.py
fflorent: The PR is already quite big, I'll open an issue to suggest that instead to not grow more its scope.

src/migrations/python.py
Josue-T: From what I tested and my understanding the issue happen maybe because pip found on the system side `python3-greenlet 3.1.0-1+b2` and that's why he is not happy because it require `greenlet>=3.1.1`. We can force to install a different version in the venv which override the default one, but maybe it's a bit less implicit for pip.

But now I'm a bit surprised that `ihatemoney` need `--system-site-packages` because this package don't seem to require system package. I mean generally the package that need this option have debian python dependencies (in addition to the default `python3-venv`) but it seem not the case for ihatemoney.

src/migrations/python.py
```diff
@@ -18,18 +18,52 @@
...
 
-import os
+from functools import cached_property
 from logging import getLogger
+import os
+from pathlib import Path
+from typing import Callable, Tuple
+try:
+    from pip import __version__ as PIP_VERSION
+except Exception:
```
Salamandar:
```suggestion
except ImportError:
```

Salamandar (approved):

src/migrations/python.py
fflorent: Also the disclaimer is outdated, it mentions explicitly bookworm, so in any way, it needs a refresh:

https://github.com/YunoHost/yunohost/blob/83df92b76061a332aef8b05554251d5c2ee9fec8/locales/en.json#L718

Josue-T (approved):

src/migrations/python.py
fflorent: Issue opened here for discussion: https://github.com/YunoHost/issues/issues/2867

PR #2356, opened 2026-08-21
fix(dyndns): update conditions in case of single dns_exposure

### LINKED ISSUE #2865: DynDNS still tries to update IP, ignoring DNS exposure setting
### Describe the bug

In case of DNS exposure setting not being set to `both`, the IP update is systematically triggered even though the exposed IP has not changed.

Many thanks to @artlog for pointing out this unwanted behavior.

### Context

- YunoHost version: 12+

### To reproduce

1. Set DNS exposure setting to IPv4 or IPv6.
2. Setup a DynDNS domain, let it sync once.
3. Even though the one IP has not changed, since the excluded one is compared to `None`, the update is triggered.

### Expected behavior

The DynDNS script should be a tad more smart and actually ignore the IPv4 or IPv6 if not exposed.

### Logs

```
sudo yunohost settings get misc.network.dns_exposure
ipv6
```

Despite this setting and IPv6 not being changed, the client still pushes the update.

https://paste.yunohost.org/raw/lohemimexi

tituspijean:
## The problem

**Related issues:**
Closes [#2865](https://github.com/YunoHost/issues/issues/2865)

**Related PRs:**

## Solution

Import the dns_exposure setting and use it to actually determine if an update is needed.

We could also go one step further and conditionally check the IPs depending on the dns_exposure.
:arrow_right:  *Done in [5c346a0](https://github.com/YunoHost/yunohost/pull/2356/commits/5c346a03a0f1f04913e0be97435a0b9a4beb2e8f)*

## PR Status
Untested

### Code TODOs

 - [ ] Update related repo (like yunohost-admin, yunohost-portal, package-linter, etc.)
 - [ ] Write data or settings migration
   - [ ] Update/write unit tests

### Tests TODOs

## How to test
...

### DIFF
--- a/src/dyndns.py
+++ b/src/dyndns.py
@@ -373,6 +373,7 @@ def dyndns_update(
     import dns.update
 
     from .dns import _build_dns_conf
+    from .settings import settings_get
 
     # If domain is not given, update all DynDNS domains
     if domain is None:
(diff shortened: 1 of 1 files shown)


tituspijean:
There might be a way to factorize the tests with `dns_exposure`, but I am not seeing it tonight. :)

PR #2358, opened 2026-08-22
Add one more check to ensure that no app is declaring postgresql in t…

### LINKED ISSUE #2737: Postgresql migration is skipped sometimes
Different user that upgraded to Debian bookworm had some issues on the postgresql pool upgrade. 

For instance we had theses [this message](https://forum.yunohost.org/t/synapse-instant-messaging-server-matrix-network/4381/159?u=josue) and this issue [this issues](https://github.com/YunoHost-Apps/synapse_ynh/issues/598) related to this. This was caused by the fact that synapse updated the min version of postgreql in the last release. Probably what happen is that theses user upgraded to Debian bookworm and didn't see that the postgresql pool upgraded didn't happen and so whey were still running on postgresql 13. I also had the experience some time ago of a Yunohost upgrade to bookworm and postgresql didn't upgrade as expected. On my side I discovered the issue quite quickly and I fixed the issue by just doing the upgrade manually.

Even if this issue can be fixed quite easily by some command, it's something that need some improvements because:
- Event if there are not every day a CVE about postgresql, running an old postgresql version is really secure.
- Some user will have some issue like with synapse because the postgresql version is too old.
- Some app will not work any more but might not raise an explicit error about the fact that postgresql is too old. And the issue might be difficult to reproduce because we need to know that the impacted user run an old postgresql version.
- Maybe some app will have just some really strange bug just because only some query are not compatible with an old postgresql version.

Thinking about how to improve the situation about this I thought about different solution:
- [ ] Reimplement a specific migration for postgresql upgrade and make it more strong. A long time ago when we did a major upgrade I saw a migration related to postgresql  pool upgrade. On my side it was running well. In the last upgrade I didn't see it. I don't exactly know why it was removed. I would be interested to know if it can be re-implemented.
- [ ] Add a check in the diagnostic if the postgresql version is correct. It can be a simple check if the postgresql match what is expected and we can raise an error if it's not the case. It will not solve the issue automatically but it could be enough to avoid the issue mentioned above.
- [ ] Add a check in the provisioning for postgresql.  While we provision a psql database we can also check if it's the correct version and show an error if it's not correct.
- Maybe something else.

The idea of this issue it not to make a big development around this, but more to find a simple solution to avoid this issue because it seem to impact many users. I think we could be surprised of how many instances are running an old psql version but the user just don't know because you can run a legacy postgresql version for a long time.

Josue-T:
## The problem

**Related issues: https://github.com/YunoHost/issues/issues/2737
**Related PRs: -

## Solution

Add one more check to ensure that no app is declaring postgresql in the app resources.

**AI transparency:** Not used

## PR Status
*Ready*

### Tests TODOs

## How to test

I just replayed the migration on trixie while I've an app installed with postgresql. And I checked that the resource verification was working. (I modified the first `if os.system(ynh_deps_cmd) != 0` to ensure that the app resource check was working.

### DIFF
--- a/src/migrations/postgresql.py
+++ b/src/migrations/postgresql.py
@@ -25,7 +25,9 @@ from logging import getLogger
 
 from moulinette import m18n
 
+from ..app import app_list
 from ..tools import Migration
+from ..utils.app_utils import _get_manifest_of_app
 from ..utils.error import YunohostError, YunohostValidationError
 from ..utils.system import free_space_in_directory, space_used_by_directory
 
@@ -42,8 +44,11 @@ class PostgreSQLMigration(Migration):
     def run(self):
         ynh_deps_cmd = 'grep -A10 "ynh-deps" /var/lib/dpkg/status | grep -E "Package:|Depends:" | grep -B1 postgresql'
         if os.system(ynh_deps_cmd) != 0:
-            logger.info("No YunoHost app seem to require postgresql... Skipping!")
-            return
+            # In addition, also check that no app declare in his resource postgresql. cf
+            # https://github.com/YunoHost/issues/issues/2737
+            if not self._has_app_with_psql_resource():
+                logger.info("No YunoHost app seem to require postgresql... Skipping!")
+                return
 
         if not self.package_is_installed(f"postgresql-{self.previous_version}"):
             logger.warning(m18n.n("migration_postgresql_previous_not_installed"))
@@ -103,6 +108,13 @@ class PostgreSQLMigration(Migration):
             # See https://www.postgresql.org/docs/17/sql-altercollation.html#SQL-ALTERCOLLATION-NOTES
             self.runcmd(f"{sudocmd} psql --dbname='{database}' --command='REINDEX DATABASE {database};'")
             self.runcmd(f"{sudocmd} psql --dbname='{database}' --command='ALTER DATABASE {database} REFRESH COLLATION VERSION;'")
+
+    def _has_app_with_psql_resource(self) -> bool:
+        for app_info in app_list()['apps']:
+            app_manifest = _get_manifest_of_app(app_info['id'])
+            if app_manifest.get('resources', {}).get('database', {}).get('type', None) == 'postgresql':
+                return True
+        return False
 
     def package_is_installed(self, package_name):
         (returncode, out, err) = self.runcmd(


src/migrations/postgresql.py
fflorent: I see below we import utils running `from ..utils.my_module import whatever`, why not doing the same?

```suggestion

from moulinette import m18n

from ..app import app_list
from ..tools import Migration
from ..utils.app_utils import _get_manifest_of_app
from ..utils.error import YunohostError, YunohostValidationError
from ..utils.system import free_space_in_directory, space_used_by_directory
```

src/migrations/postgresql.py
```diff
@@ -104,6 +109,13 @@ def run_post_migration(self):
             self.runcmd(f"{sudocmd} psql --dbname='{database}' --command='REINDEX DATABASE {database};'")
             self.runcmd(f"{sudocmd} psql --dbname='{database}' --command='ALTER DATABASE {database} REFRESH COLLATION VERSION;'")
 
+    def _has_app_with_psql_resource(self) -> bool:
+        for app_id in yunohost.app.app_list()['apps']:
+            app_manifest = _get_manifest_of_app(app_id['id'])
+            if app_manifest.get('resources', {}).get('database', {}).get('type', None) == 'postgresql':
+                return True
```
fflorent: - either rename `app_id` to `app_info` which gives a more accurate idea of what it contains

```suggestion
        for app_info in yunohost.app.app_list()['apps']:
            app_manifest = _get_manifest_of_app(app_info['id'])
            if app_manifest.get('resources', {}).get('database', {}).get('type', None) == 'postgresql':
                return True
```

- Or otherwise maybe directly import and use [`_installed_apps` from app_utils](https://github.com/YunoHost/yunohost/blob/3c339cc33a55a91005cbbcbb1eb43a88f554968b/src/utils/app_utils.py#L894) which is [what calls `app_list` internally to get the list of app ids](https://github.com/YunoHost/yunohost/blob/3c339cc33a55a91005cbbcbb1eb43a88f554968b/src/app.py#L136).

fflorent (commented):
Thank you!

The code looks OK to me (except few remarks, not sure how important they are).

If I understand correctly [your point here](https://github.com/YunoHost/issues/issues/2737#issuecomment-5379403084), the migration issue occurred because despite having had coin, discourse and pgadmin requiring postgresql, the command in `ynh_deps_cmd` returned no result.

I wonder if this is not due to the fact that discourse adds the apt.postgresql.org extra repository: https://github.com/YunoHost-Apps/discourse_ynh/blob/5894d970f9a0ddd8c8c79b01636fae2d506cecc9/manifest.toml#L134

Which is discussed here: https://github.com/YunoHost-Apps/immich_ynh/issues/432#issuecomment-5422272418

On my side, that could be interesting to check what happens in a ynh-dev container when installing discourse and migrating.

PS: Also related, I wonder if the command `ynh_deps_cmd` could not be simplified: https://github.com/YunoHost/issues/issues/2737#issuecomment-5461557719

Thovi98:
> I wonder if this is not due to the fact that discourse adds the apt.postgresql.org extra repository

For the record, the psgl extra repo has been added less than a year ago (in October 2025), so after Josue's migration in March 2025 https://github.com/YunoHost-Apps/discourse_ynh/commit/c958405802ccf194653671f34050e016e72ab8c0.

Josue-T:
> Thank you!
> 
> The code looks OK to me (except few remarks, not sure how important they are).
> 
> If I understand correctly [your point here](https://github.com/YunoHost/issues/issues/2737#issuecomment-5379403084), the migration issue occurred because despite having had coin, discourse and pgadmin requiring postgresql, the command in `ynh_deps_cmd` returned no result.
> 
> I wonder if this is not due to the fact that discourse adds the apt.postgresql.org extra repository: https://github.com/YunoHost-Apps/discourse_ynh/blob/5894d970f9a0ddd8c8c79b01636fae2d506cecc9/manifest.toml#L134
> 
> Which is discussed here: [YunoHost-Apps/immich_ynh#432 (comment)](https://github.com/YunoHost-Apps/immich_ynh/issues/432#issuecomment-5422272418)
> 
> On my side, that could be interesting to check what happens in a ynh-dev container when installing discourse and migrating.
> 
> PS: Also related, I wonder if the command `ynh_deps_cmd` could not be simplified: [YunoHost/issues#2737 (comment)](https://github.com/YunoHost/issues/issues/2737#issuecomment-5461557719)

Well my understanding of the issue is that sometimes on the upgrade the dependency management is not easy to handles and maybe it could happen that the ynh dependency Debian package could be removed. And so ever if this happen, (a gree it should not happen and it's not Commons but I not sure we can excludes this cases) we need to ensure that the psql migration happen. That's the purpose of this PR.

fflorent (approved):
I could reproduce the issue by removing the `{app)-deps-ynh` deb packages of apps requiring postgresql:
- in trixie, the migration does not run
- with the patch, the migration is run as expected.

Thanks to incus snapshots and to #2361.

Also I could check how the script is run thanks to the python debugger tools.


So LGTM modulo this comment: https://github.com/YunoHost/yunohost/pull/2358/changes#r3886071986

src/migrations/postgresql.py
```diff
@@ -104,6 +109,13 @@ def run_post_migration(self):
             self.runcmd(f"{sudocmd} psql --dbname='{database}' --command='REINDEX DATABASE {database};'")
             self.runcmd(f"{sudocmd} psql --dbname='{database}' --command='ALTER DATABASE {database} REFRESH COLLATION VERSION;'")
 
+    def _has_app_with_psql_resource(self) -> bool:
+        for app_info in yunohost.app.app_list()['apps']:
```
fflorent:
```suggestion
        for app_info in app_list()['apps']:
```

fflorent (approved):

fflorent:
With the agreement of the author, I applied my suggested changes. After having checked it still works, I merge.

Thank you @Josue-T :)

PR #2361, opened 2026-08-23
migrations: Add --skip-postmigrations option

fflorent:
## The problem

As of today, testing the venv and postgresql migrations, despite being simple and fast to run, can be painful: you would have to run the whole Trixie migration, and if it fails, run it again. And again… Wait, what time is it now?

A trick is to comment lines of the Trixie migration to avoid to run those migrations, as described in this PR (section "How to test"):
https://github.com/YunoHost/yunohost/pull/2353

Quoting:
> - comment these lines to disable the automatic run of migrations following the one for trixie: https://github.com/YunoHost/yunohost/blob/83df92b76061a332aef8b05554251d5c2ee9fec8/src/migrations/0036_migrate_to_trixie.py.disabled#L343-L346

## Solution

Introduce the `--skip-postmigrations` option, to avoid the run of the `yunhost tools migrations run` command.

I am not much happy with the name, if you find a better one, I would be grateful!

**AI transparency:** not used

## PR Status
*Fully tested*

## How to test

1. Use the `ynh-dev` container
2. Switch to this PR's branch in `ynh-dev/yunohost`
3. Prepare your environment before the migration
4. Run: `./ynh-dev use-git yunohost` so the changes brought by this branch are run
5. Make a snapshot
6. Enable the Trixie migration: `mv /usr/lib/python3/dist-packages/yunohost/migrations/0036_migrate_to_trixie.py{.disabled,}`
7. And run it: `yunohost tools migrations run --accept-disclaimer --skip-postmigrations`
8. Check the migration went well and that the migrations for venv and postgresql are listed as pending: `yunohost tools migrations list`
9. ⚠️IMPORTANT: at this step, reset the `ynh-dev/yunohost` repo to the state of HEAD (changes have been brought by the migration)
10. Restore the snapshot of step 5, you may run the migration again without the `--skip-postmigrations` to validate the venv and postgresql are this time run automatically.

### DIFF
--- a/src/tools.py
+++ b/src/tools.py
@@ -801,6 +801,7 @@ def tools_migrations_run(
     auto: bool = False,
     force_rerun: bool = False,
     accept_disclaimer: bool = False,
+    skip_postmigrations: bool = False,
 ) -> None:
     """
     Perform migrations
@@ -810,6 +811,7 @@ def tools_migrations_run(
     --auto         Automatic mode, won't run manual migrations (to be used only if you know what you are doing)
     --force-rerun  Re-run already-ran migrations (to be used only if you know what you are doing)(must explicit which migrations)
     --accept-disclaimer  Accept disclaimers of migrations (please read them before using this option) (only valid for one migration)
+    --skip-postmigrations  Don't automatically run the following migrations if any. Especially useful to not run the migrations following a Debian version bump and run them manually instead (to be used only if you know what you are doing)
     """
 
     all_migrations = _get_migrations_list()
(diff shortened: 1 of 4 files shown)


src/tools.py
```diff
@@ -917,6 +919,9 @@ def get_matching_migration(target):
         operation_logger = OperationLogger("tools_migrations_migrate_forward")
         operation_logger.start()
 
+        if skip_postmigrations:
+            migration.skip_postmigrations = True
```
zamentur:
```suggestion
        migration.skip_postmigrations = skip_postmigrations
```

Salamandar (approved):

